PR #1, opened 2019-11-21
Refactoring

tomas-pluskal:
@robinschmid It's still work in progress, but feel free to take a look!


robinschmid:
I think I like the new structure. What do you think we could try to use a short + accurate version for prefixes.
- normalization --> norm_
- gapfilling --> gapfill_
- filtering --> filter_
- alignment --> align_
- featuredetection --> (no idea - maybe featuredetect_)

tomas-pluskal:
Shorter prefixes sound good. Perhaps we can use featdet for featuredetection, although it is a bit cryptic.

tomas-pluskal:
Shall we merge this now? Of course we can continue refactoring and cleaning up afterwards. I would like to start working on moving the the MSDK data model for raw data.

robinschmid:
Merged my first PR - roles are switched. hahahhaha.

tomas-pluskal:
Yay! :)

PR #8, opened 2019-11-25
Feature data structure

robinschmid:
@kaibioinfo 
@tomas-pluskal 

I have created a preview on how we could change the feature data model to make it modular. 
What do you think?

- Every new data entity needs to be provided into an enum
- Then all the data is added to FeatureDataMap 
- Each DataType 
  - knows how to format the data
  - could also know how to visualize in feature list / plots

Is there any way to change the structure so that we directly get the correct DataType from the map?


kaibioinfo:
It looks very much like CDK code, which might be a good thing, as that is a mature library.

However, I don't like type casting, and your model seems to enforce that. Furthermore, an enum is a strange way to express modularity, as it can't be extended from outside. When everyone has to add an enum key to add his own data types, he could also just add this datatype directly to the Feature class.

The way we handle this in SIRIUS is:
```java
interface DataType {
  public abstract String getFormattedString();
  // other required methods for serialization and so one
}

public interface HasExtendableDataType {
  protected HashMap<Class<DataType>, DataType> dataTypeMap;
  
  default public <T extends DataType> void set(Class<T> type, T value)  {
    dataTypeMap.put(type, value);
  } 

  default public <T extends DataType> Optional<T> get(Class<T> type, T value)  {
    T value = (T)dataTypeMap.get(type);
    return Optional.ofNullable(value);
  } 
}
```

The advantage is that the user never has to cast anything.
```java

class RTValue {
  public final long rtInSeconds;
  //...
}

feature.set(RTValue.class, new RTValue(1234));
long time = feature.get(RTValue.class).rtInSeconds;
```

robinschmid:
Sounds good to me. 
I thought that some DataTypes might be included multiple times and I wanted to use the enum to enforce the correct DataType and also to list all possible datatypes for a feature. 
However, we are going to change it to a similar structure to your proposal. 

Thanks Kai.

robinschmid:
@kaibioinfo thanks for the provided code snippets. 

#### DataType:
``` java
public abstract class DataType<T> {

  protected final T value;

  public DataType(T value) {
    this.value = value;
  }

  public T getValue() {
    return value;
  }
.....
.....
}
```

#### RowData
``` java

public class RowData {

  @SuppressWarnings({"rawtypes"})
  private HashMap<Class<? extends DataType<?>>, DataType> map = new HashMap<>();

  public <T extends DataType<?>> Optional<T> get(Class<T> type) {
    return Optional.ofNullable(map.get(type));
  }

  public void set(Class<? extends DataType<?>> key, DataType<?> data) {
    if (key.isInstance(data))
      map.put(key, data);
    // wrong data type. Check code that supplied this data
    else
      throw new WrongTypeException(key, data);
  }
}
```

This allows:
```java
get(MZType.class).map(DataType::getValue).orElse(0d);
```

src/main/java/io/github/mzmine/datamodel/fx/FeatureTableFX.java
```diff
@@ -0,0 +1,170 @@
...
+              setText(item.getFormattedString());
+              setGraphic(null);
+            }
+          }
+        }
+      });
+      // add to table
+      this.getColumns().add(col);
+    }
+  }
```
robinschmid: RowData and DataType lead to a very easy and clean way to create the columns in the TreeTableView

src/main/java/io/github/mzmine/datamodel/data/types/DetectionType.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Override
+  public Node getCellNode(TreeTableCell<RowData, DataType<?>> cell,
+      TreeTableColumn<RowData, ? extends DataType> coll) {
+    Circle circle = new Circle();
+    circle.setRadius(10);
+    circle.setFill(value.getColorFX());
+    return circle;
+  }
```
robinschmid: In the table, all DataTypes are represented by their DataType#getFormattedString method
unless they implement GraphicalCellData with this method

This example shows the FeatureStatus DataType

src/main/java/io/github/mzmine/datamodel/fx/FeatureTableFX.java
robinschmid: Columns in the FeatureTableFX are then created on the basis of a "summary" RowData instance which holds all DataTypes present in one of the rows

src/main/java/io/github/mzmine/datamodel/fx/FeatureTableFX.java
```diff
@@ -0,0 +1,170 @@
...
+      // value binding
+      TreeTableColumn<RowData, ? extends DataType> col =
+          new TreeTableColumn<>(dataType.getHeaderString());
+      col.setCellValueFactory(r -> {
+        Optional<? extends DataType> o = r.getValue().getValue().get(dataType.getClass());
+        if (o.isPresent())
+          return new ReadOnlyObjectWrapper<>(o.get());
+        else
+          return null;
+      });
```
robinschmid: The only question is how we actually listen to changes in the data model. 
The values are final. So we would need to listen to changes in the HashMap.

kaibioinfo: Values definitely should be immutable! We had some trouble in SIRIUS in the past, because I wrote in the doc they should be final, but sometimes people add some mutable values just for convenience, which made it later impossible to make copies of data types. Was a lot of work to change all the code in SIRIUS to ensure that all data types are immutable ^^

Regarding to listeners: I think the system allows for listeners quite easily. I think about something like
```java
// will always call the onChange method on the listener whenever set(Class<T>,T) is called
public void <T extends DataType> addListener(Class<T>, DataTypeListener<T>();
```
it's not fully compatible with the bean concept, though. But I'm not that familiar with that anyways. But you have to consider what's the best way to integrate all these stuff into javafx.

robinschmid:
The table is now constructed from ModularFeatureListRow and ModularFeature. 
- All columns are created within the DataType classes.
- Sample specific columns are created within FeaturesType<Map<RawDataFile, ModularFeature>>

#### Example of some new things that we can achieve
image

#### Without the bar chart
image

tomas-pluskal:
@robinschmid Looks awesome, this is going to be a huge step forward for MZmine!

robinschmid:
### First real data FeatureListFX
ModularADAPChromatogramBuilder creates:
- modularFeature = new ModularFeature(chromatogram)
- new ModularFeatureListRow(id, raw, modularFeature)

image

robinschmid:
### Latest version.
image

robinschmid:
ModularFeatureList has a list of row column bindings now.
These bindings can for example bind the MZType of row to the average of all MZType of all features.
- average
- sum
- min
- max
- range
- count

The NumberDataTypes create the actual bindings.

In the screencap, all row values are actually calculated with bindings (except for ID and Comment)

robinschmid:
This should be ready for merging. Can you merge and review the data structure? 
Sadly, I will need to focus on writing the next months.

Functions:
Feature table has 
- rowTypes, featureTypes (as columns)
- RowBindings to bind values from all features to the row (average mz, max height, span all ranges)

FeatureListRow and Feature
- all data types automatically add a new property to the map<DataType, Property<?>>
    - the property is final (to track changes), only the wrapped value should be changed by the set method
- bindings for row data types are added
- CommentType is the ónly editable column so far but works fine
    - by defining a StringConverter in the DataType
 
We cannot use DoubleProperty, IntegerProperty, and other primitive wrappers - we would want to stick to ObjectProperty<Double> ... 

### TODO
We should create a folder for test classes - @tomas-pluskal can you do this?
I have one test class in datamodel/data/test this would need to move. 

Switch to the javafx main window.

We should create a new interface for the modular featurelist, row, and feature and then replace the old interfaces so that every method points to the new module - then change all to use the new model.

Remove the module ModularADAPChromatogramBuilder.

tomas-pluskal:
Thanks Robin! It seems the gradle build currently fails:
````
> Task :compileJava
/home/travis/build/mzmine/mzmine3/src/main/java/io/github/mzmine/datamodel/data/ModularDataModel.java:220: error: method getTypeColumn in interface ModularDataModel cannot be applied to given types;
    DataType type = getTypeColumn(tclass);
                    ^
  required: Class<? extends DataType<T>>
  found:    Class<CAP#1>
  reason: cannot infer type-variable(s) T
    (argument mismatch; Class<CAP#1> cannot be converted to Class<? extends DataType<T>>)
  where T is a type-variable:
    T extends Property<?> declared in method <T>getTypeColumn(Class<? extends DataType<T>>)
  where CAP#1 is a fresh type-variable:
    CAP#1 extends DataType<?> from capture of ? extends DataType<?>
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
````

robinschmid:
Weird. I can run it. Do you have an idea how I can fix it? Maybe just change the Class<? extends DataType<T>> to Class<? extends DataType>

tomas-pluskal:
This seems to compile fine:
````
  default <T extends Property<?>> void remove(Class<? extends DataType<T>> tclass) {
    DataType type = getTypeColumn(tclass);
    if (type != null)
      getMap().remove(type);
  }
````

tomas-pluskal:
Could you please also reformat the files as I mentioned in the other PR?

robinschmid:
I checked it. However, the DataType<T> was important to get the correct type for the get() and getValue() methods

tomas-pluskal:
I don’t understand the last comment. In the previous comment I just changed the signature of the remove() method and that fixed the compiler error.

robinschmid:
Sorry I commented without seeing your comment/refreshing my browser.

tomas-pluskal:
Now there is a conflict in MZmineModulesList.java

robinschmid:
Done. So the problem with the ModulesList still remains. This is a bummer. No idea how we could change this. I thought git would understand if we break the lines with comments - but it seems to fail still.

PR #12, opened 2019-12-02
File chooser dialogs using JavaFX

tomas-pluskal:
I finally found how to run JavaFX code from the Swing threads using **FutureTask**. Should have realized this earlier. Anyway, I changed most of the file chooser dialogs to JavaFX versions and it works quite nice!


src/main/java/io/github/mzmine/util/components/FileNameComponent.java
```diff
@@ -1,101 +0,0 @@
...
-import javax.swing.JFileChooser;
-import javax.swing.JPanel;
-import javax.swing.JTextField;
-
-import io.github.mzmine.main.MZmineCore;
-
-/**
- * Component with textbox for filename and a button to lookup the file
- */
-public class FileNameComponent extends JPanel implements ActionListener {
```
robinschmid: this was an old duplicate I guess?

PR #23, opened 2019-12-15
JavaFX main window

tomas-pluskal:
Still work in progress but good for preview


robinschmid:
You are removing the module list?
How do you plan to replace it?

tomas-pluskal:
Yes, I switched to lazy module initialization - the instance of each module is created when it is needed. It speeds up the startup a bit. And we will have better control over the main menu, which is now defined in the FXML file: https://github.com/tomas-pluskal/mzmine3/blob/javafxgui/src/main/java/io/github/mzmine/gui/mainwindow/MainMenu.fxml

Issue #26, opened 2019-12-17
a problem after merging a new module for sorting predicted formulas


kkcau9118:
### Basic information

* My operating system: Windows10
* My MZmine version: MZmine3

### What happened

Dear MZmine developer,

This is Kai, again... First, I would like to really thank the MZmine developers for helping me solve several problems. Now I have a new problem. Today I merged a new module for sorting predicted formulas (which was contributed by Robin) in my MZmine3 repository and then cloned this modified MZmine3 to my Eclipse for running. However, when I did the step of formula prediction (by clicking Feature list method/identification/formula prediction) ,it looked that nothing happened and there was no task progress bar running on the screen and finally no predicted formulas were observed. It means that the function of formula prediction did not work. Could you please tell me how to figure it out? I have attached the java codes in the text file.
[20191217_codes of formula prediction.txt](https://github.com/mzmine/mzmine3/files/3974367/20191217_codes.of.formula.prediction.txt)

tomas-pluskal:
@robinschmid Seems like a NPE in the FormulaPredictionPeakListTask constructor.

@kkcau9118 Could you add screenshots of the exact parameter values you used to run the module? Including all the sub-module settings please (like the isotope pattern filter etc.).

kkcau9118:
> @robinschmid Seems like a NPE in the FormulaPredictionPeakListTask constructor.
> 
> @kkcau9118 Could you add screenshots of the exact parameter values you used to run the module? Including all the sub-module settings please (like the isotope pattern filter etc.).

Hi,

I just run this module again and it still did not work. Here is the text file of java code and the screenshots (I set the mass tolerance m/z to 0 m/z or 3 ppm, since I prefer to only use the unit of ppm in this step).
[20191217_02_codes of formula prediction.txt](https://github.com/mzmine/mzmine3/files/3975024/20191217_02_codes.of.formula.prediction.txt)
![image](https://user-images.githubusercontent.com/57988218/71026825-43da7000-210a-11ea-9a0e-ee928043bf42.png)
![image](https://user-images.githubusercontent.com/57988218/71026838-4b017e00-210a-11ea-8233-f33de485c2fd.png)
![image](https://user-images.githubusercontent.com/57988218/71026857-53f24f80-210a-11ea-85a7-60e4d1f8ac91.png)

tomas-pluskal:
Should be now fixed in master

robinschmid:
[...]

PR #28, opened 2019-12-18
Fixed another NPE in formula prediction

robinschmid:
Same fix for checkIsotopes to prevet NullPointerException


robinschmid:
comparable to #27 and issue #26 

sorry guys. This slipped through my attention. I included a check for the sorting parameters but missed the check on the isotope and MSMS parameters

tomas-pluskal:
Looks good.

Another reason to be using objects instead of primitive types - here the NPE happens during the unboxing conversion from Double to double because the Double is null. If we were using Double everywhere, there would be no exception.

robinschmid:
Not completely true... Most of the libraries use primitives (JFreeChart, Guava, appache.commons.math) ... at one point there will be unboxing and there will be an NPE. However, we could defiitely limit the exceptions in MZmine code with this.

PR #43, opened 2020-03-11
Theme implementation to SpectraPlot and TICPlot

SteffenHeu:
Hi Tomas,

I moved most of the appearance settings from the individual plot's constructors to the theme settings.

Best
Steffen


tomas-pluskal:
Awesome, thanks a lot!

tomas-pluskal:
The palette setup dialog looks a little chaotic, especially when resized:
Screen Shot 2020-03-11 at 21 12 38
How about adding some labels to the components and arranging them into a grid?
Also, it currently accepts empty Name, which I think it shouldn't.

tomas-pluskal:
I am not sure why we need the "Add default" button. The default palettes should always be there and non-changeable. You can also remove the "Color palettes (color blindness mode)" parameter.

SteffenHeu:
grafik
This is how it looks now.

I think removing the blindness parameter should be a different PR.

tomas-pluskal:
I think you can remove the blindness parameter in this PR, and instead make three (not one) default palettes, which correspond to the different settings of the old blindness parameter.

Regarding adding colors, perhaps it would make sense to place the Color, Add color, and Remove color components right under the color palette. The Name parameter can be in the top. And the Accept and Cancel buttons can be in the bottom. 
By the way, how does the Cancel button work? If you add a color to the palette and then click Cancel, the color is removed again?

SteffenHeu:
Regarding the cancel button: 
Yes, the palette gets cloned for the picker dialog. If it gets changed and accepted it is replaced, if canceled nothing happens.

SteffenHeu:
grafik

tomas-pluskal:
Thanks, it's starting to look fantastic!

When I click "New" it creates an empty palette with no name, and clicking Edit on it throws an exception. Perhaps when you create a new palette you should give it a default name ("New palette") and add at least 1 color to it (black?).

When I edit the palette and remove the name, the Cancel button gives an error. I think Cancel should always work.
Screen Shot 2020-03-19 at 09 17 55

For consistency, it would be nice to use a ButtonBar for the Accept and Cancel buttons  and move them to the bottom of the window, same as in the ParameterSetupDialog.
Screen Shot 2020-03-19 at 09 21 09

Also please give the dialog a title.

tomas-pluskal:
It's a personal preference, but I would prefer to avoid abbreviations (cpy, pos, neu, neg) and instead spell out the full words.

robinschmid:
Looks great. Can you drag and drop the colors to change the order?

SteffenHeu:
> 
> 
> Looks great. Can you drag and drop the colors to change the order?

Yes you can.

Without the abbreviations the component would get pretty large I think.

tomas-pluskal:
Thanks @SteffenHeu !
Please set the minimum size of the palette editing Stage, otherwise it can be resized too small.
Screen Shot 2020-03-20 at 22 53 51
Also would be nice to add some padding around the Ok Cancel buttons and separate them by a horizontal line (in other words, use identical layout to the ParameterSetupDialog).

Removing all colors from the palette leads to exceptions and weird behavior.

By the way, with your current code I see these errors during MZmine startup. Please check the code for loading/saving of the values.
````
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: java.lang.IllegalArgumentException: Invalid color specification
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: Could not load positve/negative colors of Normal. Setting default colors
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: java.lang.IllegalArgumentException: Invalid color specification
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: Could not load positve/negative colors of Deuteranopia. Setting default colors
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: java.lang.IllegalArgumentException: Invalid color specification
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: Could not load positve/negative colors of Protanopia. Setting default colors
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: java.lang.IllegalArgumentException: Invalid color specification
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: Could not load positve/negative colors of Tritanopia. Setting default colors
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: java.lang.IllegalArgumentException: Invalid color specification
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: Could not load positve/negative colors of Deuternopia. Setting default colors
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: java.lang.IllegalArgumentException: Invalid color specification
[22:50:28|WARNING|io.github.mzmine.util.color.SimpleColorPalette]: Could not load positve/negative colors of . Setting default colors
````

**Comment by SteffenHeu** [comment]
@tomas-pluskal The warnings you are receiving are intendet behaviour. They should be caused, because you have older versions of the color palettes saved in your xml files, that did not contain the neg/pos/neu colors. So the values cannot be read from the xml.

The exception is handled here: https://github.com/SteffenHeu/mzmine3/blob/bbef859edeb36e993e4b22057c18c84474fdb2e0/src/main/java/io/github/mzmine/util/color/SimpleColorPalette.java#L288
`public void loadFromXML(Element xmlElement) {
    this.setName(xmlElement.getAttribute(NAME_ATTRIBUTE));
    String text = xmlElement.getTextContent();

    String pos = xmlElement.getAttribute(POS_ATTRIBUTE);
    String neg = xmlElement.getAttribute(NEG_ATTRIBUTE);
    String neu = xmlElement.getAttribute(NEU_ATTRIBUTE);

    Color clrPos, clrNeg, clrNeu;

    try {
      text = text.substring(1, text.length() - 1);
      text = text.replaceAll("\\s", "");
      String[] clrs = text.split(",");
      for (String clr : clrs) {
        delegate.add(Color.web(clr));
      }
    } catch (Exception e) {
      logger.warning(e.toString());
      logger.warning("Could not load color palette " + name + ". Setting default colors.");
      this.addAll(
          ColorsFX.getSevenColorPalette(Vision.DEUTERANOPIA, true));
    }

    try {
      clrPos = Color.web(pos);
      clrNeg = Color.web(neg);
      clrNeu = Color.web(neu);
    } catch (Exception e) {
      logger.warning(e.toString());
      logger.warning(
          "Could not load positve/negative colors of " + name + ". Setting default colors.");
      clrPos = ColorsFX.getPositiveColor(Vision.DEUTERANOPIA);
      clrNeg = ColorsFX.getNegativeColor(Vision.DEUTERANOPIA);
      clrNeu = ColorsFX.getNeutralColor();
    }`

Should i remove the logging of the error message itself and just log that the color was not loaded?

File: src/main/java/io/github/mzmine/util/color/SimpleColorPalette.java [hunk 1]
```diff
@@ -204,31 +274,57 @@ public SimpleColorPalette clone() {
...
+
+    try {
+      text = text.substring(1, text.length() - 1);
+      text = text.replaceAll("\\s", "");
+      String[] clrs = text.split(",");
+      for (String clr : clrs) {
+        delegate.add(Color.web(clr));
+      }
+    } catch (Exception e) {
+      logger.warning(e.toString());
```
tomas-pluskal: When logging an exception (Throwable), please use this method:
https://docs.oracle.com/en/java/javase/13/docs/api/java.logging/java/util/logging/Logger.html#log(java.util.logging.Level,java.lang.String,java.lang.Throwable)
That way you can log the error message and the exception at the same time.

src/main/java/io/github/mzmine/parameters/parametertypes/colorpalette/ColorPalettePreviewField.java
```diff
@@ -83,30 +95,44 @@ private void setRectangles() {
...
+
         Point2D exit = new Point2D(e.getSceneX(), e.getSceneY());
-        
+
         double x = this.sceneToLocal(exit).getX();
-        int newIndex = (int) (x / RECT_HEIGHT + .5);
+        double y = this.sceneToLocal(exit).getY();
+        int rows = (int) ((RECT_HEIGHT * palette.size()) / getWidth() + 1);
+
+        logger.info("rows: " + rows);
```
tomas-pluskal: Please clean up these messages, they are not informative enough to be logged at INFO level

src/main/java/io/github/mzmine/parameters/parametertypes/colorpalette/ColorPalettePreviewField.java
```diff
@@ -83,30 +95,44 @@ private void setRectangles() {
...
+
+        x = (x < 0) ? 0 : x;
+        y = (y < 0) ? 0 : y;
+        y = (y / RECT_HEIGHT <= rows) ? y : rows;
+        logger.info("y: " + y);
```
tomas-pluskal: Also here

tomas-pluskal:
Awesome improvements! 

I noticed this warning when I opened the dialog:
[9:04:37|INFO|io.github.mzmine.gui.mainwindow.MainMenuController]: Showing the Preferences dialog
````
[9:04:37|WARNING|io.github.mzmine.parameters.parametertypes.colorpalette.ColorPaletteComponent]: Value of ColorPaletteComponent was set to a value not contained in the items. This might lead to unexpected behaviour.
````
I just opened the dialog after starting MZmine, so I think this should not cause a warning.

By the way, you could indicate the drag&drop procedure visually using the setDragView() and setOpacity() functions. See here:
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/util/javafx/DraggableListCell.java#L48
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/util/javafx/DraggableListCell.java#L62

tomas-pluskal:
I also noticed these log messages, which probably should be more informative: 
````
[10:41:08|INFO|io.github.mzmine.util.color.SimpleColorPalette]: [palette: null]
````

SteffenHeu:
> By the way, you could indicate the drag&drop procedure visually using the setDragView() and setOpacity() functions. See here:
> https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/util/javafx/DraggableListCell.java#L48
> https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/util/javafx/DraggableListCell.java#L62

Thanks @tomas-pluskal, that made me understand how the drag & drop works properly. I will change the whole drag & drop implementation now to make it proper!

tomas-pluskal:
Thanks @SteffenHeu, I think it looks great now! 

I noticed the "Show data points" button shows the data points in white color, could you please check why that happens?
Screen Shot 2020-03-25 at 22 48 54

tomas-pluskal:
Thanks @SteffenHeu ! Are you still working on this or is it ready for merging?

SteffenHeu:
I'd say it's ready. We might find some bugsor inconsistencies at some point, but they can be fixed then.

tomas-pluskal:
Thanks @SteffenHeu ! It's a great addition.

PR #44, opened 2020-03-13
Removed unnecassry files

knakul853:
In this PR I tried to delete unnecessary previous files that don't have any dependencies and are not being used.


tomas-pluskal:
Great, thanks!

PR #47, opened 2020-03-15
I have changed the PeriodTable from Swing to Javafx

FourierX9:


src/main/java/io/github/mzmine/util/dialogs/DialogController.java
```diff
@@ -0,0 +1,1318 @@
...
+
+    private String result = "Unknown";
+    private int atomicNumber = 0;
+    private String elementSymbol = new String();
+
+
+
+
+
+    private void elementNameTranslator(int rowIndex, int colIndex) {
```
knakul853: Hey I think we can compress this with some data structures.this si too big and naive implementation.
Let me know if you have any suggestions.I think using loop and map and some other 
Equivalent data structures we can compree this

tomas-pluskal: I agree, this implementation is very impractical.
@FourierX9 please use the Elements enum from CDK (http://cdk.github.io/cdk/2.2/docs/api/org/openscience/cdk/config/Elements.html) to obtain the periodic table data. I think the enum contains all the data needed to populate the periodic table.

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
```diff
@@ -0,0 +1,1318 @@
...
+            return GT.get("Gas");
+        else if (serie.equals("Liquid"))
+            return GT.get("Liquid");
+        else if (serie.equals("Solid"))
+            return GT.get("Solid");
+        else
+            return GT.get("Unknown");
+    }
+
+private static class GT {
```
tomas-pluskal: What is this?

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
```diff
@@ -0,0 +1,1318 @@
...
+                        + DialogController.GT.get("State:") + " " + phaseTranslator(PeriodicTable.getPhase(elementSymbol))
+                        + "<br> " + DialogController.GT.get("Electronegativity:") + " "
+                        + (PeriodicTable.getPaulingElectronegativity(elementSymbol) == null ? DialogController.GT.get("undefined")
+                        : PeriodicTable.getPaulingElectronegativity(elementSymbol))
+                        + "<br>" + "</FONT></html>"
+        );
+        fontWebview_center.getEngine().loadContent(sb_Center.toString());
+    }
+
+    @FXML
```
tomas-pluskal: What is the purpose of these functions?

src/main/java/io/github/mzmine/gui/mainwindow/MainMenuController.java
```diff
@@ -1,178 +1,184 @@
...
-          // load file
-          ProjectOpeningTask newTask = new ProjectOpeningTask(f);
-          MZmineCore.getTaskController().addTask(newTask);
-        }
-      });
-      recentProjectsMenu.getItems().add(item);
-    });
-  }
-}
+/*
```
tomas-pluskal: This pull request contains unnecessary formatting changes. Please use the Google Code Style for java and reformat your code. The formatting configuration files for Eclipse and IDEA are in the mzmine3 repository.

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
```diff
@@ -0,0 +1,1318 @@
+package io.github.mzmine.util.dialogs;
```
tomas-pluskal: Please add the license header to all new .java files (copy from other files). Rename this class to PeriodTableDialogController.

src/main/java/io/github/mzmine/util/dialogs/PeriodicTableDialog.java
```diff
@@ -1,74 +1,46 @@
...
-    setLayout(new BorderLayout());
+    private DialogController periodicTable;
+    private IIsotope selectedIsotope;
 
-    periodicTable = new PeriodicTablePanel();
-    periodicTable.addCDKChangeListener(this);
-    add(BorderLayout.CENTER, periodicTable);
 
-    pack();
+    public void start(Stage primaryStage) throws Exception {
```
tomas-pluskal: Remove this start() method and instead make `PeriodicTableDialog` extend `Stage`

FourierX9: Did you mean as changing the function name from start() to PeriodicTableDialog()?

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
FourierX9: Got it!

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
FourierX9: Got it thanks!

src/main/java/io/github/mzmine/util/dialogs/PeriodicTableDialog.java
tomas-pluskal: No I mean remove this function completely, it is not needed,
The class PeriodicTableDialog should extend Stage.

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
FourierX9: I want to use this function to load the StringBuilder into the Webview. I am thinking that those two stringbuilder can be combined together?

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
FourierX9: I reused the class GT because I want to reuse some other old code like the html code in the StringBuilder so that the appearance can be similar with the old one and this class GT was called in other code.

src/main/java/io/github/mzmine/util/dialogs/PeriodicTableDialog.java
FourierX9: Got it! I will try it out.

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
tomas-pluskal: The class GT has no function in your code. And this function
````
 private static String get(String s) {
        return s;
   }
````
also does absolutely nothing. You should remove them.

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
tomas-pluskal: To be honest, I think the WebView solution is not so great here. I would prefer if you just use standard (non HTML-based) JavaFX components like Label. The layout of the information doesn't need to 100% correspond to the previous Swing version.

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
FourierX9: Got it

src/main/java/io/github/mzmine/util/dialogs/DialogController.java
FourierX9: Sure I will use Label

src/main/java/io/github/mzmine/util/dialogs/PeriodicTableDialog.java
FourierX9: I've looked up to some resources, but I cannot figure out how can I load the fxml files after I deleted the start function? Is there any websites that I can look up to?

tomas-pluskal: See this one for example: https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/modules/dataprocessing/id_onlinecompounddb/ResultWindowFX.java

FourierX9:
Thanks for all the comments! I have updated the code. Hope you all the best!

tomas-pluskal:
Thanks, it's looking pretty good now! I'll make a few final adjustments myself.

PR #50, opened 2020-03-16
Swing to javafx in SQLColumnSettingsComponent

itissandeep98:
This is the task assigned to me by @tomas-pluskal .
In this I have translated all component from Swing to JavaFX in:
io.github.mzmine.modules.io.sqlexport.SQLColumnSettingsComponent. I tried to stay as close to original swing code and just provide with alternative JavaFX code, also I have created a new File SQLRowObject.java which is used to store data of every row


pom.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tomas-pluskal: This file should not be included in the PR

.gitignore
```diff
@@ -10,3 +10,5 @@ mzmine.log*
 /target
 /build
 /bin
+
+sample-2.39.mzmine
```
tomas-pluskal: This change should not be included in the PR

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettings.java
```diff
@@ -32,6 +35,7 @@
   List<String> columnNames = new ArrayList<String>();
   List<SQLExportDataType> columnTypes = new ArrayList<SQLExportDataType>();
   List<String> columnValues = new ArrayList<String>();
+  ObservableList<SQLRowObject> list;
```
tomas-pluskal: If you store all the information in the ObservableList<SQLRowObject>, then you can remove the columnNames, columnTypes, and columnValues lists, because they are redundant.

tomas-pluskal: Also the list should be declared private final.

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettingsComponent.java
```diff
@@ -18,152 +18,122 @@
...
+            }
+        });
+
+
+        //Todo: change the below swing code to JavaFX
+//        DefaultCellEditor dataTypeEditor = new DefaultCellEditor(dataTypeCombo);
+//        columnsTable.setDefaultEditor(SQLExportDataType.class, dataTypeEditor);
+
+        // Add buttons
+        VBox buttonsPanel=new VBox(20);
```
tomas-pluskal: Better to use ButtonBar instead of Vbox (https://openjfx.io/javadoc/13/javafx.controls/javafx/scene/control/ButtonBar.html)

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLRowObject.java
```diff
@@ -0,0 +1,41 @@
+package io.github.mzmine.modules.io.sqlexport;
```
tomas-pluskal: Add the GPL license header to all new .java files

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLRowObject.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class SQLRowObject {
+    String Name;
```
tomas-pluskal: By Java conventions the attribute names should start with lowercase letter (name, type, value not Name, Type, Value). And you should declare them private.

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettingsComponent.java
itissandeep98: But according to your screenshot, the buttons were ordered in a vertical fashion and by using A Button Bar it will result in horizontal Buttons.
image
instead of vertical buttons
image

tomas-pluskal: You are right, sorry for that. Please leave a VBox then.

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettings.java
```diff
@@ -65,62 +66,64 @@ public String getColumnName(int col) {
 
   @Override
   public synchronized int getRowCount() {
```
tomas-pluskal: This class extends AbstractTableModel, but that is not needed anymore, since JavaFX doesn't use table models. You can remove the extends clause and remove the methods that implement TableModel, like getRowCount, getColumnCount, isCellEditable, getValueAt.. any method that you are not directly using in your code.

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettings.java
```diff
@@ -29,9 +32,7 @@
    * 
    */
   private static final long serialVersionUID = 1L;
```
tomas-pluskal: You can remove the serialVersionUID, that is only needed for Swing classes because those are declared Serializable.

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLRowObject.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along with MZmine; if not,
+ * write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301
+ * USA
+ */
+
+package io.github.mzmine.modules.io.sqlexport;
+
+public class SQLRowObject {
+    String name;
```
tomas-pluskal: Declare the fields as private.

tomas-pluskal:
It is starting to look good! 
One important point is that the table needs to be editable.

itissandeep98:
> See my comments to the PR

From my side I have made each cell of the tables editable, I just can't check for it because for that I need some database login details which has this kind of data

tomas-pluskal:
> > See my comments to the PR
> 
> From my side I have made each cell of the tables editable, I just can't check for it because for that I need some database login details which has this kind of data

Actually you don't need any database to test this. The data in the table is supposed to be editable by the user, the database connection only happens after the dialog is closed.

itissandeep98:
>Actually you don't need any database to test this. The data in the table is supposed to be editable by the user, the database connection only happens after the dialog is closed.
I mean everytime i open this export sql i see this kind of table
image
is this supposed to happen?  does this table should be filled with data already?

tomas-pluskal:
> > is this supposed to happen?  does this table should be filled with data already?

Yes, it is supposed to happen. The purpose of the table is to configure the export to SQL database. The table should be filled by the user by manual editing.

itissandeep98:
>The table should be filled by the user by manual editing.

Ok got it, will implement it. Just one more question that in column "Export data type" every cell needs to be a choice box or it will be a text field like the rest of the column?

tomas-pluskal:
> > The table should be filled by the user by manual editing.
> 
> Ok got it, will implement it. Just one more question that in column "Export data type" every cell needs to be a choice box or it will be a text field like the rest of the column?

The Export data type should be a combo box. Try running MZmine 2.53 to see how exactly it worked in the original Swing version (https://github.com/mzmine/mzmine2/releases/tag/v2.53).

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettings.java
```diff
@@ -23,7 +23,7 @@
 
 public class SQLColumnSettings {
 
-  private final ObservableList<SQLRowObject> tableData = FXCollections.observableArrayList();;
+  private ObservableList<SQLRowObject> tableData = FXCollections.observableArrayList();
```
tomas-pluskal: Not sure why you removed the final keywords?

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettingsComponent.java
```diff
@@ -68,23 +65,15 @@ public SQLColumnSettingsComponent() {
...
-                editFocusedCell();
-            }
-        });
-
 
-        //Todo: change the below swing code to JavaFX
-//        DefaultCellEditor dataTypeEditor = new DefaultCellEditor(dataTypeCombo);
-//        columnsTable.setDefaultEditor(SQLExportDataType.class, dataTypeEditor);
+        //Editors on each cell
+        columnName.setCellFactory(param -> new SQLStringCellEditor(value));
```
tomas-pluskal: Is it necessary to create a new SQLStringCellEditor here? I think you can just use the default TextFieldTableCell, see https://stackoverflow.com/questions/19335196/how-to-make-javafx-tableview-cells-editable

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettingsComponent.java
```diff
@@ -68,23 +65,15 @@ public SQLColumnSettingsComponent() {
...
+        columnType.setCellFactory(param ->new SQLExportTypeCellEditor(value));
+        columnValue.setCellFactory(param -> new SQLStringCellEditor(value));
+        columnsTable.setOnMouseMoved(event -> setValue(getValue())); // making table update on every mouse movement over table
```
tomas-pluskal: This seems like brutal hacking. What if the user uses keyboard shortcuts instead of using the mouse?

src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettingsComponent.java
tomas-pluskal: Also see https://stackoverflow.com/questions/50598258/how-to-make-combobox-editable-in-a-tablecell-in-javafx for the second column

itissandeep98: The above links you shared have a very shorter version of doing the same thing I did. I learned a new way to make the tables editable, Thanks for sharing them

tomas-pluskal:
I think it looks good now, thanks!

PR #51, opened 2020-03-19
Swing to Javafx InterpolatingLookupPaintScaleSetupDialog

Deepak123bharat:


src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleRow.java
```diff
@@ -0,0 +1,25 @@
+package io.github.mzmine.util.interpolatinglookuppaintscale;
```
tomas-pluskal: Please add the license header to all new .java files

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleRow.java
```diff
@@ -0,0 +1,25 @@
...
+
+
+import java.awt.*;
+
+public class InterpolatingLookupPaintScaleRow {
+
+
+    private Double key;
```
tomas-pluskal: If the fields don't change, declare them as final and add the @Immutable annotation to the class declaration.

src/main/java/io/github/mzmine/modules/dataanalysis/bubbleplots/RTMZAnalyzerWindow.java
```diff
@@ -83,12 +87,19 @@ public RTMZAnalyzerWindow(AbstractXYZDataset dataset, PeakList peakList,
 
     colorButton.setOnAction(e -> {
       SwingUtilities.invokeLater(() -> {
-        InterpolatingLookupPaintScaleSetupDialog colorDialog =
-            new InterpolatingLookupPaintScaleSetupDialog(null, plot.getPaintScale());
-        colorDialog.setVisible(true);
 
-        if (colorDialog.getExitCode() == ExitCode.OK)
-          plot.setPaintScale(colorDialog.getPaintScale());
+        Platform.runLater(new Runnable() {
```
tomas-pluskal: You can remove the SwingUtilities.invokeLater() and also the Platform.runLater() blocks, because the colorButton.setOnAction() handler will already be triggered on the JavaFX thread

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -0,0 +1,257 @@
...
+import javafx.scene.Scene;
+import javafx.scene.control.*;
+import javafx.scene.control.Button;
+import javafx.scene.control.Label;
+import javafx.scene.control.ScrollPane;
+import javafx.scene.control.TextField;
+import javafx.scene.layout.*;
+import javafx.stage.Stage;
+
+public class InterpolatingLookupPaintScaleSetupDialogController extends Stage implements Initializable {
```
tomas-pluskal: Usually the Stage class and the FXML controller class are two different classes. You put them together into the same class, which creates some confusion (for example you need two different constructors). See how it is done here: https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/modules/dataprocessing/id_formulaprediction/ResultWindowFX.java

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -0,0 +1,257 @@
...
+
+
+    private static final long serialVersionUID = 1L;
```
tomas-pluskal: Remove the serialVersionUID as that is only required for Swing classes

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialog.java
```diff
@@ -42,177 +42,177 @@
 import io.github.mzmine.util.ExitCode;
 
 public class InterpolatingLookupPaintScaleSetupDialog extends JDialog
-    implements ActionListener, ListSelectionListener {
```
tomas-pluskal: The old Swing dialog can be removed

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogTableModel.java
```diff
@@ -26,7 +26,7 @@
 public class InterpolatingLookupPaintScaleSetupDialogTableModel extends AbstractTableModel {
```
tomas-pluskal: The TableModel class can also be removed

Deepak123bharat: Ok, I will remove it.

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -180,14 +117,26 @@ public void updateItem(Color item, boolean empty) {
             }
         });
 
-        tableLookupValues.setItems(ObservableTableList);
+    }
+
+    private static TreeMap<Double, Color> lookupTable = new TreeMap<Double, Color>();
```
tomas-pluskal: This field should not be static

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -180,14 +117,26 @@ public void updateItem(Color item, boolean empty) {
...
+    private Logger logger = Logger.getLogger(this.getClass().getName());
+    private static ExitCode exitCode = ExitCode.CANCEL;
```
tomas-pluskal: This field should not be static

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -180,14 +117,26 @@ public void updateItem(Color item, boolean empty) {
...
+    private javafx.scene.paint.Color bColor = javafx.scene.paint.Color.WHITE;
+
+
+    private static ObservableList<InterpolatingLookupPaintScaleRow> ObservableTableList = FXCollections.observableArrayList();
```
tomas-pluskal: This field should not be static. Also, by convention names of fields start with lower case letters in Java. You can also make this field final, because for each window there will be only one observable list.

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -0,0 +1,207 @@
...
+    private Button buttonCancel;
+
+    @FXML
+    private ScrollPane scrollpaneLookupValues;
+
+    @FXML
+    private TableView<InterpolatingLookupPaintScaleRow> tableLookupValues;
+
+    @FXML
+    private TableColumn<InterpolatingLookupPaintScaleRow, Double> Value;
```
tomas-pluskal: Should be named value (lower case letter)

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -0,0 +1,207 @@
...
+
+    @FXML
+    private TableColumn<InterpolatingLookupPaintScaleRow, Color> Color;
```
tomas-pluskal: Should be named color (lower case letter). Since it refers to table column, better name it colorColumn.

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
tomas-pluskal: Since it refers to table column, better name it valueColumn.

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
Deepak123bharat: done

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogFX.fxml
```diff
@@ -0,0 +1,52 @@
...
+          <children>
+              <AnchorPane fx:id="panelOKCancelButton" layoutX="283.0" layoutY="321.0" prefHeight="58.0" prefWidth="217.0">
+              <children>
+                <Button fx:id="buttonOK" layoutX="31.0" layoutY="16.0" mnemonicParsing="false" onAction="#actionPerformed" prefHeight="26.0" prefWidth="51.0" text="Ok" />
+                  <Button fx:id="buttonCancel" layoutX="109.0" layoutY="16.0" mnemonicParsing="false" onAction="#actionPerformed" prefHeight="26.0" prefWidth="66.0" text="Cancel" />
+              </children>
+            </AnchorPane>
+            <AnchorPane fx:id="panelList" layoutX="55.0" layoutY="95.0" prefHeight="139.0" prefWidth="425.0">
+               <children>
+                  <ScrollPane fx:id="scrollpaneLookupValues" prefHeight="214.0" prefWidth="459.0">
```
tomas-pluskal: Actually, you can remove this ScrollPane, because in JavaFX the TableView already has a scroll pane included

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogFX.java
```diff
@@ -0,0 +1,88 @@
...
+import java.awt.*;
+import java.io.IOException;
+import java.util.TreeMap;
+
+
+public class InterpolatingLookupPaintScaleSetupDialogFX extends Stage {
+
+    private InterpolatingLookupPaintScaleSetupDialogController controller;
+
+    private static ObservableList<InterpolatingLookupPaintScaleRow> ObservableTableList = FXCollections.observableArrayList();
```
tomas-pluskal: This field should not be static, but it should be final. The observable list instance should not change once it is created. Also, again - the name should start with lower case letter.

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogFX.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final TreeMap<Double, Color> lookupTable = new TreeMap<Double, Color>();
```
tomas-pluskal: This field should not be static.

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogFX.java
```diff
@@ -0,0 +1,88 @@
...
+
+
+
+    public InterpolatingLookupPaintScaleSetupDialogFX(Object parent,InterpolatingLookupPaintScale paintScale){
```
tomas-pluskal: The parent parameter is not used, so you can remove it

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -0,0 +1,207 @@
...
+
+    private  TreeMap<Double, Color> lookupTable = new TreeMap<Double, Color>();
+    private Logger logger = Logger.getLogger(this.getClass().getName());
+    private  ExitCode exitCode = ExitCode.CANCEL;
+    private javafx.scene.paint.Color bColor = javafx.scene.paint.Color.WHITE;
+
+
+    private ObservableList<InterpolatingLookupPaintScaleRow> ObservableTableList = FXCollections.observableArrayList();
+
+    public void addObservableList(TreeMap<Double, Color> lookupTable, ObservableList<InterpolatingLookupPaintScaleRow> ObservableTableList){
```
tomas-pluskal: It makes more sense to use a get() method here, not add() method. You can use the get() method to obtain the list instance from the controller and then you can modify the list items freely.

Deepak123bharat: ok done

tomas-pluskal: Check again my comments carefully. The observable list should be final, and the get() method should return the reference to this list, not change it.

Deepak123bharat: I want to add constructor items to the controller class so I used the addObservableList method. now its changed.

tomas-pluskal:
Great! Now the code looks pretty good. What remains is to fix some issues with the functionality:
Screen Shot 2020-03-20 at 22 14 01

- The window should have minimum width and height set.
- The buttons should be placed in a ButtonBar at the bottom of the window, similarly to other dialogs in MZmine.
- Clicking the Add/Modify button without setting the Value throws an exception.
- Clicking on a value in the table should automatically update the Value textbox and the Color. Alternatively, you could make the Color field in the table editable using a ColorTableCell (https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/util/components/ColorTableCell.java)
- The table currently shows up with a horizontal scroll bar, which is not needed.

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -187,8 +188,18 @@ public void actionPerformed(ActionEvent event) {
...
+        try {
+            Double.parseDouble(str);
+            return true;
+        } catch (NumberFormatException e) {
+            return false;
+        }
+    }
 
-
+    private javafx.scene.paint.Color awtColorToJavaFX(Color c) {
```
tomas-pluskal: Actually, we have a utility class FxColorUtil in package io.github.mzmine.util.javafx that does this, so you can just use that class.

src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogController.java
```diff
@@ -187,8 +188,18 @@ public void actionPerformed(ActionEvent event) {
         }
     }
 
+    public boolean isDouble(String str) {
```
tomas-pluskal: It is not a nice code behavior to check something by throwing an exception. 
You can use this Guava method instead:
https://guava.dev/releases/18.0/api/docs/com/google/common/primitives/Doubles.html#tryParse(java.lang.String)

tomas-pluskal: More info here: https://stackoverflow.com/questions/3543729/how-to-check-that-a-string-is-parseable-to-a-double

tomas-pluskal:
Good, now the functionality is close to the original Swing version. It still has some bugs - when I click Add/Modify to add a new value, it is not always added with the selected color in Select Color, but sometimes it is added with the last color that was clicked on in the table. 
In any case, it is better to move the color picker into the table, like I wrote above.
Please try changing the color column in the table to ColorTableCell as shown here: 
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/modules/visualization/fx3d/Fx3DStageController.java#L184

ColorTableCell:
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/util/components/ColorTableCell.java

tomas-pluskal:
Great! Now why not make the Value column user-editable too, and then we don't need the Modify function at all and instead we just keep the Add function. 

You can use this approach to make the Value editable:
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/modules/io/sqlexport/SQLColumnSettingsComponent.java#L86

Also, it is better to constrain the table columns to always fit the whole table, using this setting:
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/gui/mainwindow/MainWindow.fxml#L121

tomas-pluskal:
Awesome, thanks!

PR #52, opened 2020-03-22
fixed duplication issue in element table

knakul853:


src/main/java/io/github/mzmine/parameters/parametertypes/elements/ElementsTableComponent.java
```diff
@@ -32,10 +33,15 @@
...
 import org.openscience.cdk.interfaces.IIsotope;
 
+import java.util.HashSet;
+import java.util.Set;
+
 public class ElementsTableComponent extends FlowPane {
 
   private final ObservableList<ElementsValue> elementsValues = FXCollections.observableArrayList();
   private final TableView<ElementsValue> elementsValueTable = new TableView();
+  private final Set<String>checkDuplicate = new HashSet<>();
```
tomas-pluskal: It is a dirty solution to create another Set. Now your data is duplicated in the elementsValues ObservableList and in the checkDuplicate Set. That's just going to cause problems in the future.

src/main/java/io/github/mzmine/parameters/parametertypes/elements/ElementsTableComponent.java
```diff
@@ -107,14 +115,16 @@ public void setElements(MolecularFormulaRange elements) {
...
       return;
-
     for (IIsotope isotope : elements.isotopes()) {
       int minCount = elements.getIsotopeCountMin(isotope);
       int maxCount = elements.getIsotopeCountMax(isotope);
-      String s = isotope.getSymbol();
-      elementsValues.add(new ElementsValue(isotope, String.valueOf(maxCount), String.valueOf(minCount)));
-
+      ElementsValue elementsValue = new ElementsValue(isotope, String.valueOf(maxCount), String.valueOf(minCount));
+     if(!checkDuplicate.contains(elementsValue.getIsotope().getSymbol())){
```
tomas-pluskal: Instead of using another Set, just check if the elementsValues List already contains an entry with the same symbol or not.

tomas-pluskal:
You added a workaround, but why were the duplicate entries being added to the table in the first place? I mean, where did the duplicate entries come from?

knakul853:
```  public ElementsTableComponent createEditingComponent() {
    ElementsTableComponent editor = new ElementsTableComponent();
    editor.setElements(value);
    return editor;
  }```

```  @Override
  public void setValueToComponent(ElementsTableComponent component,
      MolecularFormulaRange newValue) {
    component.setElements(newValue);
  }```

do we really need to set element in both places? i think we can replace     ```editor.setElements(value);```

**Comment by tomas-pluskal** [comment]
> ```
> public ElementsTableComponent createEditingComponent() {
>     ElementsTableComponent editor = new ElementsTableComponent();
>     editor.setElements(value);
>     return editor;
>   }
> ```
> 
> ```
> @Override
>   public void setValueToComponent(ElementsTableComponent component,
>       MolecularFormulaRange newValue) {
>     component.setElements(newValue);
>   }
> ```
> 
> do we really need to set elements in both places? I think we can replace
> 
> ```
> editor.setElements(value);
> ```

I agree, you can remove editor.setElements(value)

File: src/main/java/io/github/mzmine/parameters/parametertypes/elements/ElementsTableComponent.java [hunk 3]
```diff
@@ -107,14 +114,14 @@ public void setElements(MolecularFormulaRange elements) {
 
     if (elements == null)
       return;
-
     for (IIsotope isotope : elements.isotopes()) {
```
tomas-pluskal: I think you need to call elementsValues.clear() before this iteration cycle.

tomas-pluskal:
Looks good, thanks!

PR #53, opened 2020-03-23
Transormation of id_Sirius module from swing to javafx

knakul853:
In this PR I will be converting the implementation of ```id_sirius``` module from Swing to JavaFX.


knakul853:
Testing is not completed yet.

build.gradle
```diff
@@ -199,7 +199,7 @@ application {
                                  "-XX:InitialRAMPercentage=30", 
                                  "-XX:MinRAMPercentage=80",
                                  "-XX:MaxRAMPercentage=80",
-                                 "-Djava.library.path=lib",
+                                 "-Djava.library.path=tmp",
```
tomas-pluskal: This change should not be included in the PR

knakul853: ok.

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindowController.java
```diff
@@ -0,0 +1,226 @@
+package io.github.mzmine.modules.dataprocessing.id_sirius;
```
tomas-pluskal: Please add the license header

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindowFX.java
```diff
@@ -0,0 +1,40 @@
+package io.github.mzmine.modules.dataprocessing.id_sirius;
```
tomas-pluskal: Please add the license header

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindowFX.java
```diff
@@ -0,0 +1,40 @@
...
+import javafx.scene.Parent;
+import javafx.scene.Scene;
+import javafx.stage.Stage;
+import java.io.IOException;
+import java.util.List;
+
+public class ResultWindowFX extends Stage {
+    private ResultWindowController controller;
+
+    public ResultWindowFX(PeakListRow peakListRow, Task searchTask)
```
tomas-pluskal: You should set minWidth and minHeight in the constructor

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindowFX.java
```diff
@@ -0,0 +1,40 @@
...
+import io.github.mzmine.taskcontrol.Task;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Parent;
+import javafx.scene.Scene;
+import javafx.stage.Stage;
+import java.io.IOException;
+import java.util.List;
+
+public class ResultWindowFX extends Stage {
+    private ResultWindowController controller;
```
tomas-pluskal: Make the controller final. Always make fields final unless they are supposed to change.

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/SingleRowIdentificationTask.java
```diff
@@ -145,15 +141,28 @@ public String getTaskDescription() {
...
+              "SIRIUS/CSI-FingerID identification of " + massFormater.format(parentMass) + " m/z");
+      resultWindowFX.show();
+      return resultWindowFX;
+
+    });
+    Platform.runLater(query);
 
+    try {
+      query.get();
+    } catch (InterruptedException e) {
```
tomas-pluskal: If exception happens in this code, you should terminate the task. You can join the catch blocks into one.

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindowFX.java
knakul853: Hey if I add final it says I have to make it a constructor parameter, that I can't do.
let me know if I'm missing something

tomas-pluskal: I guess that's because you are doing the assignment to the controller variable inside the try {} block in the constructor, so the compiler complains that the assignment might not actually happen if there is an exception. You can  change the try/catch block to only contain the `root.load();` code.

tomas-pluskal:
It's looking good. What remains is to show the chemical structures in the Chemical Structure column.

Screen Shot 2020-03-28 at 07 56 05

Perhaps you can use this component: 
````
io.github.mzmine.modules.visualization.molstructure.Structure2DComponent
````

**Comment by knakul853** [comment]
Thanks.i was stuck whole days to show the structure.but learn a lot new things🙂

**Comment by tomas-pluskal** [comment]
Looks good! A few additional comments:
- The Elements table in the parameter setup is too long, make the vertical size smaller. Also you can move the Remove button under the Add button.
- Change the default TableView message "No content in table" to something like "Searching with SIRIUS..."
- Try making the chemical structure background transparent instead of white, so it doesn't show white rectangle when the table row is selected.
- You can fix the table columns to always occupy 100% of the table, even when the window is resized. See CONSTRAINED_RESIZE_POLICY here: https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/util/interpolatinglookuppaintscale/InterpolatingLookupPaintScaleSetupDialogFX.fxml#L26 (this change should be applied also to the chemical formula search window)
- The Display DB links button doesn't work
- The chemical structure should be enlarged on mouse hover (similar to the old MZmine 2 implementation)

Screen Shot 2020-04-01 at 10 03 25

File: gradle/wrapper/gradle-wrapper.properties [hunk 7]
```diff
@@ -1,5 +1,11 @@
+#Fri Mar 27 15:17:03 IST 2020
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+<<<<<<< HEAD
```
tomas-pluskal, line 5: Seems like the conflict was not merged properly

knakul853:
``` Display DB links button doesn't work``` this is expected as I have provided null in place of jbutton, bcoz the implementation of DBframe is in the swing. After this finished I would change the DBframe table.

**Comment by tomas-pluskal** [comment]
I think it is better to show a floating window with the large structure (like in MZmine 2), rather than changing the size of the existing structure in the table cell.

File: src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindowController.java [hunk 8]
```diff
@@ -0,0 +1,298 @@
...
+                protected void updateItem(Node item, boolean empty)
+                {
+                    super.updateItem(item, empty);
+                    if (item == null || empty) {
+                        return;
+                    }
+                    setItem(item);
+                    Tooltip tooltip = new Tooltip();
+                    item.resize(300, 300);
+                    tooltip.setGraphic(item);
```
tomas-pluskal, line 165: I think you cannot use the same 2DStructureComponent for the cell and for the tooltip, one Node can only have one parent. That's why the structure doesn't show in the table.

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindowController.java
```diff
@@ -0,0 +1,298 @@
...
+@FXML
+private void addIdentityOnClick(ActionEvent ae){
+    SiriusCompound compound = compoundsTable.getSelectionModel().getSelectedItem();
+    if(compound == null)
+    {
+        MZmineCore.getDesktop().displayMessage(null,
+                "Select one result to add as compound identity");
+        return;
+    }
+    peakListRow.addPeakIdentity(compound, false);
```
tomas-pluskal, line 198: After the identity is added, you can hide the window

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindowController.java
```diff
@@ -0,0 +1,298 @@
...
+}
+
+    /**
+     * Method sets value of clipboard to `content`
+     *
+     * @param content - Formula or SMILES string
+     * @param errorMessage - to print in a message if value of `content` is null
+     */
+    private void copyToClipboard(String content, String errorMessage) {
+        if (content == null) {
```
tomas-pluskal, line 246: When dealing with strings, it is better to use `if (Strings.isNullOrEmpty(content)) ...`
Because empty string is also bad.

tomas-pluskal:
What's the `doge.png` for? :)

knakul853:
I was trying to display some image on tooltip.by mistake I pushed.will delete it.

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindowController.java
```diff
@@ -0,0 +1,297 @@
...
+        clipboard.setContents(stringSelection, null);
+    }
+
+    /**
+     * Update content of a table using swing-thread
+     *
+     * @param compound
+     */
+    public void addNewListItem(@Nonnull final SiriusCompound compound) {
+        Platform.runLater(new Runnable() {
```
tomas-pluskal, line 261: Use the lambda notation 
`Platform.runLater(() -> compounds.add(compound));`

tomas-pluskal:
Screen Shot 2020-04-06 at 14 00 15

The structure under the mouse pointer should  be enlarged. Right now, the selected structure is always enlarged, even when the mouse pointer is hovering over different structure.

knakul853:
seems working let me know if the size is good.

tomas-pluskal:
Looks great! You can now take a look at the "Display DB Links" button. By the way, the "DBs" column is supposed to show multiple databases (for most compounds). In your implementation, it shows only one. Take a look at that issue, too.

tomas-pluskal:
By the way, have you tried using the PopOver control instead of Tooltip (https://controlsfx.bitbucket.io/org/controlsfx/control/PopOver.html)? The PopOver control has the advantage that the user can detach it into a separate window.

knakul853:
@Tomas Pluskal Problem remained same either using popover or tooltip. that is why I duplicate the node.Now we can use whatever we want.

knakul853:
How could I add the dbframe task in the main task. Creating a new task leading to the thread exception.?

tomas-pluskal:
I think DBFrame should not be a VBox, you need to make it a standalone window, maybe use a Popup? https://docs.oracle.com/javase/8/javafx/api/javafx/stage/Popup.html

knakul853:
Open browser action still stuck and freezes the window.

tomas-pluskal:
Use `MZmineCore().getDesktop().openWebpage(url)` instead

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/table/db/DBFrame.java
```diff
@@ -17,128 +17,133 @@
...
-    // Initiate loading of DB links into new table
-    SwingUtilities.invokeLater(new Runnable() {
-      public void run() {
-        model.addElement(compound);
+  private static final Logger logger = LoggerFactory.getLogger(DBFrame.class);
+  private final TableView<SiriusDBCompound> dbTable = new TableView();
+  private final ObservableList<SiriusDBCompound> compounds = FXCollections.observableArrayList();
+    VBox vBox = new VBox();
+
+  public DBFrame(SiriusCompound compound, Button openBrowser) {
```
tomas-pluskal, line 62: Not sure why the Button is a constructor parameter, it doesn't make much sense

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/table/db/DBFrame.java
```diff
@@ -17,128 +17,134 @@
...
+          return;
+        }
+       //  Generate url
+      try
+      {
+        URL url = selectedCompound.generateURL();
         if (url == null)
           throw new RuntimeException("Unsupported DB");
-        if (Desktop.isDesktopSupported()) {
+        if (Desktop.isDesktopSupported())
```
tomas-pluskal, line 111: remove this check, the MZmine.getDesktop() has nothing to do with the AWT desktop

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/table/db/DBFrame.java
```diff
@@ -17,128 +17,134 @@
...
+    DBLink[] links = annotation.getDBLinks();
+    if (links == null)
+      return;
+
+    for (DBLink link : links) {
+      compounds.add(new SiriusDBCompound(link.name, link.id));
     }
   }
+
+  public void display(){
```
tomas-pluskal, line 141: This function should be removed. DBFrame already extends Popup, which is a window, so there is no point in creating another Stage for it. If you want to use a Stage, then DBFrame should extend Stage and not Popup. The instance of the Scene should be created in the constructor.

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/table/SiriusCompound.java
```diff
@@ -175,22 +177,34 @@ public String getSMILES() {
    *
    * @return unique names of the databases
    */
-  public String[] getDBS() {
+  public SimpleObjectProperty<Node> getDBS() {
```
tomas-pluskal, line 180: SiriusCompound is a class that describes the compound, it should not be creating JavaFX Nodes. I know it works, but it is dirty. This class should just return the data (String[]) and the corresponding Nodes should be created in the classes that deal with GUI components.

knakul853: Hey, I was looking the way to put an array of string in a single cell and I looked on the web's various ways but didn't get the result as expected. So I came up with this on my own. I have fixed this and also accommodated my requirements as the table required.Please let me know anything that i should  try;

tomas-pluskal:
Screen Shot 2020-04-09 at 15 18 27

The DBs column now takes way too much space. Try to find a way how to show the same information in a more compact way.  You can use a multiline Label, see here: https://stackoverflow.com/questions/15977295/control-for-displaying-multiline-text

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/table/db/DBFrame.java
```diff
@@ -17,128 +17,133 @@
...
+    if (links == null)
+      return;
+
+    for (DBLink link : links) {
+      compounds.add(new SiriusDBCompound(link.name, link.id));
     }
   }
+
+    @Override
+    protected void show() {
```
tomas-pluskal, line 145: Stop and think about this method for a moment :)

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindow.fxml
```diff
@@ -13,9 +13,6 @@
           <placeholder>
             <Label text="Searching with SIRIUS..." />
           </placeholder>
-          <columnResizePolicy>
-              <TableView fx:constant="CONSTRAINED_RESIZE_POLICY" />
-          </columnResizePolicy>
```
knakul853: hey, i removed this because it won't let you resize the column. Since here we might need that.Let me know what you think

tomas-pluskal: CONSTRAINED_RESIZE_POLICY should not prevent you from resizing columns. It only guarantees that the sum of the widths of all columns is equal to the width of the table.
https://openjfx.io/javadoc/14/javafx.controls/javafx/scene/control/TableView.html#CONSTRAINED_RESIZE_POLICY

src/main/java/io/github/mzmine/modules/dataprocessing/id_sirius/ResultWindow.java
```diff
@@ -167,8 +167,8 @@ public void actionPerformed(ActionEvent e) {
       int realRow = compoundsTable.convertRowIndexToModel(row);
       final SiriusCompound compound = listElementModel.getCompoundAt(realRow);
 
-      DBFrame dbFrame = new DBFrame(compound, browse);
```
tomas-pluskal, line 170: You can just remove the ResultWindow class, since it is not needed anymore.

tomas-pluskal:
Ok, looks great, thanks!

PR #58, opened 2020-04-02
Echartviewer update

itissandeep98:
In this PR I have added following features to the context menu of EchartViewer
- added an option to reset zoom
- added an option to set ranges of axes 
- added an option to export to EPS and EMF 
- added an option to copy the chart image to the clipboard
- added an option to print using the printer

### DIFF
--- a/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
+++ b/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
@@ -34,13 +34,12 @@ import io.github.mzmine.util.SaveImage;
 import io.github.mzmine.util.SaveImage.FileType;
 import io.github.mzmine.util.dialogs.AxesSetupDialog;
 import io.github.mzmine.util.io.XSSFExcelWriterReader;
-import java.awt.Graphics;
+import java.awt.Graphics2D;
 import java.awt.Image;
 import java.awt.Toolkit;
 import java.awt.datatransfer.DataFlavor;
 import java.awt.datatransfer.Transferable;
 import java.awt.datatransfer.UnsupportedFlavorException;
-import java.awt.print.PageFormat;
 import java.awt.print.Printable;
 import java.awt.print.PrinterException;
 import java.awt.print.PrinterJob;
@@ -52,6 +51,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javafx.event.ActionEvent;
 import javafx.event.EventHandler;
+import javafx.scene.control.Alert;
+import javafx.scene.control.Alert.AlertType;
 import javafx.scene.control.ContextMenu;
 import javafx.scene.control.Menu;
 import javafx.scene.control.MenuItem;
@@ -199,7 +200,35 @@ public class EChartViewer extends ChartViewer {
       });
 
       addMenuItem(getContextMenu(), "Print", event -> {
-
+        Printable content = (graphics, pageFormat, pageIndex) -> {
+          if (pageIndex > 0) {
+            return Printable.NO_SUCH_PAGE;
+          }
+          Graphics2D g2d = (Graphics2D) graphics;
+          g2d.translate(pageFormat.getImageableX(), pageFormat.getImageableY());
+          graphics.drawImage(chart.createBufferedImage(500, 500), 100, 100, null);
+          return Printable.PAGE_EXISTS;
+        };
+        PrinterJob job = PrinterJob.getPrinterJob();
+        if (job.getPrintService() != null) {
+          job.setPrintable(content);
+          boolean doPrint = job.printDialog();
+          if (doPrint) {
+            try {
+              job.print();
+            } catch (PrinterException e) {
+              Alert alert = new Alert(AlertType.ERROR);
+              alert.setTitle("Printing Service");
+              alert.setHeaderText("Error in Printing the Chart");
+              alert.showAndWait();
+            }
+          }
+        } else {
+          Alert alert = new Alert(AlertType.ERROR);
+          alert.setTitle("Printing Service");
+          alert.setHeaderText("No Printing Service Found");
+          alert.showAndWait();
+        }
       });
     }
 
--- a/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
+++ b/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
@@ -18,10 +18,46 @@
 
 package io.github.mzmine.gui.chartbasics.gui.javafx;
 
+import io.github.mzmine.gui.chartbasics.gestures.ChartGestureHandler;
+import io.github.mzmine.gui.chartbasics.gestures.interf.GestureHandlerFactory;
+import io.github.mzmine.gui.chartbasics.graphicsexport.GraphicsExportModule;
+import io.github.mzmine.gui.chartbasics.graphicsexport.GraphicsExportParameters;
+import io.github.mzmine.gui.chartbasics.gui.javafx.menu.MenuExportToClipboard;
+import io.github.mzmine.gui.chartbasics.gui.javafx.menu.MenuExportToExcel;
+import io.github.mzmine.gui.chartbasics.gui.swing.ChartGestureMouseAdapter;
+import io.github.mzmine.gui.chartbasics.gui.wrapper.ChartViewWrapper;
+import io.github.mzmine.gui.chartbasics.listener.AxesRangeChangedListener;
+import io.github.mzmine.gui.chartbasics.listener.AxisRangeChangedListener;
+import io.github.mzmine.gui.chartbasics.listener.ZoomHistory;
+import io.github.mzmine.main.MZmineCore;
+import io.github.mzmine.util.SaveImage;
+import io.github.mzmine.util.SaveImage.FileType;
+import io.github.mzmine.util.dialogs.AxesSetupDialog;
+import io.github.mzmine.util.io.XSSFExcelWriterReader;
+import java.awt.Graphics;
+import java.awt.Image;
+import java.awt.Toolkit;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
```diff
@@ -129,17 +146,117 @@ public EChartViewer(JFreeChart chart, boolean graphicsExportMenu, boolean dataEx
...
+          @Override
+          public Object getTransferData(DataFlavor flavor)
+              throws UnsupportedFlavorException, IOException {
+            if (!DataFlavor.imageFlavor.equals(flavor)) {
+              throw new UnsupportedFlavorException(flavor);
+            }
+            return image;
+          }
+        };
+        Toolkit.getDefaultToolkit().getSystemClipboard().setContents(clipboardWriter, null);
```
tomas-pluskal, line 199: Here you used the AWT class to access the clipboard. Since we are moving everything to JavaFX, I think it would be better to use the corresponding JavaFX class:
https://openjfx.io/javadoc/13/javafx.graphics/javafx/scene/input/Clipboard.html

src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
```diff
@@ -129,17 +146,117 @@ public EChartViewer(JFreeChart chart, boolean graphicsExportMenu, boolean dataEx
...
+      });
+
+      addMenuItem(getContextMenu(), "Print", event -> {
+        Printable content = (graphics, pageFormat, pageIndex) -> {
```
tomas-pluskal, line 203: Same as above - let's use JavaFX classes for printing instead of AWT classes:
https://examples.javacodegeeks.com/desktop-java/javafx/javafx-print-api/

tomas-pluskal:
@robinschmid Please check if you have any suggestions

src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
itissandeep98: Yes i was trying that too, but i was having trouble converting the buffered image to javafx image.

tomas-pluskal: Try
`Image image = SwingFXUtils.toFXImage(capture, null);`

itissandeep98: but aren't we trying to remove the swing dependencies?

tomas-pluskal: SwingFXUtils is actually a JavaFX class :)

tomas-pluskal: I believe JFreeChart currently can only create a BufferedImage, not JavaFX image, so we need to convert it.

src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
```diff
@@ -129,17 +142,84 @@ public EChartViewer(JFreeChart chart, boolean graphicsExportMenu, boolean dataEx
...
+        content.putImage(image);
+        Clipboard.getSystemClipboard().setContent(content);
+      });
+
+      addMenuItem(getContextMenu(), "Print", event -> {
+        ImageView image = new ImageView(
+            SwingFXUtils.toFXImage(getChart().createBufferedImage(500, 500), null));
+        PrinterJob job = PrinterJob.createPrinterJob();
+        if (job != null) {
+          boolean doPrint = job.showPrintDialog(new Stage());
```
tomas-pluskal, line 184: Here you should not use `new Stage()` for a parent. Instead, obtain a reference to the current window, by calling `this.getScene().getWindow()`

src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
```diff
@@ -129,17 +142,84 @@ public EChartViewer(JFreeChart chart, boolean graphicsExportMenu, boolean dataEx
...
+      });
+
+      addMenuItem((Menu) getContextMenu().getItems().get(0), "EPS..",
+          event -> handleSave("EMF Image", "EMF", ".emf", FileType.EMF));
+
+      addMenuItem((Menu) getContextMenu().getItems().get(0), "EMF..",
+          event -> handleSave("EPS Image", "EPS", ".eps", FileType.EPS));
+
+      addMenuItem(getContextMenu(), "Copy Chart to Clipboard", event -> {
+        Image image = SwingFXUtils.toFXImage(getChart().createBufferedImage(500, 500), null);
```
tomas-pluskal, line 173: Avoid making multiple function calls on the same line - here you are calling toFXImage, getChart, and createBufferedImage. If an exception happens, it is much easier to debug if the calls are on separate lines. It is also easier to read such code for other people who work on the same codebase.

tomas-pluskal: Second comment: 500,500 should not be hardcoded here. Add a static final constant field to the class, and include a comment that explains the reason why 500 was chosen as the constant value.

src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
```diff
@@ -129,17 +142,84 @@ public EChartViewer(JFreeChart chart, boolean graphicsExportMenu, boolean dataEx
...
+        addExportMenu(graphicsExportMenu, dataExportMenu);
+      }
+
+      addMenuItem(getContextMenu(), "Reset Zoom", event -> {
+        xAxis.setAutoRange(true);
+        yAxis.setAutoRange(true);
+      });
+
+      addMenuItem(getContextMenu(), "Set Range on Axis", event -> {
+        AxesSetupDialog dialog = new AxesSetupDialog(new Stage(), chart.getXYPlot());
```
tomas-pluskal, line 162: Here also you should not use `new Stage()` for parent, see my other comment below

src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
```diff
@@ -129,17 +142,84 @@ public EChartViewer(JFreeChart chart, boolean graphicsExportMenu, boolean dataEx
...
+        dialog.show();
+      });
+
+      addMenuItem((Menu) getContextMenu().getItems().get(0), "EPS..",
```
tomas-pluskal, line 166: The expression `(Menu) getContextMenu().getItems().get(0)` works but is unnecessary complicated. Better to make a private class variable for the Export menu. You can change the addMenu() function to return Menu instead of void. Or you can remove the addMenu() function completely, I don't really think it is needed.

src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
itissandeep98: >500,500 should not be hardcoded here. Add a static final constant field to the class, and include a comment that explains the reason why 500 was chosen as the constant value.

Actually there is no specific reason for choosing 500. Should I make the size of the image equal to the size of the window or leave it to some constant value?

tomas-pluskal: Okay, you can make it equal to the window size. The user can use the graphics export dialog if different size is needed.

src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/EChartViewer.java
```diff
@@ -129,17 +144,92 @@ public EChartViewer(JFreeChart chart, boolean graphicsExportMenu, boolean dataEx
...
+
+      addMenuItem(exportMenu, "EPS..",
+          event -> handleSave("EMF Image", "EMF", ".emf", FileType.EMF));
+
+      addMenuItem(exportMenu, "EMF..",
+          event -> handleSave("EPS Image", "EPS", ".eps", FileType.EPS));
+
+      addMenuItem(getContextMenu(), "Copy Chart to Clipboard", event -> {
+        BufferedImage bufferedImage = getChart()
+            .createBufferedImage((int) this.getScene().getWidth(),
```
tomas-pluskal, line 178: Actually, it makes more sense to use `this.getWidth()` instead of `this.getScene().getWidth()`, because the scene contains also other components than the chart. If you use `this.getWidth()` the exported image will have the same size as the image on the screen, which is a reasonable behavior.

tomas-pluskal:
Looks great, thanks!

robinschmid:
https://github.com/mzmine/mzmine3/pull/58/files#diff-a141e5ff4687f291c438f6a5fe602affR150-R206
Sorry for the late comment. And thanks for the great work.
You have to be careful with your additions to the constructor. 
1. if the chart is null in the constructor, the menus are never added to the EChartView. I think they should always be added. If the chart is null we could do setEnabled(false)
2. If the chart is changed later by calling the setChart method, your menu actions in the constructor will still use the old chart (domainAxis/rangeAxis). You should access the chart inside the action listeners by calling the getChart method from ChartView.

itissandeep98:
>If the chart is null we could do setEnabled(false)

I don't understand what exactly do you mean by this?

robinschmid:
Lets say we create a ChartView - there is the case where the chart in the constructor is null and we pass the chart later (or we change the JFreeChart later) with setChart().

In the case the chart of a ChartView is null, we might want to disable (grey-out) all menu items or the whole menu. setEnabled is the way we typically did it in swing. JavaFX should have something similar.

PR #59, opened 2020-04-04
fixed npe when applied to mirror plot

SteffenHeu:
When working on the spectral search, i noticed a crash during theme application if it was applied to a mirror plot. This was due to the mirror plot having no range axis but subplots with their own range axes.


tomas-pluskal:
Looks good, thanks!

PR #60, opened 2020-04-06
Swing to javafx mzhistogram

itissandeep98:
This PR is for the conversion of io/github/mzmine/modules/visualization/mzhistogram/chart/HistogramPanel.java from Swing to JavaFX.
there is a lot of code change because the direct alternative of swing components in java was not giving optimal results. In short, the code related to UI change is as follows
* I have created 4 HBox in place of earlier JPanels 
* A VBox is containing all those 4 HBox and a label with proper spacing
* That VBox is the bottom pane of the main Pane 

Moreover, I have removed the dependency of DelayedDocumentListener.java and added an equivalent code for it.


src/main/java/io/github/mzmine/modules/visualization/mzhistogram/chart/HistogramDialog.java
```diff
@@ -56,16 +56,14 @@ public HistogramDialog(String title, String xLabel, HistogramData data, double b
...
     histo = new HistogramPanel(xLabel, data, binWidth);
 
-    SwingNode swingNode = new SwingNode();
-    mainPane.setCenter(swingNode);
-    SwingUtilities.invokeLater(() -> {
-      swingNode.setContent(histo);
+    setMinWidth(1050);
+    setMinHeight(700);
+    setScene(mainScene);
+    Platform.runLater(() -> {
```
tomas-pluskal, line 65: Here it is not necessary to use Platform.runLater(), because the Stage constructor must always be called on the JavaFX thread, anyway.

src/main/java/io/github/mzmine/modules/visualization/mzhistogram/chart/HistogramPanel.java
```diff
@@ -400,108 +312,112 @@ private void updateHistograms() {
...
-                logger.error("", e);
-                lbStats.setText("ERROR");
-              } catch (ExecutionException e) {
-                logger.error("", e);
-                lbStats.setText("ERROR");
-              }
-            }
-          }.execute();
+          try {
+            Platform.runLater(() -> {
```
tomas-pluskal, line 326: Here Platform.runLater() is not the best solution, because the background calculation will be performed on the JavaFX thread and therefore block the GUI. It might be best to use a thread pool executor, see here: https://stackoverflow.com/questions/29733004/properly-doing-multithreading-and-thread-pools-with-javafx-tasks

tomas-pluskal:
Looks great, thanks! I added a few comments.

src/main/java/io/github/mzmine/modules/visualization/mzhistogram/chart/HistogramPanel.java
```diff
@@ -315,16 +232,14 @@ public void setData(HistogramData data, double binWidth) {
         try {
           bws = Precision.toString(bw, 4);
         } catch (Exception e) {
+          logger.error("", e);
```
tomas-pluskal, line 235: Please add some descriptive error messages when calling the logger

src/main/java/io/github/mzmine/modules/visualization/mzhistogram/chart/HistogramPanel.java
itissandeep98: Yes i thought of that too, but it was giving an IllegalStateExeception as we are trying to update the chart which is on javafx thread

tomas-pluskal: You can do the calculation on the thread pool executor, and then update the chart from the threadpool executor thread using Platform.runLater(). The key point is to do the CPU-intensive calculation on a different thread and the GUI updates on the JavaFX thread.

itissandeep98: are you suggesting something like this?
image

tomas-pluskal: Yes, exactly

itissandeep98:
One more thing, The file Histogramdialog.java can be merged easily into EHistogramdialog.java. should i merge them?

tomas-pluskal:
> One more thing, The file Histogramdialog.java can be merged easily into EHistogramdialog.java. should i merge them?

Sounds good!

The GUI is now much more responsive after the threading update, I like it.

PR #62, opened 2020-04-11
Spectralsearchfx

SteffenHeu:
Hey Tomas, 

this converts the spectral match to javafx.

If you want to be able to sleep well at night, don't look at how the export works now :).

Best
Steffen


tomas-pluskal:
Thanks Steffen!
What is the Structure2DComponentAWT for? It seems like you don't use it.

SteffenHeu:
I do. In the export function. :D

tomas-pluskal:
Ah, I see. But I don't get the motivation behind it.
If you need to convert the JavaFX node into a Swing node for export, why don't you just wrap it in a JFXPanel? https://openjfx.io/javadoc/14/javafx.swing/javafx/embed/swing/JFXPanel.html

SteffenHeu:
I tried that, It resulted in a blank export.
I also tried the PrintJob to just export a fx node as pdf, but that converted it to an image in a pdf.

tomas-pluskal:
Got it! I guess there is no way to completely avoid Swing at the moment. Let's hope someone implements a better way to export JavaFX nodes to vector objects soon.

@robinschmid Could you briefly test this PR? I guess you have the most experience in using the spectral search module.

robinschmid:
It seems to work. Although, one time it crashed when I opened the  results dialog (I ran it on the list and opened it when it was still running).

And I cannot open the setup dialog from the results window. it tells me that it does not run from the FX thread:
`
Exception in thread "AWT-EventQueue-0" java.lang.IllegalStateException: Not on FX application thread; currentThread = AWT-EventQueue-0
	at com.sun.javafx.tk.Toolkit.checkFxUserThread(Toolkit.java:291)
	at com.sun.javafx.tk.quantum.QuantumToolkit.checkFxUserThread(QuantumToolkit.java:446)
	at javafx.stage.Stage.<init>(Stage.java:254)
	at javafx.stage.Stage.<init>(Stage.java:240)
	at io.github.mzmine.parameters.dialogs.ParameterSetupDialog.<init>(ParameterSetupDialog.java:113)
	at io.github.mzmine.parameters.dialogs.ParameterSetupDialog.<init>(ParameterSetupDialog.java:104)
	at io.github.mzmine.parameters.impl.SimpleParameterSet.showSetupDialog(SimpleParameterSet.java:179)
	at io.github.mzmine.modules.visualization.spectra.spectralmatchresults.SpectraIdentificationResultsWindow.lambda$0(SpectraIdentificationResultsWindow.java:95)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.AbstractButton.doClick(AbstractButton.java:369)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI.doClick(BasicMenuItemUI.java:1020)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI$Handler.mouseReleased(BasicMenuItemUI.java:1064)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6632)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3342)
	at java.desktop/java.awt.Component.processEvent(Component.java:6397)
	at java.desktop/java.awt.Container.processEvent(Container.java:2263)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5008)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2321)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4840)
	at java.desktop/java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4918)
	at java.desktop/java.awt.LightweightDispatcher.processMouseEvent(Container.java:4547)
	at java.desktop/java.awt.LightweightDispatcher.dispatchEvent(Container.java:4488)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2307)
	at java.desktop/java.awt.Window.dispatchEventImpl(Window.java:2762)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4840)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:772)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:721)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:715)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:389)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:85)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:95)
	at java.desktop/java.awt.EventQueue$5.run(EventQueue.java:745)
	at java.desktop/java.awt.EventQueue$5.run(EventQueue.java:743)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:389)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:85)
	at java.desktop/java.awt.EventQueue.dispatchEvent(EventQueue.java:742)
	at java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:203)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:124)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:113)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:109)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.desktop/java.awt.EventDispatchThread.run(EventDispatchThread.java:90)
`

robinschmid:
The results window is great - I cannot test the export to graphics function though

SteffenHeu:
> It seems to work. Although, one time it crashed when I opened the results dialog (I ran it on the list and opened it when it was still running).
> 
> And I cannot open the setup dialog from the results window. it tells me that it does not run from the FX thread:


I must have missed some functionality then. From where did you open the results dialog? I was only aware of the button next to the spectra window, which runs the search and automatically opens the result window.

robinschmid:
oh ok. you can run the module from 
1. the feature list methods/identification/spectral somewhere.
2. right click context menu on a feature list row (in table)
3. from a single spectrum

the first option does not open the result window directly as it runs on multiple rows.

I got the exception when trying to open the dialog within the results window to enable the export buttons.

SteffenHeu:
@tomas-pluskal / @robinschmid can you run anything from the Peak list table context menu? For me it crashes on every operation. Is it supposed to work at the moment?

robinschmid:
I was able to run some things from the context menu.
However, we should not invest too much time in the old table. We definitely need to change soon to the new feature model and FX table.

tomas-pluskal:
I agree with Robin. The Swing to JavaFX transition is almost completed now (with a great help from many GSoC student candidates), and the new feature table is the next big thing :)

tomas-pluskal:
@robinschmid Do you consider this PR ready to merge?

SteffenHeu:
I have not had time to test my latest commit. I can check it during the weekend, so robin has to test it twice.

SteffenHeu:
Feel free to test now, @robinschmid.

robinschmid:
Works now and I had no exceptions so far. Only two things
- The pdf export looks pretty good at first - but somehow the metadata (SMILES etc) are missing.
- the zoom is broken. Maybe this is a general thing with the charts at this point... However, somehow the bar width is increasing on zoom in
image

tomas-pluskal:
I wonder why the chemical structure is rendered differently in the window and in the exported PDF (see the example that robin posted above).

robinschmid:
Oh yeah I did not even realize this. Seems like the style of  the structure is different. Labels are very small in the JavaFX window version.
I think this is still a Swing component right? Maybe it is caused by swing within javafx

SteffenHeu:
I do not have a clue, why the bar is different. I did not change the plots. I did some refactoring in the creation, but it uses the same source. This should be caused in the renderer, shouldn't it?

tomas-pluskal:
> Edit 3: The renderer/the dataset did indeed behave differently. Manually overriding the getX ect. methods in the dataset fixed that. However, this is super weird.
> So the java fx-environment might also explain, why the mirror chart behaves differently. (In the screenshot the swing one has a joined domain axis, the javafx one does not.)

It might be worthing posting an issue on the jfreechart-fx repo to discuss these problems

robinschmid:
Thats pretty weird. Lets hope that JFreeChart works fine in JavaFX and that we do not find more problems. I think it is still the best chart library.

The CustomTextPane was a mess from the start. we should change it to full JavaFX

SteffenHeu:
I am not using it in the fx pane, just in the export-awt one.

robinschmid:
Some weird things are going on with JFreeChart in JavaFX.
I have found a dirty solution to the gap. If you setGap(-9) there seems to be no gap (on my two screens). But you will have to do this everytime after you apply a theme.

https://github.com/robinschmid/mzmine3/commit/9a1e018c6fef7c117b84d167b6d7277997ba1c9d

robinschmid:
JFreeChart in JavaFX in general is weird. All the Gestures are kind of broken

SteffenHeu:
> 
> 
> Some weird things are going on with JFreeChart in JavaFX.
> I have found a dirty solution to the gap. If you setGap(-9) there seems to be no gap (on my two screens). But you will have to do this everytime after you apply a theme.
> 
> [robinschmid@9a1e018](https://github.com/robinschmid/mzmine3/commit/9a1e018c6fef7c117b84d167b6d7277997ba1c9d)

For me, this leads to the following behavior:
grafik
After creation it is fine (see bottom plot). When you do right click->export graphics the plot reverts to the old gap (top chart). When you then apply the theme in the export dialog, the gap looks like it was doubled (0s don't align anymore).

robinschmid:
Thats weird. Is the chart placement in graphics export similar to the results window? (the surrounding layouts / panes?) Maybe the gap is considered correctly in the graphics export window.
The theme usually sets the gap to the default values - we should add the gap to the theme and apply it to combined plots.

SteffenHeu:
> Thats weird. Is the chart placement in graphics export similar to the results window? (the surrounding layouts / panes?) Maybe the gap is considered correctly in the graphics export window.

Export: surrounded by a border pane. - Results: In a grid pane. 
I tried wrapping the chart in a boder pane to put that into a grid pane. Did not work.

> The theme usually sets the gap to the default values - we should add the gap to the theme and apply it to combined plots.
I tried that already, it does not work.

I tried some more things:
- Removing all the setGap during the plot creation itself, only do it during theme application. - Does not work
- Since applyTheme in the export dialog works for some reason, i  thought it might have to do with a swing panel not updating while it is not show. So i added an extra button in the SpectralMatchPanelFX to manually trigger the theme application which would then update the jfreechart while it's already on screen. -> Does not work.

I don't have any clue what could be the reason for this odd behaviour.

tomas-pluskal:
The problem is not in jfreechart-fx, but in `EStandardChartTheme`. 
The `EStandardChartTheme` class extends `StandardChartTheme` and in its apply() method it calls super.apply(), see here:
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/gui/chartbasics/chartthemes/EStandardChartTheme.java#L144

The problem is that the super class sets axis offset individually for each embedded subplot:

https://github.com/jfree/jfreechart/blob/master/src/main/java/org/jfree/chart/StandardChartTheme.java#L1274

And the EStandardChartTheme doesn't fix the offset of the subplots afterwards (it only sets the axis offset for the main combined plot, which is not enough).

SteffenHeu:
But that does not explain why it behaved differently in the first place, does it?

tomas-pluskal:
> But that does not explain why it behaved differently in the first place, does it?

I have not investigated that. It's probably because of some minor difference between EChartViewer and EChartPanel..

src/main/java/io/github/mzmine/gui/chartbasics/chartthemes/EStandardChartTheme.java
```diff
@@ -37,25 +40,27 @@
...
   private static final Color DEFAULT_CROSS_HAIR_COLOR = Color.BLACK;
 
   private static final boolean DEFAULT_CROSS_HAIR_VISIBLE = true;
   private static final Stroke DEFAULT_CROSS_HAIR_STROKE = new BasicStroke(1.0F,
-      BasicStroke.CAP_BUTT, BasicStroke.JOIN_BEVEL, 1.0f, new float[] {5.0F, 3.0F}, 0.0F);
+      BasicStroke.CAP_BUTT, BasicStroke.JOIN_BEVEL, 1.0f, new float[]{5.0F, 3.0F}, 0.0F);
 
-  private static final RectangleInsets DEFAULT_AXIS_OFFSET =
-      new RectangleInsets(5.0, 5.0, 5.0, 5.0);
+  private static final RectangleInsets DEFAULT_AXIS_OFFSET = RectangleInsets.ZERO_INSETS;
```
tomas-pluskal, line 62: Not sure about this approach. Now DEFAULT_AXIS_OFFSET and MIRROR_PLOT_AXIS_OFFSET are redundant, and your change would remove the offset for all charts, not just mirror plots. Is that really desirable?

SteffenHeu: I was not sure, if we had intended the axis offset before. I'll put it back.

SteffenHeu: Do we want an offset on the range axis of the mirror plot aswell? (not for the export ofc)

tomas-pluskal: You can try with and without the offset and choose whichever is more visually pleasing :)

tomas-pluskal:
I'm going to merge this now, if we find some bugs or problems, we can deal with them in separate PRs.

PR #63, opened 2020-04-16
Complete Parameter setup option in JavaFX

SurajViitk:


tomas-pluskal:
Looks great!!

A few comments and suggestions:
- When adding a new parameter, the parameter name should be required. Currently the dialog accepts empty name.
- The Add parameter dialog should have a Cancel button
- The Cancel button in the Setup project parameters dialog can be removed, because there is no practical difference between clicking OK and Cancel (or is it?)
- Instead it would be great to add a Help button, similar to other setup dialogs in the project. In the help, you can write a very short (1 paragraph) description what the dialog does and how the parameter import works (what is the imported file format). The Help was missing in MZmine 2.

tomas-pluskal:
Maybe you can also set min column width for each column, because currently they can be resized too small.

SurajViitk:
> Looks great!!
> 
> A few comments and suggestions:
> 
> * When adding a new parameter, the parameter name should be required. Currently the dialog accepts empty name.
> * The Add parameter dialog should have a Cancel button
> * The Cancel button in the Setup project parameters dialog can be removed, because there is no practical difference between clicking OK and Cancel (or is it?)
> * Instead it would be great to add a Help button, similar to other setup dialogs in the project. In the help, you can write a very short (1 paragraph) description what the dialog does and how the parameter import works (what is the imported file format). The Help was missing in MZmine 2.



> * The Cancel button in the Setup project parameters dialog can be removed, because there is no practical difference between clicking OK and Cancel (or is it?)

I found [this](https://ux.stackexchange.com/questions/17507/is-a-cancel-button-necessary-in-a-windows-form/17509#17509) in reference to cancel and close button dilemma, and I think we should follow one of these for **both** the dialogs - i) Have Cancel button to abandon any change made and close buttton to act like OK button as mentioned in guideline ii) Not have Cancel button, let close button have it's functionality. Which one (or any other) option do you recommend @tomas-pluskal ?

tomas-pluskal:
> > Looks great!!
> > A few comments and suggestions:
> > 
> > * When adding a new parameter, the parameter name should be required. Currently the dialog accepts empty name.
> > * The Add parameter dialog should have a Cancel button
> > * The Cancel button in the Setup project parameters dialog can be removed, because there is no practical difference between clicking OK and Cancel (or is it?)
> > * Instead it would be great to add a Help button, similar to other setup dialogs in the project. In the help, you can write a very short (1 paragraph) description what the dialog does and how the parameter import works (what is the imported file format). The Help was missing in MZmine 2.
> 
> > * The Cancel button in the Setup project parameters dialog can be removed, because there is no practical difference between clicking OK and Cancel (or is it?)
> 
> I found [this](https://ux.stackexchange.com/a/17510) in reference to cancel and close button dilemma, and I think we should follow one of these for **both** the dialogs - i) Have Cancel button to abandon any change made and close buttton to act like OK button as mentioned in guideline ii) Not have Cancel button, let close button have it's functionality. Which one (or any other) option do you recommend @tomas-pluskal ?

Use option ii) for the main dialog and option i) for the "Add parameter" dialog.

tomas-pluskal:
Have you figured out the problem with the Help button?

SurajViitk:
> Have you figured out the problem with the Help button?

Looking into it

tomas-pluskal:
Thanks a lot!

PR #65, opened 2020-04-21
Adding Jpackage

knakul853:
In this PR I will be adding [jpackage](https://badass-jlink-plugin.beryx.org/releases/latest/#_jpackage) :

previous behavior:-  MZmine is using the org.beryx.runtime Gradle plugin to build a runtime image that includes the Java runtime, the dependencies (jars), the compiled classes, resources, and a startup script. That all happens when you run the ./gradlew script.<br>
but after the release of Java 14, we can use jpackage, which allows the build to produce a native application package for Windows, macOS, and Linux.


build.gradle
```diff
@@ -248,7 +248,19 @@ runtime {
                "java.security.jgss",
                "jdk.unsupported",
                "jdk.unsupported.desktop"]
-    
+    jpackage {
+        jpackageHome = '/usr/lib/jvm/jdk-14'
```
tomas-pluskal, line 252: This is not going to work universally. Can you set the location based on the java runtime used for the build?

build.gradle
```diff
@@ -248,7 +248,19 @@ runtime {
...
+        outputDir = 'mzmine-packaging'
+        if(org.gradle.internal.os.OperatingSystem.current().windows) {
+            installerType = 'exe'
+            skipInstaller = true
```
tomas-pluskal, line 256: It's not obvious what these options do, maybe add some brief comments to them?

build.gradle
```diff
@@ -4,15 +4,15 @@ plugins {
...
+    id "org.beryx.runtime" version "1.8.1"
 }
 
 group = "io.github.mzmine"
 version = "3.0"
 description = "MZmine"
 sourceCompatibility = "12"
 targetCompatibility = "12"
-defaultTasks "runtimeZip"
+defaultTasks 'clean',"runtimeZip"
```
tomas-pluskal, line 15: If the jpackage task is working, we can make it default

build.gradle
```diff
@@ -248,7 +248,19 @@ runtime {
...
-    
+    jpackage {
+        jpackageHome = '/usr/lib/jvm/jdk-14'
+        outputDir = 'mzmine-packaging'
+        if(org.gradle.internal.os.OperatingSystem.current().windows) {
+            installerType = 'exe'
+            skipInstaller = true
+            imageOptions = ['--win-console']
+            installerName = "MZmine"
+            appVersion = "3"
```
tomas-pluskal, line 259: Use ${version} instead

build.gradle
```diff
@@ -4,15 +4,15 @@ plugins {
...
     id "org.openjfx.javafxplugin" version "0.0.8"
-    id "org.beryx.runtime" version "1.8.0"
+    id "org.beryx.runtime" version "1.8.1"
 }
 
 group = "io.github.mzmine"
 version = "3.0"
 description = "MZmine"
 sourceCompatibility = "12"
 targetCompatibility = "12"
```
tomas-pluskal, line 14: We should change the sourceCompatibility and targetCompatibility to 14, to make sure JDK 14 is used for build

build.gradle
```diff
@@ -248,7 +248,19 @@ runtime {
                "java.security.jgss",
                "jdk.unsupported",
                "jdk.unsupported.desktop"]
-    
+    jpackage {
+        jpackageHome = '/usr/lib/jvm/jdk-14'
+        outputDir = 'mzmine-packaging'
```
tomas-pluskal, line 253: It is not necessary to change the outputDir, just leave default value

tomas-pluskal:
Screen Shot 2020-04-22 at 11 43 24

It works on macOS. Some things need improving:
- package icon
- target package type should be DMG on macOS
- package naming should be MZmine-3.0.dmg (currently mzmine3-3.0.dmg)

tomas-pluskal:
We also need to fix the logging configuration. Perhaps the log file should be saved to user's home directory as `.mzmine.log`

build.gradle
knakul853: Actually I figured it out that we don't need this we can configure in our idea

tomas-pluskal: Generally speaking, we use the command-line build via gradlew to make the official MZmine releases, not the IDE like IDEA

build.gradle
```diff
@@ -4,15 +4,15 @@ plugins {
...
 
 group = "io.github.mzmine"
 version = "3.0"
 description = "MZmine"
-sourceCompatibility = "12"
-targetCompatibility = "12"
-defaultTasks "runtimeZip"
+sourceCompatibility = "14"
+targetCompatibility = "14"
+defaultTasks 'clean', "runtimeZip", "jpackage","jpackageImage"
```
tomas-pluskal, line 15: What is the practical difference between jpackage and jpackageImage?
I think you should remove the clean and runtimeZip targets if we use jpackage as default.

build.gradle
```diff
@@ -224,7 +224,7 @@ startScripts {
 }
 
 /*
- * Build the JVM runtime using jlink
+ * Build the JVM runtime using runtime
```
tomas-pluskal, line 227: Actually this comment was correct, the JVM runtime is built using jlink (even if you use the badass-runtime plugin)

build.gradle
```diff
@@ -248,19 +248,33 @@ runtime {
                "java.security.jgss",
                "jdk.unsupported",
                "jdk.unsupported.desktop"]
-    
+    jpackage {
+
+        if(org.gradle.internal.os.OperatingSystem.current().windows) {
```
tomas-pluskal, line 253: Why are these options only applied on Windows?

build.gradle
```diff
@@ -248,19 +248,33 @@ runtime {
...
     imageZip = file(relativePath(imageDir) + ".zip")
 }
-
 /*
  * Copy the resources and logging properties
  */
 tasks.runtime.doLast {
 
-    // Remove the "release" file produced by jlink
+    // Remove the "release" file produced by runtime
```
tomas-pluskal, line 276: The original comment was correct

build.gradle
knakul853: It is not necessary for windows we can remove these.

build.gradle
```diff
@@ -248,19 +248,27 @@ runtime {
                "java.security.jgss",
                "jdk.unsupported",
                "jdk.unsupported.desktop"]
-    
+    jpackage {
+            //lets you generate only the platform-specific application
+            skipInstaller = true
+
+            //list of additional options to be passed to the jpackage executable when creating the application installer.
+            imageOptions = ['--icon', 'src/main/resources/MZmineIcon.png']
```
tomas-pluskal, line 256: Seems like it is not accepting the icon, I see this error:
````
WARNING: Using incubator modules: jdk.incubator.jpackage
The specified icon "src/main/resources/MZmineIcon.png" is not an ICNS file and will not be used. The default icon will be used in its place.
````

File: build.gradle [hunk 12]
```diff
@@ -248,19 +248,27 @@ runtime {
                "java.security.jgss",
                "jdk.unsupported",
                "jdk.unsupported.desktop"]
-    
+    jpackage {
+            //lets you generate only the platform-specific application
+            skipInstaller = true
```
tomas-pluskal, line 253: Instead of skipInstaller, you should set the installerType, but it needs to be set based on platform.
on macOS: "dmg"
on Windows: "msi"
on Linux: null (generate all packages)

tomas-pluskal:
I think it looks great now! 
The last thing we need to take care of is the logging configuration. 
By default the log is written to `log/mzmine.log`, see here: https://github.com/mzmine/mzmine3/blob/master/src/main/resources/logging.properties#L45
This doesn't work anymore once the application is packaged with jpackage.
We should save the log file to the user home directory as a hidden file, similarly to how the config file is saved:
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/main/MZmineConfiguration.java#L41
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/main/impl/MZmineConfigurationImpl.java#L240

tomas-pluskal:
Please also add these options on Windows to make sure the application icon is created:
`installerOptions = ['--win-menu', '--win-shortcut']`

tomas-pluskal:
Also would be useful to add `--win-dir-chooser` and `--win-upgrade-uuid <string>`, see the Windows-specific options here: https://openjdk.java.net/jeps/343

tomas-pluskal:
Complete documentation for jpackage:
[packaging-tool-user-guide.pdf](https://github.com/mzmine/mzmine3/files/4543021/packaging-tool-user-guide.pdf)

knakul853:
Thanks for doc.

build.gradle
```diff
@@ -248,19 +248,33 @@ runtime {
                "java.security.jgss",
                "jdk.unsupported",
                "jdk.unsupported.desktop"]
-    
+    jpackage {
+        if(org.gradle.internal.os.OperatingSystem.current().windows) {
+            installerType = "msi"
+            //list of additional options to be passed to the jpackage executable when creating the application installer.
+            imageOptions = ['--icon', 'src/main/resources/MZmineIcon.ico']
+            installerOptions = ['--win-menu', '--win-shortcut','--win-dir-chooser','--win-upgrade-uuid <string>']
```
tomas-pluskal, line 256: The `<string>` part needs to be replaced with actual UUID, check the jpackage documentation

src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -50,14 +35,36 @@
...
  * MZmine main class
  */
 public final class MZmineCore {
 
   private static Logger logger = Logger.getLogger(MZmineCore.class.getName());
+  private static Handler fileHandler;
+
+  static {
+    try {
+      fileHandler = new FileHandler("log/mzmine.log");
```
tomas-pluskal, line 62: I don't quite understand why you're making the FileHandler here? The logging is configured in the logging.properties file so you should update that file instead..

build.gradle
```diff
@@ -248,19 +248,33 @@ runtime {
                "java.security.jgss",
                "jdk.unsupported",
                "jdk.unsupported.desktop"]
-    
+    jpackage {
+        if(org.gradle.internal.os.OperatingSystem.current().windows) {
+            installerType = "msi"
+            //list of additional options to be passed to the jpackage executable when creating the application installer.
+            imageOptions = ['--icon', 'src/main/resources/MZmineIcon.ico']
+            installerOptions = ['--win-menu', '--win-shortcut','--win-dir-chooser','--win-upgrade-uuid']
```
tomas-pluskal, line 256: Did you check the docs about the win-upgrade-uuid option?

src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -81,6 +80,8 @@ public static void main(String args[]) {
      * Configure the logging properties before we start logging
      */
     MZmineLogging.configureLogging();
+    InputStream configFile = MZmineCore.class.getResourceAsStream("/logging.properties");
```
tomas-pluskal, line 83: The logging.properties file should be automatically loaded on application start if it is placed in the resources folder. Is it not loaded automatically?

knakul853: I followed some tutorials, didn't know this fact. : )

build.gradle
```diff
@@ -248,19 +248,33 @@ runtime {
                "java.security.jgss",
                "jdk.unsupported",
                "jdk.unsupported.desktop"]
-    
+    jpackage {
+        if(org.gradle.internal.os.OperatingSystem.current().windows) {
+            installerType = "msi"
+            //list of additional options to be passed to the jpackage executable when creating the application installer.
+            imageOptions = ['--icon', 'src/main/resources/MZmineIcon.ico']
+            installerOptions = ['--win-menu', '--win-shortcut','--win-dir-chooser','--win-upgrade-uuid', 'aabbccdd-eeff-1122-3344-556677889900']
```
tomas-pluskal, line 256: How did you decide the UUID?
Use https://www.uuidgenerator.net/ instead.
Also, make the UUID a constant that is defined in the beginning of the file (`ext` section) and use the constant here.

tomas-pluskal:
This section can be removed from build.gradle
````

    // Remove the "release" file produced by jlink
    delete file(relativePath(imageDir) + "/release")

    copy {
        from("log")
        into(relativePath(imageDir) + "/log")
    }
    copy {
        from("manual.pdf")
        into(relativePath(imageDir))
    }
    copy {
        from("LICENSE.txt")
        into(relativePath(imageDir) + "/legal/MZmine/")
    }
````
Because if we package the application with jpackage, this is not relevant anymore.

File: build.gradle (hunk 17 above)
Comment (knakul853) [reply build.gradle]: I generated with ```UUID.randomUUID()```

Comment (tomas-pluskal) [reply build.gradle]: Please add that information in a comment in the build.gradle file

File: src/main/java/io/github/mzmine/main/MZmineCore.java [hunk 18]
```diff
@@ -67,10 +63,13 @@
   private static Map<Class<?>, MZmineModule> initializedModules =
       new Hashtable<Class<?>, MZmineModule>();
 
+  public MZmineCore() throws IOException {
```
tomas-pluskal, line 66: This change should be removed

src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -67,10 +63,13 @@
...
+  }
+
   /**
    * Main method
    */
-  public static void main(String args[]) {
+  public static void main(String args[]) throws IOException {
```
tomas-pluskal, line 72: This change should be removed

tomas-pluskal:
What is the status of the logging? Have you tested it?

knakul853:
Yes have tested https://pastebin.com/5wDs0rwW is content of logger

tomas-pluskal:
> Yes have tested https://pastebin.com/5wDs0rwW is content of logger

Your pull request doesn't update the logging.properties file, is something missing here?

src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -67,6 +62,9 @@
   private static Map<Class<?>, MZmineModule> initializedModules =
       new Hashtable<Class<?>, MZmineModule>();
 
+  public MZmineCore(){
```
tomas-pluskal, line 65: Please remove this constructor completely.

knakul853:
> > Yes have tested https://pastebin.com/5wDs0rwW is content of logger
> 
> Your pull request doesn't update the logging.properties file, is something missing here?

it's weird by branch have the updated logging.properties don't know why it is not reflecting here

knakul853:
checked the logging properties since there is no new +(addition) or - (deletion) have made that is why it was not updating. for checking ,i just removed one line and it is reflecting in PR https://github.com/mzmine/mzmine3/pull/65/commits/359f1d16bb35ae012c0c79787c9e4ee031a4fada

tomas-pluskal:
> checked the logging properties since there is no new +(addition) or - (deletion) have made that is why it was not updating. for checking ,i just removed one line and it is reflecting in PR [359f1d1](https://github.com/mzmine/mzmine3/commit/359f1d16bb35ae012c0c79787c9e4ee031a4fada)

Did you read carefully what I said about the logging in my comment 9 days ago?

knakul853:
Do you mean making this hidden?

knakul853:
Sorry. I thought first to create a log as it was generated earlier and then just make it hidden in home directory.

tomas-pluskal:
> Sorry. I thought first to create a log as it was generated earlier and then just make it hidden in home directory.

On macOS the user cannot access the log folder, because it is embedded in the application package.

On Windows the application is installed into Program Files, and that is not a good location to store log files either.

That's why we need to change the location of the log file

build.gradle
```diff
@@ -63,6 +63,9 @@ ext {
     poiVersion = "4.1.0"
     batikVersion = "1.11"
     controlsfxVersion = "11.0.1"
+    //The UUIDs generated by https://www.uuidgenerator.net/ site are provided AS IS without warranty of any kind,
```
tomas-pluskal, line 66: This comment is true, but it would be useful to explain *why* the uuid is here, what is the purpose.

knakul853: ok.

knakul853:
In that case, we can store as  %h/java%u.log 
https://github.com/mzmine/mzmine3/blob/master/src/main/resources/logging.properties#L36

knakul853:
> > Sorry. I thought first to create a log as it was generated earlier and then just make it hidden in home directory.
> 
> On macOS the user cannot access the log folder, because it is embedded in the application package (https://en.wikipedia.org/wiki/Bundle_(macOS)#macOS_application_bundles).
> 
> On Windows the application is installed into Program Files, and that is not a good location to store log files either.
> 
> That's why we need to change the location of the log file

In that case, we can store as  %h/java%u.log 
https://github.com/mzmine/mzmine3/blob/master/src/main/resources/logging.properties#L36

knakul853:
dropping %u since it will make a unique log whenever we try to run the application.

knakul853:
Ok that's great

tomas-pluskal:
I think it looks good now!

Annexhc:
Has anyone tested this on linux? I have just installed java 14, cloned the latest master branch, and tried to build. I get the following error:
./gradlew 

FAILURE: Build failed with an exception.

* What went wrong:
Could not initialize class org.codehaus.groovy.runtime.InvokerHelper

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 739ms

Here are my gradle settings:
------------------------------------------------------------
Gradle 6.3
------------------------------------------------------------

Build time:   2020-03-24 22:12:49 UTC
Revision:     <unknown>

Kotlin:       1.3.70
Groovy:       2.5.10
Ant:          Apache Ant(TM) version 1.10.7 compiled on September 1 2019
JVM:          14 (Oracle Corporation 14+36-1461)
OS:           Linux 5.6.10-3-MANJARO amd64

Java version:
java -version
openjdk version "14" 2020-03-17
OpenJDK Runtime Environment (build 14+36-1461)
OpenJDK 64-Bit Server VM (build 14+36-1461, mixed mode, sharing)

SteffenHeu:
On windows I was getting this error:
IntelliJ: Could not target platform: 'Java SE 14' using tool chain: 'JDK 13 (13)'
I had to manually set the JAVA_HOME environment variable to Java 14 although all settings in IntelliJ and Gradle were set to Java 14. 

Now i am getting this exception when executing gradlew:
> Task :jpackage FAILED
WARNING: Using incubator modules: jdk.incubator.jpackage
Can not find WiX tools (light.exe, candle.exe)
Download WiX 3.0 or later from https://wixtoolset.org and add it to the PATH.
Error: Invalid or unsupported type: [msi]


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':jpackage'.
> Process 'command 'C:\Program Files\Java\openjdk-14.0.1/bin/jpackage.exe'' finished with non-zero exit value 1

Annexhc:
I was able to build it with intellij, thanks @SteffenHeu

tomas-pluskal:
Running the application from the IDE (Eclipse, IDEA, etc.) should be completely unrelated to the jpackage build via gradle. You should be only using gradle when building the final package for deployment.

Annexhc:
Yes, but intellij allows you to build via gradle which I did for testing. Anyway, it still failes using gradle in terminal

tomas-pluskal:
@Annexhc Try updating your JDK to 14.0.1 and see if that changes anything

tomas-pluskal:
@Annexhc Seems like it could be a classpath issue:
http://gradle.1045684.n5.nabble.com/Could-not-initialize-class-org-codehause-groovy-runtime-InvokerHelper-td4452034.html

tomas-pluskal:
@Annexhc Can also try upgrading gradle to 6.4

Annexhc:
Thanks, will try

PR #69, opened 2020-05-13
minor UI changes and handle close button

SurajViitk:
1) UI change - Button were not visible on reducing window size.
2) On closing the window, database search task is now stopped.
3) On updating to jdk14, I was getting an error on running ./gradlew
`Error: Invalid or unsupported type: [rpm]`
so I added it, I'm not sure but there might be problems with rpm based linux distros like CentOS after adding this?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -241,7 +241,6 @@ runtime {
             imageOptions = ['--icon', 'src/main/resources/MZmineIcon.icns']
             installerOptions = ['--license-file', 'LICENSE.txt']
         } else {
-            installerType = "deb"
             // additional options for jpackage
             imageOptions = ['--icon', 'src/main/resources/MZmineIcon.png']
             installerOptions = ['--linux-shortcut',
--- a/build.gradle
+++ b/build.gradle
@@ -241,6 +241,7 @@ runtime {
             imageOptions = ['--icon', 'src/main/resources/MZmineIcon.icns']
             installerOptions = ['--license-file', 'LICENSE.txt']
         } else {
+            installerType = "deb"
             // additional options for jpackage
             imageOptions = ['--icon', 'src/main/resources/MZmineIcon.png']
             installerOptions = ['--linux-shortcut',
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/mzmine/modules/dataprocessing/id_onlinecompounddb/SingleRowIdentificationTask.java
```diff
@@ -122,6 +122,12 @@ public void run() {
     Platform.runLater(() -> {
               resultWindowFX= new ResultWindowFX(peakListRow, searchedMass, this);
               resultWindowFX.show();
+              //close button handle
+              resultWindowFX.getScene().getWindow().setOnCloseRequest(e->{
+                if (getStatus() != TaskStatus.FINISHED) {
```
tomas-pluskal, line 127: This is wrong, because for example if the task was CANCELED you will incorrectly switch it to FINISHED.
Use this instead:
```
if ((getStatus() == TaskStatus.WAITING) || (getStatus() == TaskStatus.RUNNING)) {
   cancel();
}
```

build.gradle
```diff
@@ -241,6 +241,7 @@ runtime {
             imageOptions = ['--icon', 'src/main/resources/MZmineIcon.icns']
             installerOptions = ['--license-file', 'LICENSE.txt']
         } else {
+            installerType = "deb"
```
tomas-pluskal, line 244: This should be removed from this PR.

src/main/java/io/github/mzmine/modules/dataprocessing/id_onlinecompounddb/SingleRowIdentificationTask.java
```diff
@@ -122,6 +122,12 @@ public void run() {
     Platform.runLater(() -> {
               resultWindowFX= new ResultWindowFX(peakListRow, searchedMass, this);
               resultWindowFX.show();
+              //close button handle
+              resultWindowFX.getScene().getWindow().setOnCloseRequest(e->{
+                if (getStatus() == TaskStatus.WAITING || getStatus() == TaskStatus.PROCESSING) {
+                  setStatus(TaskStatus.FINISHED);
```
tomas-pluskal, line 128: should use `cancel()` here, not `setStatus(TaskStatus.FINISHED);`

SurajViitk: Sorry, overlooked that

PR #71, opened 2020-05-24
Add method in FormulaUtils.java to obtain mz values for ionic formula strings

lukasz-fiszer:
Add `FormulaUtils.calculateMzRatio(String ionicFormula)` method,
it is based on CDK's exisiting methods serving as a quick implementation to obtain mz values from the ionic formula strings.
It is going to be used in the new mass calibration module (for obtaining m/z ratios of standard calibrants given by their ionic formula strings) but is also likely of some use in the wider MZmine codebase


src/main/java/io/github/mzmine/util/FormulaUtils.java
```diff
@@ -161,6 +161,47 @@ public static String formatFormula(@Nonnull Map<String, Integer> parsedFormula)
...
+  public static double calculateMzRatio(String ionicFormula) {
+    IChemObjectBuilder builder = SilentChemObjectBuilder.getInstance();
+    IMolecularFormula mf = MolecularFormulaManipulator.getMolecularFormula(ionicFormula, builder);
+
+    int charge = 1;
+    char lastChar = ionicFormula.charAt(ionicFormula.length() - 1);
+    if(lastChar == '-') {
+      charge = -1;
+    }
+    if(mf.getCharge() != null) {
```
tomas-pluskal, line 194: Not sure what this does, actually?

src/main/java/io/github/mzmine/util/FormulaUtils.java
```diff
@@ -161,6 +161,47 @@ public static String formatFormula(@Nonnull Map<String, Integer> parsedFormula)
...
+      charge = mf.getCharge();
+    }
+
+    double mass = MolecularFormulaManipulator.getMass(mf, MolecularFormulaManipulator.MonoIsotopic);
+    mass -= charge * electronMass;
+
+    double mz = Math.abs(mass / charge);
```
tomas-pluskal, line 201: Any division always have to be checked in case the divider (charge) is 0

src/main/java/io/github/mzmine/util/FormulaUtils.java
lukasz-fiszer: As requested, when no charge is given in the string, +1 is assumed. For simple strings where charge is either -1 or +1, square brackets do not have to be used and the charge symbol is assumed to be placed at the end of the string. So if last character is '-' then it is -1 charge, if charge symbol is not given or it is '+' then the charge is +1.
In case square brackets are used and charge is given at the end as`(\d+)[\+\-]` molecular formula mf object will have that charge parsed into, so we can use `mf.getCharge()`.
For strings with no square brackets, `mf.getCharge()` is null, so we first check for simple case with no square brackets, and then check if mf charge is not null and if so we use it.
This method is a little hackish, but it seems to do its jobs while we want to implemented the mass calibration module.

src/main/java/io/github/mzmine/util/FormulaUtils.java
lukasz-fiszer: The `charge` should not be 0 at that point
unless mf charge is not null and also 0 but should not be the case as cdk seems to set charge to null if it will not find square brackets and charge at the end
but we can also use the following to make sure, 
```suggestion
    if(mf.getCharge() != null && mf.getCharge() != 0)
```

src/main/java/io/github/mzmine/util/FormulaUtils.java
tomas-pluskal: Okay, could you add this brief explanation as comments in the code? Because it is not clear from the code.

In case the formula does not end with + or - character, I think we should throw an IllegalArgumentException. That way we make sure the users only use this method for ionized formulas.

src/main/java/io/github/mzmine/util/FormulaUtils.java
tomas-pluskal: CDK sets the charge to null in its current version, but nothing guarantees that future CDK versions will behave the same. So it's always worth adding a check. Assumptions in code are bad, they just generate future bugs.

PR #76, opened 2020-06-03
Fix the rpm / deb build error on travis and linux

nilshoffmann:


tomas-pluskal:
Thanks Nils, you're right.

PR #77, opened 2020-06-06
Exportdialogfix

SteffenHeu:
Hey Tomas, 

I noticed the issue that the theme was not properly applied in the export dialog anymore. Especially the titles were not updated. 
I also removed some deprecated parameters, so the charts are exported as-is.

Additionally I increased the size of the color palette picker dialog a little bit, because the scroll bar was always showing.

Best
Steffen


tomas-pluskal:
Thanks, Steffen!

PR #78, opened 2020-06-07
Mass calibration module

lukasz-fiszer:
Mass calibration module


build.gradle
```diff
@@ -153,7 +154,7 @@ dependencies {
     implementation "org.json:json:20190722"
     implementation "com.panemu:tiwulfx:3.0" // For DetachableTabPane
     implementation "com.github.oshi:oshi-core:4.5.2" // because the version that is pulled by SIRIUS crashes
-
+    implementation "org.jfree:jcommon:1.0.24"
```
tomas-pluskal, line 157: Why do you need jcommon?

tomas-pluskal:
A few formal requirements:
- Add the license header to all new .java files.
- We use Google java code style, so please reformat your source codes accordingly. The repository has formatting definitions for Eclipse and IDEA so it's easy.
- The .png and .txt files that (I assume) you use for testing should not be included in the PR, it's better to keep them outside of the repository (otherwise, it just adds more mess to git history).

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationModule.java
```diff
@@ -0,0 +1,93 @@
...
+  @Override
+  @Nonnull
+  public ExitCode runModule(@Nonnull MZmineProject project, @Nonnull ParameterSet parameters,
+                            @Nonnull Collection<Task> tasks) {
+
+    String standardsListFilename = null;
+    StandardsListExtractor standardsListExtractor;
+    try {
+      standardsListFilename = parameters.getParameter(MassCalibrationParameters.standardsList).getValue().getName();
+      standardsListExtractor = StandardsListExtractor.createFromFilename(standardsListFilename);
```
tomas-pluskal, line 63: This should be moved into the MassCalibrationTask. The runModule() method is executed on the JavaFX thread, so it should not be performing any real data manipulations (freezes the GUI). Everything should be done in the task.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/errormodelling/BiasEstimator.java
```diff
@@ -0,0 +1,115 @@
...
+
+			index++;
+		}
+
+	}
+
+	protected static ArrayList<Double> getErrors(Sheet sheet)
+	{
+		System.out.println();
+		System.out.println(sheet.getSheetName());
```
tomas-pluskal, line 89: Use the logging system instead of System.out.println

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsList.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+/**
+ * Class for maintaining a list of standard molecules
+ * expected to appear in the dataset
+ */
+public class StandardsList {
```
tomas-pluskal, line 31: Do you need a specific class for this? Isn't it better to use a simple `List<Standard>`?

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsListSpreadsheetExtractor.java
```diff
@@ -0,0 +1,139 @@
...
+   *
+   * @return new standards list object
+   * @throws IllegalArgumentException thrown when sheet index is out of range available
+   * @throws RuntimeException         thrown when required column names are missing
+   */
+  public StandardsList extractStandardsList() {
+    logger.info("Extracting standards list " + filename + " sheet index " + sheetIndex);
+
+    if (this.extractedData != null) {
+      logger.info("Using cached list");
```
tomas-pluskal, line 79: logger.info should be only for important messages

tomas-pluskal:
Please don't use Excel file parsing. We use CSV files for these things across MZmine, so use a CSV format also for the standards.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationModule.java
lukasz-fiszer: Should not we check whether the file is valid before we open tasks? If it happens that the file is corrupted, the standards list will not load and that will cancel all tasks open for the module giving a lot of errors popping up in logs and console output. Checking the file once before starting the tasks seems like a nice place for it.
Also it does not perform any heavier data handling, just creates spreadsheet object. Standards list is loaded and cached on the first call to `extractStandardsList()` which happens within task run method.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsList.java
lukasz-fiszer: We have just used a standard calibrants list of hundreds of items, but later on we might have biggest lists and dataset of higher resolution giving orders of magnitude more matching attempts between mass peaks and calibrants from the list. We might use a more specialized data structure for this, also standards list is pretty much an entity of its own, with its own search methods.
So I decided to have a StandardsList class encapsulating that. Other causes include self-documentation and it seems easier to toss the list around in the codebase.
Though I agree it is a pretty small class as of now, but it is likely gonna grow as we demand more and move on with the module.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsListSpreadsheetExtractor.java
lukasz-fiszer: Changed to logger.fine, left logger.info just when printing the summary of the file. Also when it happens that some rows were skipped, eg: misformatted, I used logger.warning to print how many were skipped.

build.gradle
lukasz-fiszer: It is left from the experimental implementation, but it still seems to be required for DistributionPlot

lukasz-fiszer:
Thanks for the comments @tomas-pluskal 
As I mentioned the branch still needed a refactor and small changes, now it should be updated. The additional files were left from the experimental implementation and its own branch. I will clean up unnecessary files in later commit.
Speaking of file format, I was given xlsx file so went with that. I wanted to add csv either way so will add it later on.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationModule.java
tomas-pluskal: You should think of the task as a self-sufficient unit of work. The task is submitted to the task controller, and in principle it may be executed after a long delay or it may be executed by a different computer on a computing cluster (we don't have that now but in the future). So all the input checks should be done within the task, as the situation between the runModule() method and the execution of the task can change dramatically - even filesystem location and availability can change. 

If you like, you can keep a simple *additional* check in the runModule() method, but that should be only minimal, e.g. using File.exists() and File.canRead() methods, no parsing. And the checks needs to be repeated in the task's run() method, anyway, for the above mentioned reasons.

Parsing the standards list does count as heavy data handling. Imagine the list has a million rows, and the file is loaded from a network filesystem that is connected via a VPN from a server on a different continent (happens to me all the time). It will freeze the GUI.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsList.java
tomas-pluskal: Ok

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationModule.java
lukasz-fiszer: Alright, thank you for explaining the issue. The distributed processing argument makes a lot of sense.
But should we make sure that we load the list only once per module run, so that all tasks use the same list? Would all the tasks from the same module run be processed with access to the same storage?
On one hand it seems wasteful to load the same list multiple times, on the other hand it might be intentional that each task loads it within it own execution context. I assume, at least for now, we could make the list load only once and be cached for subsequent uses. This could be done for instance by synchronizing the `extractStandardsList()` method.

lukasz-fiszer:
I have cleaned up old unnecessary files and moved standards list initalization to module task class. Also added thread safe caching of extracted data from standards list and added csv extractor. 
Should be ready for a merge as a first version of the module.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
```diff
@@ -0,0 +1,75 @@
...
+  public static final RawDataFilesParameter dataFiles = new RawDataFilesParameter();
+
+  public static final MassListParameter massList = new MassListParameter();
+
+  public static final FileNameParameter standardsList = new FileNameParameter("Standards list",
+          "File with a list of standard molecules expected to appear in the dataset",
+          FileSelectionType.OPEN);
+
+  public static final DoubleParameter tolerance = new DoubleParameter("Range tolerance",
+          "Range tolerance is the max distance allowed between errors to be included in the range.",
```
tomas-pluskal, line 44: This parameter needs a better description. What kind of discance? What are the units of this value? What does it mean "included in the range"?

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
```diff
@@ -0,0 +1,75 @@
...
+          NumberFormat.getNumberInstance(), 0.4, 0.0, Double.POSITIVE_INFINITY);
+
+  public static final DoubleParameter mzRatioTolerance = new DoubleParameter("mz ratio tolerance",
```
tomas-pluskal, line 47: We have a specific parameter type called MZToleranceParameter, so please use that instead of DoubleParameter

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
```diff
@@ -0,0 +1,75 @@
...
+          "Max difference between actual mz peaks and standard calibrants to consider a match",
+          NumberFormat.getNumberInstance(), 0.01, 0.0, Double.POSITIVE_INFINITY);
+
+  public static final DoubleParameter retentionTimeSecTolerance = new DoubleParameter("Retention time tolerance",
```
tomas-pluskal, line 51: Please use RTToleranceParameter

build.gradle
```diff
@@ -75,6 +76,7 @@ dependencies {
     implementation "org.jfree:jfreechart-fx:1.0.1"
     implementation "org.jfree:fxgraphics2d:1.8"
     implementation "org.jfree:jfreesvg:3.3"
+    implementation "org.jfree:jcommon:1.0.24"
```
tomas-pluskal, line 79: Why did you add jcommon?

tomas-pluskal:
Sorry for the late response. Please merge the current master, cause there is a conflict in build.gradle. Not sure why you added the jcommon library, I think it should not be needed, it's a relic, jfreechart doesn't use it anymore. The code overall looks good, but the module needs a help file (HTML, see help files of other modules) with a brief explanation how it works and exact meaning of all parameter values.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
```diff
@@ -0,0 +1,84 @@
...
+  public static final DoubleParameter rangeSize = new DoubleParameter("Most populated error range size",
+          "The maximum length of the range that contains the most errors, in PPM m/z ratio.",
+          NumberFormat.getNumberInstance(), 2.0, 0.0, Double.POSITIVE_INFINITY);
+
+  public static final MZToleranceParameter mzRatioTolerance = new MZToleranceParameter("mz ratio tolerance",
+          "Max difference between actual mz peaks and standard calibrants to consider a match," +
+                  " max of m/z and ppm is used", 0.001, 5, true);
+
+  public static final RTToleranceParameter retentionTimeSecTolerance = new RTToleranceParameter("Retention time tolerance",
+          "Max retention time difference in seconds between mass peaks and standard calibrants to consider a match.");
```
tomas-pluskal, line 62: Don't mention seconds in the description, just say "Max retention time difference"

src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/RTTolerance.java
```diff
@@ -20,31 +20,52 @@
...
+  }
 
-    final double absoluteTolerance = isAbsolute ? tolerance : rtValue * tolerance;
+
+  public Range<Double> getToleranceRange(final double rtValue) {
+    // rtValue is given in minutes
+    double absoluteTolerance;
+    switch (unit) {
+      case SECONDS:
+        absoluteTolerance = tolerance / 60;
```
tomas-pluskal, line 59: This should be tolerance * 60, shouldn't it?

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
```diff
@@ -0,0 +1,84 @@
...
+  public static final MassListParameter massList = new MassListParameter();
+
+  public static final FileNameParameter standardsList = new FileNameParameter("Standards list",
+          "File with a list of standard calibrants (ionic formula and retention time)" +
+                  " expected to appear in the dataset", FileSelectionType.OPEN);
+
+  public static final BooleanParameter filterDuplicates = new BooleanParameter("Filter out duplicate errors",
+          "If checked, the distribution of errors will be filtered to remove duplicates");
+
+  public static final DoubleParameter tolerance = new DoubleParameter("Range tolerance",
```
tomas-pluskal, line 48: This parameter still needs a better description. "Range tolerance" doesn't really give any intuitive idea to a mass spectrometrist what range is it talking about. Can you explain more clearly?

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
```diff
@@ -0,0 +1,84 @@
...
+          "Range tolerance is the max distance allowed between errors to be included in the same range," +
+                  " in PPM m/z ratio.",
+          NumberFormat.getNumberInstance(), 0.4, 0.0, Double.POSITIVE_INFINITY);
+
+  public static final DoubleParameter rangeSize = new DoubleParameter("Most populated error range size",
```
tomas-pluskal, line 53: The same comment here - try to expand the description. Also, "in PPM m/z ratio" is a bit confusing. Is the value relative (in PPM) or absolute (in m/z)?

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/errormodeling/BiasEstimator.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Class containing methods estimating distribution bias
+ * <p>
+ * used with distributions of measurement errors to estimate systematic error of the measurements
+ * <p>
+ * for instance: after extracting a certain substantial subset of all mass measurement PPM errors,
+ * the measured mass spectrum bias can be estimated by taking the arithmetic mean of considered errors
+ */
+public class BiasEstimator {
```
tomas-pluskal, line 32: The Apache Math library already has functions for these basic math calculations, so it is better to avoid reinventing the wheel: https://commons.apache.org/proper/commons-math/javadocs/api-3.4/org/apache/commons/math3/stat/StatUtils.html#mean(double[])

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsListSpreadsheetExtractor.java
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * StandardsListExtractor for xls and xlsx spreadsheets
+ * uses sheet at specified index, first sheet available by default
+ * expects fixed case-insensitive column names
+ * 'retention time (min)' and 'ion formula'
+ * for storing needed data
+ */
+public class StandardsListSpreadsheetExtractor implements StandardsListExtractor {
+
+  protected static final String retentionTimeColumnName = "retention time (min)";
```
tomas-pluskal, line 38: Using the column names to extract the values is a bit too restrictive. E.g., somebody might call it "ret. time" instead of "retention time (min)" and your module won't work anymore. I think it is better to simply define the format as first column = RT and second column = ion formula, and ignore the headers.

tomas-pluskal:
I think the code looks pretty mature now. Please take a look at the comments I added and once you fix those issues we can merge this and focus on the preview components.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/errormodeling/BiasEstimator.java
tomas-pluskal: Also the Guava library: https://guava.dev/releases/snapshot/api/docs/com/google/common/math/Stats.html#meanOf-java.lang.Iterable-

src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/RTTolerance.java
lukasz-fiszer: I would say it is correct. `absolute tolerance` should be in minutes, so if the tolerance parameter in constructor is given in seconds, then it should be divided by 60 to get minutes. As a sanity check, if tolerance of 60 seconds is given, it should be just a minute. This is used to build the range with `rtValue` given in minutes.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/errormodeling/BiasEstimator.java
lukasz-fiszer: This class was meant for keeping related code together. When we add error size vs m/z value regression and shift the mass peaks with different values, this class is likely going to get an update. Also, because we are estimating bias from a list of errors, when the list is empty we want a bias of 0 not an exception or `Double.NaN` as the other methods give.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsListSpreadsheetExtractor.java
lukasz-fiszer: I have raised the issue of column naming (including the units of the columns) on slack before.
Column names are documented in help and in code docs. We could allow different units (min/sec) in parentheses and different naming (retention time/ret time/rt and so on) if needed, but for now fixed column names seemed enough.
I am not as convinced towards the variant where we fix columns by their position.
First, it is less portable, as other use cases for the standards library file could already dictate a certain column ordering. In both cases user needs to check the docs/help file to see the expected format, but it is easier to rename/duplicate a column then to maintain a certain order of columns.
Also with column names it would be easier to allow for m/z values to be given in another column (taking priority over ion formula if it was present). I have spoken with Albert recently and it seems like a valuable addition.
Also if we actually allow m/z values, then column names additionally defend against the case when RT and m/z columns would get swapped, which might not be trivial to detect and the module would likely casually run but with gibberish data.
I have updated the extractor classes to explicitly mention required columns names if an exception occurs when checking the file.
Please comment on what is your opinion on that, if you still want to use fixed column positions I can add it, but I feel more convinced towards the method we use currently.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/errormodeling/BiasEstimator.java
lukasz-fiszer: I have updated it a little

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsListSpreadsheetExtractor.java
tomas-pluskal: In other column MZmine modules we use defined column order, not column names. I would like to use the same approach here for consistency. In addition, relying on column names means you are relying on the user not making even a small difference when naming the columns, and that is just asking for problems. Please change it to defined column order, thanks.

src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/RTTolerance.java
tomas-pluskal: You're right, sorry for that.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
tomas-pluskal: Perhaps you can name this "Max m/z error range size (ppm)", that makes it more obvious what the parameter is for.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
```diff
@@ -0,0 +1,90 @@
...
+                  " expected to appear in the dataset", FileSelectionType.OPEN);
+
+  public static final BooleanParameter filterDuplicates = new BooleanParameter("Filter out duplicate errors",
+          "If checked, the distribution of errors will be filtered to remove duplicates");
+
+  public static final DoubleParameter tolerance = new DoubleParameter("Error range tolerance",
+          "Error range tolerance is the max distance allowed between errors to be included in the same range." +
+                  " This is used when extending the most populated error range, if next closest error is within that" +
+                  " tolerance, the range is extended to contain it. The process is repeated until no new errors can" +
+                  " be included that range. The tolerance is the absolute difference in PPM errors of m/z ratio." +
```
tomas-pluskal, line 52: "The tolerance is the absolute difference in PPM errors of m/z ratio" - not sure I understand the meaning of this. PPM is a relative value, not absolute.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsListSpreadsheetExtractor.java
lukasz-fiszer: When the column names would not match, an exception would get thrown and a very explicit error message with required column names would appear. But the consistency case changes the situation a little and adds to the fixed column position format. Updated as requested, retention time is taken from first column and ion formula string is taken from the second column.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
lukasz-fiszer: Yes, PPM is relative value, but the tolerance (used for extending the error range) is just a difference/distance between two PPM errors. It is an absolute value/difference used with the collection of values that also happen to have a prior meaning that is relative.
For instance, if a range's upper endpoint is at 0.4 (PPM error of the m/z match), if then next biggest error that appears in the distribution is at 0.55 and the tolerance is 0.2, the bigger error is within that fixed tolerance over the relative PPM errors and as such it would get included in the range, that is the range would get extended to contain it, having a newly established upper endpoint at 0.55. The range extension process continues until the range cannot be extended with such tolerance, that is no errors are within the tolerance from the range.
The method contains some subtleties that are not trivial to be communicated in a sentence or two of a short description. The help file contains more details, but it is likely going to get a retouch when we finish with the module as details might still change. When the live preview is added to the module the range extraction method and the parameters are going to be more clear.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
lukasz-fiszer: When using the tolerance to extend the range later on, that range size might get bigger than what would have been called max at this stage. This is still correct for that part, but for the whole run of the module it is not really max and it might get more confusing for the user.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
lukasz-fiszer: I have changed the wording in the description a little to make it more clear.

src/main/java/io/github/mzmine/modules/dataprocessing/id_cliquems/cliquemsimplementation/ComputeCliqueModule.java
```diff
@@ -428,6 +428,7 @@ public AnClique getClique(boolean filter, MZTolerance mzdiff, RTTolerance rtdiff
    * @return AnClique object
    */
   public AnClique getClique() {
-    return getClique(true, new MZTolerance(0,5),new RTTolerance(false,0.0001), 0.0001, .000001);
+    return getClique(true, new MZTolerance(0,5),
```
tomas-pluskal, line 431: This change should not be included in the PR, because it will create a conflict for someone else who is working on the CliqueMS module.

lukasz-fiszer:
Updated

tomas-pluskal:
Please merge to current master to resolve the conflict in RawDataElementName_2_5.java

lukasz-fiszer:
Remerged

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsListCsvExtractor.java
```diff
@@ -0,0 +1,97 @@
...
+   * Extracts standards list caching underlying list data
+   *
+   * @return new standards list object
+   * @throws IOException thrown when issues extracting from the file occur
+   */
+  public StandardsList extractStandardsList() throws IOException {
+    logger.fine("Extracting standards list " + filename);
+
+    if (this.extractedData != null) {
+      logger.fine("Using cached list");
```
tomas-pluskal, line 70: Please remove the caching function. It can cause problems, because the file might not be static but can be updated by the user. You would have to track last modification time etc... It's better to avoid caching altogether, I think.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
```diff
@@ -0,0 +1,221 @@
...
+import java.util.TreeMap;
+
+public class MassCalibrationParameters extends SimpleParameterSet {
+
+  public static final RawDataFilesParameter dataFiles = new RawDataFilesParameter();
+
+  public static final MassListParameter massList = new MassListParameter();
+
+  public enum MassPeakMatchingChoice {
+    STANDARDS_LIST("Standards list"), UNIVERSAL_CALIBRANTS("Universal calibrants");
```
tomas-pluskal, line 45: I think the selection of the calibration list could be moved to this enum - that way you can remove one additional combo box.
Options:
- Custom standards list
- Universal calibrants from Keller et al., Anal Chim Acta 2008, positive mode
- Universal calibrants from Keller et al., Anal Chim Acta 2008, negative mode
- Universal calibrants from Hawkes et al., Limnol Oceanogr Methods 2020, positive mode
- Universal calibrants from Hawkes et al., Limnol Oceanogr Methods 2020, negative mode
- Universal calibrants merged, positive mode
- Universal calibrants merged, negative mode

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/standardslist/StandardsListCsvExtractor.java
lukasz-fiszer: This was needed for when caching was used in the extractor factory. Currently it is not used, as extraction happens only once per module run and each extractor instance is created separately. But when the boolean flag to use caching is set on in extractor factory, then it would share the same instance among task runs and then the caching might get used. 
So as we currently do not use that, the user might edit the file between module runs and the new contents would be used. But we can use caching later by just changing the boolean flag in the method call.

src/main/java/io/github/mzmine/modules/dataprocessing/masscalibration/MassCalibrationParameters.java
lukasz-fiszer: With current codebase it required much less changes to introduce multiple universal calibrants list choices with a separate combo and then use a fixed mapping from the choice to a string pointing to the java resource containing the list. Also, it is safer to use separate parameters for nested combo parameter, so it would then require duplication of the mz tolerance parameter. We could make it into a single combobox but that would require unncessarily more changes in module code.

tomas-pluskal:
I'm going to merge this now, to make reviewing of the deisotoping module easier. Further updates can be merged in separate PRs.

PR #80, opened 2020-06-18
reverted EStandardChartTheme back to java 11

SteffenHeu:
Hi Tomas,

as discussed i removed the java 14 patterns and added the null check that was causing problems in other projects.

Best
Steffen


tomas-pluskal:
Thanks Steffen! I'll update the build script accordingly.

PR #83, opened 2020-06-27
Main window gui update

SteffenHeu:
Since the changes in this branch are getting huge, i figured it might be worth to open it as a PR already.
Feel free to leave comments on the code.

I recently changed quite a lot in the TICPlot, would be cool if someone could check that.

Best
Steffen


SteffenHeu:
grafik

That's how it looks like now

- Raw data files get assigned a color for visualisation
  - used in the raw data tree, SpectraPlot and TICPlot
  - colors are currently taken from the user defined color palette, can be changed later on
- Spectra plot and TIC plot 
  - now have properties for the selected item so listeners can be added
  - colors as specified by raw data ; label colors are changed, too
- Renamed PeakDataSet to FeatureDataSet 
- ChromatogramAndSpectrumVisualizer
  - packs a TICPlot and SpectraPlot
  - makes use of the properties
    - clicks in TICPlot shop the spectrum at that RT
      - spectrum at that RT
      - FeatureDataSet of the base peak of selected scan for all raw data files
    - clicks in Spectrum plot show FeatureDataSet of the clicked signal in all raw data files - different than showing XIC 
  - built with properties, automatically updated if e.g. m/z range, m/z tolerance, selected raw data files and so on change.
  - the "TIC" plot can visualise all kinds of chromatograms
    - BPC
    - TIC
    - XIC by setting a specific range - only updated if "update chromatogram" is specifically pressed
- RawDataOverview
  - raw file is visualised by ChromatogramAndSpectrumVisualizer
  - Also shows a table view of the scan data
  - Table view is only loaded if the specific tab of the raw data is selected
  - selection in the table causes the SpectraPlot to show the single scan of that file, TICplot is not updated (maybe change the marker?)
- Main window
  - new layout, tab pane for raw files and feature lists
  - visualisation can be added as a new tab in the main tab pane via `MZmineCore.getDesktop().addTab(Tab)`
  - MZmineTab is an abstract class that describes what should happen if raw file/feature list selection changes
    - only called by the main window if the tab is selected
    - updating can be enabled or disabled by the checkbox in the tab name
  - in the future every visualisation module should implement MZmineTab so they can be easily added and updated by listeners without contributors needing to implement their own listeners
  - generally the new gui is more single-window-focused than the old one, however it should be easy to implement the option to open MZmineTabs in new windows

tomas-pluskal:
Note: there is some merge conflict in src/main/java/io/github/mzmine/parameters/dialogs/ParameterSetupDialog.java

SteffenHeu:
I wanted to rebase the whole branch on top of the master after review, so i didn't merge it with the master branch.

tomas-pluskal:
Okay. Here are a few comments:
- Resizing the main window to smaller size breaks the task controller - there should be some minimum size.
- Selecting a scan in the table moves the scan "pointer" (vertical line) in the TIC plot, but it doesn't update the shaded XIC area in the same way as when I click in the TIC plot using a mouse. I thought it would be nice to have it updated as well based on what is selected in the table.
- You could add a bit more transparency to the shaded XIC area
- The XIC option is only possible for TIC plot but not for BPC plot. It should be possible for both, there is no reason to restrict it only to TIC plot.
- Perhaps you can remove the "Report" tab from this PR and let's just focus on the Raw data overview. Additional functions and tabs can be added in future PRs.
- Is there a way to manually define the color of each raw data file? I could not find it.
- When I click on Scan # column, the values are sorted alphabetically, not numerically (1, 10, 100, 1000, 1001,...)
- Currently aligned feature lists are still added to the Feature lists tab, not to the Aligned feature lists tab
- Can we add keyboard shortcuts for left arrow and right arrow to circle through scans (previous scan, next scan)?

SteffenHeu:
Progress checklist: 
[x] Resizing the main window to smaller size breaks the task controller - there should be some minimum size.
[x] Selecting a scan in the table moves the scan "pointer" (vertical line) in the TIC plot, but it doesn't update the shaded XIC area in the same way as when I click in the TIC plot using a mouse. I thought it would be nice to have it updated as well based on what is selected in the table.
[] You could add a bit more transparency to the shaded XIC 
[] The XIC option is only possible for TIC plot but not for BPC plot. It should be possible for both, there is no reason to restrict it only to TIC 
[] Perhaps you can remove the "Report" tab from this PR and let's just focus on the Raw data overview. Additional functions and tabs can be added in future 
[] Is there a way to manually define the color of each raw data file? I could not find 
[] When I click on Scan # column, the values are sorted alphabetically, not numerically (1, 10, 100, 1000,
[] Currently aligned feature lists are still added to the Feature lists tab, not to the Aligned feature lists 
[] Can we add keyboard shortcuts for left arrow and right arrow to circle through scans (previous scan, next scan)?

src/main/java/io/github/mzmine/gui/MZmineGUI.java
```diff
@@ -111,8 +112,8 @@ public void start(Stage stage) {
     }
 
     stage.setTitle("MZmine " + MZmineCore.getMZmineVersion());
-    stage.setMinWidth(600);
-    stage.setMinHeight(400);
+    stage.setMinWidth(1200);
```
tomas-pluskal, line 115: I think this is way to big for **minimum** width and height. It's fine to set these as default values, but not minimum values. It would cause problems to people who work on laptops with lower screen resolution.

src/main/java/io/github/mzmine/gui/mainwindow/MainWindowControllerOld.java
```diff
@@ -0,0 +1,431 @@
...
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.BorderPane;
+import javafx.scene.layout.StackPane;
+import javafx.scene.paint.Color;
+import javafx.util.Duration;
+
+/**
+ * This class controls the main window of the application
+ */
+public class MainWindowControllerOld {
```
tomas-pluskal, line 73: I guess we can remove this class, or is it used for anything?

src/main/java/io/github/mzmine/gui/mainwindow/MainWindowOld.fxml
```diff
@@ -0,0 +1,186 @@
...
+<?import javafx.scene.control.TableView?>
+<?import javafx.scene.control.Tooltip?>
+<?import javafx.scene.image.Image?>
+<?import javafx.scene.image.ImageView?>
+<?import javafx.scene.layout.BorderPane?>
+<?import javafx.scene.layout.StackPane?>
+<?import javafx.scene.Scene?>
+<?import org.controlsfx.control.StatusBar?>
+
+<Scene fx:id="mainScene" xmlns="http://javafx.com/javafx/8"
```
tomas-pluskal, line 22: We can remove this file as well

tomas-pluskal:
The PR looks good to me, I added just a few minor comments.

src/main/java/io/github/mzmine/gui/MZmineGUI.java
SteffenHeu: How low would like the minimum width to be? It is currently limited by the minimum sizes of the plots + the raw list view which is about 200 (files) + 300 (tic plot) + the spectrum plot which gets cut off since it has no minimum set. 
Are there really laptops with less than 1200 x 800 px that can handle mzmine?

tomas-pluskal: People can always run into edge cases (e.g., remote access). I have no problem with setting the default size to 1200, but the software should be still usable when the available screen size is small, even if some functions (e.g. the spectrum plot) cannot be used. I would leave the previous minimum values.

tomas-pluskal:
Thanks Steffen, let's merge this now and deal with bugs as they appear :)

PR #84, opened 2020-06-28
Added auto sizing to ParameterSetupDialogWithScanPreview.java 

SteffenHeu:
Hi Tomas, 
this automatically sizes the ParameterSetupDialogWithScanPreview. It resizes a little bit more if you move the dialog afterwards, but it is still better than before.

Best
Steffen


tomas-pluskal:
Thanks, Steffen!

PR #85, opened 2020-06-28
Fix parameter component listeners

SteffenHeu:
I rudimentary reimplemented the addListenersToComponent method we used in swing for java fx. To be merged after #84 

Best
Steffen


tomas-pluskal:
Looks good to me :)

PR #87, opened 2020-07-13
Double Parameter Default value

SurajViitk:
Double parameter default value is here given priority over loading values from xml.

### DIFF
--- a/src/main/java/io/github/mzmine/parameters/parametertypes/DoubleComponent.java
+++ b/src/main/java/io/github/mzmine/parameters/parametertypes/DoubleComponent.java
@@ -30,13 +30,14 @@ public class DoubleComponent extends FlowPane {
   private final Double maximum;
   private final TextField textField;
 
-  public DoubleComponent(int inputsize, Double minimum, Double maximum, NumberFormat format) {
+  public DoubleComponent(int inputsize, Double minimum, Double maximum, NumberFormat format, Double defvalue) {
     this.minimum = minimum;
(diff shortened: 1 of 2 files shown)


tomas-pluskal:
Hey Suraj, I think this is not correct. The default value should not be given priority. It should only be used in case when there is no value present in the config file. If the config file already contains a value (which is the last used value), that value should be used.

SurajViitk:
I've removed the part where default values was given priority over xml stored values. Now if no value is present in xml, MZMine will use default value as the prompt text for the parameters text fields.

src/main/java/io/github/mzmine/parameters/parametertypes/DoubleComponent.java
```diff
@@ -30,13 +30,14 @@
...
 
-  public DoubleComponent(int inputsize, Double minimum, Double maximum, NumberFormat format) {
+  public DoubleComponent(int inputsize, Double minimum, Double maximum, NumberFormat format, Double defvalue) {
     this.minimum = minimum;
     this.maximum = maximum;
 
     textField = new TextField();
     textField.setTextFormatter(new TextFormatter<>(new NumberStringConverter(format)));
     textField.setPrefWidth(inputsize);
+    textField.setPromptText(String.valueOf(defvalue));
```
tomas-pluskal, line 40: I think you should just set the value of the textField, not the prompt text. Prompt text will not be treated as a value when the user confirms the dialog.

PR #92, opened 2020-08-22
Implemented some rawDataTree context menu options

roman4750:


src/main/java/io/github/mzmine/gui/mainwindow/MainWindowController.java
```diff
@@ -19,11 +19,14 @@
 package io.github.mzmine.gui.mainwindow;
 
 import com.google.common.collect.Ordering;
+import com.sun.glass.ui.Application;
```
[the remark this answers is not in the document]
roman4750: I`ve removed all unnecessary imports

src/main/java/io/github/mzmine/gui/mainwindow/MainWindowController.java
```diff
@@ -415,17 +428,120 @@ public void handleExportFile(Event event) {
...
+
+    ObservableList<RawDataFile> rows = rawDataTree.getSelectionModel().getSelectedItems();
+
+    // Only one file must be selected
+    if (rows == null || rows.size() != 1 || rows.get(0) == null) {
+      return;
+    }
+
+    // Creating new popup window
+    Stage popup = new Stage();
```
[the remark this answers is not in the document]
roman4750: Changed, should look better now

src/main/java/io/github/mzmine/gui/mainwindow/MainWindowController.java
```diff
@@ -163,8 +177,18 @@ public void initialize() {
 
     rawDataTree.getSelectionModel().setSelectionMode(SelectionMode.MULTIPLE);
     // rawDataTree.setShowRoot(true);
+    rawDataTree.setEditable(false);
```
tomas-pluskal, line 180: Why not set the list to editable mode from the beginning - setEditable(true)? That way the user can also edit the name using a mouse click.

src/main/java/io/github/mzmine/gui/mainwindow/MainWindowController.java
```diff
@@ -178,6 +202,51 @@ protected void updateItem(RawDataFile item, boolean empty) {
...
+      public void cancelEdit() {
+        if(getItem() == null) {
+          return;
+        }
+
+        super.cancelEdit();
+
+        setText(getItem().getName());
+        setGraphic(new ImageView(rawDataFileIcon));
+        rawDataTree.setEditable(false);
```
tomas-pluskal, line 233: See above

src/main/java/io/github/mzmine/gui/mainwindow/MainWindowController.java
```diff
@@ -178,6 +202,51 @@ protected void updateItem(RawDataFile item, boolean empty) {
...
+          return;
+        }
+
+        super.commitEdit(item);
+
+        item.setName(textField.getText());
+
+        setText(item.getName());
+        setGraphic(new ImageView(rawDataFileIcon));
+        rawDataTree.setEditable(false);
```
tomas-pluskal, line 248: See above

src/main/java/io/github/mzmine/gui/mainwindow/MainWindowController.java
```diff
@@ -415,17 +484,79 @@ public void handleExportFile(Event event) {
...
+      return;
+    }
+
+    ObservableList<RawDataFile> rows = rawDataTree.getSelectionModel().getSelectedItems();
+    // Only one file must be selected
+    if (rows == null || rows.size() != 1) {
+      return;
+    }
+
+    rawDataTree.setEditable(true);
```
tomas-pluskal, line 497: See above

src/main/java/io/github/mzmine/gui/mainwindow/MainWindowController.java
roman4750: I did it by design, because double click event is used to show chromatogram.

tomas-pluskal: I think the editing mode is not started with a double click, but with click-and-wait.

roman4750: What should be a desirable behaviour? If setEditable is always true, then renaming of the file and showing chromatogram appear at the same time when user clicks on the row.

tomas-pluskal: Desirable behavior:
Double click = open chromatogram window
Single click and wait = edit name

roman4750: Now the behaviour should be correct, but i still use rawDataTree.setEditable() method to avoid collision between showing chromatogram and going to edit mode, because I think, that other solutions would be overcomplicated.

tomas-pluskal:
Looks great, thanks!

PR #95, opened 2020-09-04
Visualization to tabs

roman4750:


src/main/java/io/github/mzmine/modules/visualization/fx3d/Fx3DTabController.java
```diff
@@ -208,8 +215,20 @@ public void initialize() {
...
 
+  public void loadController() {
+    FXMLLoader loader = new FXMLLoader((getClass().getResource("Fx3DTab.fxml")));
+    loader.setRoot(this);
+    loader.setController(this);
+    try {
+      loader.load();
+      logger.finest("Tab has been successfully loaded from the FXML loader.");
+    } catch (Exception e) {
+      e.printStackTrace();
```
SteffenHeu, line 229: please use logger.log instead

src/main/java/io/github/mzmine/modules/visualization/fx3d/Fx3DTabController.java
```diff
@@ -67,19 +71,18 @@
...
 import javafx.scene.transform.Translate;
-import javafx.stage.Stage;
 import javafx.util.Duration;
 
 /**
  * @author akshaj The controller class of the Fx3DVisualizer which handles all user actions and
  *         shows the plot along with the table.
  */
-public class Fx3DStageController {
+public class Fx3DTabController extends MZmineTab {
```
SteffenHeu, line 80: @Tomas do you think having the class act as controller and MZmineTab works in the long term? I vaguely remember reading something somewhere about i cannot recall what it was. Having a separate Tab class would keep the code cleaner and more uniform, as that's what we did for the other classes. Furthermore, the ability to simply create a new instance of the tab with the constructor is not possible atm, you have to call the loadController method to load the fxml as well

src/main/java/io/github/mzmine/modules/visualization/fx3d/Fx3DTabController.java
```diff
@@ -208,8 +215,20 @@ public void initialize() {
     scene3D.setPickOnBounds(true);
 
     // Add the Windows menu
-    WindowsMenu.addWindowsMenu(scene);
+    //WindowsMenu.addWindowsMenu(scene);
+  }
 
+  public void loadController() {
```
SteffenHeu, line 221: i think "loadController" is misleading, because this method loads the fxml and sets itself as controller (and root). If we keep the layout as is, "loadFXML" or so might be more appropriate.

src/main/java/io/github/mzmine/modules/visualization/fx3d/Fx3DTabController.java
tomas-pluskal: Can do both, I personally like to print out the stack traces because they contain more info for debugging

src/main/java/io/github/mzmine/modules/visualization/fx3d/Fx3DTabController.java
tomas-pluskal: Yeah, I agree, the controller should be a separate class from the Tab class.

roman4750:
I've divided tab class from the controller class and fixed mistakes.

src/main/java/io/github/mzmine/modules/visualization/fx3d/Fx3DVisualizerTab.java
```diff
@@ -0,0 +1,82 @@
+package io.github.mzmine.modules.visualization.fx3d;
```
tomas-pluskal, line 1: Please add the license header to all new .java files (copy & paste from any other file)

src/main/java/io/github/mzmine/modules/visualization/twod/TwoDVisualizerTab.java
```diff
@@ -198,4 +210,38 @@ void updateTitle() {
...
+
+  @Nonnull
+  @Override
+  public Collection<? extends ModularFeatureList> getAlignedFeatureLists() {
+    return Collections.emptyList();
+  }
+
+  @Override
+  public void onRawDataFileSelectionChanged(Collection<? extends RawDataFile> rawDataFiles) {
+    // TODO
```
tomas-pluskal, line 234: Can this be implemented easily?

roman4750: I tried to implement it as it's done for example in chromatogram or neutralloss modules, but there is no class implementing default renderer for the plot in the twod, that's why I've got problems with changing datasets and their renderers. I decided to finish this after I'll port other Stages to MZmineTabs. Maybe by that time I'll come up with a solution without implementation of the renderer or implement it if it will be required.

tomas-pluskal: Sounds good, thanks.

PR #102, opened 2020-10-02
Modular feature list

roman4750:
I am working on the Modular feature list and parts related to it.


roman4750:
I've started port from PeakList to ModularFeatureList. State of the MZmine according to my latest commit is such, that only modular ADAP chromatogram builder and modular isotope groupper modules are working(upon condition, that almost everything else depending on the PeakList is hided from the compiler).

SteffenHeu:
One thing that i found difficult to do, when i was working on the modular feature table was the visibility of column types. Ideally, the user should be able to set the visibility for each DataType as RowType and FeatureType. (In the regular menu you would have to set it for every raw data file, so every row type column and it does not get saved) 
However, since some columns have sub columns that are not listed as DataType column, such as the area charts, we cannot access them easily. One solution would be to hardcode it, but i would consider that kinda against the point, as the whole data model was constructed around being modular. 
I'm talking about this bit of code here:
https://github.com/roman4750/mzmine3/blob/88f5f91e3d3b6cf0419315a1232148054b149a96/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableFX.java#L304

Maybe adding a GraphicalType to the data model, that does not inherit values but just the column in the feature table might be an elegant solution? What do you think, @robinschmid, @tomas-pluskal, @roman4750?

roman4750:
> 
> 
> One thing that i found difficult to do, when i was working on the modular feature table was the visibility of column types. Ideally, the user should be able to set the visibility for each DataType as RowType and FeatureType. (In the regular menu you would have to set it for every raw data file, so every row type column and it does not get saved)
> However, since some columns have sub columns that are not listed as DataType column, such as the area charts, we cannot access them easily. One solution would be to hardcode it, but i would consider that kinda against the point, as the whole data model was constructed around being modular.
> I'm talking about this bit of code here:
> https://github.com/roman4750/mzmine3/blob/88f5f91e3d3b6cf0419315a1232148054b149a96/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableFX.java#L304
> 
> Maybe adding a GraphicalType to the data model, that does not inherit values but just the column in the feature table might be an elegant solution? What do you think, @robinschmid, @tomas-pluskal, @roman4750?

Thanks for the pointing this out. I think, that such problem is caused by the fact, that Area Bars, Area Share and Shapes are not FeatureTypes nor RowTypes and we need to solve this on a bit deeper level. In my opinion we could solve this by adding a new map in FeatureList, that would contain columns, that are related to all the RawDataFiles in FeatureType and are not RowTypes. Then we can create new DataType for each column(Area Bars, Area Share and Shapes) and add these columns in FeatureTableFx addColumn method simmilar way as FeatureTypes are added. On the one hand such solution is a bit overcomplicated, but on the other hand it doesn`t break the consistancy of DataTypes concept and makes mentioned columns easily accessible. What do you think about that, @robinschmid, @SteffenHeu?

robinschmid:
That's a good question.
Maybe we could even just extract the graphical representation of the features and turn them into individual RowTypes - in the end, they are very similar to the average mz and other columns in the RowType map.

roman4750:
> Maybe we could even just extract the graphical representation of the features and turn them into individual RowTypes - in the end, they are very similar to the average mz and other columns in the RowType map.

I think, it is a good idea.

src/main/java/io/github/mzmine/datamodel/data/FeatureListRow.java
```diff
@@ -215,9 +216,9 @@
   public void setID(int id);
 
   @Nullable
-  PeakList getPeakList();
+  ModularFeatureList getFeatureList();
 
-  void setPeakList(@Nonnull PeakList peakList);
+  void setFeatureList(@Nonnull ModularFeatureList flist);
```
SteffenHeu: Isn't it a bit inconsistent to require a ModularFeatureList here instead of a feature list? 
Previously, we were using the interfaces instead of the implementing classes in MZmine. I know it is more convenient to require ModularFeatureList, because we now use it as a data model. However, some of these implementations are inconsistent now, because sometimes you require or return the implementing class and sometimes the interface.

src/main/java/io/github/mzmine/datamodel/data/FeatureNew.java
```diff
@@ -201,8 +201,8 @@ default Integer getParentChromatogramRowID() {
   }
 
   @Nullable
-  PeakList getPeakList();
+  FeatureList getFeatureList();
 
-  void setPeakList(@Nonnull PeakList peakList);
+  void setFeatureList(@Nonnull FeatureList featureList);
```
SteffenHeu: In contrast to what is done in FeatureListRow (see my comment https://github.com/mzmine/mzmine3/pull/102/commits/7c32d42b7dd57f90f07c1d85fa81529376894770#r511078555), here the interface is used as a parameter.

src/main/java/io/github/mzmine/datamodel/data/FeatureListRow.java
SteffenHeu: Nevermind my comment, if that has already been fixed ;)

roman4750: My commits are messy, sorry for that:)

src/main/java/io/github/mzmine/datamodel/data/ModularFeatureList.java
```diff
@@ -59,13 +83,49 @@ public ModularFeatureList(String name, @Nonnull List<RawDataFile> dataFiles) {
     descriptionOfAppliedTasks = new ArrayList<>();
     dateCreated = DATA_FORMAT.format(new Date());
 
+    // Type columns will be created in the same sequence as they are initialized
```
SteffenHeu, line 86: Is it necessary to add all column types individually? Shouldn't the types be added based on the present DataTypes in a ModularFeature when it is added to the feature list? @robinschmid

robinschmid: The idea is to add the RowTypes and FeatureTypes before entering data to define the columns in the feature list. 
We should not add the types right here in the ModularFeatureList class but rather in the modules that fill in the data.

The minimal feature list I can think of only consists of MZType and IntensityType (e.g., generated for direct injection analysis). Therefore, we should not add types presumptively.

roman4750:
I've ported crucial classes dependend on the FeatureList and the class itself to the modular version and updated feature table. It is not done completely yet, but, I think, that it doesn't make sense to polish it, until it is not merged:). As far as non-modular classes might not be used in the future and not everything is done, the biggest part of modules and some classes(see screenshot below) must be hidden from the compiler in order to run mzmine.
зображення

tomas-pluskal:
Please merge the current master branch to fix the merge conflicts.
It seems some classes are missing, maybe you need to add them? E.g. ArithmeticUtils, ExpandableType...

tomas-pluskal:
Sorry, please ignore my previous comment about missing classes - my Eclipse just needed to refresh the source folder. Still need to merge the master, though.

SteffenHeu:
> 
> 
> I've ported crucial classes dependend on the FeatureList and the class itself to the modular version and updated feature table. It is not done completely yet, but, I think, that it doesn't make sense to polish it, until it is not merged:). As far as non-modular classes might not be used in the future and not everything is done, the biggest part of modules and some classes(see screenshot below) must be hidden from the compiler in order to run mzmine.
> зображення

Can you elaborate, why it is necessary to exclude these classes from the compiler currently? Is it just a matter of porting them to the modular classes in the future?

roman4750:
> Can you elaborate, why it is necessary to exclude these classes from the compiler currently? Is it just a matter of porting them to the modular classes in the future?
Yes, it's just a matter of porting them to the modular classes. But today we discussed with Tomáš, that I'll port them and other modules before merging pull request.

tomas-pluskal:
Please change MZmineModuleCategory.FEATUREPICKING to MZmineModuleCategory.FEATUREDETECTION. Thanks!

roman4750:
> Please change MZmineModuleCategory.FEATUREPICKING to MZmineModuleCategory.FEATUREDETECTION. Thanks!

Ok.

tomas-pluskal:
Hey Roman, please rename TICPlotType.BASEFEATURE back to TICPlotType.BASEPEAK. That's because "base peak plot" is a specific term, it is not related to feature detection.

roman4750:
> Hey Roman, please rename TICPlotType.BASEFEATURE back to TICPlotType.BASEPEAK. That's because "base peak plot" is a specific term, it is not related to feature detection.

Ok, sure. Are there any other occurences of "peak", where it can not be replaced with "feature"?

tomas-pluskal:
> > Hey Roman, please rename TICPlotType.BASEFEATURE back to TICPlotType.BASEPEAK. That's because "base peak plot" is a specific term, it is not related to feature detection.
> 
> Ok, sure. Are there any other occurences of "peak", where it can not be replaced with "feature"?

Well, that's hard to generalize. "Feature" should be something that is the result of feature detection. "Peak" is typically a local maximum in 2-dimensional data. "Base peak" is a specific term that means the highest signal in the spectrum (= 2-dimensional data).

roman4750:
> Well, that's hard to generalize. "Feature" should be something that is the result of feature detection. "Peak" is typically a local maximum in 2-dimensional data. "Base peak" is a specific term that means the highest signal in the spectrum (= 2-dimensional data).

Ok, thanks.

PR #109, opened 2020-11-30
Simplexylinechart

SteffenHeu:
This is supposed to be a more generec plot class for XY plots, so we can easily create a new chart of a dataset we develop. It's stil work in progress, but already used for the MobilogramVisualizer in #108. 
I will probably add additional functionality in the process, but this can give you a rough idea, how it works.

Briefly, any kind of class can implement the interfaces. 
**PlotDatasetProvider** implements all necessary methods from other sub interfaces, but is the most practical if range and domain values and series keys are provided by the same class.
The interface's methods can then be used to create a **ColoredXYDataset** or the class can be passed to the plot directly since it implements PlotDatasetProvider.

Label and tooltip generators don't work yet.

Of course, the ui controls are dataaset dependant so they still have to be implemented individually. 

Basic demo of the plot here:
[MZmine 0.0 2020-12-01 00-07-53.zip](https://github.com/mzmine/mzmine3/files/5619400/MZmine.0.0.2020-12-01.00-07-53.zip)

All i had to do was implement **PlotDatasetProvider** in my Mobilogram interface here: https://github.com/SteffenHeu/mzmine3/blob/66319f8d8ab9f889f460bb29b64d38cdfde93a8c/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_mobilogrambuilder/Mobilogram.java
The getDomainValue and getRangeValue methods then just call the mobility and intensity methos in the mobilogram interface.

Let me know,
 what you guys think


tomas-pluskal:
What would be nice to have is a unified component to control layers (datasets) in the plot. Take a look at the 3D visualizer bottom part. Do you think such dataset table could be included in this Simplexylinechart component? That would be a great step forward towards unifying the visualizers.

Screen Shot 2020-12-02 at 10 06 02

SteffenHeu:
Yeah, i think so.
At first, I will try to get the IMS feature picking running, though. This class was just a byproduct of the process because it is really annoying that we have to create a new plot and dataset class for everything atm.

Did you have time to check the way it's implemented atm? Do you think that's feasible for every plot? 
We'd have to work something out to be able to use datasets that still have to be calculated, like the ticdataset atm.

src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/template/providers/PlotDatasetProvider.java
```diff
@@ -0,0 +1,24 @@
...
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with MZmine; if not,
+ * write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301
+ * USA
+ */
+
+package io.github.mzmine.gui.chartbasics.gui.javafx.template.providers;
+
+public interface PlotDatasetProvider extends DomainValueProvider<Number>,
```
tomas-pluskal, line 21: Perhaps this should be named PlotDataProvider, because it provides values, not datasets

tomas-pluskal:
Overall it looks fine, but the approach goes against what we discussed on Slack. We discussed that objects like features, chromatograms (mobilograms?) should primarily be stored on disk to reduce the memory usage. However, JFreeChart needs to have the data in memory for fast displaying and zooming. That's why the common workflow is to load raw data (e.g. DataPoint[] from StorableScans) from disk into memory -> TICDataSet. By creating the datasetprovider, you are essentially saying that the objects (like Scan, Feature, Mobilogram) should directly provide data to the charts, but that will be too slow if the data is stored on disk.

SteffenHeu:
You're right. I think we could add a preloadData() method to the interfaces, that will load the data into the ram and refer to that data with the interface methods.
That should work, or am i missing something?

tomas-pluskal:
Well, it would work, but you are essentially slowly reinventing the AbstractXYDataset :) 
You also need the listeners in AbstractXYDataset, to let the plot know when the data is updated.

tomas-pluskal:
Also, the AbstractXYDataset can be collected by the garbage collector when the plot is closed, but if you use the preloadData() method then there is no clear mechanism to remove the data again - if you add a forgetData() method then you always need to be sure you call it when the plot is closed.... sounds like a nightmare

SteffenHeu:
Ok, then we have to be really clever or just use this plot for visualisation of precalculated datasets.

Maybe a dataset factory could be possible that precalculates a dataset from an object that implements the PlotDatasetProvier interface.

`ColoredXYDataset dataset = DatasetFactory(PlotDatasetProvider source);`

The dataset could be instantiated right away and the factory starts a thread to load the data. Since the data for the dataset is then stored in a new object, the garbage collector should be able to collect, or?

tomas-pluskal:
Yes, that is possible.. or just `ColoredXYDataset dataset = new ColoredXYDataset dataset(PlotDataProvider source)`

tomas-pluskal:
Hey Steffen, since the templates and providers are meant to be used by all other modules, it would be good to add brief Javadoc descriptions to the classes describing the purpose and usage. It's not very clear from the name like ValueCountProvider what that class actually does and how it should be used.

Second comment: the package name io.github.mzmine.gui.chartbasics.gui.javafx.template.providers seems a bit deep (e.g. contains gui twice), how about moving it somewhere like io.github.mzmine.gui.chartbasics.providers?

SteffenHeu:
Should i move the whole package including the chart class then?

tomas-pluskal:
Yes, please

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/DatasetsChangedListener.java
```diff
@@ -0,0 +1,10 @@
+package io.github.mzmine.gui.chartbasics.simplechart;
```
tomas-pluskal, line 1: Something missing here? :)

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/DatasetControlPane.java
```diff
@@ -0,0 +1,138 @@
...
+        new Callback<CellDataFeatures<XYDataset, Color>, ObservableValue<Color>>() {
+          @Override
+          public ObservableValue<Color> call(CellDataFeatures<XYDataset, Color> param) {
+            if (param.getValue() instanceof ColorPropertyProvider) {
+              return ((ColoredXYDataset) param.getValue()).fxColorProperty();
+            }
+            return null;
+          }
+        });
+    colColor.setOnEditCommit(new EventHandler<CellEditEvent<XYDataset, Color>>() {
```
tomas-pluskal, line 95: I think you can use the lambda notation here, like you use it in other places:
````
colColor.setOnEditCommit(event -> {
        ((ColorPropertyProvider) event.getRowValue()).fxColorProperty().set(event.getNewValue());
        chart.getChart().fireChartChanged();
      }
    );
````
Or not?

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/DatasetsChangedListener.java
SteffenHeu: Perhaps :D

tomas-pluskal:
Looks good, thanks!

PR #111, opened 2020-12-02
fixed EChartViewer export to clipboard if first dataset was removed

SteffenHeu:
fixed EChartViewer export to clipboard if first dataset was removed
ScanSelection from List<Scan> to List<? extends Scan> so frames can actually be processed


tomas-pluskal:
Thanks!

PR #115, opened 2020-12-17
added raw data import changes from other branch

SteffenHeu:
NOTE: Empty scans are created even if they don't contain any data points, since this might fix a problem for @Annexhc . This will consume ~1/3 more ram, so we have to find something clever in the future.


robinschmid:
Seems fine - merged

PR #116, opened 2020-12-17
List to set/collection for IMSRawDataFile and Frame

SteffenHeu:
to be merged after #115

should improve performance

### DIFF
--- a/src/main/java/io/github/mzmine/datamodel/Frame.java
+++ b/src/main/java/io/github/mzmine/datamodel/Frame.java
@@ -20,6 +20,7 @@ package io.github.mzmine.datamodel;
 
 import com.google.common.collect.Range;
 import java.util.List;
+import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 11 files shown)


src/main/java/io/github/mzmine/datamodel/Frame.java
```diff
@@ -35,13 +36,22 @@
   @Nonnull
   public MobilityType getMobilityType();
 
-  public List<Integer> getMobilityScanNumbers();
+  /**
+   *
+   * @return Unsorted set of sub spectrum numbers.
+   */
+  public Set<Integer> getMobilityScanNumbers();
```
robinschmid, line 43: in the new version this is going to be changed to an array of MobilityScans right?

robinschmid:
sounds good I guess

src/main/java/io/github/mzmine/datamodel/Frame.java
SteffenHeu: Since the mobilityscans are currently stored in a Map<Integer, Scan> returning as a set seems the most practical. If we allocated an array it would be inefficient.

SteffenHeu: Currently the scans are stored in a Map<Integer, Scan>, so creating an array would be inefficient, since we can just return map.keySet(). 
And since every frame will store each subspectrum in the future, we will probably have a map there as well, because otherwise we would have to seach an array of Spetra for the specific scan number.

robinschmid: So the main point is we are going to change it in the future. Merging it now and then we change

PR #118, opened 2020-12-18
Datamodel change

SteffenHeu:
- Removed getFragmentScanNumbers() from Scan
- Precursor information for ion mobility stored in ImsMsMsInfo for Frames with mslevel = 2
- IMSRawDataFile.getScan returns the frames (Bit of a dirty hack atm, can be fixed in the future)
  - old modules should be able to work with these average spectra
  - getFrame[..] methods have not been removed!
- MobilityMassSpectrum added
  - StorableMobilityMassSpectrum only stores its frame (ptr), highest data point (ptr + 2 doubles), number (int), storageId(int) in ram and caches a Range<Double> if it was accessed once.
    - the numDp are accessed via this hack: `return ((IMSRawDataFileImpl)getFrame().getDataFile()).getDataPointsLengths().get(storageId);`
    - RT accessed via frame
    - MobilitySpectrum number <-> mobility is stored once in the IMSRawDataFile per segment => millions of doubles less


tomas-pluskal:
Just a cosmetic detail: I think MobilityMassSpectrum should be called MobilityScan, because it represents the result of instrument scanning the mass range.

SteffenHeu:
I didn't want to use the word 'scan' because it could imply it implemented 'Scan'. So should i call it MobilityScan?

I want to add one more method before merging, though.

tomas-pluskal:
￼I think `MobilityScan implements MassSpectrum` is fine.

Annexhc:
> Just a cosmetic detail: I think MobilityMassSpectrum should be called MobilityScan, because it represents the result of instrument scanning the mass range.

I am very much in favor of MobilityScan!

SteffenHeu:
@robinschmid what do you think about the MobilityScan <-> MobilityMassSpectrum situation?

SteffenHeu:
Note: this breaks project loading for ims raw data files since mobility scans are not stored. However, this does not limit us from proceeding

SteffenHeu:
I will have to add masslist + detection support. Might take another hour or so

robinschmid:
About the masslist support etc
Maybe the best option to add broad support to access the scans is to change the ScanSelectionParameter
We typically access the scans from there with getMatchingScans or something. You can then automatically check if we are looking at Frames and select all sub scans

SteffenHeu:
I will add getMatchingScan methods for mobility spectra (have that in some branch already), but returning all subscans when a frame is inserted will lead to non-ims modules to retrieve the mobility sub-scans, which is not a good idea in my opinion, because it will break almost everything.

SteffenHeu:
seems to work.

SteffenHeu:
Ansgar and i would need a review + merge so we can continue properly, please.

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/MassDetectionParameters.java
```diff
@@ -130,6 +138,18 @@ public ExitCode showSetupDialog(boolean valueCheckRequired) {
       MZmineCore.getDesktop().displayMessage(null, msg);
     }
 
+    RawDataFile[] files = dataFiles.getValue().getMatchingRawDataFiles();
+    Optional<RawDataFile> opt = Arrays.stream(files)
+        .filter(file -> (file instanceof IMSRawDataFile)).findAny();
+    if (opt.isPresent() && !massDetectorName.startsWith("Centroid")) {
+      String msg = "MZmine thinks you are running a profile mass detector on an Ion mobility raw "
+          + "data file. Only the centroid mass detector supports mobility scan peak detection."
+          + " Do you want to continue any way?";
```
tomas-pluskal, line 147: "any way" -> "anyway"

src/main/java/io/github/mzmine/modules/dataprocessing/filter_scanfilters/mean/MeanFilter.java
```diff
@@ -92,7 +92,7 @@ public Scan filterScan(Scan sc, ParameterSet parameters) {
     // Create filtered scan
     Scan newScan = new SimpleScan(sc.getDataFile(), sc.getScanNumber(), sc.getMSLevel(),
         sc.getRetentionTime(), sc.getPrecursorMZ(), sc.getPrecursorCharge(),
-        sc.getFragmentScanNumbers(), newDataPoints, MassSpectrumType.CENTROIDED, sc.getPolarity(),
+        /*sc.getFragmentScanNumbers(),*/ newDataPoints, MassSpectrumType.CENTROIDED, sc.getPolarity(),
```
tomas-pluskal, line 95: I think you can remove the getFragmentScanNumbers() completely, instead of commenting out. MZmine 3 is a fresh start, so we don't need to keep the history.

src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/RawDataFileOpenHandler_3_0.java
```diff
@@ -327,9 +327,9 @@ public void endElement(String namespaceURI, String sName, String qName) throws S
 
     if (qName.equals(RawDataElementName_3_0.FRAME.getElementName())) {
 
-      final StorableFrame storableScan = new StorableFrame(newRawDataFile, currentStorageID,
+      /*final StorableFrame storableScan = new StorableFrame(newRawDataFile, currentStorageID,
```
tomas-pluskal, line 330: You're not using the StorableFrame anymore?

SteffenHeu: Due to the change in data model, the project import for ims is broken, and will need to be fixed in the future. Currently constructing modules for processing is more important to continue than having a project import (imo). But i will fix the project load later on. 

Or maybe we should switch to a new one, directly? What do you think?

tomas-pluskal:
It’s fine for now!
Overall the PR looks good to me.

PR #122, opened 2020-12-20
data type independent smoothing

SteffenHeu:
@Annexhc might also be needed for deconvolution


tomas-pluskal:
Is this work in progress? It seems the `convolve()` function is not used anywhere.

SteffenHeu:
I'm using this method in my mobilogram deconvolution (different branch).
And i will probably use it to deconvolve the Ion mobility traces from Ansgar.
So it would be nice, if you could merge it. Or i just copy it into the new Ion mobility traces deconvolution branch.

tomas-pluskal:
Ok no problem!

PR #131, opened 2020-12-28
Auto add row bindings

robinschmid:
This PR adds automatic row binding.

All feature types know their default row binding and those are added automatically to the featurelist

Add MZType to a feature will automatically add MZType to the RowTypes and add the binding

### DIFF
--- a/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
+++ b/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
@@ -18,6 +18,19 @@
 
 package io.github.mzmine.datamodel.features;
 
+import com.google.common.collect.Range;
+import com.microsoft.schemas.office.visio.x2012.main.RowType;
+import io.github.mzmine.datamodel.FeatureIdentity;
+import io.github.mzmine.datamodel.FeatureInformation;
+import io.github.mzmine.datamodel.IsotopePattern;
+import io.github.mzmine.datamodel.Scan;
+import io.github.mzmine.datamodel.features.types.*;
+import io.github.mzmine.datamodel.features.types.exceptions.TypeColumnUndefinedException;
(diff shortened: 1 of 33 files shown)


src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
```diff
@@ -58,6 +45,14 @@
...
 import io.github.mzmine.util.SortingProperty;
 import io.github.mzmine.util.spectraldb.entry.SpectralDBFeatureIdentity;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Stream;
```
SteffenHeu, line 55: @tomas-pluskal can you check the eclipse/intellij code formatters? I feel like Intellij puts java.* on the bottom and eclipse on top. This leads to a lot of whitespace changes and sometimes even merge conflicts, since @Annexhc uses eclipse.

SteffenHeu (commented):
looks good to me

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
tomas-pluskal: I think the code formatter does not deal with imports, only with the actual code. In Eclipse those are two separate functions - "format code" and "organize imports". So it is not surprising that Eclipse and IDEA do it differently. Whitespace changes are annoying but as long as it is just in the imports, I think we can simply ignore them...

SteffenHeu: Ahh ok, i thought the formatter would also take care of that.

PR #133, opened 2020-12-28
Imsrawdataoverview

SteffenHeu:
still work in progress
grafik


SteffenHeu:
grafik

I know its an incredibly dirty way to adjust the plots, but it's the only thing i could think of. What do you think?

SteffenHeu:
grafik

There's one small thing i want to fix tomorrow and then i'd say it's done.

SteffenHeu:
From my side it's ready now.

src/main/java/io/github/mzmine/modules/visualization/rawdataoverviewims/Threads.java
```diff
@@ -0,0 +1,210 @@
...
+import io.github.mzmine.util.color.SimpleColorPalette;
+import java.awt.Color;
+import java.text.NumberFormat;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import javafx.application.Platform;
+import javax.annotation.Nonnull;
+
+class Threads {
```
tomas-pluskal, line 42: What is the Threads class for? The individual classes could be defined in their own individual files, no?

src/main/java/io/github/mzmine/modules/visualization/rawdataoverviewims/IMSRawDataOverviewTask.java
```diff
@@ -0,0 +1,59 @@
...
+  public double getFinishedPercentage() {
+    return 0;
+  }
+
+  @Override
+  public void run() {
+    setStatus(TaskStatus.PROCESSING);
+    MZmineTab tab = new IMSRawDataOverviewTab(parameterSet);
+    tab.onRawDataFileSelectionChanged(Set.of(file));
+    MZmineCore.getDesktop().addTab(tab);
```
tomas-pluskal, line 56: This design is a bit strange - it seems like the only purpose of this Task is to add the new tab to the desktop. Tasks should be for data processing, so the tab can just be added immediately in the runModule() method

src/main/java/io/github/mzmine/modules/visualization/rawdataoverviewims/Threads.java
SteffenHeu: Yes they could. I thought it was unnecessary to make three classes for such rudimentary tasks and they are all somehow related. Should i split it up any way?

tomas-pluskal: I think short and simple classes are better for readability. You can move them to a separate subpackage.

tomas-pluskal:
Looks good, thanks!

PR #134, opened 2020-12-30
Memory-mapped storage for raw data

tomas-pluskal:
Work in progress. The MemoryMapStorage class is functional, now I need to modify Scan and other classes to use DoubleBuffers instead of DataPoint[] arrays.


src/main/java/io/github/mzmine/util/javafx/groupablelistview/GroupableListViewEntity.java
```diff
@@ -100,15 +101,16 @@ public boolean isValue() {
   }
 
   public boolean isGroupHeader() {
-    assertVariables(true);
+    // assertVariables(true);
```
roman4750, line 104: I would leave that assert, now, when toString() is not called in assertVariables(), it will not lead to infinite recursive call.

src/main/java/io/github/mzmine/util/MemoryMapStorage.java
```diff
@@ -0,0 +1,285 @@
...
+ */
+public class MemoryMapStorage {
+
+  private final Logger logger = Logger.getLogger(this.getClass().getName());
+
+  /**
+   * One temporary file can store STORAGE_FILE_CAPACITY bytes. We need to fit within 2GB limit for a
+   * single MappedByteBuffer. 1 GB per file seems like a good start.
+   */
+  private static final long STORAGE_FILE_CAPACITY = 1_000_000_000L;
```
robinschmid, line 69: this is important as larger files often break back up systems right?

src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
```diff
@@ -85,15 +82,18 @@
   private final TreeMap<Integer, Integer> dataPointsLengths;
 
   // Temporary file for scan data storage
-  private File dataPointsFileName;
-  private RandomAccessFile dataPointsFile;
+  private File storageFileName;
+  private RandomAccessFile storageFile;
```
robinschmid, line 86: Why is this not part of the MemoryMapStorage?

src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
```diff
@@ -146,19 +151,21 @@ public RandomAccessFile getDataPointsFile() {
...
       throw new IOException("Cannot open another data points file, because one is already open");
     }
 
-    this.dataPointsFileName = dataPointsFileName;
-    this.dataPointsFile = new RandomAccessFile(dataPointsFileName, "rw");
+    this.storageFileName = dataPointsFileName;
+    this.storageFile = new RandomAccessFile(dataPointsFileName, "rw");
+    FileChannel fileChannel = storageFile.getChannel();
+    // this.storageMemoryMap = fileChannel.map(FileChannel.MapMode.READ_WRITE, 0,
+    // Integer.MAX_VALUE);
```
robinschmid, line 162: At this stage we are not using the memory map in RawDataFileImpl?

src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
```diff
@@ -407,12 +414,12 @@ public double getDataMaxTotalIonCurrent(int msLevel) {
 
   public synchronized int storeDataPoints(DataPoint dataPoints[]) throws IOException {
 
-    if (dataPointsFile == null) {
+    if (storageFile == null) {
```
robinschmid, line 417: This still needs to be changed to use the memory map right?

robinschmid:
I think most of my comments are resolved if @tomas-pluskal confirms that RawDataFileImpl still contains a lot of legacy code that needs to be cleaned

src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
robinschmid: I guess its legacy code that needs to be removed

tomas-pluskal:
Yes, you're right, RawDataFileImpl hasn't been cleaned up yet, most of the code for storing DataPoints[] is not used anymore. I'll remove it.

src/main/java/io/github/mzmine/util/MemoryMapStorage.java
tomas-pluskal: That's one reason. The second reason is that MappedByteBuffer has an inherent 2GB limit, I believe.

PR #135, opened 2020-12-30
Main window menu updates

roman4750:


tomas-pluskal:
I would rename ListViewGroups to GroupableListView

src/main/java/io/github/mzmine/gui/mainwindow/MainWindowController.java
```diff
@@ -693,19 +693,19 @@ public void addTab(Tab tab) {
     if (tab instanceof MZmineTab) {
       ((MZmineTab) tab).updateOnSelectionProperty().addListener(((obs, old, val) -> {
         if (val) {
-          if (!((MZmineTab) tab).getRawDataFiles()
+          if (((MZmineTab) tab).getRawDataFiles() != null && !((MZmineTab) tab).getRawDataFiles()
```
SteffenHeu, line 696: Actually, this should never be null, as it should return an empty collection by default. 
Double checking is good, but the exception should not have happened.

roman4750: Yeah, but RawDataOverviewPane getFeatureLists() doesn't override `@Nonnull` from MZmineTab, so it caused NPE:
```
  @Override
  public Collection<? extends FeatureList> getFeatureLists() {
    return null;
  }
```
I'm gonna fix it

tomas-pluskal:
Looks good, thanks!

PR #136, opened 2020-12-31
Fix feature creation for aligned feature lists

robinschmid:
This fixes:
- feature creation in general
- default values for empty DataType properties in ModularFeature.get methods
- chromatogram building, deconvolution, deisotoping, feature list rows filter, feature alignment, gap filling, and GNPS FBMN work now

For now I have removed creation of graphs in the FeatureShapesType - we need to change to JFreeChart and Steffens classes to reuse the datasets

Ready for merge from my side


tomas-pluskal:
Great, thanks!

PR #137, opened 2020-12-31
Null checks for modular feature model

robinschmid:
Another try to add the correct null checks to the helper methods in ModularFeature(Row)

The case:
Properties might be empty (p.getValue()==null) if an undetected modular feature is added as a placeholder
Properties are null (p==null) if the type was not added to the row or feautre


tomas-pluskal:
👍🏻

PR #142, opened 2021-01-03
fixed batch mode list having empty cells

SteffenHeu:
Cells in the batch mode list were empty although there were modules in the queue, this fixes it.


tomas-pluskal:
Thanks Steffen!

PR #143, opened 2021-01-03
Remove scan numbers

robinschmid:
Praise the lord as he has taken the burden of boxing and unboxing scan numbers from us.

This PR removes the use of scan numbers - we use references to scans now.
Tested:
- Raw data import
- project import
- mass detection, chromatogram building, deconvolution

Chromatogram building feels way faster now. Not tested though

This is the memory difference when we run Chromatogram building on 3 files with around 8000 features total: (I will upload a comparison to the old scan numbers mzmine later):
image


robinschmid:
I still need to rename some methods / variables from scanNumbers to scans but I did not want to rename too much to not break your stuff

robinschmid:
Oh I actually have it here:
### Old version with scan numbers
image

### New version without scan numbers --> scans
image

PR #144, opened 2021-01-04
Featuretablecharts

SteffenHeu:
Since i'm adding some stuff to the feature model, please feel free to comment on my changes early, since I don't think I have completely understood the model.

### DIFF
--- a/src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
+++ b/src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
@@ -51,6 +51,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import javafx.beans.property.ListProperty;
 import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.Property;
@@ -58,6 +59,7 @@ import javafx.collections.FXCollections;
 import javafx.collections.MapChangeListener;
 import javafx.collections.ObservableList;
 import javafx.collections.ObservableMap;
+import javafx.scene.Node;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
@@ -72,6 +74,9 @@ public class ModularFeature implements Feature, ModularDataModel {
   private @Nonnull ModularFeatureList flist;
   private final ObservableMap<DataType, Property<?>> map =
       FXCollections.observableMap(new HashMap<>());
+
+  // buffert col charts and nodes
+  private final Map<String, Node> buffertColCharts = new HashMap<>();
 
 
   public ModularFeature(@Nonnull ModularFeatureList flist) {
@@ -206,6 +211,14 @@ public class ModularFeature implements Feature, ModularDataModel {
         set(AsymmetryFactorType.class, af);
       }
     }
+  }
+
+  public Node getBufferedColChart(String colname) {
+    return buffertColCharts.get(colname);
+  }
+
+  public void addBufferedColChart(String colname, Node node) {
+    buffertColCharts.put(colname, node);
   }
 
   @Override
--- a/src/main/java/io/github/mzmine/datamodel/features/types/FeatureShapeMobilogramType.java
+++ b/src/main/java/io/github/mzmine/datamodel/features/types/FeatureShapeMobilogramType.java
@@ -1,16 +1,16 @@
 /*
  * Copyright 2006-2020 The MZmine Development Team
- * 
+ *
  * This file is part of MZmine.
- * 
+ *
  * MZmine is free software; you can redistribute it and/or modify it under the terms of the GNU
  * General Public License as published by the Free Software Foundation; either version 2 of the
  * License, or (at your option) any later version.
- * 
+ *
  * MZmine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
  * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
  * Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU General Public License along with MZmine; if not,
  * write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301
  * USA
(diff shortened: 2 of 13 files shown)


SteffenHeu:
To keep PRs small, you can review & merge this already, if you like.

grafik

robinschmid:
Review mode activated

robinschmid:
Oh man one chart for each feature is going to be quite a load for aligned feature lists with multiple samples.
But lets see - I mean its the users choice what to do with that option and it for sure is very informative

src/main/java/io/github/mzmine/datamodel/features/types/FeatureShapeMobilogramType.java
```diff
@@ -23,47 +23,50 @@
...
 import javafx.scene.Node;
 import javafx.scene.control.TreeTableCell;
 import javafx.scene.control.TreeTableColumn;
 import javafx.scene.layout.StackPane;
 import javax.annotation.Nonnull;
 
-public class FeatureShapeMobilogramType extends DataType<MapProperty<RawDataFile, ModularFeature>>
-    implements GraphicalColumType<MapProperty<RawDataFile, ModularFeature>> {
+public class FeatureShapeMobilogramType extends DataType<BooleanProperty>
+    implements GraphicalColumType<Boolean> {
```
robinschmid, line 42: I would suggest to use the LinkedDataType similar to the FeatureShapeType.
The boolean is just there to have a property to fire up the change listeners. 
Then in DataPointsType.getDefaultRowBindings the FeatureShapeType is automatically added to the ModularFeatureListRow and bound to any changes in the data points type.

This way you do not have to add the FeatureShapeType manually - and if anything changes (we add a feature to the list or change the data points) the chart is updated.

/**
 * Contains no data - listens to changes in all {@link ModularFeature} in a {@link
 * ModularFeatureListRow} On change of any {@link DataPointsType} - the chart is updated
 *
 * @author Robin Schmid (https://github.com/robinschmid)
 */
public class FeatureShapeType extends LinkedDataType
    implements GraphicalColumType<Boolean>

src/main/java/io/github/mzmine/datamodel/features/types/FeatureShapeMobilogramType.java
```diff
@@ -23,47 +23,50 @@
...
-    return "Mobilogram";
+    return "Mobilograms";
   }
 
   @Override
-  public MapProperty<RawDataFile, ModularFeature> createProperty() {
-    return new SimpleMapProperty<RawDataFile, ModularFeature>();
+  public BooleanProperty createProperty() {
+    return new SimpleBooleanProperty();
   }
```
robinschmid, line 53: Remove if you use LinkedDataType

src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
```diff
@@ -208,6 +213,14 @@ public ModularFeature(@Nonnull ModularFeatureList flist, Feature f) {
     }
   }
 
+  public Node getBufferedColChart(String colname) {
+    return buffertColCharts.get(colname);
+  }
+
+  public void addBufferedColChart(String colname, Node node) {
+    buffertColCharts.put(colname, node);
```
robinschmid, line 221: Same for the feature list row. we also need a place to clear the buffered charts to free the memory. Maybe after the last feature list tab is closed

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/DeconvolutionTask.java
```diff
@@ -249,7 +249,7 @@ private FeatureList resolvePeaks(final FeatureList originalFeatureList, RSession
     DataTypeUtils.addDefaultChromatographicTypeColumns(resolvedFeatureList);
     if (originalFeatureList.getRawDataFile(0) instanceof IMSRawDataFile) {
       resolvedFeatureList.addRowType(new FeatureShapeIonMobilityRetentionTimeType());
-      resolvedFeatureList.addRowType(new FeatureShapeIonMobilityRetentionTimeHeatMapType());
+      resolvedFeatureList.addFeatureType(new FeatureShapeIonMobilityRetentionTimeHeatMapType());
       resolvedFeatureList.addRowType(new FeatureShapeMobilogramType());
```
robinschmid, line 253: Those should actually all be removed. 
If your types are based on LinkedDataType you can simply use the DataType that forms the basis of your chart and create a default row binding.
FeatureShapeType for example is only dependent on all the DataPointsTypes from all features. So if we have a feature list without the DataPointsType - no FeatureShapes
If we have DataPointsType for the features we automatically add the FeatureShapes as part of the defaultRowBinding of DataPointsType

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_ionmobilitytracebuilder/IonMobilityTraceBuilderTask.java
```diff
@@ -401,9 +401,8 @@ private void buildModularFeatureList(SortedSet<IIonMobilityTrace> ionMobilityTra
           new ModularFeatureListRow(featureList, featureId, rawDataFile, modular);
       newRow.set(MobilityType.class, ionTrace.getMobility());
       newRow.set(FeatureShapeIonMobilityRetentionTimeType.class, newRow.getFeaturesProperty());
-      newRow.set(FeatureShapeMobilogramType.class, newRow.getFeaturesProperty());
-      newRow.set(FeatureShapeIonMobilityRetentionTimeHeatMapType.class,
-          newRow.getFeaturesProperty());
+      newRow.set(FeatureShapeMobilogramType.class, true);
```
robinschmid, line 404: Unless you actually want to use the value - it does not make any sense to set a shape to true. The example of FeatureShapeType is farely simple

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_ionmobilitytracebuilder/IonMobilityTraceBuilderTask.java
```diff
@@ -388,10 +388,10 @@ private void buildModularFeatureList(SortedSet<IIonMobilityTrace> ionMobilityTra
         new ModularFeatureList(rawDataFile + " " + suffix, rawDataFile);
     // ensure that the default columns are available
     DataTypeUtils.addDefaultChromatographicTypeColumns(featureList);
-    featureList.addFeatureType(new FeatureShapeIonMobilityRetentionTimeType());
+    featureList.addRowType(new FeatureShapeIonMobilityRetentionTimeType());
     featureList.addFeatureType(new FeatureShapeIonMobilityRetentionTimeHeatMapType());
-    featureList.addFeatureType(new FeatureShapeMobilogramType());
-    featureList.addFeatureType(new MobilityType());
+    featureList.addRowType(new FeatureShapeMobilogramType());
```
robinschmid, line 393: Similar to previous comment. You do not need to add this manually if you use the bindings

SteffenHeu:
> 
> 
> Oh man one chart for each feature is going to be quite a load for aligned feature lists with multiple samples.
> But lets see - I mean its the users choice what to do with that option and it for sure is very informative

Yeah, but it does not make sense to have this plot as a row type. Because summing all row values does not make sense since you wouldn't be able to tell any kind of isomers from that plot.
So we should either have it as feature type or not have it.

src/main/java/io/github/mzmine/datamodel/features/types/FeatureShapeMobilogramType.java
SteffenHeu: How do you only add the MobilogramType when the data points are of type RetentionTimeMobilityDataPoint or the raw data file is an IMSRawDataFile?

SteffenHeu:
I added some of your suggestions, but the row binding is only possible for the mobilogram chart and not for the Heatmap because the heatmap would be a feature type, or?

robinschmid:
From my side, this can be merged. We have discussed all the future changes, once the MemoryMapData is here.

tomas-pluskal:
👍🏻

PR #149, opened 2021-01-13
Export modular feature list to CSV

robinschmid:
This exports a modular feature list to a CSV file.
It is completely different to the other CSV export. 

- it exports all DataTypes (except for the tag: NoTextType)
- uses formatted values from the DataType
- headers from DataTypes and sub columns

TODO:
- add another header with accession IDs or harmonized names for DataTypes
- change the joining of ListDataTypes to the selected separator
- use the full raw data file name or add the file path as a DataType (for removed name prefixes)


robinschmid:
The results. Feature data is represented by **DATAFILE:file_name:data_type**

image

SteffenHeu:
That's sweet!
Do you think we can use this to save and load feature lists from the project?

robinschmid:
We can discuss the project loading saving in issue #150 
This csv does not cover all data types (only the ones that produce text) and we export formatted text

tomas-pluskal:
I think it would make sense to use the Apache Commons CSV library - it can take care of encodings, quoting values etc. But since you already wrote the code, perhaps it's best to leave it as it is now :)

robinschmid:
No we should definitely use some sort of library here. Was not sure which one to use though.

tomas-pluskal:
The CSV import module is using the OpenCSV library, so that one is already included in MZmine. 
See http://opencsv.sourceforge.net/apidocs/com/opencsv/CSVWriter.html

tomas-pluskal:
So, should we merge this or do you want to rewrite it using a library?

robinschmid:
We can merge it from my side and add changes later if @SteffenHeu agrees.
Lets track proposed changes in #156

PR #153, opened 2021-01-14
Smoothing for LC-MS and LC-IMS-MS

SteffenHeu:
- The smoothing module was completely refactored.
- added copy methods for the data series

For IMS data, i'm smoothing the summed intensities which are used to construct a regular 2d-tic. Furthermore, every mobilogram is smoothed on it's own, although the summed mobilograms are not smoothed. For now, this is intended behaviour, I'll see later on, what works during mobility deconvolution.

Feel free to leave reviews before merging. I also added/changed a bit in the ModularFeature and ModularFeatureList.

### DIFF
--- a/src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
+++ b/src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
@@ -118,12 +118,13 @@ public class FeatureDataUtils {
     DoubleBuffer intensities = series.getIntensityValues();
     List<? extends Scan> scans = series.getSpectra();
     double lastIntensity = intensities.get(0);
+    float lastRT = scans.get(0).getRetentionTime();
     for (int i = 1; i < series.getNumberOfValues(); i++) {
       final double thisIntensity = intensities.get(i);
-      final float lastRT = scans.get(i - 1).getRetentionTime();
       final float thisRT = scans.get(i).getRetentionTime();
       area += (thisRT - lastRT) * (thisIntensity + lastIntensity) / 2.0 /* 60d*/;
       lastIntensity = thisIntensity;
+      lastRT = thisRT;
     }
     return area;
   }
--- a/src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
+++ b/src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
@@ -117,8 +117,8 @@ public class FeatureDataUtils {
     float area = 0f;
     DoubleBuffer intensities = series.getIntensityValues();
     List<? extends Scan> scans = series.getSpectra();
-    double lastIntensity = 0d;
-    for (int i = 0; i < series.getNumberOfValues(); i++) {
+    double lastIntensity = intensities.get(0);
+    for (int i = 1; i < series.getNumberOfValues(); i++) {
       final double thisIntensity = intensities.get(i);
       final float lastRT = scans.get(i - 1).getRetentionTime();
       final float thisRT = scans.get(i).getRetentionTime();
--- a/src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
+++ b/src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
@@ -1,0 +1,144 @@
+/*
+ *  Copyright 2006-2020 The MZmine Development Team
+ *
+ *  This file is part of MZmine.
+ *
+ *  MZmine is free software; you can redistribute it and/or modify it under the terms of the GNU
+ *  General Public License as published by the Free Software Foundation; either version 2 of the
+ *  License, or (at your option) any later version.
+ *
+ *  MZmine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ *  the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ *  Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with MZmine; if not,
+ *  write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301
+ *  USA
+ */
+
+package io.github.mzmine.datamodel.featuredata;
+
+import com.google.common.collect.Range;
+import io.github.mzmine.datamodel.MassSpectrum;
+import io.github.mzmine.datamodel.Scan;
+import io.github.mzmine.datamodel.features.ModularFeature;
+import io.github.mzmine.datamodel.features.types.numbers.AreaType;
+import io.github.mzmine.datamodel.features.types.numbers.IntensityRangeType;
+import io.github.mzmine.datamodel.features.types.numbers.MZRangeType;
+import io.github.mzmine.datamodel.features.types.numbers.RTRangeType;
+import java.nio.DoubleBuffer;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class FeatureDataUtils {
+
+  public static Range<Float> getRtRange(IonTimeSeries<? extends Scan> series) {
+    final List<? extends Scan> scans = series.getSpectra();
+    return Range
+        .closed(scans.get(0).getRetentionTime(), scans.get(scans.size() - 1).getRetentionTime());
+  }
+
+  public static Range<Double> getMzRange(MzSeries series) {
+    double min = Double.MAX_VALUE;
+    double max = Double.MIN_VALUE;
+
+    for (int i = 0; i < series.getNumberOfValues(); i++) {
+      final double mz = series.getMZ(i);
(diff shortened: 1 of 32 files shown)


src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
```diff
@@ -0,0 +1,167 @@
...
+
+  public static float calculateArea(IonTimeSeries<? extends Scan> series) {
+    if (series.getNumberOfValues() <= 1) {
+      return 0f;
+    }
+    float area = 0f;
+    DoubleBuffer intensities = series.getIntensityValues();
+    List<? extends Scan> scans = series.getSpectra();
+    for (int i = 1; i < series.getNumberOfValues(); i++) {
+      final double lastIntensity = intensities.get(i - 1);
```
tomas-pluskal, line 121: With this implementation each intensity is retrieved twice (once as "last" and once as "this"). Could be optimized to retrieve each value only once.

SteffenHeu: You're right, good point.

tomas-pluskal:
Looks good to me, anyone else wants to review?

src/main/java/io/github/mzmine/util/DataPointUtils.java
```diff
@@ -44,12 +44,41 @@
...
+    for (int i = 0; i < mzValues.capacity(); i++) {
       data[0][i] = mzValues.get(i);
       data[1][i] = intensityValues.get(i);
     }
     return data;
   }
-}
\ No newline at end of file
+
+  public static double[] getDoubleBufferAsArray(DoubleBuffer values) {
```
robinschmid, line 56: Where is this used?`We should definitely add some Java doc to layout the use cases and that people should try to use the DoubleBuffers directly - if possible

src/main/java/io/github/mzmine/util/DataPointUtils.java
```diff
@@ -44,12 +44,41 @@
     assert mzValues.capacity() == intensityValues.capacity();
```
robinschmid, line 44: Where is this used?`We should definitely add some Java doc to layout the use cases and that people should try to use the DoubleBuffers directly - if possible

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_smoothing/SmoothingTask.java
```diff
@@ -120,87 +118,43 @@ public void run() {
...
+              smoothedSeries = smoother.smooth(filterWeights);
+
+            } else if (featureData instanceof IonMobilogramTimeSeries) {
+              IonSpectrumSeriesSmoothing<IonMobilogramTimeSeries> smoother = new IonSpectrumSeriesSmoothing<>(
+                  ((IonMobilogramTimeSeries) featureData), newPeakList.getMemoryMapStorage(),
+                  (List<? extends MassSpectrum>) origPeakList
+                      .getSeletedScans(feature.getRawDataFile()));
+              smoothedSeries = smoother.smooth(filterWeights);
+            } else {
+              logger.info("Could not smooth feature, unknown ion series type.");
```
robinschmid, line 149: maybe throw exception? or at least change to warning level logging

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_smoothing/SmoothingTask.java
```diff
@@ -111,7 +109,7 @@ public void run() {
       final double[] filterWeights = SavitzkyGolayFilter.getNormalizedWeights(filterWidth);
 
       // Create new feature list
-      newPeakList = new ModularFeatureList(origPeakList + " " + suffix, origPeakList.getRawDataFiles());
+      newPeakList = origPeakList.createEmptyCopyWithSameTypes(origPeakList + " " + suffix);
```
robinschmid, line 112: I think it is better to copy the whole feature list and change the feature data in this task.
Further down, you use the new ModularFeature(oldFeature, flist) constructor which copies the feature anyways - however, you do not take care of FeatureListRows. I supsect that the manual annotations and other RowTypes are not copied right now.

The featurelsit.copy method is fast and we can make sure that we do not miss anything.

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_smoothing/SmoothingTask.java
```diff
@@ -120,87 +118,43 @@ public void run() {
 
           // Create a new peak-list row.
           final int originalID = row.getID();
-          final FeatureListRow newRow = new ModularFeatureListRow(newPeakList, originalID);
+          final ModularFeatureListRow newRow = new ModularFeatureListRow(newPeakList, originalID);
```
robinschmid, line 121: See earlier comment. newRow does not inherit all values from  the old row

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_smoothing/SmoothingTask.java
```diff
@@ -120,87 +118,43 @@ public void run() {
...
+                      .getSeletedScans(feature.getRawDataFile()));
+              smoothedSeries = smoother.smooth(filterWeights);
+            } else {
+              logger.info("Could not smooth feature, unknown ion series type.");
+              continue;
             }
+
+            ModularFeature newFeature = new ModularFeature(newPeakList, feature);
+            newFeature.set(FeatureDataType.class, smoothedSeries);
+            FeatureDataUtils.recalculateIonSeriesDependingTypes(newFeature);
```
robinschmid, line 155: See earlier comment. newFeature is a copy of the old feature - this is pretty much what the featureList.copy method does

SteffenHeu:
@robinschmid should take a short look at the feature list and the feature, I think.

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
```diff
@@ -461,31 +489,49 @@ public void setDateCreated(String date) {
...
+    for (FeatureListAppliedMethod proc : this.getAppliedMethods()) {
+      flist.addDescriptionOfAppliedTask(proc);
+    }
+
+    selectedScans.forEach(flist::setSelectedScans);
+    return flist;
+  }
+
+  public ModularFeatureList createEmptyCopyWithSameTypes(String title) {
+    ModularFeatureList flist = new ModularFeatureList(title, getRawDataFiles());
```
robinschmid, line 512: See comment in SmoothingTask. I think we do not need such functionality. The risk of loosing any information is high.

src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
```diff
@@ -272,7 +317,7 @@ public DataPoint getDataPoint(Scan scan) {
     if (index < 0) {
       return null;
     }
-    return getDataPoints().get(index);
+    return new SimpleDataPoint(getFeatureData().getMZ(index), getFeatureData().getIntensity(index));
```
robinschmid, line 320: We should at least deprecate this method or try to remove it completely

src/main/java/io/github/mzmine/util/DataPointUtils.java
robinschmid: I just never know when it will show multiple lines and when it only shows one :D

SteffenHeu: When copying the series which store the data. The MemoryMapStorage stores arrays, so we have to load them first and then store the new array.

src/main/java/io/github/mzmine/util/DataPointUtils.java
SteffenHeu: During smoothing you keep the old mzs, but store new intensities. So loading just a single buffer is useful,too.

src/main/java/io/github/mzmine/util/DataPointUtils.java
SteffenHeu: > 
> 
> Where is this used?`We should definitely add some Java doc to layout the use cases and that people should try to use the DoubleBuffers directly - if possible

True. However, I don't exactly know, how the memory mapped file works and if it's faster to query all points at once at some point.

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
SteffenHeu: Will remove again.

src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
SteffenHeu: The problem is that the Feature interface still has it. As soon was we remove that, we can remove this method. I went through all usages from ModularFeature in #154.

tomas-pluskal:
Thanks Steffen!

PR #154, opened 2021-01-15
Imsvisualiserfix

SteffenHeu:
In addition to #153:
- fixes the ims visualiser (I know you dont like the idea of a CachedFrame, so im open to ideas, but this is the easiest solution)
- Fixes ScanUtils.findBasePeak, by that, most of the old plot datasets are fixed.
  - plot datasets preferably use the new data type, for generation. However, i did not change their internal data storage.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,8 +11,7 @@ group = "io.github.mzmine"
 version = "3.0"
 description = "MZmine"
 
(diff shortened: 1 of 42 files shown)


tomas-pluskal:
Great, you were faster than me :)

tomas-pluskal:
I'll look into the CachedFrame thing.

PR #155, opened 2021-01-16
Fix RT-deconvolution for LC-MS and LC-IMS-MS

SteffenHeu:
Hey,
follow up to #153 and #154.
This fixes the deconvolution modules since they were broken due to the move to the new feature data model.
LC-MS can be processed as usual.
LC-IMS-MS can be processed as actual IMS data or, if build with the regular ADAP chromatogram builder, as regular "2d" LC-MS data.

grafik


SteffenHeu:
Ideally, the PeakResolver interface is updated to take 
DoubleBuffer/double[] intensities and List<? extends MassSpectrum> in the future. Then we can use the same modules to resolve the rt and mobility dimension and dont have to create intermediate legacy obects of the Feature interface.

SteffenHeu:
Something to discuss: 
Currently, the deconvolution is designed in a way to intentionally not transfer any information previously assigned to rows features, the reason being that before deconvolution a EIC can have multiple features.
These should not be targted of any annotation or so anyway, and information like rt and mz windows will have to be reprocessed anyway.
Further, we can prevent wrong annotations due to wrong module usage.

tomas-pluskal:
> Something to discuss:
> Currently, the deconvolution is designed in a way to intentionally not transfer any information previously assigned to rows features, the reason being that before deconvolution a EIC can have multiple features.
> These should not be targted of any annotation or so anyway, and information like rt and mz windows will have to be reprocessed anyway.
> Further, we can prevent wrong annotations due to wrong module usage.

I think this is the right way to do it.

robinschmid:
Sounds resonable.

PR #158, opened 2021-01-19
Fix deconvolution in batch mode and join aligner

robinschmid:
Some fixes:

- join aligner was checking for the data files at the wrong spot - the check is already done 
- fixed chromatogram deconvolution in batch mode (GroupMS2Parameters are now GroupMS2SubParameters - without a FeatureListsParameter)
-  found some uncommited CSV export changes (oops)

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
@@ -182,13 +182,6 @@ public class JoinAlignerTask extends AbstractTask {
     // Iterate source feature lists
     for (ModularFeatureList featureList : featureLists) {
 
-      // this should not be possible
-      for (RawDataFile file : featureList.getRawDataFiles()) {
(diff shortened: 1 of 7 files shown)


src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
```diff
@@ -182,13 +182,6 @@ public void run() {
     // Iterate source feature lists
     for (ModularFeatureList featureList : featureLists) {
 
-      // this should not be possible
```
SteffenHeu, line 185: Did i miss something when i added this?

SteffenHeu (approved):

src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
robinschmid: A few lines above, you create the aligned feature list with allDataFiles - so they are always present. 
You also aready checked if there are data file duplicates

tomas-pluskal:
Thanks Robin!

PR #161, opened 2021-01-22
added double click events to data types

SteffenHeu:
Added double click events to the feature table. The DataType specifies how it wants to handle the double clicks. Selection of multiple rows is not supported since it's kinda impossible to double click on two rows.

What do you think of this implementation, @robinschmid ?


SteffenHeu:
I also cleaned up the code of FeatureTableFX. I will push it, after i get your thoughts on this, since there are already a lot of whitespace changes.

tomas-pluskal:
It looks good to me.

PR #163, opened 2021-01-23
Provider refactoring

SteffenHeu:
Hey, 

i moved all providers to a single package, so it can be seen easily, which kinds of providers we already have. 
By that, i noticed there was a duplicate already :D


tomas-pluskal:
👍🏻

PR #168, opened 2021-01-29
MassSpectrum refactor

SteffenHeu:
Refactored MobilityScan to not have a DoubleBuffer
Importing 15 min, ~8000 frames, ~8 mio mobility scans before
pre optimisation

After:
Post optimization int pasePeakIndex

If we removed the HashSet<MassList> we'd also save ~500 mb + 180 mb ram. ^^


SteffenHeu:
Also removed the hash set from SimpleMobilityScan: 
(before mass detection)
grafik

src/main/java/io/github/mzmine/datamodel/impl/BuildingMobilityScan.java
```diff
@@ -0,0 +1,242 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * During raw data import, we need to cache the m/z and intensity values of mobility scans, so we
+ * can store all m/zs and intensities of a frame in a single double buffer to save memory.
+ * Therefore, we only implement the basic functionality here to limit the usage of this class to the
+ * loading process.
+ */
+public class BuildingMobilityScan implements MobilityScan {
```
tomas-pluskal, line 47: Why is this called BuildingMobilityScan? I mean, why not use just SimpleMobilityScan?

SteffenHeu: Because SimpleMobility scan now does not store it's own mz and intensity values anymore, so I needed a carrier for that. And the carrier is needed, because for storing, all mz and intensity values for all mobility scans that belong to one frame need to be stored in ram, so then can all be put into the same double buffer.

src/main/java/io/github/mzmine/datamodel/Frame.java
```diff
@@ -110,7 +110,7 @@ default int getPrecursorCharge() {
     return 0;
   }
 
-  void addMobilityScan(MobilityScan originalMobilityScan);
+  void setMobilityScans(List<BuildingMobilityScan> originalMobilityScan);
```
tomas-pluskal, line 113: BuildingMobilityScan should not be part of the data model - it is an implementing class. The data model should be defined only using interfaces.

SteffenHeu: I can make it
`void addMobilityScan(MobilityScan originalMobilityScan);`
again, if you like. Originally, i wanted to remove that method from the Frame interface and only declare it in SimpleFrame, so it's clear it shall not be used during data processing, but i seem to have forgotten that.
What do you think of that solution?

tomas-pluskal:
I think the fact that MobilityScan doesn't extend MassSpectrum is conceptually wrong. I believe there is a better way to achieve what you are trying to achieve here, by using a single shared DoubleBuffer per raw file. Each Scan or MobilityScan can store just an offset and length of the double arrays, that is 8 bytes of memory per scan. I'll look into it more in detail, ok?

tomas-pluskal:
I mean, we can merge this now and I can update the code later.

src/main/java/io/github/mzmine/datamodel/Frame.java
tomas-pluskal: Either solution works for me :)

SteffenHeu:
But mobility scan extends mass spectrum? 

`public interface MobilityScan extends MassSpectrum, Comparable<MobilityScan> {`

tomas-pluskal:
> But mobility scan extends mass spectrum?
> 
> `public interface MobilityScan extends MassSpectrum, Comparable<MobilityScan> {`

Ah okay, sorry, then it was my misunderstanding.

SteffenHeu:
I think having just one double buffer per file might be a problem for IMS data. I'm sure there can be more than 2E9 datapoints  in some cases.

SteffenHeu:
> 
> 
> > But mobility scan extends mass spectrum?
> > `public interface MobilityScan extends MassSpectrum, Comparable<MobilityScan> {`
> 
> Ah okay, sorry, then it was my misunderstanding.

No problem. ;) I just dont have SimpleMobilityScan extend AbstractMassSpectrum and AbstractStorableSpectrum anymore, because the had bloaty fileds which add up with regard to ram usage for SimpleMobilityScans.

tomas-pluskal:
I'll merge this and later I'll take a closer look if I can make some additional improvements in terms of RAM usage.

PR #173, opened 2021-02-02
Spectra merging

SteffenHeu:
This can merge spectra and is now used to merge MS/MS spectra of PASEF acquisition.
The merged spectra for PASEF are assigned to the features:
grafik
One Spectrum is produced per frame.

Example of merged MSMS:
grafik

There is a test module for IMS data you can play around with under Visualization -> Spectra Merging. I would remove that module before merging, though.
grafik

Merging itself works similar to the ADAP chromatogram builder, sorting by intensity first and then assigning datapoints to mz ranges.


tomas-pluskal:
Ok, let's merge the master and remove the test module.

Regarding the MergedMassSpectrum and MergedMsMsSpectrum interfaces - what are they needed for? Adding more interfaces into the data model will require more work for the visualization modules and project saving/loading modules to properly handle them. An alternative would be to just produce a simple MassSpectrum as a result of the merging process, but I don't know if you really need the extra information for something.

SteffenHeu:
MergedMsMsSpectrum is mainly a tag with some extra information you might need for database entries. I didn't want to just make it a scan because it was merged from multiple scans (or MobilityScans in that case).
Fom my point of view it only has to extend the Scan interface for campatibility reasions with the feature model and older modules, because they require a Scan as fragmentScan/allMs2Scans.

MergedMassSpectrum can be removed, i just wanted to make clear that there _could_ be another layer in betwee.

tomas-pluskal:
I understand your thinking, but generally I would recommend to not add features just because they might be needed in the future. It's all about maintainability and lack of human resources... :)

SteffenHeu:
Ok, so remove MergedMassSpectrum and keep MergedMsMsSpectrum?

tomas-pluskal:
And what is MergedMsMsSpectrum used for?

SteffenHeu:
In PASEF (and i guess other IMS-DDA) acquisition, multiple MS/MS mobility scans are recorded in a single frame for a single precursor. Since these scans usually have low overall intensities and since it's inconvenient to look at 25 MS/MS spectra, they are merged to a single merged spectrum. 
MergedMsMsSpectrum represents such a merged spectrum, keeping track of the original list of source spectra, the merging type (max, summed, avg intensities), the m/z centering method and the collision energy the spectrum was acquired at. This basically represents a ImsMsMsInfo, but as i can read from other vendors' mzml files, we might need different implementations of ImsMsMsInfo, so i decided it's better to safe it in the merged spectrum directly.

tomas-pluskal:
But is it necessary to have an interface for that? I mean, what if you have a SimpleMergedMsMsSpectrum class that implements Scan, would that not be enough?

SteffenHeu:
But then these methods wouldn't be accessible from the data model/interface alone and the programmer would have to use the actual class, which we wanted to avoid during processing, or?

tomas-pluskal:
What I'm looking for is some example where these methods are needed.

SteffenHeu:
As far as i'm aware colission energies are necessary for proper database entries, for which we have a few modules. These wouldnt need additional user interaction then, but could directly access the stored colission energy, making it less error prone.

As for the source spectra and merging methods, i personally like to know what happened to my data and why it might look different than when processed with other software/or different runs. 

For LC-coeluting compunds and partial ims-separation for example, it could be useful to reprocess the merged ms/ms spectra only taking one half of the source spectra for one compound.

tomas-pluskal:
Ok, so how about you leave it as it is now (MergedMassSpectrum and MergedMsMsSpectrum) and we evaluate it again later based on how the visualization and database submission modules turn out.

SteffenHeu:
Ok :+1:

SteffenHeu:
I'll remove the preview now so we can merge

tomas-pluskal:
👍🏻

PR #176, opened 2021-02-03
Feature list summary

SteffenHeu:
SimpleFeatureListAppliedMethod saves the actual module and a clone of the parameter set.

Added the dialog again.
Parameter name and value are tab separated, i did not see the benefit in making a table now.

Will implement the same thing for raw data files now.


SteffenHeu:
Raw data file methods are also listed now:

grafik

Listing the import is difficult to do, because these tasks don't have a parameter set in their processing.

Only thing left to do here is sending this to the batch mode module.

SteffenHeu:
Ok, done:

grafik

src/main/java/io/github/mzmine/datamodel/RawDataFile.java
```diff
@@ -153,4 +154,7 @@ default Scan getScanAtNumber(int scanNumber) {
   default Scan getScan(int i) {
     return getScans().get(i);
   }
+
+  @Nonnull
+  ObservableList<FeatureListAppliedMethod> getAppliedMethods();
```
robinschmid, line 159: We can keep the name FeatureListAppliedMethod now and change it later to AppliedMethod when everything else is merged?

robinschmid: Oh saw your TODO comments below never mind

src/main/java/io/github/mzmine/datamodel/features/FeatureList.java
```diff
@@ -154,32 +166,33 @@
...
    * @param mzRange m/z range
    */
   public ObservableList<Feature> getFeaturesInsideMZRange(RawDataFile file, Range<Double> mzRange);
 
   /**
    * Returns all features in a given m/z & retention time ranges
    *
-   * @param file Raw data file
+   * @param file    Raw data file
    * @param rtRange Retention time range
```
robinschmid, line 186: Somehow we need to make sure that the style is the same.
Is this also done in the eclipse style?

src/main/java/io/github/mzmine/datamodel/features/SimpleFeatureListAppliedMethod.java
```diff
@@ -1,32 +1,57 @@
...
+   * with the actual parameter set so these parameters can be reused. ~SteffenHeu
+   */
+  /*public SimpleFeatureListAppliedMethod(String description, String parameters) {
+    this.description = description;
+    this.parameters = parameters;
+  }*/
+
+//  public SimpleFeatureListAppliedMethod(String description) {
+//    this.description = description;
+//  }
```
robinschmid, line 54: This should be removed

src/main/java/io/github/mzmine/project/impl/SimpleRawDataFileAppliedMethod.java
```diff
@@ -0,0 +1,80 @@
...
+
+package io.github.mzmine.project.impl;
+
+import io.github.mzmine.datamodel.features.FeatureList.FeatureListAppliedMethod;
+import io.github.mzmine.main.MZmineCore;
+import io.github.mzmine.modules.MZmineModule;
+import io.github.mzmine.parameters.ParameterSet;
+import javax.annotation.Nonnull;
+
+public class SimpleRawDataFileAppliedMethod implements FeatureListAppliedMethod {
```
robinschmid, line 27: Is there any difference to SimpleFeatureListAppliedMethod?
Maybe just use one and give it an enum tag target = FeatureList / RawDataFile

Or the parameters will have a FeatureListParameter or DataFIleParameter

src/main/java/io/github/mzmine/datamodel/features/FeatureList.java
SteffenHeu: Don't think so. I loaded the intellij style in the mzmine repository.

src/main/java/io/github/mzmine/project/impl/SimpleRawDataFileAppliedMethod.java
SteffenHeu: Not different at all, i just wanted to keep it separated in case we need that at some point so we don't have to refactor everything. Enum tag is fine with me too, but will need another constructor parameter.

Just basing this on the parameters does not make sense, as some feature list methods use raw data file paramters, too.

src/main/java/io/github/mzmine/datamodel/RawDataFile.java
SteffenHeu: TODO: Go through all the TODOs  and do all the TODOs we left there as TODO

tomas-pluskal:
I resolved a minor merge conflict in imports. Otherwise it looks good!

PR #179, opened 2021-02-06
Scan selection load fix

SteffenHeu:
fixed scan selection not being loaded correctly due to misplaced Objetcs.requireNonnull


tomas-pluskal:
👍🏻

PR #182, opened 2021-02-08
Warning messages for ims data

SteffenHeu:
This adds an opt out warning message if a module that is untested/restriced with ims data is used to process it. By default, every module is untested.

Default untested warning message:
grafik

If a module has been tested with ims data and bears cetrain restrictions, it can override a method and display a specific warning message:
grafik

At the same time, modules can only support ims data:
grafik

The opt out values are saved in a hidden parameter in the preferences and can thus only be deleted, by going to the config directly. If someone feels like doing a component for that, we could change it.

### DIFF
--- a/src/main/java/io/github/mzmine/modules/batchmode/BatchSetupComponent.java
+++ b/src/main/java/io/github/mzmine/modules/batchmode/BatchSetupComponent.java
@@ -44,7 +44,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import javafx.collections.FXCollections;
 import javafx.scene.control.Button;
 import javafx.scene.control.ChoiceDialog;
 import javafx.scene.control.ComboBox;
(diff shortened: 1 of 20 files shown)


src/main/java/io/github/mzmine/modules/batchmode/BatchSetupComponent.java
```diff
@@ -243,7 +242,7 @@ protected void updateItem(MZmineProcessingStep<MZmineProcessingModule> item,
 
         // Add step to queue.
         batchQueue.add(step);
-        currentStepsList.setItems(FXCollections.observableArrayList(batchQueue));
+        currentStepsList.setItems(batchQueue);
```
SteffenHeu, line 245: Was there a reason the batch queue wasn't set directly? This lead to the drag & drop gestures not being represented in the underlying batch queue.

tomas-pluskal: Not sure :)

PR #184, opened 2021-02-08
Mzmlframe merging

SteffenHeu:
A simple tool to merge mobility scans into a summed frame spectrum. 

For now, i have decided to not use the original raw data file and the original frames to store the merged data. However, this is debatable. Nevertheless, since there is no data in those frames prior to merging and the bruker library merging does the merge on request, too, i decided to implement it this way for now.

grafik


tomas-pluskal:
Sounds great to me, at least this approach is universal and doesn't depend on the bruker library.

PR #185, opened 2021-02-08
Isotopic peaks

SteffenHeu:
After deisotoping, the row types were missing.

I know this is quite a dirty fix, but it works until we have the row grouping. Also, isotopic peaks now consider the mobility dimension.


tomas-pluskal:
👍🏻

PR #187, opened 2021-02-12
Remove mass list

robinschmid:
I removed all uses of the mass list parameter
- each scan now has exactly one MassList
- all MassLists of all RawDataFIles, that are generated by the same module call, are added to the same MemoryMapStorage
- fixed a few NPE

### Tested:
- Import, MassDetection, ADAP chromatogram builder, local minimum search, deisotoping, join aligner, peak finder gap filling


SteffenHeu:
https://github.com/robinschmid/mzmine3/pull/3

tomas-pluskal:
Be gone, mass lists! :-p

PR #188, opened 2021-02-12
Ram only feature lists

SteffenHeu:
As requested by Robin, i added a tag to keep all feature data in ram, for server/workstation processing.

Note this is currently unsafe to enable if merged msms spectra are created, because the MemoryMapStorage of the feature list is used for that operation.
There are two solutions: 
1. Make AbstractStorableSpectrum support MemoryMapStorage = null to store data points in ram
2. Create a central buffer where merged spectra will be saved to.

I'll leave that for @robinschmid to implement :) 

Edit: btw this also fixes creation of a new MemoryMapStorage if the user clicked in the raw data overview


robinschmid:
Will look at it tomorrow. Thanks steffen

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
```diff
@@ -28,27 +29,23 @@
...
+  public static final DateFormat DATA_FORMAT = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
+  // bindings for values
+  private final List<RowBinding> rowBindings = new ArrayList<>();
+  // unmodifiable list
+  private final ObservableList<RawDataFile> dataFiles;
+  private final ObservableMap<RawDataFile, List<? extends Scan>> selectedScans;
   // storage of data points in case features are edited
   private MemoryMapStorage memoryMapStorage = null;
-
+  private boolean forceFeatureDataIntoRam = false;
```
robinschmid, line 40: I would set this as a global parameter.
People start MZmine with an argument and then everything is done on RAM

robinschmid: As this is an advanced option - I thought about using startup arguments instead of preferences

SteffenHeu: Feel free to submit that as a pull request to this branch

tomas-pluskal:
Wouldn't it be cleaner to have a storage backend (interface) and MemoryMapStorage and RAMStorage as two different implementing classes?

SteffenHeu:
I think that would be over-enginieering it a bit, as long as we are using DoubleBuffers. 
For now, they are an easy way to achieve storing data either on disc by means of the memory mapped file, or storing it in ram by wrapping the double buffer around an array. If we decided to not use the double buffer internally, we could think about that.

I added this utitlity class as part of the ims optimisations, since i found myself implementing the same thing over and over again when it came to ims data (storing multiple arrays in a single buffer). The implementations of Frame and FrameMassList do not yet use this utility class, but i will port them to use it at some point in the next weeks.
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/datamodel/featuredata/impl/StorageUtils.java

This class also comes with a method to store the data either in ram or on disc, so we don't have to use the if+try block.
` 
public static DoubleBuffer storeValuesToDoubleBuffer(@Nullable final MemoryMapStorage storage, @Nonnull final double[] values) {

    DoubleBuffer buffer;
    if (storage != null) {
      try {
        buffer = storage.storeData(values);
      } catch (IOException e) {
        e.printStackTrace();
        buffer = DoubleBuffer.wrap(values);
      }
    } else {
      buffer = DoubleBuffer.wrap(values);
    }
    return buffer;
  }
`

PR #189, opened 2021-02-13
Fix FBMN and GC-MS on GNPS

robinschmid:
This PR is dependent on #187 

- added the legacy CSV export module for compatibility
- FBMN and GC-MS GNPS exports the feature table to legacy CSV and the full table with the modular CSV export


tomas-pluskal:
Looks good to me.

PR #191, opened 2021-02-16
Ims series optimisation

SteffenHeu:
Optimises ram usage of  ion mobility series with regard to double buffers. 

Similar to MobilityScans and the soon to come mass lists, all mobilogram data for an ion mobility trace is stored in a single double buffer.


src/main/java/io/github/mzmine/datamodel/featuredata/impl/ModifiableSpectra.java
```diff
@@ -0,0 +1,9 @@
+package io.github.mzmine.datamodel.featuredata.impl;
```
tomas-pluskal, line 1: License header? :)

PR #194, opened 2021-02-19
Fix local minimum deconvolution 

robinschmid:
Chromatographic threshold needed number of all scans of a chromatogram to calc quantile
(still need to merge master)


src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/ResolvingUtil.java
```diff
@@ -157,4 +160,27 @@
...
+  /**
+   * All scans that were used to build a chromatogram or the maximum number of mobility scans in all
+   * Frames used to construct the mobilogram
+   *
+   * @param data
+   * @param dimension
+   * @param selectedScans
+   * @return
+   */
+  public static int selectedTotalScans(@Nonnull IonTimeSeries<? extends Scan> data,
```
SteffenHeu, line 172: Should be called getTotalNumberOfScansInDimension to be more descriptive

robinschmid: done

robinschmid:
Merged into master

robinschmid:
There still remains the question what kind of data the other deconvolution modules need.
Currently we only supply the detected data points. I think we should split the modules more and leave the data retrieval to each module.
### Local minimum search needs:
- only data points where the signal was detected in the scan
- insert zeros in holes between two signals?
- the total number of scans to determine the chromatographic threshold (quantiles)

### Noise amplitude deconvolution needs:
- only data points where signal was detected (all non zero data points)

### Other algorithms need:
- Other algorithms might need the full chromatogram with zeros at scans with missing signals

SteffenHeu:
I agree. However, since this fixes some stuff, we should merge this now and keep that in mind. Add it to the Project maybe?

SteffenHeu:
Furthermore, keep in mind that it was implemented this way to be able to resolve the mobility and rt dimension with the same module.

robinschmid:
We definitely need to keep it in mind. mobility and rt should be no problem.

PR #198, opened 2021-02-22
Test for MemoryMapStorage creation and discard (disabled by default)

robinschmid:
A small test that you can activate manually to check the MemoryMapStorage
- discard 
- creation of thousands of files

### DIFF
--- a/src/main/java/io/github/mzmine/util/MemoryMapStorage.java
+++ b/src/main/java/io/github/mzmine/util/MemoryMapStorage.java
@@ -37,26 +37,24 @@ import javax.annotation.Nonnull;
  * storeData() function stores an array into the underlying temporary file and returns a buffer. The
  * buffer is directly bound to the memory-mapped portion of the file so the data can be directly
  * accessed without loading it into another intermediate primitive type array.
- *
+ * <p>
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/mzmine/util/MemoryMapStorage.java
```diff
@@ -70,11 +68,28 @@
...
   /**
    * The file that we are currently writing into.
    */
   private MappedByteBuffer currentMappedFile = null;
 
+  public MemoryMapStorage() {
+    this(null);
+  }
+
+  public MemoryMapStorage(File tmpDir) {
```
SteffenHeu, line 83: The overall temp dir can be set via -Djava.io.tmpdir=/path/to/tmpdir JVM settings. Do you think we should separate the memory map storage from there?

robinschmid:
For now this is only a test. You can try the behavior on different Systems.

src/main/java/io/github/mzmine/util/MemoryMapStorage.java
robinschmid: For now this was only for test purposes. 
I guess we do not need this option

PR #202, opened 2021-02-22
Low garbage scan data access

robinschmid:
Hey everyone,

I have implemented a class to loop over all scans in a raw data file and access either the raw data or the centroid mass lists. 

- RawDataFile keeps count of the maxNumberOfRawDataPoints / maxNumberOfCentroidDataPoints (over all scans)
- the new class uses a dataFile, scan selection, and RAW/CENTROID flag
- Iterates over all scans and uses only two arrays (mz, intensity)
- gets the mzs and intensities with the getIntensityValues(double[] dst) methods
- It implements the MassSpectrum interface - so no need to change the mass detectors, I only changed the general MassDetectorTask

What do you think?
My idea is that it will promote low garbage + memory efficient access to Scan/MassList data. And we can control it and add some security to it.


src/main/java/io/github/mzmine/datamodel/impl/MemoryEfficientDataAccess.java
```diff
@@ -0,0 +1,135 @@
...
+import io.github.mzmine.datamodel.MassList;
+import io.github.mzmine.datamodel.RawDataFile;
+import io.github.mzmine.datamodel.Scan;
+import io.github.mzmine.parameters.parametertypes.selectors.ScanSelection;
+import io.github.mzmine.util.exceptions.MissingMassListException;
+
+/**
+ * @author Robin Schmid (https://github.com/robinschmid)
+ */
+public class MemoryEfficientDataAccess {
```
SteffenHeu, line 29: It does not yet implement MassSpectrum, or am i blind?

src/main/java/io/github/mzmine/datamodel/listeners/MassListChangedListener.java
```diff
@@ -0,0 +1,37 @@
...
+
+import io.github.mzmine.datamodel.MassList;
+import io.github.mzmine.datamodel.Scan;
+
+/**
+ * MassList was changed
+ *
+ * @author Robin Schmid (https://github.com/robinschmid)
+ */
+public interface MassListChangedListener {
```
SteffenHeu, line 28: I don't really like the Idea of having yet another listener for a single task that is only executed in a single module. Seems like unnecessary ram consumption to me. I think having a simple method in the mass detection task or so updating the max number of data points is way more efficient.

src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
```diff
@@ -60,16 +63,21 @@
   private String dataFileName;
 
   private final Hashtable<Integer, Range<Double>> dataMZRange = new Hashtable<>();
-  private final Hashtable<Integer, Range<Float>> dataRTRange = new Hashtable<>();;
+  private final Hashtable<Integer, Range<Float>> dataRTRange = new Hashtable<>();
+  ;
```
SteffenHeu, line 67: Your formatter shows us we have some leftover semicolons here :D

src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
```diff
@@ -263,10 +295,22 @@ public double getDataMaxTotalIonCurrent(int msLevel) {
   }
 
 
-
   @Override
   public synchronized void addScan(Scan newScan) throws IOException {
     scans.add(newScan);
+    if (newScan.getNumberOfDataPoints() > maxRawDataPoints) {
```
SteffenHeu, line 301: This is imo how it should be handled for mass lists during mass detection, too.

src/main/java/io/github/mzmine/datamodel/impl/MemoryEfficientDataAccess.java
```diff
@@ -0,0 +1,135 @@
...
+
+  public enum DataType {
+    RAW, CENTROID;
+  }
+
+  protected final RawDataFile dataFile;
+  protected final DataType type;
+  private final Scan[] scans;
```
SteffenHeu, line 37: About this idea in general: If we use it, it should also be able to handle MobilityScans. List<? extends MassSpectrum> here maybe?

src/main/java/io/github/mzmine/datamodel/impl/MemoryEfficientDataAccess.java
robinschmid: oh thats weird it somehow did not push. 
I pushed again

src/main/java/io/github/mzmine/datamodel/listeners/MassListChangedListener.java
robinschmid: No idea we can also just directly call RawDataFile.massListChanged(); and set the maxNumberOfCentroidDataPoints to zero

I thought maybe there are multiple methods that want to listen for changes. Like charts could register and update if things are changed.

SteffenHeu: In which case would that be needed? The mass detector preview does it manuallly anyway.

src/main/java/io/github/mzmine/datamodel/impl/MemoryEfficientDataAccess.java
robinschmid: I thought about using a factory style and different implementations
EfficientDataAccess.of(RawDataFile, ScanSelection, DataType.CENTROID/RAW)
EfficientDataAccess.of(FeatureList, HandleZero.INCLUDE/EXCLUDE)

src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
robinschmid: We have plenty of modules that create new mass lists and this should be built in somewhere. 
But we can just set the maxNumberOfCentroidDataPoints to -1 each time a mass list is changed in a RawDataFIle.

src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
robinschmid: done

src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
SteffenHeu: Ok. Just keep in mind that adding listeners to all mobility scans is a huge ram consumption. Convenience is nice, but we have  to keep both sides in mind.

robinschmid:
I removed the MassListChangedListener and now we call applyMassListChanged(Scan, oldMasses, newMasses) directly on the RawDataFile to reset some precomputed values.
+ cleaned up. 
Ready for merge from my side

src/main/java/io/github/mzmine/datamodel/impl/MemoryEfficientDataAccess.java
SteffenHeu: Ok, so there should be an 
`class MemoryEfficientScanDataAccess implements EffiecientDataAccess`
and
`class MemoryEffiecientMobilityScanDataAccess implements EfficientDataAccess`
and
`class DataAccessFactory`
?
And i can see you also want to use feature lists. Then we should think of an interface and it's methods now so we can handle all those cases and don't have to refactor multiple times after merging this.

src/main/java/io/github/mzmine/datamodel/data_access/FeatureDataAccess.java
```diff
@@ -0,0 +1,364 @@
...
+      switch (type) {
+        case ONLY_DETECTED -> {
+          featureData.getMzValues(mzs);
+          featureData.getIntensityValues(intensities);
+          currentNumberOfDataPoints = featureData.getNumberOfValues();
+        }
+        case INCLUDE_ZEROS -> {
+          // add detected data points and zero for missing values
+          allScans = flist.getSeletedScans(feature.getRawDataFile());
+          List<Scan> detectedScans = feature.getScanNumbers();
```
SteffenHeu, line 265: This should be featureData.getSpectra(). Is getScanNumbers not deprecated yet?

src/main/java/io/github/mzmine/datamodel/data_access/FeatureDataAccess.java
```diff
@@ -0,0 +1,364 @@
...
+          int detectedIndex = 0;
+          for (int i = 0; i < intensities.length; i++) {
+            if (allScans.get(i) == detectedScans.get(detectedIndex)) {
+              intensities[i] = featureData.getIntensity(detectedIndex);
```
SteffenHeu, line 269: If we want this to be more performant, we should read the whole array/double buffer at once and then loop over the array instead of querieing each intensity.
Before, we (I) should update the feature model to not expose the double buffers to the user, similarly to the raw data model.

src/main/java/io/github/mzmine/datamodel/featuredata/MzSeries.java
```diff
@@ -33,6 +33,14 @@
    */
   DoubleBuffer getMZValues();
 
+
+  /**
+   *
+   * @param dst results are reflected in this array
+   * @return All m/z values of detected data points.
+   */
+  double[] getMzValues(double[] dst);
```
SteffenHeu, line 42: Ok, nevermind my earlier comment. Thanks for doing that ;)

src/main/java/io/github/mzmine/datamodel/impl/SimpleMergedMsMsSpectrum.java
```diff
@@ -187,9 +186,20 @@ public MassList getMassList() {
...
+    // we are not going into any details if this.massList equals massList
+    // do not call listeners if the same object is passed multiple times
+    if (this.massList == massList) {
+      return;
+    }
+    MassList old = this.massList;
     this.massList = massList;
+
+    if (rawDataFile != null) {
+      rawDataFile.applyMassListChanged(this, old, massList);
```
SteffenHeu, line 201: Since a merged spectrum is not part of the original data file, this call should have no effect, or?

src/main/java/io/github/mzmine/datamodel/data_access/FeatureDataAccess.java
```diff
@@ -0,0 +1,364 @@
...
+/**
+ * Access the chromatographic data of features in a feature list sorted by scan ID (usually sorted
+ * by retention time)
+ *
+ * @author Robin Schmid (https://github.com/robinschmid)
+ */
+public class FeatureDataAccess implements IonTimeSeries<Scan> {
+
+  /**
+   * Different types to handle feature data: {@link #ONLY_DETECTED}: Use only detected data points;
```
SteffenHeu, line 45: ONLY_DETECTED should be more specific:
ONLY_DETECTED: only data points currently assigned to the feature. Features might include leading an/or trailing zeros depending on the state of processing. Leading and trailing zeros are added during chromatogram detection, bit might be removed during feature resolving or other processing steps.

robinschmid (commented):
somehow I cannot comment directly on yours

src/main/java/io/github/mzmine/datamodel/data_access/FeatureDataAccess.java
robinschmid: Yeah I was trying to keep the garbage low. We could try to have a second final array to do this.

robinschmid: Maybe we just change this in the future?

src/main/java/io/github/mzmine/datamodel/impl/SimpleMergedMsMsSpectrum.java
robinschmid: No idea. But I think as long as a MergedSpectrum can have a data file - we should call the change method

robinschmid: But I also do not really know where you use the MergedMSMSSpec

src/main/java/io/github/mzmine/datamodel/data_access/FeatureDataAccess.java
robinschmid: done, uses batch read now

robinschmid:
Should be ready for merging now. @SteffenHeu

SteffenHeu:
Looks good :+1:

Issue #206, opened 2021-02-23
One MemoryMapStorage per module call


robinschmid:
We still need to create a way to safely control the creation of MemoryMapStorage
The idea is to bundle all RawDataFiles, MassLists, and FeatureLists that were created by the same module call to limit the number of files.

In Windows 10 I had the issue that we could only create 8 temp files with 1GB each on ~8 GB free disk space.

### TODO
- Maybe keep track of all MemoryMapStorage instances in a List / Map to delete them if not needed anymore
- Have a way to create a single memory map for a module call

SteffenHeu:
#209

Issue #204, opened 2021-02-23
Removing temporary files on exit


SteffenHeu:
Temporary files should be deleted on exit since they take up large amounts of disc space.
Is there anything speaking against either adding this:
https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/main/TmpFileCleanup.java 
to the shutdown hook or (in case there is a problem with shutdown hooks, i remember reading something like that somewhere) adding this
`stage.setOnHidden(e -> Platform.runLater(TmpFileCleanup::new));`
to MZmineGUI?

SteffenHeu:
#208 (not activated)

PR #208, opened 2021-02-23
Remove temp files on exit

SteffenHeu:
In reference to #204.

I did some digging and understood why the we have the cleanup on startup. However, the disk space should be freed on exit of MZmine. Therefore, i did some more research and found a way to remove a temp file of a byte buffer. 

Nevertheless, I have the feeling you guys might not like it because of the usage of sun.misc.Unsafe. :D Furthermore, I think this will currently only delete one file per MemoryMapStorage. We would need to keep track of all MappedByteBuffers created by MemoryMapStorage.

Also, the deletion method uses reflection to access a private field of the MemoryMapStorage, which is a hack, i totally agree, but the field is also meant to not be accessed in any other case during runtime. So i thought the hackiest solution might be the best one.

Keeping track of all MappedByteBuffers  could be incorporated with #206.

In case the deletion of a MappedByteBuffer is only a windows problem, i could add a platform check.

What do you guys think?

### DIFF
--- a/src/main/java/io/github/mzmine/main/TmpFileCleanup.java
+++ b/src/main/java/io/github/mzmine/main/TmpFileCleanup.java
@@ -18,20 +18,34 @@
 
 package io.github.mzmine.main;
 
+import io.github.mzmine.datamodel.MZmineProject;
+import io.github.mzmine.datamodel.features.ModularFeatureList;
+import io.github.mzmine.project.ProjectManager;
+import io.github.mzmine.util.MemoryMapStorage;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.RandomAccessFile;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
+import java.nio.MappedByteBuffer;
 import java.nio.channels.FileLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.Nullable;
+import sun.misc.Unsafe;
 
-class TmpFileCleanup implements Runnable {
+public class TmpFileCleanup implements Runnable {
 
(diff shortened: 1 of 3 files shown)


robinschmid:
Did you try it? Maybe with my test class?
Maybe you could access the List of files from MemoryMapStorage somehow?

Do you have a reference where you found this solution?

src/main/java/io/github/mzmine/main/TmpFileCleanup.java
```diff
@@ -76,4 +95,91 @@ public boolean accept(File dir, String name) {
...
+  }
+
+  /**
+   * Taken from https://stackoverflow.com/a/48821002
+   *
+   * @return Instance {@link Unsafe} or null.
+   * @author https://github.com/SteffenHeu
+   */
+  @Nullable
+  private synchronized Unsafe initUnsafe() {
```
SteffenHeu, line 165: Reference here, @robinschmid

SteffenHeu:
> 
> Maybe you could access the List of files from MemoryMapStorage somehow?
> 

I think it's not about the files list, it's more about the instances of currentMappedFile (=MappedByteBuffer). But I'm not completely into that topic. @tomas-pluskal would have to comment on that, perhaps.

tomas-pluskal:
I don't understand why you need to use reflection to access the MemoryMapStorage field... Why don't you just add a MemoryMapStorage.dispose() method that would do the cleanup and move all your Unsafe code into this method?

SteffenHeu:
That was the first implementation I did, but i did not want to expose that functionality to the general user, because it will just delete the files, no matter if they are are still used or not (I think). But i can revert that if you like.

I'm currently working on an implementation to use storages efficiently, so i can combine both approaches there.

SteffenHeu:
#209

PR #215, opened 2021-02-26
Clean up ModularFeature(Row) and use objects instead of Double.NaN 

robinschmid:
Changed all the data accession to objects (Double/Float) and switched from Double.NaN to null for missing values


SteffenHeu:
I don't know it we should do this now. Can casting between doubles, Doubles, Floats and floats lead to wrong values due to wrong conversion?

robinschmid:
Yeah it is also not safe for now that everything is fixed. 
I was hoping it was easier :D

SteffenHeu:
We should probably do this when we remove the properties.

PR #216, opened 2021-02-26
Set preferences file and batch by arguments

robinschmid:
- fixes batch mode in headless mode
- set preferences file by command line arguments

### DIFF
--- a/src/main/java/io/github/mzmine/main/MZmineCore.java
+++ b/src/main/java/io/github/mzmine/main/MZmineCore.java
@@ -57,6 +57,7 @@ import java.util.Properties;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javafx.application.Application;
+import javafx.application.Platform;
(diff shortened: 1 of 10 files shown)


src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -318,6 +339,28 @@ private static void setTempDirToPreference() {
...
+  public static boolean isHeadLessMode() {
+    return headLessMode;
+  }
+
+  /**
+   *
+   * @param r runnable to either run directly or on the JavaFX thread
+   */
+  public static void runLater(Runnable r) {
+    if(isHeadLessMode()) {
```
SteffenHeu, line 356: could be if(isHeadLessMode() || Platform.isFxApplicationThread()) or?

robinschmid: Done

PR #220, opened 2021-03-01
More efficient ADAPChromatogramBuilder

robinschmid:
I tried to optimize a bit the ADAPChromatogramBuilder and reduce the created objects. 

- optimized the "list all data points of all scans" part
- removed unused arrays
- maybe optimized the add zeros method (at least it feels optimized)
- clean up methods

Seems to work fine for my LC-MS test data.

### DIFF
--- a/src/main/java/io/github/mzmine/util/FeatureConvertors.java
+++ b/src/main/java/io/github/mzmine/util/FeatureConvertors.java
@@ -26,6 +26,7 @@ import io.github.mzmine.datamodel.Frame;
 import io.github.mzmine.datamodel.IMSRawDataFile;
 import io.github.mzmine.datamodel.RawDataFile;
 import io.github.mzmine.datamodel.Scan;
+import io.github.mzmine.datamodel.featuredata.FeatureDataUtils;
 import io.github.mzmine.datamodel.featuredata.IonMobilitySeries;
 import io.github.mzmine.datamodel.featuredata.IonMobilogramTimeSeries;
 import io.github.mzmine.datamodel.featuredata.IonTimeSeries;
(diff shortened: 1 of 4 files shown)


SteffenHeu:
Looks good.

What do you think about removing the area calc from ADAPChromatogram and using
`FeatureDataUtils.recalculateIonSeriesDependingTypes(ModularFeature feature)` 
everywhere to make it consistent across all modules?

robinschmid:
Is it already redundant?

SteffenHeu:
Don't think so, but we could put it in the FeatureConvertor.

src/main/java/io/github/mzmine/util/FeatureConvertors.java
```diff
@@ -131,20 +128,12 @@ static public ModularFeature ADAPChromatogramToModularFeature(
...
-    Range<Double> mzRange = Range.closed(chromatogram.getRawDataPointsMZRange().lowerEndpoint(),
-        chromatogram.getRawDataPointsMZRange().upperEndpoint());
-    Range<Float> intensityRange =
-        Range.closed(chromatogram.getRawDataPointsIntensityRange().lowerEndpoint().floatValue(),
-            chromatogram.getRawDataPointsIntensityRange().upperEndpoint().floatValue());
-    modularFeature.set(MZRangeType.class, mzRange);
-    modularFeature.set(RTRangeType.class, rtRange);
-    modularFeature.set(IntensityRangeType.class, intensityRange);
+    // recalculate data dependet types
+    FeatureDataUtils.recalculateIonSeriesDependingTypes(modularFeature, CenterMeasure.AUTO);
```
SteffenHeu, line 132: Auto is not implemented in the math utils

```
 public static double calcCenter(CenterMeasure measure, double[] values, double[] weights,
      Weighting weightTransform, Double noiseLevel, Double maxWeight) {
    switch (measure) {
      case AVG:
        return calcWeightedAvg(values, weights, weightTransform, noiseLevel, maxWeight);
      case MEDIAN:
        return calcMedian(values);
      default:
        return Double.NaN;
    }
  }
```

robinschmid: oh

robinschmid: used the default then

PR #222, opened 2021-03-04
Keep features with < 3  datapoints

robinschmid:
- Keep features <3 data points
- efficient subset of IonTimeSeries (Feature data)

- Checks if subset is a real subset of spectra in case of error
- Checks the order in case of error


robinschmid:
Also added a minimum number of data points to the local minimum search

PR #223, opened 2021-03-04
Align by mobility

SteffenHeu:
- join aligner weights mobility for alignment if wanted
- center function test for mz calculation
- feature list summary shows values of optional parameters
- fixed bugs in ion mobility trace builder, added advanced parameters


SteffenHeu:
- removed tool tips from feature table
- added tooltips to parameter setup dialog

PR #228, opened 2021-03-05
Test workflow in batch mode

robinschmid:
- added option to run MZmine in headless mode and keep it open    -r  or --running
- added option to leep all data (scans, features, etc) in RAM   -m  or --memory
- added small test files for LC-MS with DDA
- added tests for advanced data import and chromatogram builder

### TODO
- add tests for next steps


robinschmid:
- clean up of join alginer works like a charm
- fixed gap-filling that removed the new list instead of the original feature list

PR #232, opened 2021-03-11
Add main module categories

robinschmid:
Just an idea how we could manage the module categories.

We also need to clean them of course


SteffenHeu:
looks good

SteffenHeu:
Maybe 
`case NORMALIZATION, DATAANALYSIS -> "Data analysis";`
could be called "Feature processing" since normalisation does not really analyse?
And "Identification" could be called "Annotation". Then stuff like the CCS calculation could be put there

robinschmid:
Thats true. Its just an idea and I wanted to put it online if you need it for the batch mode

PR #239, opened 2021-03-24
TSF Import

SteffenHeu:
- TSF import
- Maldi TIMS import
- Update/fix image builder
- Update ims trace builder to be used for imaging files
- Alignment of maldi + lc TIMS files by mobility + mz (join aligner)


robinschmid:
Great

PR #240, opened 2021-03-24
Feature table

SteffenHeu:
- Feature table raw file icons displayed in accourdance to raw file color
grafik

- Column visibility applied recursively

- Default color palette added to PaintScalePalette dialog:
grafik


SteffenHeu:
Fixed periodic table dialog, elements are now actually added to the ElementsComponent.

PR #248, opened 2021-03-31
increased control contrast, styled tree view

SteffenHeu:
TreeView style matches ListView
Background of Tree-, List-, TreeTable-, TableView use the odd cell background colour instead of the button bg colour, so doesnt suddenly change once a file/flist is loaded.
Increased contrast on buttons/controls in general.
grafik


SteffenHeu:
EDIT: What do you think about this:
grafik
Light

grafik
Dark

robinschmid:
looks great.
Can you make the headers smaller? Would be nice to safe a little space with the height of the tabs

SteffenHeu:
grafik
Like this? I also dont want it to look bunched

robinschmid:
Yeah

PR #249, opened 2021-04-01
Chart performance

SteffenHeu:
ion traces + images rendered as image in feature table


robinschmid:
BufferedImages are first added to the FeatureTable
On click those are changed to interactive charts - faster load time
Currently they are not changed back. This mean they stay as charts

PR #250, opened 2021-04-01
Recursivebuilder performance

SteffenHeu:
increased performance of recurse builder using parallel stream


robinschmid:
Thanks for using lists in cases where sets ar inferior.

PR #256, opened 2021-04-09
jdk 16, gradle 7 rc

SteffenHeu:


SteffenHeu:
I think lgtm doesn't use gradle RCs. So we have to wait until the actual release.

robinschmid:
You are faster than I. 
I also read about the release of gradle 7 and wanted to try it

SteffenHeu:
Oh i just saw the actual gradle 7 version released yesterday. Then it might work now if i adjust the script. Before i was trying to use the release candidate, which didnt work.

tomas-pluskal:
It seems JavaFX now produces a warning when its classes are present on the classpath (instead of the module path, which is the preferred way). See the first issue in the Release Notes: https://github.com/openjdk/jfx/blob/master/doc-files/release-notes-16.md

MZmine is currently not a modular application, but perhaps we should convert it into one. There will be some obstacles on the way.

SteffenHeu:
How high of a priority should it be?

tomas-pluskal:
I have already noticed some issues during the JavaFX Application startup, but it is not something burning.

robinschmid:
Do you mean the jigsaw module system of JDK9? 
As far as I understand - no one uses it :D
Its main advantage is that it modularizes the JDK and you can include only the needed parts of it into your runtime.

PR #262, opened 2021-04-24
image to csv export

SteffenHeu:
- CSV export for images for via the FeatureTableContextMenu. Exported via a matrix.
- Made MobilityScan extend Scan, but it's all default implementations and none of them need extra objects. 
- Add option to show the best mobility scan of IMS traces.


tomas-pluskal:
Hey Steffen, the `clustering-20130919.jar` file should not be needed anymore, the dependenci is updated in build.gradle

SteffenHeu:
Oops, my bad. I thought it was the other way round. I'll remove it.

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -0,0 +1,199 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.text.NumberFormat;
+import java.util.Collection;
+import java.util.List;
+import java.util.logging.Logger;
+
+public class ImageToCsvExportTask extends AbstractTask {
```
robinschmid, line 46: some small java doc to explain that this exports imaging data for features to CSV

robinschmid (changes requested):

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -0,0 +1,199 @@
...
+    return processed / (double) features.size();
+  }
+
+  @Override
+  public void run() {
+    setStatus(TaskStatus.PROCESSING);
+
+    logger.fine(() -> "Determining maximum export dimensions...");
+    final List<ImagingRawDataFile> distinctFiles =
+        (List<ImagingRawDataFile>) (List<? extends RawDataFile>) features
```
robinschmid, line 91: double cast?

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -0,0 +1,199 @@
...
+            .stream().map(ModularFeature::getRawDataFile).distinct()
+            .filter(file -> file instanceof ImagingRawDataFile).toList();
```
robinschmid, line 93: Maybe add .map(f -> (ImagingRawDataFile) f) this shoudl make the cast obsolete

src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
```diff
@@ -96,25 +97,24 @@
...
-      }
-      if (mz > max) {
-        max = mz;
-      }
+    if (series.getNumberOfValues() == 0) {
+      return Range.singleton(Float.NaN);
     }
-    return Range.closed((float) min, (float) max);
+    return Range.singleton((float) series.getMobility(0))
+        .span(Range.singleton((float) series.getMobility(series.getNumberOfValues() - 1)));
```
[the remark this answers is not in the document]
SteffenHeu: This way we don't have too loop trhough all values

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -0,0 +1,199 @@
...
+
+    // create a buffer with maximum dimensions
+    final int absMaxX = distinctFiles.stream()
+        .mapToInt(file -> file.getImagingParam().getMaxNumberOfPixelX()).max().orElse(0) + 1;
+    final int absMaxY = distinctFiles.stream()
+        .mapToInt(file -> file.getImagingParam().getMaxNumberOfPixelY()).max().orElse(0) + 1;
+
+    // invert x and y so we can easily loop while saving
+    final double[][] dataMatrix = new double[absMaxY][];
```
robinschmid, line 102: new double[absMaxY][absMaxX]  this should be fine - no need for a loop creating arrays with the same length

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -0,0 +1,199 @@
...
+        continue;
+      }
+
+      final ImagingRawDataFile rawDataFile = (ImagingRawDataFile) f.getRawDataFile();
+      final ImagingParameters imagingParam = rawDataFile.getImagingParam();
+
+      for (int i = 0; i < data.getNumberOfValues(); i++) {
+        ImagingScan scan = data.getSpectrum(i);
+        // x and y inverted!
+        dataMatrix[scan.getCoordinates().getY()][scan.getCoordinates().getX()] = data
```
robinschmid, line 130: is scan.getCoordinates always 0 based? Otherwise you would have to subtract the minimumX minmumY coordinate

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -0,0 +1,199 @@
...
+    try (BufferedWriter writer = Files
+        .newBufferedWriter(file.toPath(), StandardCharsets.UTF_8)) {
+
+      for(int y = 0; y < dataMatrix.length; y++) {
+        StringBuilder b = new StringBuilder();
+        for (int x = 0; x < dataMatrix[y].length; x++) {
+          if(Double.compare(dataMatrix[y][x], 0d) != 0) {
+            b.append(dataMatrix[y][x]);
+          }
+          b.append(", ");
```
robinschmid, line 172: only separate by "," and rather use your String sep - that is defined in the parameters.

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+        writer.append(b.substring(0, b.length() - 2));
```
robinschmid, line 174: the -2 was from the trailing ", " (with a space) now you would need to use sep.length.
However, I would just check if x<dataMatrix[y].length-1 before appending the separation string.

src/main/java/io/github/mzmine/util/files/FileAndPathUtil.java
```diff
@@ -414,4 +429,30 @@ public static File getPathOfJar() {
...
+    int i = 1;
+    File uniqueFile = new File(
+        noExtension.getAbsolutePath() + "(" + i + ")." + extension);
+    while (uniqueFile.exists()) {
+      i++;
+      uniqueFile = new File(
+          noExtension.getAbsolutePath() + "(" + i + ")." + extension);
+      if(i > 50) {
+        Random r = new Random(System.currentTimeMillis());
+        i = r.nextInt();
```
robinschmid, line 453: thats weird.
nextint might return some negative value and then you begin to loop until i>50.
You could make sure that you get a random int >50 here

src/main/java/io/github/mzmine/util/scans/SpectraMerging.java
```diff
@@ -303,14 +323,42 @@ public static MergedMsMsSpectrum getMergedMsMsSpectrumForPASEF(@Nonnull final Im
...
+  public static MergedMassSpectrum extractSummedMobilityScan(@Nonnull final ModularFeature f,
+      @Nonnull final MZTolerance tolerance, @Nonnull final Range<Float> mobilityRange,
+      @Nullable final MemoryMapStorage storage) {
+    final MobilityScan bestMobilityScan = IonMobilityUtils.getBestMobilityScan(f);
+
+    if (!(f.getFeatureData() instanceof IonMobilogramTimeSeries series)) {
+      return null;
+    }
+
+    final List<MobilityScan> scans = series.getMobilograms().stream()
```
robinschmid, line 343: wait this works? You can access pattern matched variable series outside of the if statement? 
I mean you do if( .... series) { return }  and then you access it

src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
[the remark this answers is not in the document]
robinschmid: Yes but why use span on two singletons if you could also use Range.closed(series.getMobility(0), series.getMobility(series.getNumberOfValues() - 1))

You could check 
if(size == 1) then Range.singleton
else Range.closed()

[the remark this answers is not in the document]
SteffenHeu: Only SummedIntensitySeries are guaranteed to be sorted in ascending order. IonMobilitySeries might be sorted in descending order due to the nature of TIMS.
Math.min/Math.max might be more performant tht generating two temporary objects, though, so I'll adjust it to that.

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
SteffenHeu: yes :D

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
SteffenHeu: smart

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
SteffenHeu: yes, see javadoc for Coordinates.

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -0,0 +1,199 @@
...
+    for (int i = 0; i < dataMatrix.length; i++) {
+      dataMatrix[i] = new double[absMaxX];
+    }
+
+    for (ModularFeature f : features) {
+      if (isCanceled()) {
+        return;
+      }
+
+      ArrayUtils.fill2D(dataMatrix, 0d);
```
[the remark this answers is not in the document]
SteffenHeu: We don't quantify in images and 0d values are not saved.

src/main/java/io/github/mzmine/util/scans/SpectraMerging.java
SteffenHeu: Java 16 feature for mapMulti i think

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
robinschmid: ok

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -163,15 +166,17 @@ private boolean writeImageToCsv(double[][] dataMatrix, ModularFeature f, File di
...
-      for(int y = 0; y < dataMatrix.length; y++) {
+      for (int y = 0; y < dataMatrix.length; y++) {
         StringBuilder b = new StringBuilder();
         for (int x = 0; x < dataMatrix[y].length; x++) {
-          if(Double.compare(dataMatrix[y][x], 0d) != 0) {
+          if (Double.compare(dataMatrix[y][x], 0d) != 0) {
             b.append(dataMatrix[y][x]);
           }
-          b.append(", ");
+          if(x < dataMatrix[y].length) {
```
robinschmid, line 175: this should be length -1 to not add the last separation

src/main/java/io/github/mzmine/modules/io/export_image_to_csv/ImageToCsvExportTask.java
```diff
@@ -163,15 +166,17 @@ private boolean writeImageToCsv(double[][] dataMatrix, ModularFeature f, File di
...
+            b.append(sep);
+          }
         }
-        writer.append(b.substring(0, b.length() - 2));
+        writer.append(b.toString());
         writer.newLine();
```
robinschmid, line 180: You could do b.newLine() and then only one writer.append(b.toString());
But not really important I think

src/main/java/io/github/mzmine/util/files/FileAndPathUtil.java
```diff
@@ -450,7 +450,7 @@ public static File getUniqueFilename(final File parent, final String fileName) {
           noExtension.getAbsolutePath() + "(" + i + ")." + extension);
       if(i > 50) {
         Random r = new Random(System.currentTimeMillis());
-        i = r.nextInt();
+        i = r.ints().filter(number -> number > 50).findFirst().orElse(0);
```
robinschmid, line 453: Still not sure. Feels unsafe.
Why not just loop through all ints until you find the next free file name?
If Random gives you Integer.MAX_VALUE and you add i++ in the next while loop you will throw an exception.

rand.nextInt(Integer.MAX_VALUE-52)+51 should do the trick. But then again - why?

robinschmid:
Thanks Steffen

PR #263, opened 2021-04-26
Plot updates, scatter pie charts

SteffenHeu:
- Remove MultiDatasetXYZScatterPlot
  - functionality integrated into SimpleXYZScatterPlot
- Mobility-m/z heatmap uses pie chart, can now plot aligned feature lists
  - added pie dataset, renderer, and provider to draw scatter-pie-charts
  - rows are rendered as pie, overall size = summed row intensity, individual slices are data file dependant
grafik
- the pie dataset, chart and renderer can be used in SimpleXYZScatterPlot
- added method to merge frames (but not feasibly to plot for now, too much data points)
- RegionSelectionWrapper now has buttons for loading and saving regions to an extra file

- SimpleXYZScatterPlot:
  - if dataset(s) with a paint scale are used, the paintscale of the first dataset will be used as a legend by default (still need to add this as javadoc)
  - a custom legend paint scale can be set
  - if the dataset is not a PaintScaleProvider, the default methods of the renderers are used. (as in pie datasets)


robinschmid (changes requested):

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/datasets/ColoredXYZPieDataset.java
```diff
@@ -0,0 +1,182 @@
...
+      maxZ = Math.max(zValue, maxZ);
+
+      for(int j = 0; j < sliceIdentifiers.length; j++) {
+        summedZValues[i] += getZValue(j, i);
+      }
+    }
+
+    domainRange = Range.closed(minDomain, maxDomain);
+    rangeRange = Range.closed(minRange, maxRange);
+    zRange = Range.closed(minZ, maxZ);
```
robinschmid, line 158: Same here. Range 0-1 if computedItemCount == 0

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/features/RowToMobilityMzHeatmapProvider.java
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  @Override
+  public double getDomainValue(int index) {
+    return rows.get(index).getAverageMZ();
+  }
+
+  @Override
+  public double getRangeValue(int index) {
+    return Objects.requireNonNullElse(rows.get(index).getAverageMobility(), 0f).doubleValue();
```
robinschmid, line 153: getAverageMobility is currently float - so null check does not work.
Why is the range method different to the domain method?

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/RegionSelectionWrapper.java
```diff
@@ -136,10 +178,83 @@ private void initSelectionPane() {
...
 
   public GridPane getSelectionControls() {
     return selectionControls;
   }
+
+  private void loadRegionsFromFile() {
+    MZmineCore.runLater(() -> {
+      final FileChooser chooser = new FileChooser();
+      chooser.getExtensionFilters()
+          .add(new ExtensionFilter("MZmine regions file", "*.mzmineregion"));
```
robinschmid, line 191: maybe change format name to something xml - makes it obvious that this is in xml format
mzmineregionxml

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYChart.java
```diff
@@ -241,7 +242,7 @@ public synchronized XYDataset removeDataSet(int index) {
     }
     plot.setDataset(index, null);
     plot.setRenderer(index, null);
-    notifyDatasetsChangedListeners();
+//    notifyDatasetChangeListeners();
```
robinschmid, line 245: Maybe there should be an notify option as a method argument

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYZScatterPlot.java
```diff
@@ -81,35 +89,39 @@
...
+  private XYShapeAnnotation currentRegionAnnotation;
 
   /**
    * Needs to be stored in case a separate legend canvas is used, so we can redraw the legend when
    * the canvas is resized.
    */
-  private PaintScaleLegend currentLegend = null;
+  private Title currentLegend = null;
+  @Nullable
+  private final ObjectProperty<PaintScale> legendPaintScale = new SimpleObjectProperty<>(null);
```
robinschmid, line 124: this property cannot be null - the annotations works on the property object and not on its value right?
Does it need to be a property (for JavaFX binding) or could it also be an Optional?

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYZScatterPlot.java
```diff
@@ -144,53 +155,100 @@ public SimpleXYZScatterPlot(@Nonnull String title) {
...
       getChart().fireChartChanged();
     });
   }
 
   /**
-   * @param dataset the dataset. null to clear the plot. Removes all other datasets.
+   * @param dataset the main dataset. null to clear the plot. Removes all other datasets.
    */
   public void setDataset(@Nullable ColoredXYZDataset dataset) {
 //    assert Platform.isFxApplicationThread();
```
robinschmid, line 233: remove this comment? 
and it should not be asserted at this point

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYZScatterPlot.java
```diff
@@ -144,53 +155,100 @@ public SimpleXYZScatterPlot(@Nonnull String title) {
...
 
     removeAllDatasets();
+    if (dataset == null) {
+      return;
+    }
+
     plot.setDataset(dataset);
```
robinschmid, line 240: synchronized is not needed if you then assert its on the FX Thread.
So you already force it to be on one thread - there is no way for concurrent modification ;)
However, do we really need to run it on the fx thread?

As far as I remember you change the dataset and then you still need to notify jfreechart to repaint the chart with the fire change event method. It should be fine to set the dataset outside of JavaFX and then call the change event

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYZScatterPlot.java
```diff
@@ -208,8 +266,9 @@ public void setDataset(T dataProvider) {
    */
   public synchronized int addDataset(XYZDataset dataset, XYItemRenderer renderer) {
     assert Platform.isFxApplicationThread();
```
robinschmid, line 268: copied from above:
synchronized is not needed if you then assert its on the FX Thread. So you already force it to be on one thread - there is no way for concurrent modification ;) However, do we really need to run it on the fx thread?
As far as I remember you change the dataset and then you still need to notify jfreechart to repaint the chart with the fire change event method. It should be fine to set the dataset outside of JavaFX and then call the change event

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYZScatterPlot.java
```diff
@@ -416,41 +502,37 @@ private void initializePlot() {
...
-      }
+
+    if(getXYPlot().indexOf((XYDataset) event.getDataset()) == 0) {
+      updateLegend();
     }
+
     MZmineCore.getConfiguration().getDefaultChartTheme().applyToLegend(chart);
-    chart.fireChartChanged();
+    if (chart.isNotify()) {
+      chart.fireChartChanged();
```
robinschmid, line 532: ok sounds about right. So this way you do not need to force the addDataset / setDataset methods on the fx thread. This part should handle it

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYZScatterPlot.java
SteffenHeu: JFreechart automatically notifies. However, when rendering the chart to an image, we don't want it to be on the jfx thread. So it's kinda hacky right now and i don't know what the best solution is.

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/features/RowToMobilityMzHeatmapProvider.java
SteffenHeu: Since we want to change to Float in the future, i wanted to make it future safe.

src/main/java/io/github/mzmine/datamodel/impl/SimpleMergedMassSpectrum.java
```diff
@@ -169,6 +171,9 @@ public MassList getMassList() {
 
   @Override
   public synchronized void addMassList(final @Nonnull MassList massList) {
+    if(massList == null) {
+      logger.info("blub");
```
robinschmid, line 175: remove logging

src/main/java/io/github/mzmine/util/scans/SpectraMerging.java
```diff
@@ -340,6 +337,7 @@ public static MergedMsMsSpectrum getMergedMsMsSpectrumForPASEF(@Nonnull final Im
           mzIntensities[0], mzIntensities[1],
           spectrum.getPrecursorMZ(), spectrum.getPrecursorCharge(), spectrum.getCollisionEnergy(),
           spectrum.getMSLevel(), sourceSpectra, mergingType, cf);
+//      mergedMsMsSpectrum.addMassList(new ScanPointerMassList(mergedMsMsSpectrum));
```
robinschmid, line 340: remove this line if you do not need it

robinschmid:
Thanks Steffen

PR #264, opened 2021-04-30
fix standard workflow to test jpackage

robinschmid:
Standard workflows are now test jpackage in gradle.build


tomas-pluskal:
Looks good to me, thanks!

PR #267, opened 2021-05-08
advanced tdf import

SteffenHeu:
added advanced tdf import

@robinschmid selection of .d folders does not work in your dialog. am i missing something?


tomas-pluskal:
Looks good to me.

PR #271, opened 2021-05-14
remove autorange filter on chart gestures

robinschmid:
Removed the autorange filter on zooming. 
We need to come up with a good way to set boundaries. Maybe "do not zoom to negative values" or similar

### DIFF
--- a/src/main/java/io/github/mzmine/gui/chartbasics/ChartLogics.java
+++ b/src/main/java/io/github/mzmine/gui/chartbasics/ChartLogics.java
@@ -1,16 +1,16 @@
 /*
  * Copyright 2006-2020 The MZmine Development Team
- * 
+ *
  * This file is part of MZmine.
- * 
+ *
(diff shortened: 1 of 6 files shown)


SteffenHeu (commented):
Thanks, Robin!
Was the same method used to restrict zooming on the domain axis?

src/main/java/io/github/mzmine/gui/chartbasics/ChartLogics.java
```diff
@@ -671,11 +671,13 @@ public static void offsetAxis(ValueAxis axis, double offset) {
   }
 
   public static Range keepRangeWithinAutoBounds(ValueAxis axis, Range range) {
+    // was not good for all charts - so better remove this function for now and think of a better way
+    return range;
```
SteffenHeu, line 675: What about
`return new Range(Math.max(range.getLowerBound(), 0d), Math.max(range.getUpperBound(), 1d));`

SteffenHeu:
Looks good to me

PR #273, opened 2021-05-17
Thermo raw file import on Mac & Linux

tomas-pluskal:
Using Mono and the ThermoRawFileParser library


tomas-pluskal:
This allows direct import of Thermo RAW files on 3 platforms. Unfortunately it doesn't work at the moment, I believe due to a bug in ThermoRawFileParser that I reported here: https://github.com/compomics/ThermoRawFileParser/issues/118

PR #289, opened 2021-06-17
Ims gapfill

SteffenHeu:
Multithreaded peak finder supports ims gap filling


SteffenHeu:
I reused the methods from the old Gap class in the ImsGap, but this required introduction of the GapDataPoint interface, so it supports mobilograms, too. 

I also cleaned the calculation of the properties like m/z, rt and so on for normal gap filling to use the FeatureDataUtils.

tomas-pluskal:
Looks good to me, but there is a merge conflict now

tomas-pluskal:
Thanks Steffen!

PR #291, opened 2021-06-23
Fixes: Contextmenu MSMS, Common Prefix, NPE

SteffenHeu:
- if a row type is selected in the feature table, "show most intense ms/ms" will not be disabled, but show the most intense msms of the row
- common prefix removal always removed prefixes, no matter which button was pressed
- npe in a chart provider


tomas-pluskal:
Thanks Steffen!

PR #299, opened 2021-07-08
fixed sirus MS/MS merge module for export

SteffenHeu:
In IMS multiple MS/MS spectra are acquired per MS2 frame per precursor. We merge all MS2 mobility scans during MS2 assignment and therefore get one summed MS2 spectrum per precursor per MS2 frame. Since it is an "artificial" spectrum, none of these summed MS/MS have a real "scan number" since they are not represented in the raw data file, which lead to issues during Sirius' MS/MS merge. 
This PR fixes the issue by extracting the mobility scans in the same mobility range from the preceeding/succeeding MS1 scan to assess the spectra chimerity with the sirius merge module.

EDIT: On some occasions this warning is displayed: 
[18:04:08|WARNING|io.github.mzmine.modules.tools.msmsspectramerge.MsMsSpectraMergeModule]: Scan -1 cannot be merged: it seems to belong to a different feature.

I noticed that it is due to the detected precursor charge in the raw data. Should we leave it as it is?
differentprecursorcharge


kaibioinfo:
Looks good to me. Do I understand that correctly, that also the MS1 scans can be merged from multiple IMS frames? At least the code reads like you select all MS1 scans preceeding the IMS frames and merge them into one scan. Your fix solve the problem with the scan numbers, so that should be no problem. It just sounds strange and extremely computational resource demanding to merge multiple MS1 scans (if that ever happens, maybe in your code you will always just have one MS1 scan and everything is fine).

PR #303, opened 2021-07-13
Fix multi tdf import

SteffenHeu:
Exception happened when importing multiple tdf files via All MS Data Import. The error was due to jdbc class. I synchronised the access and the TDF reader is now instantiated instead of a static class


tomas-pluskal:
Thanks Steffen!

PR #308, opened 2021-07-14
All ms ms - added mobilogram

SteffenHeu:
The "Show all MS/MS" window now also shows a mobilogram for the precursor. 

1. A raw mobilogram is summed over the retention time of the feature and the features' m/z **range** +- 5 ppm, (just like the EIC). 
2. The actual mobilogram of the feature is also displayed.

grafik

 The blue marker in the mobilogram shows the width of MS/MS window.


tomas-pluskal:
Great!

PR #315, opened 2021-07-16
Disable project menus

robinschmid:


tomas-pluskal:
This could be disabled directly in the FXML :)

PR #316, opened 2021-07-16
Mass list visualiazation for SpectraPlot in RawDataOverview

roman-bushuiev:
image

### DIFF
--- a/src/main/java/io/github/mzmine/modules/visualization/chromatogramandspectra/ChromatogramAndSpectraVisualizer.java
+++ b/src/main/java/io/github/mzmine/modules/visualization/chromatogramandspectra/ChromatogramAndSpectraVisualizer.java
@@ -26,6 +26,9 @@ import java.util.Hashtable;
 import java.util.List;
 import java.util.Objects;
 import java.util.logging.Logger;
+import javafx.beans.property.BooleanProperty;
+import javafx.scene.control.CheckBox;
+import javafx.scene.layout.HBox;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jfree.chart.plot.ValueMarker;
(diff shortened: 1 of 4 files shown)


SteffenHeu:
I think the mass list should only be displayed on demand and not by default, as this is is a "raw" data overview, mainly because it makes the plots look even more crowded.
Or even make it change between mass list or raw data.

A checkbox could be added next to the "AUTO" combo box.

tomas-pluskal:
I agree with @SteffenHeu , it would be nice to have a checkbox for showing the mass lists. Maybe the mass list can be shown not as peaks, but as filled dots (showing the position of each detected ion).

roman-bushuiev:
I agree, I'll add it :)

roman-bushuiev:
@tomas-pluskal I think the peaks are looking better than dots because even for the smallest size of dots it is too messy. (If I've understood it correctly)
image

tomas-pluskal:
It doesn't look so messy to me. I would keep the dots same color with the raw data. What do others think?

SteffenHeu:
Sorry for the late reply. I think if the dots are in the same colour as the raw data, it's really hard to distinguish it. 
I think sometimes we have to much data in the plots, so it becomes indistinguishable. Maybe displaying just the mass list or the raw data is better. Most of the time, the average user probably just want's to look at the spectra and does not really care if it's the mass list or the raw data. 
Maybe another combo box, that lets you select between "raw data", "mass list", and "both" is the best option to please everyone. And for the "both" case i would stick to the regular "theme" we use for mass lists, so just using a different colour. If it's just the mass list then it's good to use the raw data colour, i think.

tomas-pluskal:
I agree that when it's the same color, it is indistinguishable when you look at the spectrum as a whole. But I think that is fine, because of what you said - `Most of the time, the average user probably just want's to look at the spectra and does not really care if it's the mass list or the raw data.`
The only time when the user wants to see the mass list is when examining the mass detection results after zooming in on some specific peaks, and in that case the user can clearly distinguish the dots from the raw data trace.

roman-bushuiev:
Ok, I'll do it today and update the PR in the evening

roman-bushuiev:
What do you think about it?
image
image

tomas-pluskal:
I personally like it. I'll merge this for now so we have a way to show the mass list at the workshop. We can discuss the design and colors afterwards.

PR #317, opened 2021-07-16
Fixes

SteffenHeu:
- made the presentation mode a parameter in the preferences
- removed unnecessary processing in ImsGap
- Added the ADDUCT field to the sirius export


robinschmid:
This will conflict with my changes to the sirius export.
The correct field should be ION=
instead of adduct

PR #320, opened 2021-07-18
Wizard

SteffenHeu:
- adjusted Oribtrap parameters
- added licenses
- force disable of isotope mass detection in auto mass detector

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/auto/AutoMassDetector.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/auto/AutoMassDetector.java
@@ -20,6 +20,7 @@ package io.github.mzmine.modules.dataprocessing.featdet_massdetection.auto;
 
 import io.github.mzmine.datamodel.MassSpectrum;
 import io.github.mzmine.datamodel.MassSpectrumType;
+import io.github.mzmine.modules.dataprocessing.featdet_massdetection.DetectIsotopesParameter;
 import io.github.mzmine.modules.dataprocessing.featdet_massdetection.MassDetector;
 import io.github.mzmine.modules.dataprocessing.featdet_massdetection.centroid.CentroidMassDetector;
 import io.github.mzmine.modules.dataprocessing.featdet_massdetection.centroid.CentroidMassDetectorParameters;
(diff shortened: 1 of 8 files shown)


src/main/java/io/github/mzmine/modules/tools/batchwizard/defaults/DefaultMsParameters.java
```diff
@@ -13,12 +31,12 @@
       1E2, 1E3, new MZTolerance(0.005, 10), new MZTolerance(0.005, 3), new MZTolerance(0.005, 8));
 
   public static final DefaultMsParameters defaultOrbitrapPositiveParameters = new DefaultMsParameters(
-      1E4, 5E3, 5E4, new MZTolerance(0.005, 5), new MZTolerance(0.005, 3),
-      new MZTolerance(0.005, 5));
+      1E5, 3E3, 3E5, new MZTolerance(0.0015, 10), new MZTolerance(0.0015, 5),
```
tomas-pluskal, line 34: Robin set the 1E5 noise level as default for the workshop, to make processing faster. But I think 1E4 is better for real applications 1E4 is better to catch small peaks and isotopes.

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/auto/AutoMassDetector.java
```diff
@@ -44,16 +62,21 @@
   }
 
   private ParameterSet getExactParam(ParameterSet autoParam) {
-    exactMassDetectorParameters = new ExactMassDetectorParameters();
+    exactMassDetectorParameters = (ExactMassDetectorParameters) (new ExactMassDetectorParameters())
+        .cloneParameterSet();
     exactMassDetectorParameters.getParameter(ExactMassDetectorParameters.noiseLevel)
         .setValue(autoParam.getParameter(AutoMassDetectorParameters.noiseLevel).getValue());
+    exactMassDetectorParameters.setParameter(ExactMassDetectorParameters.detectIsotopes, false);
```
tomas-pluskal, line 69: The detectIsotopes is actually a nice feature for the high-res Orbitrap data, might be worth setting it to true (with default C, H, N, O, S, P elements)

SteffenHeu:
Also changed the default paint scale and made all color palettes available on first load

SteffenHeu:
Ok. I will change it.

What about the tolerances, should they also stay narrower as they were before, or should htey be widened?

tomas-pluskal:
The tolerance looks good to me

SteffenHeu:
Ok. I will remove P though, as it is monoisotopic.

PR #325, opened 2021-07-21
changed "..." buttons to "Setup"

SteffenHeu:


tomas-pluskal:
Wouldn't it be better to use something like "Browse" or "Select" instead of "Setup"? We are not really setting up filenames.

src/main/java/io/github/mzmine/parameters/parametertypes/filenames/FileNameComponent.java
```diff
@@ -53,7 +52,7 @@ public FileNameComponent(int textfieldcolumns, List<File> lastFiles, FileSelecti
     // on click - set file name to textField
     btnLastFiles = new LastFilesButton("last", file -> txtFilename.setText(file.getPath()));
 
-    Button btnFileBrowser = new Button("...");
+    Button btnFileBrowser = new Button("Select");
```
robinschmid, line 55: For file names we use select - so everything fine with me

robinschmid:
Seems fine with me.

tomas-pluskal:
There is a merge conflict

PR #328, opened 2021-07-27
Bugfixes

SteffenHeu:
- fixed loading/saving bug in RTTolerance 
- set correct num threads on tdf import
-  double click on FeatureShapeType provides same visualisation as "Show XIC" from menu
- dirty fix for no new tabs showing
- fix possible exception in ChromatogramVisualizerModule
- added FeatureUtils#rowToString 
- remove quality parameter calc after ims feature picking
- added try/catch block to FeatureTableDoubleClickTask


tomas-pluskal:
Thanks Steffen!

PR #334, opened 2021-07-29
Featuremodel objects

SteffenHeu:
All applicable methods return objects. Also hcanged the return types of double to Float where only a float value was stored anyway.

This depends on the annotations PR. Can pnly be merged if that pr is merged as is.


tomas-pluskal:
yay!

robinschmid:
Did you run it through your pipeline?

PR #340, opened 2021-08-02
MSDK removal

tomas-pluskal:
Since the MSDK project is stalled or dead, it makes updating some parts of MZmine very difficult because the code depends on code that is in MSDK jars. This PR completely removes the dependency on MSDK. There are still a few classes left from the MSDK data model, but I moved them into MZmine's datamodel package and those will gradually be removed, too.


tomas-pluskal:
Note: this also removes the SIRIUS module, because it doesn't work anyway and updating it would take too much work (I tried).  Exporting the data into SIRIUS is a better way to go, so the user can enjoy its full power including CANOPUS etc.

robinschmid:
I ran through my IIMN pipeline and it all works.
Will go through the changes later

robinschmid:
There is an MZmine/ Project file - maybe put in gitignore?
image

SteffenHeu:
In which modules do we need the MSDK types? If it's just the import, it would probably be better to put them in these packages to not confuse new users.

Can the msdk classes be packed into a module to restict their usage for all classes that don't need them?

robinschmid:
I think there are multiple modules that use the MSDK data structures. Not too many though.

tomas-pluskal:
It is my goal to remove the MSDK classes completely, I just need to need it in stages. This PR is just the first stage.

tomas-pluskal:
> There is an MZmine/ Project file - maybe put in gitignore?
> image

I don't think the Project file is included in the PR, it's maybe your local file?

src/main/java/io/github/mzmine/datamodel/RawDataFile.java
```diff
@@ -152,6 +153,7 @@
 
   ObservableList<Scan> getScans();
 
+  ObservableList<Chromatogram> getChromatograms();
```
SteffenHeu, line 156: Is this only for the import via msdk? So far we don't support reading chromatograms from raw files in mzmine.

tomas-pluskal: We don't at the moment, but we should in the future.. I thought it's good to keep it in there, since the code is already written in the mzML import module.

SteffenHeu: That's true. Maybe we should not make a new interface for it, though. The Chromatogram looks like the Feature interface to me, with reduced functionality. Also, i don't know if the List should be Observable, because that implies that the list might be subject to change, which it should not since it's loaded from the raw data.

tomas-pluskal: The Chromatogram might look like a Feature, but it actually represents raw data. In mzML, SRM spectra can be represented by chromatograms, because each spectrum contains only a single data point so it would be a waste of space to have all the <spectrum> tags. Maybe I can rename the Chromatogram to something different... like SRMChromatogram?

PR #341, opened 2021-08-03
Fast aligner

SteffenHeu:
Parallelised solution of the join aligner.

All rows are aligned to the full base feature list at once. The alignment happens iteratively.

In each iteration:
- The feature list with the most unaligned rows is chosen as a base list. 
  - the (remaining) features of this feature list are chosen as a base and added to the aligned feature list
- the rows from the new base list are removed from the unaligned rows
- the remaining unaligned rows are aligned to the full aligned feature list (this is done because the row m/z and rt might change after an alignment)
- the rows without an alignmed are used as unaligned rows for the next interation

Realigning on the full feature list is up for debate! Can also just align on the rows of the newly chosen base feature list. What do you think @robinschmid, @tomas-pluskal ?


tomas-pluskal:
I think we should not have two modules (Join aligner and Parallel Join aligner) that do exactly the same thing, that is confusing for the user. If this new module generates the same results as the old one, we should remove the old one.

SteffenHeu:
I agree. The results are very similar but not exactly the same, however, because the alignment happens for all raw data files in parallel and not consecutively.

tomas-pluskal:
> Should i delete the normal join aligner then?

I think so.

SteffenHeu:
Good?

robinschmid:
The join aligner produces reproducible results when run on the same feature lists in the same sorting....
I guess the fast aligner does not. This is a problem for some people - although I do not know how reproducible we are overall.

SteffenHeu:
Did you try it?
I just ran the aligner 4 times in a row and got the same number of features every time.

grafik

SteffenHeu:
Only the scoring happens in parallel. The actual assignment is sequential.

robinschmid:
I also tested it and it seems to prodcue reproducible results.

SteffenHeu:
Yes i tried that, the total number of rows is different, but not by much.

robinschmid:
This would be fine by me  then

PR #344, opened 2021-08-17
Fixes

SteffenHeu:
- fixed class cast exception in feature resolver
- fixed BatchLastFeatureLists for embedded parameter sets (added a tag interface)


SteffenHeu:
- fixed exception in tdf import in case of empty spectra 
- fix multiple additions of isomer type to flist

PR #347, opened 2021-08-23
MassQL support

robinschmid:
MassQL support in the RawDataOverview and FeatureTable

Filters for example for precursor mz, fragment ions, etc

For now we use the GNPS MassQL web API and a simple engine to support the main functionality.

```
QUERY scaninfo(MS2DATA) WHERE MS2MZ=318.244:TOLERANCEPPM=20
```


robinschmid:
Needs to be reworked with better support and use cases

PR #349, opened 2021-08-24
Streamline setting color for input files.

davidmasek:
Color for input files can now be selected with a single mouse click.

image


tomas-pluskal:
Thanks David, it looks great!

A few comments:
1. It would be ideal to directly show the palette of colors (as used in the ColorPicker component) in the menu:
Screen Shot 2021-08-25 at 15 13 51
JavaFX has the [CustomMenuItem class](https://openjfx.io/javadoc/16/javafx.controls/javafx/scene/control/CustomMenuItem.html), which allows embedding a Node into a menu, but unfortunately it is not going to work with the ColorPicker component directly. You would have to make a copy of the internal [PickerColorBox class](https://github.com/openjdk/jfx/blob/master/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ColorPickerSkin.java#L579), which is private. Still, it might be worth the extra effort for more streamlined color selection.
2. I think the hexadecimal RGB codes are not very meaningful for MZmine users, so we can remove them.
3. Please check if it works correctly when multiple files are selected in the MS data files list.

tomas-pluskal:
Note: you can check the MyCustomColorPicker code shown here: https://stackoverflow.com/a/27180647 
I guess that's basically what I meant.

davidmasek:
@tomas-pluskal 
For 1 and 2: I've looked into it and had a some trouble "porting" the code from the original color picker to display the color and I think it would basically be simpler to create the grid with colors ourselves. Based on the mentioned stackoverflow link I've created a demo which produces this: 
clrpicker
I think this would provide good workflow when display on hover over the "set color" menu item. It provides some basic colors for quick selection (color box at the top) or full color picker for users that want truly custom colors. I'm sending the picture to get some feedback before finishing. The required next steps are to add some button for confirmation (only required when selecting the custom color) and incorporating this as a menu item. It might also be nice to add second row (or more rows) of "preselected colors" that is configurable from code.

As for the 3rd point - the original version only allowed selecting colors when single file is selected so I've kept the checks and the menu item is disabled when more/less than 1 file is selected. What is the desired behavior here?

tomas-pluskal:
That looks great! As for point 3, I think the user may want to set the same color for a group of selected files (e.g., biological replicates), so please try to add that function, too.

roman-bushuiev:
Hi David, there is also a function to explicitly group selected files with `right click -> Group files`, which I was working on a while ago. If you would like to set the same color for each file in the group, then I think the easiest option is to check whether the selected item is an instance of [GroupEntity](https://github.com/mzmine/mzmine3/blob/19c6c8cc2da63dba163b524737941e4678ce4e24/src/main/java/io/github/mzmine/util/javafx/groupablelistview/GroupEntity.java) and then get all group files with [getGroupItems](https://github.com/mzmine/mzmine3/blob/19c6c8cc2da63dba163b524737941e4678ce4e24/src/main/java/io/github/mzmine/util/javafx/groupablelistview/GroupableListView.java#L216). If you have any troubles with the [GroupableListView ](https://github.com/mzmine/mzmine3/blob/19c6c8cc2da63dba163b524737941e4678ce4e24/src/main/java/io/github/mzmine/util/javafx/groupablelistview/GroupableListView.java)or anything else in general, feel free to contact me, maybe I would be able to help because I was starting to work on mzmine without mass spectrometry background as well.:smiley:

davidmasek:
Updates:
- added the custom color picker,
- enabled changing multiple files,
- changing grouped items also works (right click on the group name).

@tomas-pluskal  I think it should be ready to merge.

As a somewhat side note I saw both `rawDataList.getSelectedValues()` and `rawDataList.getSelectionModel().getSelectedItems()` used in the controller and I'm not sure what the difference is.

roman-bushuiev:
> As a somewhat side note I saw both rawDataList.getSelectedValues() and rawDataList.getSelectionModel().getSelectedItems() used in the controller and I'm not sure what the difference is.

`rawDataList.getSelectedValues()` should return only selected values (not group headers), while `rawDataList.getSelectionModel().getSelectedItems()` should return all selected items.

tomas-pluskal:
Thanks David. Could you please copy&paste the GPL license statement to all the new .java files?

tomas-pluskal:
Thanks David!

PR #351, opened 2021-08-25
update sirius export

SteffenHeu:
- all correlated features are exported, including IIN (annotation is provided)
- correlated peaks also exported if no ms2 merging is applied


src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTask.java
```diff
@@ -516,13 +556,87 @@ private void writeCorrelationSpectrum(BufferedWriter writer, Feature feature) th
...
+
+  /**
+   * Adds the isotopic peaks of this row to the list of data points.
+   */
+  private void addIsotopePattern(@NotNull Feature feature, @NotNull List<DataPoint> dps,
+      @Nullable IonIdentity identity, @Nullable IsotopePattern ip) {
+    if (ip != null) {
+      for (int i = 0; i < ip.getNumberOfDataPoints(); i++) {
+        // make sure to not export the molecular ion twice. Mass might change a bit due to smoothing
+        if (mzTol.checkWithinTolerance(feature.getMZ(), ip.getMzValue(i)) && identity != null) {
```
robinschmid, line 622: this check for identity!=null is weird here as it has nothing to do with the logic. 
You could check before calling the method and then get rid of the IonIdentity argument.

robinschmid, line 622: Also not sure about checking for the feature.getMZ() here..... I think we should just export the isotope pattern as it is. 
feature.getMZ() could also be the +2 peak of the pattern depending on your settings on how to pick the representative m/z and then you will discard a lot of signals maybe.

robinschmid (changes requested):

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTask.java
SteffenHeu: What do you mean? It only checks if there's a signal with the same m/z as the feature and and does not export it, if there's an ion identity for it, because it will be exported via the annotated export of the grouped features.

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTask.java
```diff
@@ -516,13 +556,87 @@ private void writeCorrelationSpectrum(BufferedWriter writer, Feature feature) th
...
+      final IonNetwork network = identity != null ? identity.getNetwork() : null;
+      for (final FeatureListRow groupedRow : group.getRows()) {
+        // only write intensities of the same file, otherwise intensities will be distorted
+        final Feature sameFileFeature = groupedRow.getFeature(file);
+        if (sameFileFeature == null
+            || sameFileFeature.getFeatureStatus() == FeatureStatus.UNKNOWN) {
+          continue;
+        }
+
+        if (row.equals(groupedRow) || group.isCorrelated(row, groupedRow)) {
```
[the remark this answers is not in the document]
SteffenHeu: We're both right and wrong. 
It's supposed to be row.equals(groupedRow) because i'm not sure if future implementations of group.isCorrelated will return true for the same row.
And you're right regarding the isotope pattern ;D

PR #353, opened 2021-08-29
fix ModuleComboParameter cloning

SteffenHeu:
When setting a new processing step to the parameter, the internal processing step (pair of module+param) was not updated. This lead to issues when cloning the parameter.


tomas-pluskal:
Looks good to me, thanks Steffen!

PR #354, opened 2021-08-29
Resolver update

SteffenHeu:
I rewrote the Resolver interface. Again. :D

This implementation should be clearer and easier to implement by future resolvers.
Furthermore, the ADAP Resolver now supports resolving mobility dimension.

@robinschmid For this to work properly, i had to adapt the implementation of the FeatureDataAccess to allow copying of all RT and intensity values into external arrays, so please have a look. I know it's not what we wanted to use the accesses for, but making another implementation that allows it is questionable. And for the resolvers it's nice to have all values available.

Still need to test fi the ADAP resolver produces the same results as in MZmine 2.


SteffenHeu:
ADAP Chrom Building        ADAP Resolving                     
#MZMine 3                9457                                 11111
#MZmine 2                9453                                  11103

Since the results of chrom building differ a bit, it's hard to say if it had an influence on the resolving, but it looks like its pretty much the same.

Screenshot of a randomly selected peak:
grafik

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/FeatureResolverTask.java
```diff
@@ -421,6 +399,7 @@ public void cancel() {
     }
   }
 
+  @Deprecated
   private FeatureList resolvePeaks(final ModularFeatureList originalFeatureList,
```
robinschmid, line 403: Maybe comment why deprecated and what its replaced by

robinschmid:
Not sure if I like the use of FeatureDataAccess... its meant to be a memory efficient way to iterate over all data points in all features / scans. 
Giving access to the underlying mz and intensity arrays is inconsistent and involves additional copies.

src/main/java/io/github/mzmine/datamodel/data_access/FeatureFullDataAccess.java
```diff
@@ -175,4 +175,14 @@ public Feature nextFeature() {
     return feature;
   }
 
+  @Override
+  public double[] getIntensityValues(double[] dst) {
+    return intensities;
```
robinschmid, line 180: This might be a problem if people do not assign the return value but rather work with the dst array. But on the other hand they should not do this in the first place.

src/main/java/io/github/mzmine/datamodel/data_access/FeatureFullDataAccess.java
```diff
@@ -175,4 +175,14 @@ public Feature nextFeature() {
...
+  }
+
+  @Override
+  public double[] getMzValues(double[] dst) {
+    return mzs;
```
robinschmid, line 185: This might be a problem if people do not assign the return value but rather work with the dst array. But on the other hand they should not do this in the first place.

src/main/java/io/github/mzmine/datamodel/data_access/FeatureFullDataAccess.java
```diff
@@ -175,4 +175,14 @@ public Feature nextFeature() {
     return feature;
   }
 
+  @Override
+  public double[] getIntensityValues(double[] dst) {
```
robinschmid, line 179: Add javadoc to the two methods as their are clearly discouraged because of the possibility of inconsistend data / array length

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/AbstractResolver.java
```diff
@@ -0,0 +1,240 @@
...
+      final int numValues = series.getNumberOfValues();
+      if (xBuffer == null || xBuffer.length < numValues) {
+        xBuffer = new double[numValues];
+        yBuffer = new double[numValues];
+      }
+
+      Arrays.fill(xBuffer, 0d);
+      Arrays.fill(yBuffer, 0d);
+      IonMobilityUtils.extractMobilities(series, xBuffer);
+      series.getIntensityValues(yBuffer);
```
robinschmid, line 216: I think you should always also asign the return value - just as a good practice as the method itself will create a new array if the provided is too small:
`yBuffer = series.getIntensityValues(yBuffer)`

PR #355, opened 2021-08-29
Add uppercase extension version to netCDF filter

davidmasek:
Adds uppercase version (".CDF") as valid file extension for netCFD files. This allows importing sample netCFD data (from https://drive.google.com/drive/folders/1HlMT0WPPjJ1VKCbz6qXYWvnsegcjn1pG), without changing to "any" allowed extension.

While this is minor change, I believe it could improve the experience for newcomers who use the sample data.

### DIFF
--- a/src/main/java/io/github/mzmine/modules/io/import_rawdata_netcdf/NetCDFImportParameters.java
+++ b/src/main/java/io/github/mzmine/modules/io/import_rawdata_netcdf/NetCDFImportParameters.java
@@ -30,7 +30,7 @@ import javafx.stage.FileChooser.ExtensionFilter;
 public class NetCDFImportParameters extends SimpleParameterSet {
 
   private static final List<ExtensionFilter> extensions = List.of( //
-      new ExtensionFilter("netCDF files", "*.cdf", "*.netcdf", "*.nc"), //
+      new ExtensionFilter("netCDF files", "*.cdf", "*.CDF", "*.netcdf", "*.nc"), //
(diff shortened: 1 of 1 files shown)


tomas-pluskal:
Thanks David. I guess this is only relevant on case-sensitive filesystems like Linux ext3 and such.. In any case, it's the correct way to do it. But why not add the other extensions too? (*.NETCDF, *.NC).

davidmasek:
I guess I got a bit confused when looking at how other modules handle this, as the module for MZML data actually uses mixed case (".mzML") for the extension definition. So I decided to add only the extension I actually saw in use.

I think the best solution would be to ignore case altogether, however it looks like this isn't possible with the `ExtensionFilter` class (and javafx by default). It should be easy to subclass the `ExtensionFilter` and create something like `CaseInsensitiveExtensionFilter`. I'm not sure how much of a use case it has, but given that it should be really simple I'd do it. What do you think?

tomas-pluskal:
Good idea, you should try it. It seems a bit stupid that the default ExtensionFilter is case sensitive...

davidmasek:
I missed that `ExtensionFilter` is `final`. I added uppercase versions for the module to at least be consistent. I don't think this currently has a better solution (using javafx).

@tomas-pluskal 
I suggest this to be merged as-is and we can return to this if it becomes an issue in the future.

tomas-pluskal:
Ok, sounds good.

PR #358, opened 2021-09-08
allow drag and drop in FileNamesComponent

SteffenHeu:


src/main/java/io/github/mzmine/parameters/parametertypes/filenames/FileNamesComponent.java
```diff
@@ -173,4 +175,38 @@ public void setToolTipText(String toolTip) {
...
+        StringBuilder sb = new StringBuilder(txtFilename.getText());
+        if(!sb.toString().endsWith("\n")) {
+          sb.append("\n");
+        }
+
+        filters.stream().flatMap(f -> f.getExtensions().stream())
+            .forEach(extension -> patterns.add(extension.toLowerCase().replace("*", "")));
+
+        for (File file : files) {
+          if (patterns.stream().anyMatch(filter -> file.getAbsolutePath().contains(filter))) {
```
robinschmid, line 200: you are converting the FileFilter to an extension string?
Why not use the method from the filter to check if the file matches?
file.contains is a weird way to check. Some file filters use .png and other png without the .
file.lowercase.endswith might be better but the filter should have a function

PR #360, opened 2021-09-09
add applied method on raw data import

SteffenHeu:
adds an applied method for raw data import


robinschmid:
A lot of style refoctoring. But simple change. Fine with me.

PR #361, opened 2021-09-09
Delta m/z histograms for correlated features

robinschmid:
This is a simple dialog that I needed to analyze the m/z deltas of correlated features. On the left you can see all m/z differences between all correlated feature pairs. On the right, its the m/z difference of all features correlated to an ion identity network (IIN). IIN gives us the neutral mass of the molecule and we basically calculate the m/z difference of features to this neutral mass. 

image


robinschmid:
Added more histograms for the Pearson correlation of feature shapes of grouped features.
Left: unidentified features 
Right: ion identities that belong to the same Ion Identity Network

This also provides an estimate of a good cutoff for feature shape correlation at r=0.85

image

PR #362, opened 2021-09-09
Feature table fix + Show iin chromatograms

SteffenHeu:
- fix subcolums not added to columnid map in feature table
-  double click on ion network id type opens chromatograms of all rows in iin

### DIFF
--- a/src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonNetworkIDType.java
+++ b/src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonNetworkIDType.java
@@ -18,10 +18,22 @@
 
 package io.github.mzmine.datamodel.features.types.annotations.iin;
 
+import io.github.mzmine.datamodel.RawDataFile;
+import io.github.mzmine.datamodel.features.Feature;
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonNetworkIDType.java
```diff
@@ -35,4 +47,32 @@ public String getHeaderString() {
...
+      return null;
+    }
+
+    return () -> {
+      final IonNetwork network = ionIdentity.getNetwork();
+      ModularFeature bestFeature = row.getBestFeature();
+      final RawDataFile bestRaw = bestFeature.getRawDataFile();
+
+      Map<Feature, String> labels = new HashMap<>();
+      List<Feature> features = network.keySet().stream().filter(r -> r.getFeature(bestRaw) != null)
```
robinschmid, line 66: not really important... but you could also simply map the entryStream to a map with stream.collect(Collectors.toMap(.....).
and then use the keySet() directly instead of the list of features or convert to list

robinschmid:
Good to me

PR #366, opened 2021-09-13
Adding getUniqueID() to all DataTypes

robinschmid:
DataType.getUniqueID() was added to all datatypes for this:
- key to store and read data types in xml/csv/... formats
- shall be final and not changed after we implement the saving loading of feature lists
- human-readable (retention time is still rt but I used collisional_cross_section instead of ccs, ion_mobility, ....)

Lets merge this asap and everyone can still propose changes to the ids until we have the final load/save. Then its final


SteffenHeu:
looks good to me

SteffenHeu (approved):

PR #368, opened 2021-09-15
Reactivate integration test

robinschmid:
Good news - the results from the small integration test is still the same after all this change


robinschmid:
okay integration test still failing onlne though

PR #391, opened 2021-10-24
save_import_lipid_annotation

Annexhc:


Annexhc:
> I noticed some more: The LipidAnnotationMsMsScoreType should extend the ScoreType (watch out, it's a float not a double so you have to cast it to float when setting it to the property) and the LipidMsOneErrorType can be replaced by the MzPpmDifferenceType.

I didnt go for the ScoreType, since I wanted to have custom number formatting

PR #397, opened 2021-11-01
fix peaks with overlapping ranges in adap resolver

SteffenHeu:
I noticed that resolved features in the ADAP resolver will sometimes overlap with specific wavelet settings.

Example see the blue and red features here, the blue also covers the range of the red feature.
grafik

With this fix, the ranges of the peak are not overlapping anymore.
grafik

@asmirn1 Can you have a look please? Should this rather be commited to the ADAP source?


asmirn1:
@SteffenHeu The screenshots have pretty low resolution, so I actually cannot see a difference before and after the fix. Do you have screenshots with better quality?

And would you mind attaching the MZmine project file or the raw data file so that I could reproduce this issue?

If this is an issue, I think it makes sense to move this fix into the ADAP package. I can do that.

SteffenHeu:
Hey, here are some zoomed in screenshots. I also added the rt boundaries to the chart legend so you can see the effect.
grafik

grafik

asmirn1:
Thank you @SteffenHeu. I see it now. I'll move your changes to the ADAP code.

SteffenHeu:
see #405

PR #399, opened 2021-11-01
add second preview to resolver dialog

SteffenHeu:
I always found it annoying during parameter optimisation that you have to jump back and forth between two EICs, so i added another preview chart. They both use the same feature list. Furthermore, a "good" and a "bad" EICs are are automatically selected. 
The bad EIC is determined by taking top 20 EICs (by area) and determining the lowest height/area ratio (low height and high area is most likely chromatographic noise). The good EIC is determined from the top 30 EICs (by area) and simply taking the heighest one.

grafik

Unfortunately there is currently no way to do the same for mobility dimension since we do not integrate that dimension.


Annexhc:
Thats a great feature!

PR #401, opened 2021-11-01
make BatchModeModulesList use ClassPath to get rid of manual list

SteffenHeu:
Hey @robinschmid / @tomas-pluskal, what do you think of this change? 

I used the ClassPath to enumerate all MZmineProcessingModules for the batch mode. (It's marked as `@Beta` but it's also used in the FeatureListLoadTask to enumerate all data types)


robinschmid:
Looks good to me

robinschmid:
Is this needed still or did we replace it @SteffenHeu

PR #405, opened 2021-11-05
Fix for the overlapping peaks in adap resolver #397

asmirn1:
- Move the Steffen's fix to the ADAP source code.
- Update the ADAP package to the most recent version.


SteffenHeu:
Thanks!

PR #406, opened 2021-11-07
Tdfoptimisation

SteffenHeu:
Did some magic.

Before (import + mass detection): (about 700mb total ram usage, about 580 due to the data file)
grafik

Now: (about 300 mb total ram usage)
grafik

Some processing (before):
grafik

Some processing (now):
grafik

Also made a super low ram usage implementation that loads mobility scan data on demand, but i think it's too hard to optimise to keep. Will probalby remove it after some more testing tomorrow.


Annexhc:
Looks very nice. Do you see any performance issues? Is the processing now slower?

SteffenHeu:
No, it's still blazingly fast. The data is still loaded and memory mapped as usual, just the MobilityScan and MassList are not created anymore. They are now generated on demand instead. 
This implies that MergedMsMs will have a list of unique MobilityScans, which will be generated every time a new MsMsSpectrum is created. But still uses heaps less ram than previously.

On the other hand, using the super low ram option that loads (uncompressesing and converting from raw data on demand) has huge performance implications, which is why i will probably not keep it.

robinschmid:
> No, it's still blazingly fast.

Blazingly fast is always a great description

src/main/java/io/github/mzmine/datamodel/impl/SimpleIsotopePattern.java
```diff
@@ -234,6 +234,18 @@ public Double getBasePeakIntensity() {
     }
   }
 
+  @Override
+  public void getMzValues(double[] dst, int offset) {
```
robinschmid, line 238: Do we actually need these two methods everywhere? Who is actually calling it? Maybe its enough to have it in the ion mobility specific classes? 

But you understand better the impact of that

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/MassDetectionTask.java
```diff
@@ -154,18 +153,12 @@ public void run() {
         // [mzs, intensities]
         double[][] mzPeaks = detector.getMassValues(data, massDetector.getParameterSet());
 
-        if (scan instanceof Frame) {
+        if (scan instanceof SimpleFrame frame) {
```
robinschmid, line 156: Why a SimpleFrame and not Frame?
Maybe getMobilityScanStorage should be pulled to Frame then?
or check in a different way?

src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_tdf/TDFImportModule.java
```diff
@@ -44,10 +44,9 @@
  */
 public class TDFImportModule implements MZmineProcessingModule {
 
-  private Logger logger = Logger.getLogger(this.getClass().getName());
-
   private static final String MODULE_NAME = "Bruker TDF file import";
   private static final String MODULE_DESCRIPTION = "This module imports raw data into the project.";
+  private Logger logger = Logger.getLogger(this.getClass().getName());
```
robinschmid, line 49: static

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/MassDetectionTask.java
SteffenHeu: We dont really need to expose the storage anyhwere else, so i don't see why it should be in the frame interface.

SteffenHeu: Maybe i can solve two issues at once, give me a second.

src/main/java/io/github/mzmine/datamodel/impl/SimpleFrame.java
```diff
@@ -103,10 +99,17 @@ public MobilityType getMobilityType() {
     return Range.singleton(0.0);
   }
 
+  public MobilityScanStorage getMobilityScanStorage() {
+    if(mobilityScanStorage == null) {
+      throw new IllegalStateException("Mobility scans not loaded");
```
robinschmid, line 104: Maybe add some javadoc and improve the error message?
When is it usally loaded?
Why is it null?
What do we ahve to do first?

src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_tdf/TDFImportTask.java
```diff
@@ -72,14 +75,13 @@
...
-  private final ParameterSet ms1DetectorParam;
-  @Nullable
-  private final ParameterSet ms2DetectorParam;
+  @Nullable private final MassDetector ms1Detector;
+  @Nullable private final MassDetector ms2Detector;
+  @Nullable private final ParameterSet ms1DetectorParam;
+  @Nullable private final ParameterSet ms2DetectorParam;
+
+  private final boolean denoising = false;
+  private static final double NOISE_THRESHOLD = 9E0;
```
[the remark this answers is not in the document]
SteffenHeu: It's deactivated und we have extended preferences.

src/main/java/io/github/mzmine/datamodel/impl/SimpleFrame.java
SteffenHeu: during file import, so this shouldnt occur.  It's more of an implication for us that somethign went wrong.

robinschmid:
All ready

PR #408, opened 2021-11-10
Safe name encoding for paths and filenames (raw / feature list)

robinschmid:


src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
```diff
@@ -464,7 +466,23 @@ public String getName() {
...
+        final List<String> names = new ArrayList<>(
+            project.getRawDataFiles().stream().map(RawDataFile::getName).toList());
+        names.remove(getName());
+        // make path safe
+        name = FileAndPathUtil.safePathEncode(name);
+        // handle duplicates
+        name =
+            names.contains(name) ? MZmineProjectImpl.getUniqueName(name, names) : name;
+      }
+    }
```
SteffenHeu, line 482: maybe out that in the raw data rename module? So the rename parameter is directly set to a safe name? Or call the module from here?

robinschmid: why would you want to set the parameter?
i think thats fine here?
Also it is analog to the way we do it in the FeatureList

SteffenHeu: but raw data files are reprocessed on project import so the name has to be reproducible. it does not have to be reproducible for feature lists, because they are loaded as-is.

robinschmid: Return real name - set to parameter in RawDataFIleRenameModule

robinschmid:
@SteffenHeu I did run into some problems. Renaming worked but:
I ran a standard LC-MS batch on two files, deleted a few feature lists, renamed the raw data files multiple times (also with duplicate names which were then replaced by uniques), renamed the feature lists, saved and loaded the project.

Project load only got to import the raw data files and then failed. Any idea?

SteffenHeu:
Mhh no. Tue raw data files parameter should create a RawDataFilePlaceholder with the name and path at evaluation time. So there shouldn't be a problem. Do simple rename steps work?

PR #413, opened 2021-11-12
Modification aware cosine

robinschmid:
- added modification aware cosine similarity as option to create molecular networks in MZmine
- fixed spectral library matching synchronization, MoNA import etc
- added options to display matches in network

image
image
image

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/msms/similarity/MS2SimilarityParameters.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/msms/similarity/MS2SimilarityParameters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2021 The MZmine Development Team
+ * Copyright 2006-2020 The MZmine Development Team
  *
  * This file is part of MZmine.
  *
@@ -8,12 +8,11 @@
  * License, or (at your option) any later version.
  *
  * MZmine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
- * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
- * General Public License for more details.
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
  *
  * You should have received a copy of the GNU General Public License along with MZmine; if not,
- * write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
- *
+ * write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
 package io.github.mzmine.modules.dataprocessing.group_metacorrelate.msms.similarity;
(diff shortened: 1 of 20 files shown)


src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/msms/similarity/MS2SimilarityParameters.java
```diff
@@ -37,38 +37,46 @@
...
       "Only best MS2 scan", "Compares only the best MS2 scan (or all MS2 scans)", true);
+
   public static final IntegerParameter MIN_DP = new IntegerParameter("Minimum data points (DP)",
       "Minimum data points in MS2 scan mass list", 3);
+
   public static final IntegerParameter MIN_MATCH = new IntegerParameter("Minimum matched signals",
       "Minimum matched signals or neutral losses (m/z differences)", 3);
-  public static final BooleanParameter CHECK_NEUTRAL_LOSS_SIMILARITY = new BooleanParameter(
+
+  public static final OptionalModuleParameter<NeutralLossSimilarityParameters> CHECK_NEUTRAL_LOSS_SIMILARITY = new OptionalModuleParameter<>(
```
SteffenHeu, line 70: why does this have to be an optional module? the parameterset only contains a single integer parameter.

src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/msms/similarity/MS2SimilarityTask.java
```diff
@@ -373,16 +399,47 @@ public void checkR2RMs2Similarity(R2RMap<RowsRelationship> mapSimilarity,
...
+   * @param overlapFunction different functions to determin the size of overlap
+   * @param sortedA         sorted array of data points (by intensity)
+   * @param sortedB         sorted array of data points (by intensity)
+   * @param precursorMzA    precursor mz of array sortedA, only used if useModAwareCosine is active
+   * @param precursorMzB    precursor mz of array sortedB, only used if useModAwareCosine is active
+   * @param mzTol           the tolerance to match signals
+   * @return the spectral similarity if number of overlapping signals >= minimum, else null
+   */
+  @Nullable
+  private SpectralSimilarity createMS2Sim(MZTolerance mzTol, DataPoint[] sortedA,
```
SteffenHeu, line 430: i think it should rater be createMS2Sim and createModAwareMS2Sim, highlighting the additional functionality, instead of having to explicitly exclude it.

src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/msms/similarity/MS2SimilarityParameters.java
robinschmid: It is just not very straightforward to have two parameters for this additional task. Even an OptionalParameter<IntegerParameter> would not be really nice for multiple reasons:
1. we activate a whole new matching type with the checkbox - this needs a name and description. The Integer needs a different description
2. We might add more parameters in the future... only because its just one parameter does not mean we should add very specialized parameters to the main parameters

src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/msms/similarity/MS2SimilarityTask.java
robinschmid: done

src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/msms/similarity/MS2SimilarityParameters.java
SteffenHeu: Ok, but whats the difference of CHECK_NEUTRAL_LOSS_SIMILARITY and MODIFICATION_AWARE_COSINE? I think the description does not really explain it.

PR #417, opened 2021-11-16
Add temp directory argument

robinschmid:
Currently, users would need to supply a modified config file to set the temp directory - that's very inconvenient.

Therefore, a new argument controls the temp folder and overrides the default and the one configured in the config.

--temp D:\tmpmzmine\

### DIFF
--- a/src/main/java/io/github/mzmine/main/MZmineCore.java
+++ b/src/main/java/io/github/mzmine/main/MZmineCore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2021 The MZmine Development Team
+ * Copyright 2006-2020 The MZmine Development Team
  *
  * This file is part of MZmine.
  *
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -148,21 +147,40 @@ public static void main(final String[] args) {
...
+
+    // override temp directory
+    final File tempDirectory = argsParser.getTempDirectory();
+    if (tempDirectory != null) {
+      // needs to be accessible
+      if (FileAndPathUtil.createDirectory(tempDirectory)) {
+        configuration.getPreferences().setParameter(MZminePreferences.tempDirectory, tempDirectory);
+        updateTempDir = true;
+      } else {
+        logger.log(Level.WARNING, "Cannot ");
```
SteffenHeu, line 175: Cannot?

robinschmid: done

PR #418, opened 2021-11-16
MemoryMapping option in program arguments also in configuration

robinschmid:
This brings the memory mapping options into the preferences.
If the --memory program argument is set it will override the configuration. Otherwise, the user can change the behavior at any time during the processing. All new data (spectra/features) will be handled accordingly.

### DIFF
--- a/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
+++ b/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2021 The MZmine Development Team
+ * Copyright 2006-2020 The MZmine Development Team
  *
  * This file is part of MZmine.
  *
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
```diff
@@ -118,13 +118,28 @@
...
           "Shows a warning message when a module without explicit ion mobility support is "
-              + "used to process ion mobility data."));
+          + "used to process ion mobility data."));
 
   public static final DirectoryParameter tempDirectory = new DirectoryParameter(
       "Temporary file directory", "Directory where temporary files"
-      + " will be stored. Requires a restart of MZmine to take effect",
+                                  + " will be stored. Directory should be located on a drive with fast read and write "
+                                  + "(e.g., an SSD). Requires a restart of MZmine to take effect (the program argument --temp "
+                                  + "override this parameter: --temp D:\\your_tmp_dir\\)",
```
SteffenHeu, line 127: will override / overrides

SteffenHeu (changes requested):

src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
robinschmid: done

PR #420, opened 2021-11-18
Preload libraries and keep

robinschmid:
I change the library matching to preload libraries and keep them. The whole library matching is too complex.


robinschmid:
Libraries are now imported and kept in memory. Matching is really fast now. Single spectrum against >100k library spectra (without optimization) was ~1-2 seconds. A whole feature list or selected rows is also fast. 

The whole module structure is now easier and more maintainable.

PR #421, opened 2021-11-18
fix load/save of RawDataFilesSelection after renaming

SteffenHeu:


robinschmid:
Tested and seems to fix the rename issues

PR #426, opened 2021-11-19
Remove old feature lists default

robinschmid:
Make it a default to encourage in batch mode


SteffenHeu:
I'm not sure if enabling the removal by default might throw the average user off if he cannot re-do a step because the parameters were off.
@tomas-pluskal / @Annexhc what do you think?

robinschmid:
My feeling is that this should be activated by default but no idea.

PR #430, opened 2021-11-23
Speed up isotope grouper with indexes, test 45s to 2s

robinschmid:
Speeds up the isotope grouper in my test from 45 s -> 2 s
when run on 7208 rows which are reduced to 5483 rows.

Uses two lists. One sorted by intensity for the order of main features. One sorted by m/z to only consider rows in the range and limit the number of rows accessed in the fitPattern method.

The results on my test dataset were the same rows.

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopegrouper/IsotopeGrouperTask.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopegrouper/IsotopeGrouperTask.java
@@ -35,6 +35,7 @@ import io.github.mzmine.parameters.parametertypes.tolerances.RTTolerance;
 import io.github.mzmine.parameters.parametertypes.tolerances.mobilitytolerance.MobilityTolerance;
 import io.github.mzmine.taskcontrol.AbstractTask;
 import io.github.mzmine.taskcontrol.TaskStatus;
+import io.github.mzmine.util.DataPointSorter;
 import io.github.mzmine.util.DataTypeUtils;
 import io.github.mzmine.util.FeatureListRowSorter;
 import io.github.mzmine.util.MemoryMapStorage;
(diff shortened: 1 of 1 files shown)


robinschmid:
Also interesting to see the flame graphs before and after. Because we looped over all rows for each candidate - the getAverageMz() was actually the most time-consuming method (considering it was called way to often). Now the graphs actually seem reasonable from the different shares. And we are down to ~2 seconds. So no need for further optimization.

### Before
image


### After
image

src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopegrouper/IsotopeGrouperTask.java
```diff
@@ -201,48 +199,44 @@ public void run() {
...
-        FeatureListRow p = originalRows[i];
-        isotopes[i] = new SimpleDataPoint(p.getAverageMZ(), p.getAverageHeight());
-      }
-      SimpleIsotopePattern newPattern =
-          new SimpleIsotopePattern(isotopes, IsotopePatternStatus.DETECTED, row.toString());
+      final DataPoint[] isotopes = bestFitRows.stream()
+          .map(r -> new SimpleDataPoint(r.getAverageMZ(), r.getAverageHeight()))
+          .toArray(DataPoint[]::new);
+      SimpleIsotopePattern newPattern = new SimpleIsotopePattern(isotopes,
+          IsotopePatternStatus.DETECTED, mostIntenseRow.toString());
```
SteffenHeu, line 214: Since the IsotopePattern is a MassSpectrum, we should also ensure that the first data point is of the lowest m/z. How is it currently handeled? So the either die data points or the bestFitRows should be sorted. 
If you sort the bestFitRows: You remove the bestFitRows from index 1, this might leave some leftover rows if they were sorted by m/z and a lower mass isotope was found.

robinschmid: Sorting the datapoints independently from the rows now.

PR #431, opened 2021-11-23
adapted FeatureListClearAnnotationsModule to modular data model

SteffenHeu:
also moved the menu item to the annotations menu


src/main/java/io/github/mzmine/parameters/parametertypes/ClearAnnotationsParameter.java
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * List of all annotation types
+   */
+  private static final List<DataType<?>> annotationTypes;
+  private static final String DATA_TYPE_ELEMENT = "datatype";
+  private static final String DATA_TYPE_ATTR = "type";
+
+  static {
+    final List<DataType<?>> types = new ArrayList<>();
```
robinschmid, line 51: I think we should limit our use of static constructs with reflection and this kind of heavy lifting.
You could just access all the types from DataTypes (which stores a list of singletons)

robinschmid: Maybe even convert it to a local list - this module is only used every now and then and filtering the list of datatypes for annotations should not be a big problem when the task is executed

robinschmid:
looks good

PR #438, opened 2021-11-24
toString features and rows

robinschmid:


robinschmid:
together with @SteffenHeu

PR #441, opened 2021-11-27
Find isotopes

robinschmid:
A simple module to find isotope pattern for all features in their MS1 mass lists. This is not about deciding if a signal is an isotope signal but more about finding all possible isotope signals based on elements and an MZtolerance.

The algorithm first goes back from the target signal to detect any possible preceding signals (e.g., if multiple Br isotopes are present or for large C numbers). Then we go through a second time to find all other isotope signals.

See the tests: a test for Br3C10 as a pattern works fine, regardless at which signal we start. 

### Additional changes
- Column to sort formulas by combined score
- fix sorting by absolute value for delta m/z
- show mirror match of isotope pattern
- fix style of mirror match


robinschmid:
image
image


Something is still wrong with the isotope pattern score though

robinschmid:
Correction. The isotope pattern is fine see resolved issue #442 
My parameters were wrong

robinschmid:
image

Another example of a pattern for a Cl2 containing compound. 

And matching spectral match / formula prediction:
image

src/main/java/io/github/mzmine/datamodel/data_access/ScanDataAccess.java
```diff
@@ -123,14 +108,58 @@ public double getMzValue(int index) {
...
 
+  /**
+   * Jumps to target scan. Uses hashmap that is lazily initialized. Use jumpToIndex if index is
+   * available.
+   *
+   * @param target jump to target scan
+   * @return true if target scan was available
+   */
+  public boolean jumpToScan(Scan target) {
+    return jumpToIndex(indexOf(target));
```
tomas-pluskal, line 127: What if indexOf(target) returns -1? There should be an exception or something.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderTask.java
```diff
@@ -0,0 +1,250 @@
...
+
+      if (scanRange == ScanRange.SINGLE_MOST_INTENSE) {
+        // add isotope pattern
+        candidates.sort(new DataPointSorter(SortingProperty.MZ, SortingDirection.Ascending));
+        IsotopePattern isotopePattern = new SimpleIsotopePattern(
+            candidates.toArray(new DataPoint[0]), IsotopePatternStatus.DETECTED, "Isotope finder");
+        feature.setIsotopePattern(isotopePattern);
+        detected++;
+      } else {
+        // find pattern in FWHM
```
tomas-pluskal, line 169: That's a lot of commented-out code :-D

robinschmid: 🗡️ 
Can we keep it in there for now? The idea, in the end, is to average the isotope pattern across scans in the FWHM. Currently the user has only one option though.

SteffenHeu: You could just generate a merged spectrum using the spectra merging utils and create a merged spectrum in the m/z range you're looking for while still using the same code as usual.

src/main/java/io/github/mzmine/datamodel/data_access/ScanDataAccess.java
robinschmid: so indexOf return -1, as usual
and jumpToIndex returns false if there is no scan at the specified index
same for jumpToScan

src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderTask.java
robinschmid: Would be one way.... 
but we need to ensure that the signal ratios are optimized. Thats the most important part about the isotope pattern.
Especially the small signals might only be in one scan and we need to account for that.

robinschmid:
@tomas-pluskal @SteffenHeu ready from my side - if you think I answered the comments

SteffenHeu:
Looks ok to me so far

PR #444, opened 2021-11-30
Save logging file in home folder (easier to access for users)

robinschmid:
Save the mzmine.log in the user directory - easier than Temp/Roaming/Local what ever on windows.

Also add number to allow multiple instances of MZmine to log


tomas-pluskal:
In MZmine GUI, there is an option Help - Show log file, that is not sufficient? 

This way the user's home folder will quickly get flooded by tons of mzmine log files, don't you think?

robinschmid:
Currently, the log file is only locked until MZmine exits. The next instance will then override the first log. The maximum number of logs stored is the maximum number of parallel running instances of MZmine.

The GUI is not the biggest problem... you see what is happing or if it crashed... but in headless you do not see anything - and hiding the log files in hidden folders might be weird. 
In any case, the "show log" option only shows the log of the first MZmine instance running. If people spin up multiple instances (or if one crashed in the background and still hangs around) the old logs would just be named: mzmine.log.1 etc.
The JUL logger is somehow weird not giving an option to access the file.

```
    /*
     * There doesn't seem to be any way to obtain the log file name from the logging FileHandler, so
     * it is hard-coded here for now
     */
    final Path logFilePath =
        Paths.get(System.getProperty("java.io.tmpdir") + File.separator + "mzmine.log");
```

PR #446, opened 2021-12-06
MSn tree visualizer

robinschmid:
This PR is mixed with the isotope finder PR #441 which should be merged first before reviewing this.

All spectra of level MS2, MS3, MS4 from top to bottom in one chart each.
image

image
 
The different energies are easy to compare in the charts.
image


src/main/java/io/github/mzmine/util/scans/ScanUtils.java
```diff
@@ -72,6 +78,11 @@
  */
 public class ScanUtils {
 
+  /**
+   * tolerance to compute and combine precursor m/z
+   */
+  public static final int DEFAULT_PRECURSOR_MZ_TOLERANCE = 100;
```
SteffenHeu, line 84: where and how is this used, it seems oddly high?

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
```diff
@@ -1006,6 +1018,62 @@ public static int findTopDataPoint(@NotNull double intensityValues[]) {
...
+        }
+      }
+      // sort
+      Collections.sort(result);
+      result.forEach(PrecursorIonTree::sort);
+    }
+    return result;
+  }
+
+  private static long getMzKey(double precursorMZ) {
```
SteffenHeu, line 1073: ah i get it. so this is the best solution? what about tzhe range maps we usually use?

if it's used this way, it's not really a tolerance as we usually use it. if it's only used here, maybe move it into this method or give it a more descriptive name.

SteffenHeu: maybe like PRECURSOR_ACCURACY_IN_DECIMAL_DIGITS = 3 and use 
`Math.round(Math.pow(10, PRECURSOR_ACCURACY_IN_DECIMAL_DIGITS - 1) * precursorMz)`

SteffenHeu:
also i'm missing a load/save test for this class.

robinschmid:
@SteffenHeu added test for MSnInfo - should be ready now

PR #453, opened 2021-12-16
[Tool] Formula predict

robinschmid:
Added a tool for formula prediction opening the window and updating on changes. Users can supply isotope pattern to get the isotope score.

image


robinschmid:
Needs some love

PR #459, opened 2021-12-19
linear interpolation for dia correlation

SteffenHeu:
@robinschmid what do you think?

It's some really simple stuff based on linear interpolation. 

Result:
grafik

Made it a subclass in the FeatureCorrelationUtils for now. Can probably be optimised a lot, but for now it does what it's meant to do.


SteffenHeu:
Question is, how we would want to approach building actual MS2 spectra for it. 

Easiest for us to implement would be making a MS2 feature list, but that probably gets really complicated for DIA PASEF or SWATH with acquisition windows in m/z and mobility dimension. 
Building it from actual spectra is a lot of computational effort and requires really clever caching or will lead to a lot of re-calculations. Maybe create row groups first and compute the ms2 traces for one group at once?

SteffenHeu:
todo: interpolate intensity in ms1 dimension for ms2 data points

SteffenHeu:
grafik
@robinschmid

PR #460, opened 2021-12-21
Fix mzml agilent ims import

SteffenHeu:
This fixes the agilent ims import from csv files (missing scans due to no data)

- index all mobility values in mzml -> get maximum number of mobility scans
- create a custom scan number for every mobility (starting from 0)
- get scan number by mobility value during import
- create empty scans in case there are scans missing


robinschmid:
Thanks Steffen

PR #466, opened 2021-12-23
Speed join aligner

robinschmid:
- Fixed some parts of the join aligner with @SteffenHeu 
- completely restructured a second version to compare speed

### TODO
- compare the two versions


robinschmid:
Found the discrepancy of the new method. Now the results are the same as the old one.
Ready from my side. Tests are running. @SteffenHeu

SteffenHeu:
Can merge after the comment is updated

SteffenHeu (changes requested):

PR #470, opened 2021-12-24
Options to KEEP, REMOVE, or directly run on original feature list

robinschmid:
Advanced option to directly run on the feature list to limit copying. When people select to remove the original feature list we do not really need to copy it...


robinschmid:
fixes zero height issue for gap filled features: #471

robinschmid:
fixes an issue introduced by scan data access pointing to the wrong scan in GapDataPoint #472

SteffenHeu:
looks good to me

PR #474, opened 2021-12-25
Add options to handle original feature lists

robinschmid:
- Most modules will only expose KEEP or REMOVE original feature list.
- PROCESS IN PLACE is available for: 
- Isotope grouper
- Gap fill
- Feature list rows filter
- Duplicate filter

The gain for feature list rows filter comparing PROCESS IN PLACE to the other options was a time reduction from 7 minutes --> 6 seconds. (no feature list copy)


robinschmid:
This PR replaces #426

SteffenHeu:
Is this ready @robinschmid ?

PR #475, opened 2021-12-27
Aird format support

lumiaoshan:
We have successfully completed the integration code for Aird format. After testing, using the Aird format of the same content, the loading speed of mzmine is 20-30 times that of mzML file while their analysis results are completely consistent.

Due to the good architecture of mzmine, Aird only adds dozens of lines of code to complete format compatibility.

We release a new version 2.1.1 of AirdPro: https://github.com/CSi-Studio/AirdPro/releases/tag/2.1.1
If you need all the zero intensity point and the needed cv params. You should set the conversion option with your requirements.
airdpro

Users can use this version(or above) to convert the raw files into aird format for testing.


SteffenHeu:
Hey lumaioshan,

thanks for contributing, it looks amazing!

I've looked over the code quickly and it looks good so far. One thing I'm missing is the information for MS2 scans such as precursor m/z, isolation window and collision energy. 
We have the classes DDAMsMsInfoImpl and MSnInfoImpl for that. It would be great, if you could include this information for the MS2 scans (package io.github.mzmine.datamodel.impl).

Also, if you want drag & drop import to work, you need to make a branch for the aird format in the RawDataFileTypeDetector and the AllSpectralDataImportModule.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+
+        setStatus(TaskStatus.PROCESSING);
+        logger.info("Started parsing file " + file);
+
+        DDAParser parser = new DDAParser(file.getPath());
+        AirdInfo airdInfo = parser.getAirdInfo();
+
+        try {
+            if (airdInfo == null) {
+                throw new Exception("Parsing Cancelled");
```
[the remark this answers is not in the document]
robinschmid: Oh I see its inside a try catch. 
Then you could do what I suggested or just give a more meaningful exception + message or add some logging with logging.warning("....");

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+                totalScans = airdInfo.getTotalScanCount().intValue();
+            }
+
+            List<DDAMs> cycleList = parser.readAllToMemory();
+
+            for (int i = 0; i < cycleList.size(); i++) {
+                DDAMs ms1 = cycleList.get(i);
+                HashMap<String, Object> ms1CvList = parseCVList(ms1);
+                SimpleScan ms1Scan = null;
+                if (ms1CvList != null && ms1CvList.size() != 0) {
```
robinschmid, line 110: In which case is the MS1CvList null or empty? Just wondering to understand when the else case will be triggered. Flagging all scans as profile and positive by default seems random. Depending on the case throwing an exception might be better?

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+    private final ParameterSet parameters;
+    private final Class<? extends MZmineModule> module;
+    private Logger logger = Logger.getLogger(this.getClass().getName());
+
+    private File file;
+    private MZmineProject project;
+    private RawDataFile newMZmineFile;
+    private int totalScans = 0, parsedScans;
+
+    public static final String Polarity_Type = "PolarityType"; // WINDOW_RANGE_HIGH
```
robinschmid, line 59: Maybe all those keys should be accessible from inside your library? Maybe something like MsCvKey.POLARITY_TYPE  could be an enum or just a class with final strings like here. 
This would make it future-proof in case more key are added or some are renamed.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+
+            for (int i = 0; i < cycleList.size(); i++) {
+                DDAMs ms1 = cycleList.get(i);
+                HashMap<String, Object> ms1CvList = parseCVList(ms1);
+                SimpleScan ms1Scan = null;
+                if (ms1CvList != null && ms1CvList.size() != 0) {
+                    ms1Scan = new SimpleScan(newMZmineFile, ms1.getNum(), MsLevel.MS1.getCode(),
+                            ms1.getRt(), null, ArrayUtil.fromFloatToDouble(ms1.getSpectrum().mzs()), ArrayUtil.fromFloatToDouble(ms1.getSpectrum().ints()),
+                            (MassSpectrumType) ms1CvList.get(Mass_Spectrum_Type), (PolarityType) ms1CvList.get(Polarity_Type),
+                            (String) ms1CvList.get(FILTER_STRING), Range.closed(Double.parseDouble((String) ms1CvList.get(LOWEST_MZ)), Double.parseDouble((String) ms1CvList.get(HIGHEST_MZ))));
```
robinschmid, line 114: For all (String) casts it might be better to call Object.toString() instead (e.g., ms1CvList.get(LOWEST_MZ).toString()). Then it does not break in case the datatype changes in the future.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+
+            List<DDAMs> cycleList = parser.readAllToMemory();
+
+            for (int i = 0; i < cycleList.size(); i++) {
+                DDAMs ms1 = cycleList.get(i);
+                HashMap<String, Object> ms1CvList = parseCVList(ms1);
+                SimpleScan ms1Scan = null;
+                if (ms1CvList != null && ms1CvList.size() != 0) {
+                    ms1Scan = new SimpleScan(newMZmineFile, ms1.getNum(), MsLevel.MS1.getCode(),
+                            ms1.getRt(), null, ArrayUtil.fromFloatToDouble(ms1.getSpectrum().mzs()), ArrayUtil.fromFloatToDouble(ms1.getSpectrum().ints()),
```
robinschmid, line 112: Are m/z values always stored with float precision or is there an option to store and load with double precision?

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+                parsedScans++;
+                newMZmineFile.addScan(ms1Scan);
+                if (ms1.getMs2List() != null && ms1.getMs2List().size() != 0) {
+                    for (int j = 0; j < ms1.getMs2List().size(); j++) {
+                        DDAMs ms2 = ms1.getMs2List().get(j);
+                        HashMap<String, Object> ms2CvList = parseCVList(ms2);
+                        SimpleScan ms2Scan = null;
+                        if (ms2CvList != null && ms2CvList.size() != 0) {
+                            ms2Scan = new SimpleScan(newMZmineFile, ms2.getNum(), MsLevel.MS2.getCode(),
+                                    ms2.getRt(), null, ArrayUtil.fromFloatToDouble(ms2.getSpectrum().mzs()), ArrayUtil.fromFloatToDouble(ms2.getSpectrum().ints()),
```
robinschmid, line 131: As @SteffenHeu pointed out, we need to add an DDAMsMsInfoImpl here instead of null. In case we can also load MS3 / MSn spectra we should add an MSnInfoImpl.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+                                    (MassSpectrumType) ms2CvList.get(Mass_Spectrum_Type), (PolarityType) ms2CvList.get(Polarity_Type),
+                                    (String) ms2CvList.get(FILTER_STRING), Range.closed(Double.parseDouble((String) ms2CvList.get(LOWEST_MZ)), Double.parseDouble((String) ms2CvList.get(HIGHEST_MZ))));
+                        } else {
+                            ms2Scan = new SimpleScan(newMZmineFile, ms2.getNum(), MsLevel.MS2.getCode(),
```
robinschmid, line 135: Similar to MS1, in which case is this part reached - maybe trow exception?

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+                        newMZmineFile.addScan(ms2Scan);
+                    }
+                }
+            }
+
+            newMZmineFile.getAppliedMethods().add(new SimpleFeatureListAppliedMethod(module, parameters, getModuleCallDate()));
+            project.addFile(newMZmineFile);
+
+        } catch (Throwable e) {
+            e.printStackTrace();
```
robinschmid, line 150: logger.log(Level.WARNING, "Error during aird import of file"+file.getName());
logger.log(Level.WARNING, e.getMessage(), e);

(The first is optional, the second will log the stack trace)

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+            /* we may already have set the status to CANCELED */
+            if (getStatus() == TaskStatus.PROCESSING) {
+                setStatus(TaskStatus.ERROR);
+                setErrorMessage(ExceptionUtils.exceptionToString(e));
```
robinschmid, line 154: the log above will log the stack trace you can set the error message to something readable meaningful - the reason why it failed.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+        setStatus(TaskStatus.FINISHED);
+
+    }
+
+    @Override
+    public String getTaskDescription() {
+        return "Opening file " + file;
+    }
+
+    public HashMap<String, Object> parseCVList(DDAMs ms) {
```
robinschmid, line 179: Why convert your internal CV params to a new map? This method could directly map your DDAMs scan to a SimpleScan by looping through all CV params extracting all needed values for SimpleScan

robinschmid:
Looks like a great addition. Thanks so much for the work on it. 
The whole implementation indeed looks seamless, which speaks for your aird library and for MZmine :)

I have added a few comments. That might further improve it.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: AirdPro allows users convert the raw data without PSI cv list. This may happened in some computation scene for DIA data. I will deal with this problem in a more friendly way. Thank you for your detailed suggestions.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: Aird only support for float precision.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
robinschmid: In this case there is no information stored on scan polarity or scan type (centroided or profile). In this case users will know what data they are dealing with and maybe they would have to supply the polarity and spectrum type?

lumiaoshan: This function is just used in Proteomics DIA data analysis, and assumed that users know the polarity(usually POSITIVE) and spectrum type(usually Profile) of DIA files they provide. I will add the necessary information here.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+    public static final String Polarity_Type = "PolarityType"; // WINDOW_RANGE_HIGH
+    public static final String Mass_Spectrum_Type = "MassSpectrumType"; // WINDOW_RANGE_HIGH
+    public static final String FILTER_STRING = "FilterString"; // WINDOW_RANGE_HIGH
+    public static final String LOWEST_MZ = "Lowest_Mz"; // LOWEST_MZ
+    public static final String HIGHEST_MZ = "Highest_Mz"; // HIGHEST_MZ
+
+    public AirdImportTask(MZmineProject project, File fileToOpen, RawDataFile newMZmineFile,
+                          @NotNull final Class<? extends MZmineModule> module, @NotNull final ParameterSet parameters,
+                          @NotNull Instant moduleCallDate) {
+        super(null, moduleCallDate); // storage in raw data file
```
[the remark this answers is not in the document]
lumiaoshan: done!

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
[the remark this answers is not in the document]
lumiaoshan: done!

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: done!

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: done!

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: done!

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: Aird currently does not support the calculation scenario of MSn(n>2) spectrum.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: done!

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: done!

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: done!

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: done!

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * @see Runnable#run()
+     */
+    @Override
+    public void run() {
+
+        setStatus(TaskStatus.PROCESSING);
+        logger.info("Started parsing file " + file);
+
+        DDAParser parser = new DDAParser(file.getPath());
```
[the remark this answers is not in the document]
lumiaoshan: I plan to integrate the DIA data mode for Aird in next 2 days as a try.

SteffenHeu:
I just noticed that htis will require a small adjustment to project load/save since the raw data is split into two parts, similar to imzML (Is probably a problem there, too). I will work on it, as soon as this PR is ready.

lumiaoshan:
The support for DIA data has been added in. Thanks for mzmine team's thorough advice!

tomas-pluskal:
Looks great, thanks a lot!

src/main/java/io/github/mzmine/util/RawDataFileType.java
```diff
@@ -23,18 +23,19 @@
...
-  MZDATA, //
-  NETCDF, //
-  THERMO_RAW, //
-  WATERS_RAW, //
-  ZIP, //
-  GZIP, //
-  ICPMSMS_CSV, //
-  BRUKER_TDF, //
-  BRUKER_TSF //
+    MZML, //
```
SteffenHeu, line 26: please use the intellij code formatter shipped with the repository.

lumiaoshan: done!

lumiaoshan:
Hi, I have already finished the integration for aird format. Is there any other work that I need to do?

robinschmid:
Hey sorry for the late response, I was tied up. Going through it now :)

One question, why is mavenCentral() commented out?

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    Range mzRange = null;
+    if (lowestMz != null && highestMz != null) {
+      mzRange = Range.closed(lowestMz, highestMz);
+    }
+
+    DDAMsMsInfoImpl msMsInfo = null;
+    if (msLevel == MsLevel.MS2.getCode()) {
+      msMsInfo = buildMsMsInfo(airdInfo, windowRange, parentScan);
```
robinschmid, line 260: is the MS2 precursor isolation window information available?

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
robinschmid: Thanks for clarification

robinschmid:
Fantastic work. Thank you for incorporating all my comments. The module is great. Testing it with a few files it worked like a charm. 
Three small items (from above) summarized:
1. why is mavenCentral commented out
2. do you have the MS2 precursor isolation window information - then we could add it to the MsMsInfo
3. While building your branch, intellij suggests to activate lombok. I know that we are not using it in MZmine. Are you using it in the aird library? Maybe we can try to exclude it within gradle? This should not be needed inside the class path right? What do you think @tomas-pluskal

lumiaoshan:
1. The repository "https://maven.aliyun.com/repository/public" is commonly used in China, as it mayby the fastest mirror maven repository in China. I have removed it as is is not necessary.
2. Done
3. Aird do use the lombok. I have released a new aird version which excludes the lombok package

lumiaoshan:
We are doing some result consistency test on TOF and Orbitrap data for DDA on mzmine to ensure the accuracy of Aird. I will update progress here upon finishing this work.

robinschmid:
Great thanks so much.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
```diff
@@ -203,16 +206,15 @@ private void loadAsDIA(DIAParser parser) throws IOException {
...
         parsedScans++;
         newMZmineFile.addScan(scan);
       }
     }
-
   }
 
   private SimpleScan buildSimpleScan(Spectrum spectrum, List<CV> cvList, WindowRange windowRange,
-      Integer num, float rt, int msLevel, Scan parentScan) {
+      Integer num, float rt, int msLevel, Scan parentScan, Boolean isMinute) {
```
robinschmid, line 217: Why Boolean isMinute? can it be null
So far I only see primitive booleans being passed to the method - if I dont miss anyhting.

Overall the changes look great though :)

robinschmid:
@lumiaoshan I saw you pushed some changes they are looking good. One small minor comment. 
Did you test both DIA and DDA?
Thanks so much for your work

lumiaoshan:
Sorry for late response. Our team are a lot of test on both DIA and DDA result. The EIC results are similar, but there are individual differences. Aird will have higher results on some data, some are not. There has seldom different protein point. We are continuing to investigate the reasons for the differences to see if it is just the problem of accuracy, and confirm that these accuracy have no impact on the final untarget search results.

src/main/java/io/github/mzmine/modules/io/import_rawdata_aird/AirdImportTask.java
lumiaoshan: Oh,yes, primitive is better here. Thanks for your advice

robinschmid:
Thanks for testing it.
So you are building chromatograms and finding features and the height is different? Or are all data point intensities off?

lumiaoshan:
Yes, we are building chromatograms, finding features with the ADAP algorithm and finding annotation with CAMERA method. We adjust the parameters according to the mzmine2 document, but the DDA file we use may be too large or the parameters are inappropriate. It takes about several hours to calculate one DDA file.We'll try to finish the test today

robinschmid:
Thanks for running all those tests. Do I understand the results correctly, that with different compression factors we still detect mostly the same number of extracted ion chromatograms?

We are not planning to publish any new releases for MZmine 2. So I think it does not make sense to create a PR. What do you think @tomas-pluskal 

@lumiaoshan can you elaborate on what functions of MZmine2 are not available in MZmine3?  Some parameters and some module names have changed so you would definitely need to create a new batch processing file. I was just running 800 very complex samples through MZMine3 in just one hour on a data analysis computer. Maybe some of your parameters are off or some of the modules you used are slow. 

At least for CAMERA in MZmine 3, it is not really maintained. We have metaCorrelate and Ion Identity Networking as alternatives. Or spectral library matching if you want to annotate compounds.

lumiaoshan:
1. Different precision will cause some small different EIC results. 5dp is an appropriate and sufficient precision.

2. The mass detection, ADAP Chromatogram builder is fast. But the ADAP resolver step is really slow. We use the test data(SA1.raw...) from the paper https://doi.org/10.1016/j.aca.2018.05.001.  Raw data was available at the XCMS website https://drive.google.com/drive/folders/1PRDIvihGFgkmErp2fWe41UR2Qs2VY_5G?usp=sharing_eip&ts=5b8ab35f (TripleTOF 6600 dataset ID:1197236, QE HF dataset ID: 1197351).

The analysis steps parameters:
Step1: Mass Detection
image
Step2: ADAP Chromatogram Builder
image
Step3: ADAP Resolver
image
My computer is Intel i9-10gen with 128GB memory.
The step3 needs about more than 5 hours for one file.

3. We just use CAMERA for test and have not used any other annatation method. When run CAMERA it would go for some NPE exception but in MZmine2 it wouldn't. After the fail attempt for CAMERA, we chose to compare the EIC results only.

Anyway. The work for Aird adaption is now ready. Thank you for your patient communication and responsible review. 
Our team is also focusing on the MS data analysis, including MS data compression, Proteomics with DIA, Metabolomics with DDA analysis platform. We are a relatively young team. The main technology stack is J2EE, so we have a lot to learn from mzmine team. Hoping to have deeper cooperation in the future.

robinschmid:
Definitely looking forward to collaborating more. It was a pleasure. And thank you for your thorough integration and testing. 

I think the ADAP deconvolution is great when tuned - but I think it also uses more processing time and maybe its performance has decreased because of all the changes to the internal data structures in MZmine 3. 
But maybe it is also because of the low values chosen in the chromatogram builder. I think the QExactive data has a higher noise level and they are only using 60k resolution for the MS1 scans. SO might be appropriate to use higher m/z tolerances for building chromatograms: e.g., 0.002, 15 ppm.

I am letting the workflows run one last time to then merge everything. 
Thanks @lumiaoshan

SteffenHeu:
Hi lumiaoshan,

the ADAP resolver usually takes longer to compute than other resolvers, but that long is quite unusual. But good to know, perhaps we should put some optimisation efforts there.

We usually use the local minimum resolver, because it is really fast.

tomas-pluskal:
Just curious: does the Aird format support ion mobility dimension?

tomas-pluskal:
> No. But Aird will support the ion mobility type format in the next two month with a new version.

That's great. Are you planning to also update the MZmine import module, once you have the new version with ion mobility support?

PR #476, opened 2021-12-30
Cyclic IMS support

SteffenHeu:
Draft PR until i know what's going on with the cyclic IMS data. 

In the converted mzML data, there were scans with and without mobility information. I'm guessing it's accumulated MS1/MS2 frames, but i'm not sure. 

This fix will ignore the accumulated frames and only import the mobility scans. (if the assumptions are correct)

- ignore scans without mobility information in mzml


SteffenHeu:
dealt with in #484

PR #478, opened 2022-01-03
add smiles to nist msp parser, add lipid blast support

SteffenHeu:
- the smiles tag was not imported from nist msp files
- lipid blast msp format uses tab instead of whitespace to separate mz/intensity
- lipid blast still uses precursortype over precursor_type, apperently contrary to nist


SteffenHeu:
included in #490

PR #483, opened 2022-01-04
Made import raw data files true by default, added description.

nilshoffmann:
Currently, the old way of importing mzTab(-m) files was to a) import the file as is, creating empty raw data files on the fly, or to b) import the files along with the raw data files (mzML). a) is not really a good solution any longer, since many other modules in mzmine expect the raw data / scans to be available. b) is therefor now the default, selected option. Help texts have been adapted accordingly to warn users about potential issues with other modules that expect access to scan data via raw data files.


src/main/java/io/github/mzmine/modules/io/import_features_mztab/MzTabImportParameters.java
```diff
@@ -36,7 +36,9 @@
 
   public static final BooleanParameter importrawfiles = new BooleanParameter(
       "Import raw data files?",
-      "If selected, raw data files will also be imported if they are available. If some raw data files cannot be found, empty files will be generated instead.");
+      "If selected, raw data files will also be imported if they are available. If some raw data files cannot be found, empty files will be generated instead. Please note that missing files may lead to issues with other modules that need access to scan data.",
```
SteffenHeu, line 39: since tooltips are not automatically wrapped, you can add some linebreaks.

src/main/java/io/github/mzmine/modules/io/import_features_mztabm/MzTabmImportParameters.java
```diff
@@ -37,7 +37,8 @@
 
   public static final BooleanParameter importRawFiles = new BooleanParameter(
       "Import raw files too?",
-      "If selected, raw data files will also be imported if they are available (in the msrun location in mzTab-m file or same folder as mzTab-m file in normal, or as zip or gz form). If some raw data files cannot be found, empty files will be generated instead."
+      "If selected, raw data files will also be imported if they are available (in the msrun location in mzTab-m file or same folder as mzTab-m file in normal, or as zip or gz form). If some raw data files cannot be found, empty files will be generated instead. Please note that missing files may lead to issues with other modules that need access to scan data.",
```
SteffenHeu, line 40: same

PR #484, opened 2022-01-04
CCS calibration

SteffenHeu:
- Refactored CCS calculation
- Added utilities to parse CCS reference list & construct calibration function
- Implementation done for linear relationships of CCS and mobility (DTIMS and TIMS)
- added a bunch of modules to set the ccs calibration
  - calibration is added to a raw data file via a specific module
  - therefore this should not lead to any problems with project load/safe

- Todo: implementation for Waters CCS calibration (could be done in a different PR)


src/main/java/io/github/mzmine/modules/dataprocessing/id_ccscalibration/external/ExternalCCSCalibrationTask.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    for (RawDataFile file : files) {
+      if (file instanceof IMSRawDataFile imsFile) {
+        imsFile.setCCSCalibration(calibration);
+        processed++;
+        progress = processed / (double) files.length;
+      }
+      // for reproducibility, this might take a bit longer.
+      SetCCSCalibrationModule.setCalibrationToFiles(files, calibration);
```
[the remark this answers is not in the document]
SteffenHeu: The loop above (line 65) sets the calibration to a file right away, without putting it into the applied methods. However, not adding it to the applied methods will lead to problems during project load.

Since this module might be executed in a batch and stuff on the FX thread is not executed right away, the call here sets it to the applied methods of the raw data file.

src/main/java/io/github/mzmine/modules/dataprocessing/id_ccscalibration/CCSCalibrant.java
```diff
@@ -0,0 +1,116 @@
...
+
+package io.github.mzmine.modules.dataprocessing.id_ccscalibration;
+
+import io.github.mzmine.datamodel.features.FeatureListRow;
+import java.util.Objects;
+import org.jetbrains.annotations.NotNull;
+
+public final class CCSCalibrant {
+
+  public static final double n2 = 28.006148008;
```
robinschmid, line 27: rename to N2_MASS

src/main/java/io/github/mzmine/parameters/parametertypes/CCSCalibrationComponent.java
```diff
@@ -0,0 +1,25 @@
...
+ *  You should have received a copy of the GNU General Public License along with MZmine; if not,
+ *  write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301
+ *  USA
+ */
+
+package io.github.mzmine.parameters.parametertypes;
+
+import javafx.scene.layout.BorderPane;
+
+public class CCSCalibrationComponent extends BorderPane {
```
robinschmid, line 23: why is this empty

SteffenHeu: because the calibration is not put into the gui, by the user. it is calculated or imported. The component class just exists because every parameter needs a component.

PR #486, opened 2022-01-05
add check if a batch is already running, block execution if true

SteffenHeu:
if there is already a batch task running, don't start a second one, since it will lead to issues with the BATCH_LAST_FILES and BATCH_LAST_FEATURELIST options.

@tomas-pluskal / @robinschmid ok?

### DIFF
--- a/src/main/java/io/github/mzmine/taskcontrol/impl/TaskControllerImpl.java
+++ b/src/main/java/io/github/mzmine/taskcontrol/impl/TaskControllerImpl.java
@@ -32,7 +32,7 @@ import io.github.mzmine.taskcontrol.TaskStatus;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
-import java.util.Vector;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.logging.Logger;
 
 /**
@@ -57,7 +57,7 @@ public class TaskControllerImpl implements TaskController, Runnable {
    * This vector contains references to all running threads of NORMAL priority. Maximum number of
    * concurrent threads is specified in the preferences dialog.
    */
-  private Vector<WorkerThread> runningThreads;
+  private CopyOnWriteArrayList<WorkerThread> runningThreads;
 
   /**
    * Initialize the task controller
@@ -67,7 +67,7 @@ public class TaskControllerImpl implements TaskController, Runnable {
     logger.finest("Starting task controller thread");
     taskQueue = new TaskQueue();
 
-    runningThreads = new Vector<WorkerThread>();
+    runningThreads = new CopyOnWriteArrayList<>();
 
     // Create a low-priority thread that will manage the queue and start
     // worker threads for tasks
--- a/src/main/java/io/github/mzmine/taskcontrol/impl/TaskControllerImpl.java
+++ b/src/main/java/io/github/mzmine/taskcontrol/impl/TaskControllerImpl.java
@@ -32,7 +32,7 @@ import io.github.mzmine.taskcontrol.TaskStatus;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
-import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.Vector;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/mzmine/taskcontrol/impl/TaskControllerImpl.java
```diff
@@ -256,4 +257,21 @@ public void addTaskControlListener(TaskControlListener listener) {
...
 
+  public boolean isTaskInstanceRunningOrQueued(Class<? extends AbstractTask> clazz) {
+    final WrappedTask[] snapshot = taskQueue.getQueueSnapshot();
+    for (WrappedTask wrappedTask : snapshot) {
+      if(clazz.isInstance(wrappedTask.getActualTask())) {
+        return true;
+      }
+    }
+
+    for (WorkerThread runningThread : runningThreads) {
```
robinschmid, line 268: lets change runningThreads to a CopyOnWriteArrayList to have full safety from concurrent modifications. Vector is basically deprecated since always. This should not change performance in any bit.   
https://stackoverflow.com/questions/2883140/vector-vs-collections-synchronizedlistarraylist

src/main/java/io/github/mzmine/taskcontrol/impl/TaskControllerImpl.java
```diff
@@ -256,4 +257,21 @@ public void addTaskControlListener(TaskControlListener listener) {
     listeners.add(listener);
   }
 
+  public boolean isTaskInstanceRunningOrQueued(Class<? extends AbstractTask> clazz) {
+    final WrappedTask[] snapshot = taskQueue.getQueueSnapshot();
```
robinschmid, line 261: This should be fine - its a synchronized list and therefore toArray is synchronized and should not break the iteration. Different with the Vector below though.

robinschmid (changes requested):
Great. Only change the Vector to CopyOnWriteArrayList. Use this rather than Collections.synchronizedList

SteffenHeu:
https://www.baeldung.com/java-copy-on-write-arraylist
The CopyOnWriteArrayList was created to allow for the possibility of safe iterating over elements even when the underlying list gets modified.

Because of the copying mechanism, the remove() operation on the returned Iterator is not permitted – resulting with UnsupportedOperationException.

robinschmid:
list.iterator().remove is not allowed.... but we do not use an iterator do we?
Otherwise you would need to use something like synchronizedList() instead of Vector and then toArray() before iterating over the collection

src/main/java/io/github/mzmine/taskcontrol/impl/TaskControllerImpl.java
```diff
@@ -173,8 +175,9 @@ public void run() {
       Iterator<WorkerThread> threadIterator = runningThreads.iterator();
       while (threadIterator.hasNext()) {
         WorkerThread thread = threadIterator.next();
-        if (thread.isFinished())
+        if (thread.isFinished()) {
           threadIterator.remove();
+        }
       }
```
SteffenHeu, line 181: @robinschmid  here's an iterator with remove.

SteffenHeu: we can use synchronized list, but what will it change, really? As the guy in your stackoverflow link pointed out, iteration + removal of collections (even if synced) is dangerous. That's why it is now copied.

src/main/java/io/github/mzmine/taskcontrol/impl/TaskControllerImpl.java
```diff
@@ -256,4 +261,22 @@ public void addTaskControlListener(TaskControlListener listener) {
...
 
+  public boolean isTaskInstanceRunningOrQueued(Class<? extends AbstractTask> clazz) {
+    final WrappedTask[] snapshot = taskQueue.getQueueSnapshot();
+    for (WrappedTask wrappedTask : snapshot) {
+      if (clazz.isInstance(wrappedTask.getActualTask())) {
+        return true;
+      }
+    }
+
+    var running = List.of(runningThreads.toArray(WorkerThread[]::new));
```
robinschmid, line 272: No need to wrap it in a list, you can diretly iterate over the array. toArray is synchronized

PR #482, opened 2022-01-04
Isotopefinder charge

SteffenHeu:
Added a method to deduce a charge state from a IsotopePattern found by the IsotopeFinderTask.

PR #488, opened 2022-01-06
Iso finder charge

robinschmid:
This is an alternative approach to PR #482 

- split the different mz diffs by charge - this will also remove possible false combinations between charges in isotope finder 
- set best charge state

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderTask.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderTask.java
@@ -41,8 +41,6 @@ import io.github.mzmine.taskcontrol.AbstractTask;
 import io.github.mzmine.taskcontrol.TaskStatus;
 import io.github.mzmine.util.DataPointSorter;
 import io.github.mzmine.util.IsotopesUtils;
-import io.github.mzmine.util.SortingDirection;
-import io.github.mzmine.util.SortingProperty;
 import it.unimi.dsi.fastutil.doubles.DoubleArrayList;
 import java.time.Instant;
 import java.util.ArrayList;
@@ -61,8 +59,6 @@ class IsotopeFinderTask extends AbstractTask {
   private static final Logger logger = Logger.getLogger(IsotopeFinderTask.class.getName());
   private final ModularFeatureList featureList;
 
-  private final DataPointSorter mzSorter = new DataPointSorter(SortingProperty.MZ,
-      SortingDirection.Ascending);
   // parameter values
   private final ParameterSet parameters;
   private final MZTolerance isoMzTolerance;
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderTask.java
```diff
@@ -111,25 +108,33 @@ public void run() {
     }
 
     // Update isotopesMzDiffs
-    List<Double> isoMzDiffs = IsotopesUtils.getIsotopesMzDiffs(isotopeElements, isotopeMaxCharge);
-    if (isoMzDiffs.isEmpty()) {
+    DoubleArrayList[] isoMzDiffsForCharge = IsotopesUtils.getIsotopesMzDiffsForCharge(
```
SteffenHeu, line 111: could be a double[][] if you want to optimize for performance.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderTask.java
```diff
@@ -144,26 +149,41 @@ public void run() {
...
       int scanIndex = scans.indexOf(maxScan);
       scans.jumpToIndex(scanIndex);
 
       // find candidate isotope pattern in max scan
-      List<DataPoint> candidates = IsotopesUtils
-          .findIsotopesInScan(isoMzDiffs, maxIsoMzDiff, isoMzTolerance, scans,
-              new SimpleDataPoint(mz, feature.getHeight()));
+      // for each charge state to determine best charge
+      // merge afterward to get one isotope patten with all possible isotopes
+      List<DataPoint> combinedChargesIsotopes = new ArrayList<>();
```
SteffenHeu, line 158: What's the benefit of merging multiple charge states? This will likely make the resulting isotope pattern unusable for any resulting module. I'd rather add the "best" isotope pattern (= most consecutive matches).

src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderTask.java
robinschmid: Was just to avoid List<Double>[] and double[][] which are both kind of weird

src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderTask.java
robinschmid: I want to have all possible isomers in the isotope pattern - we can then apply filtering later.
But rather have all signals without loosing important information - scoring algorithms will still work

SteffenHeu:
Approved after adding explicit disclaimer

robinschmid:
image

PR #490, opened 2022-01-06
add datamodel for ccs in spectral libraries

SteffenHeu:


src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/PercentTolerance.java
```diff
@@ -0,0 +1,67 @@
...
+
+package io.github.mzmine.parameters.parametertypes.tolerances;
+
+import com.google.common.collect.Range;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class PercentTolerance {
+
+  private final double tolerance;
```
robinschmid, line 27: Some javadoc at least describing that we are expecting the decimals. like tolerance=0.1 for 10%

robinschmid:
Approved after tests and javadoc

src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/RowsSpectralMatchTask.java
```diff
@@ -288,20 +301,54 @@ public void matchScan(List<SpectralDBEntry> entries, Scan scan) {
...
+        && ccsTolerance != null) {
+      if (ddaInfo.getPrecursorCharge() != null && (/*
+          mobScan.getDataFile().getCCSCalibration() != null // enable after ccs calibration pr is merged
+              ||*/ ((IMSRawDataFile) mobScan.getDataFile()).getMobilityType()
+          == MobilityType.TIMS)) {
+        precursorCCS = CCSUtils.calcCCS(ddaInfo.getIsolationMz(), (float) mobScan.getMobility(),
+            MobilityType.TIMS, ddaInfo.getPrecursorCharge());
+      }
+    } else if (scan instanceof MergedMsMsSpectrum merged
+        && merged.getMsMsInfo() instanceof DDAMsMsInfo ddaInfo && ccsTolerance != null) {
```
robinschmid, line 336: the ccsTolerance!=null is mentioned in 2 of 3 cases. This case would mean that we deactivated ccs matching correct? I think the whole method should just directly check this at the start and return null

robinschmid (changes requested):
Checked again after you merged the NIST parsing in. 
Found one little thing that seems to need adjustment

robinschmid:
Thanks Steffen

PR #491, opened 2022-01-06
add chart label font parameter, set in theme and renders

SteffenHeu:
Add chart label font parameter
grafik
grafik
grafik

### DIFF
--- a/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleChartUtility.java
+++ b/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleChartUtility.java
@@ -19,11 +19,15 @@
 package io.github.mzmine.gui.chartbasics.simplechart;
 
 import com.google.common.primitives.Ints;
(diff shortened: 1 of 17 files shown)


src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleChartUtility.java
```diff
@@ -94,4 +97,14 @@ public static boolean isLocalMaximum(XYDataset dataset, final int series, final
 
     return Ints.toArray(indices);
   }
+
+  public static void tryApplyDefaultChartThemeToRenderer(AbstractRenderer r) {
+    try {
+      final Font itemLabelFont = MZmineCore.getConfiguration().getDefaultChartTheme()
```
robinschmid, line 103: better do all the null checks and then use theme.applyToAbstractRenderer(); 
Then log the exception that might arise. Something might silently break here.

robinschmid:
Great for visualization

PR #494, opened 2022-01-06
Fix charttheme

SteffenHeu:
a parameter had a different name in ChartThemeParameters and ExportChartThemeParameters, leading to issues.

However, for some nontrivial reason there is a problem that the export chart theme behaviour is different than the normal parameters. (when clicking apply theme in the export dialog.)


robinschmid:
Can you delete the deprecated method public void applyToChart(JFreeChart chart) in both parametersets?

robinschmid:
Is it ready to merge?
Looks good to me

SteffenHeu:
yes

PR #495, opened 2022-01-07
update dependencies

SteffenHeu:
- updated some dependencies via gradle / intellij
- we should test if this breaks something, so far it did not for me


tomas-pluskal:
Looks good to me.

PR #499, opened 2022-01-10
Optimize ADAP chrom builder, RangeMap

robinschmid:
Chromatogram builder was using RangeSet<Double> and a HashMap<Range, Chromatogram>
I changed to single local RangeMap<Double, Chromatogram> and also cleaned up code.

No idea how to really check the maximum memory consumption or the total produced garbage.... But to me it definitely makes sense to use RangeMap instead of two separate classes.

### New RangeMap
image

### Old RangeSet + Map<>
image

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ModularADAPChromatogramBuilderTask.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ModularADAPChromatogramBuilderTask.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2020 The MZmine Development Team
+ * Copyright 2006-2021 The MZmine Development Team
  *
  * This file is part of MZmine.
  *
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ModularADAPChromatogramBuilderTask.java
```diff
@@ -417,5 +308,52 @@ public void run() {
...
+        toleranceRange.upperEndpoint());
+
+    // If both of the above ranges are null then we make the new range spaning the full
+    // mz tolerance range.
+    // If one or both are not null we need to properly modify the range of the new
+    // chromatogram so that none of the points are overlapping.
+    Double toBeLowerBound =
+        minusRange == null ? toleranceRange.lowerEndpoint() : minusRange.getKey().upperEndpoint();
+    Double toBeUpperBound =
+        plusRange == null ? toleranceRange.upperEndpoint() : plusRange.getKey().lowerEndpoint();
```
SteffenHeu, line 339: You could use the SpectraMerging.createNewNonOverlappingRange(rangeToChromMap, toleranceRange); here

robinschmid:
Will open a second PR with more clean ups of the ADAP chromatogram builder. 
Chromatograms had thousands of data types and methods but we only needed the TreeMap<Scan, DataPoint> all other datatypes are calculated later.

image

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ModularADAPChromatogramBuilderTask.java
robinschmid: below in this method there is a special case covered. I dont see when this should happen but I wanted to keep it just in case. Also this part is still concise and easy to maintain - I would keep it like this.

SteffenHeu: ok :+1:

robinschmid:
PR #504 contains these changes. Will be merged soon.

PR #500, opened 2022-01-10
set values for RowsFilter step in wizard

SteffenHeu:
- several NPE in the parameter set of the rows filter step from the batch wizard, because the "default" values were not set. 

- set default value = false for BooleanParameter
- add 
``` 
default <V, T extends UserParameter<V, ?>> void setParameter(OptionalParameter<T> optParam,
      boolean enabled, V value)
```
to ParameterSet for more convenient setting of optional parameters


robinschmid:
Looking good

PR #503, opened 2022-01-10
use positive and negative color for removed/remaining peaks

Sonja-analytik:
shoulder peak filter used orange and yellow, now using default colors


robinschmid:
Perfectly clean first PR thanks Sonja :)

PR #504, opened 2022-01-11
Clean adapChromatogramBuilder

robinschmid:
ADAPChromatogram was not optimal after moving to the new data model. 

This pull request changes the number of detected chromatograms but the old method to determine the number of connected data points > noise level seems wrong to me. 

The new implementation is easier and reflects what I understand from the parameters:

### minimum connected data points above noise level
before: intensity > noise
after: intensity >= noise

### minimum height: 
before: any data point in the chromatogram
after: the sequence of connected data points needs data point >= minimum height

### Ranges in TreeRangeMap
before: lower < value < upper bound
after:  lower <= value < upper bound

This means that if two ranges were colliding, the resulting new range that would be added to the map was leaving a hole to the next value. Now all values can be captured by ranges


SteffenHeu:
I'll look at it later.

Hey @asmirn1, what do you think of these optimisations?

robinschmid:
builds upon PR #499

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ADAPChromatogram.java
```diff
@@ -54,165 +35,65 @@
...
-  private int numOfCommittedSegments = 0;
-
-  // Isotope pattern. Null by default but can be set later by deisotoping
-  // method.
-  private IsotopePattern isotopePattern;
-  private int charge = 0;
-
-  // Victor Trevino
+  private final TreeMap<Scan, DataPoint> dataPointsMap = new TreeMap<>();
+  public int tmp_see_same_scan_count = 0;
```
SteffenHeu, line 41: where is this used?

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ModularADAPChromatogramBuilderTask.java
```diff
@@ -49,86 +49,66 @@
...
+        ADAPChromatogramBuilderParameters.minimumScanSpan).getValue();
 
     this.suffix = parameters.getParameter(ADAPChromatogramBuilderParameters.suffix).getValue();
 
     // Owen added parameters
-    this.IntensityThresh2 =
-        parameters.getParameter(ADAPChromatogramBuilderParameters.IntensityThresh2).getValue();
-    this.minIntensityForStartChrom =
-        parameters.getParameter(ADAPChromatogramBuilderParameters.startIntensity).getValue();
+    this.IntensityThresh2 = parameters.getParameter(
```
SteffenHeu, line 100: might aswell rename this to something more descriptive. minGroupIntensity?

src/test/java/FeatureFindingTest.java
```diff
@@ -395,7 +395,7 @@ void chromatogramDeconvolutionTest() throws InterruptedException {
         MinimumSearchFeatureResolverParameters.CHROMATOGRAPHIC_THRESHOLD_LEVEL, 0.8);
     generalParam.setParameter(MinimumSearchFeatureResolverParameters.dimension,
         ResolvingDimension.RETENTION_TIME);
-    generalParam.setParameter(MinimumSearchFeatureResolverParameters.MIN_ABSOLUTE_HEIGHT, 1E5);
+    generalParam.setParameter(MinimumSearchFeatureResolverParameters.MIN_ABSOLUTE_HEIGHT, 3E5);
```
SteffenHeu, line 398: Is this the only reason for the test failing or are there other reasons?

src/main/java/io/github/mzmine/util/FeatureConvertors.java
```diff
@@ -85,55 +87,30 @@
...
+    if (dataPoints.size() != scans.size()) {
       throw new IllegalArgumentException(
           "Number of data points does not match number of scan numbers");
     }
 
-    SimpleIonTimeSeries timeSeries = createSimpleTimeSeries(
-        ((ModularFeatureList) chromatogram.getFeatureList()).getMemoryMapStorage(),
-        chromatogram.getDataPoints().stream().collect(Collectors.toList()),
-        Arrays.asList(chromatogram.getScanNumbers()));
+    ModularFeature modularFeature = new ModularFeature(featureList);
```
SteffenHeu, line 100: You could use the 
`public ModularFeature(ModularFeatureList flist, RawDataFile dataFile,
      IonTimeSeries<? extends Scan> featureData, FeatureStatus featureStatus)`
constructor

src/main/java/io/github/mzmine/util/FeatureConvertors.java
```diff
@@ -85,55 +87,30 @@
...
+        new ArrayList<>(dataPoints), new ArrayList<>(scans));
     modularFeature.set(FeatureDataType.class, timeSeries);
 
     // recalculate data dependent types
     FeatureDataUtils.recalculateIonSeriesDependingTypes(modularFeature);
 
     ObservableList<Scan> allMS2 = Arrays.stream(
-            ScanUtils.findAllMS2FragmentScans(chromatogram.getDataFile(),
-                modularFeature.getRawDataPointsRTRange(), modularFeature.getRawDataPointsMZRange()))
+            ScanUtils.findAllMS2FragmentScans(dataFile, modularFeature.getRawDataPointsRTRange(),
```
SteffenHeu, line 112: i don't know if we should do this at chromatogram creation.

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ADAPChromatogram.java
robinschmid: you mean tmp_see_same_scan_count ?
I think right now its not used but we could log the number of "overlapping" data points

robinschmid:
The reason the tests are failing is that we detect a few less chromatograms - see described changes.

src/main/java/io/github/mzmine/util/FeatureConvertors.java
robinschmid: Why not? We have been doing this before. You can apply methods to chromatograms and you might stop at this point if you do direct injection

src/test/java/FeatureFindingTest.java
robinschmid: No there are a fewer chromatograms (~20 in a 1000)

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ADAPChromatogram.java
SteffenHeu: ahh ok the usage was collapsed in the github diff. we should keep it if we need to do some debugging

asmirn1:
Hi @robinschmid and @SteffenHeu 

Thank you for working on the ADAP Chromatogram builder. I have a few comments:

> minimum connected data points above noise level
> before: intensity > noise
> after: intensity >= noise

It makes sense to me.

> minimum height:
> before: any data point in the chromatogram
> after: the sequence of connected data points needs data point >= minimum height

In the ADAP Chromatogram Builder, we have two thresholds: Group intensity threshold and Min highest intensity. So I'm confused by your before and after. As I understand "before: any data point in the chromatogram" refers to the parameter Min highest intensity, and "after: the sequence of connected data points needs data point >= minimum height" refers to the parameters Group intensity threshold. Are you planning to keep only one parameter of the two?

> Ranges in TreeRangeMap
> before: lower < value < upper bound
> after: lower <= value < upper bound

It makes sense in general. However, I believe there was some reason why we used "lower < value < upper bound", but it was a few years ago.

Unfortunately, I don't have time to dig into the source code for ADAP Chromatogram builder right now. So you can go ahead with your changes. When I have time, I'll test the new algorithm.

The fact that you detect a few less chromatograms seems alarming to me though. Do you know the reason for it?

Alex

asmirn1:
@SteffenHeu I see. It makes sense to me now.

robinschmid:
@asmirn1 thanks for your feedback - I forgot to answer after discussing it with Steffen.

You have used RangeSet and for this you have to avoid overlapping or touching ranges to avoid the RangeSet from merging ranges into one.
We now use a TreeRangeMap, which does not merge ranges. 

I also think I have found the reason for slightly fewer chromatograms (~2%). The old version to check the number of consecutive data points above noise level does not check the intensity of the first data point:
```java
  public int findNumberOfContinuousPointsAboveNoise(double noise) {
    // sort the array containing all of the scan numbers of the point added
    // loop over the sorted array now.
    // if you find a point with intensity higher than noise start the count
    // if the next scan contains a point higher than the noise update the count
    // otherwise start it oer when you hit a sufficiently high point.
    // keep track of the largest count which will be returned.
    chromScanList.sort(Comparator.comparingInt(Scan::getScanNumber));

    int bestCount = 0;
    int curCount = 0;
    Scan lastScanNum;
    int scanListLength = chromScanList.size();

    Scan curScanNum;
    DataPoint curDataPoint;

    for (int i = 1; i < scanListLength; i++) {
      curScanNum = chromScanList.get(i);
      curDataPoint = dataPointsMap.get(curScanNum);

      if (curDataPoint.getIntensity() > noise) {
        lastScanNum = chromScanList.get(i - 1);
        int lastScanNumsActIndex = Arrays.binarySearch(scanNumbers, lastScanNum);
        Scan seqNextScanShouldBe = scanNumbers[lastScanNumsActIndex + 1];

        if (seqNextScanShouldBe == curScanNum) {
          curCount += 1;
          if (curCount > bestCount) {
            bestCount = curCount;
          }
        } else {
          curCount = 0;
        }

      } else {
        curCount = 0;
      }
    }
    // System.out.println("bestCount");
    // System.out.println(bestCount);

    // plus one because first point considered in advancing curcount is actualy going to be the
    // second point/
    return bestCount + 1;
  }
```

asmirn1:
@robinschmid 

> You have used RangeSet and for this you have to avoid overlapping or touching ranges to avoid the RangeSet from merging ranges into one.
> We now use a TreeRangeMap, which does not merge ranges.

Yes. This sounds right. I remember RangeSet would merge ranges.

> I also think I have found the reason for slightly fewer chromatograms (~2%). The old version to check the number of consecutive data points above noise level does not check the intensity of the first data point:

2% doesn't seem like a big change. Thank you for your work on this! The original ADAP Chromatogram builder wasn't written that well, so thank you for spending your time improving it.

robinschmid:
It definitely improved greatly on the existing chromatogram builder.

Okay will fix the tests then.

src/main/java/io/github/mzmine/util/FeatureConvertors.java
robinschmid: Done

src/test/java/FeatureFindingTest.java
robinschmid: Explained below. The old code was flawed

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ModularADAPChromatogramBuilderTask.java
robinschmid: done

robinschmid:
@SteffenHeu ready for the merge

PR #506, opened 2022-01-11
added parameter to apply mass detection to mobility scans or frames only

SteffenHeu:
- adapted mass detection parameter check (moved to check parameter values)
- fixed text wrapping for alertsfrom main window
- grafik

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/SelectedScanTypes.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/SelectedScanTypes.java
@@ -48,12 +48,4 @@ public enum SelectedScanTypes {
       case MOBLITY_SCANS -> scan instanceof MobilityScan;
     };
   }
-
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/SelectedScanTypes.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public boolean applyTo(Scan scan) {
+    return switch (this) {
+      case SCANS -> true;
+      case FRAMES -> scan instanceof Frame;
+      case MOBLITY_SCANS -> scan instanceof MobilityScan;
+    };
+  }
+
+  public boolean applyTo(SelectedScanTypes type) {
```
robinschmid, line 52: When is this used?
that looks to me like a simple equals check.

SteffenHeu: indeed

PR #510, opened 2022-01-12
Fix chart speed

robinschmid:
Fix chart speed, notify false, fix multiple tasks per chrom spec
### ChromatogramAndSpectraVisualizer
- every click in any chart was starting 2 tasks to update each plot - 4 in total
- added a small delay to simlify it
- set notify to false to only update the charts once everything is done


SteffenHeu:
I noticed it before, do you know why the click event is even triggered twice in the first place?

src/main/java/io/github/mzmine/modules/visualization/spectra/simplespectra/SpectraVisualizerTab.java
```diff
@@ -396,10 +397,14 @@ public void loadRawData(Scan scan) {
...
+    spectrumPlot.setNotifyChange(true);
+    spectrumPlot.fireChangeEvent();
   }
 
   public void loadPeaks(FeatureList selectedPeakList) {
-
+    spectrumPlot.setNotifyChange(false);
     spectrumPlot.removePeakListDataSets();
 
     if (selectedPeakList == null) {
```
SteffenHeu, line 410: if you set the notify to false and exit here, we will not get updates later on.

SteffenHeu (changes requested):
Will also have to manually test it, but the changes look good so far, only one change needed.

src/main/java/io/github/mzmine/modules/visualization/spectra/simplespectra/SpectraVisualizerTab.java
robinschmid: Done. Changed all to a try - finally

SteffenHeu:
is it ready for final review?

robinschmid:
Yes.

SteffenHeu (commented):
will do some testing now

SteffenHeu:
looks good

PR #513, opened 2022-01-13
propose code formatter change

SteffenHeu:
I found that the 'pure' google style (or at least what we adapted from the eclipse to intellij formatter conversion leads to quite ugly and not really readable code sometimes. Therefore, i adapted some small settings. 

The main change is, that it will not keep linebreaks on reformat, but will rearrange the code. I reformatted two files as an example.

### DIFF
--- a/mzmine-intellij-code-formater.xml
+++ b/mzmine-intellij-code-formater.xml
@@ -1,60 +1,12 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<code_scheme name="GoogleStyle">
+<code_scheme name="GoogleStyle" version="173">
   <option name="OTHER_INDENT_OPTIONS">
     <value>
       <option name="INDENT_SIZE" value="2" />
       <option name="CONTINUATION_INDENT_SIZE" value="4" />
       <option name="TAB_SIZE" value="2" />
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -136,9 +136,8 @@ public static void main(final String[] args) {
           getInstance().configuration.loadConfiguration(prefFile);
           updateTempDir = true;
         } catch (Exception e) {
-          logger
-              .log(Level.WARNING, "Error while reading configuration " + prefFile.getAbsolutePath(),
-                  e);
+          logger.log(Level.WARNING,
+              "Error while reading configuration " + prefFile.getAbsolutePath(), e);
```
SteffenHeu, line 140: This is an example for what i was talking about in the before version

mzmine-intellij-code-formater.xml
```diff
@@ -110,6 +118,7 @@
     </indentOptions>
   </codeStyleSettings>
   <codeStyleSettings language="JAVA">
+    <option name="KEEP_LINE_BREAKS" value="false" />
```
robinschmid, line 121: Important change

PR #518, opened 2022-01-17
Patch ccs library match

robinschmid:


src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/RowsSpectralMatchTask.java
```diff
@@ -372,20 +373,25 @@ public void matchRowToLibraries(List<SpectralDBEntry> entries, FeatureListRow ro
...
         for (int i = 0; i < scans.size(); i++) {
-          SpectralSimilarity sim = matchSpectrum(row.getAverageRT(), row.getAverageMZ(),
-              row.getAverageCCS(), rowMassLists.get(i), ident);
+          SpectralSimilarity sim = matchSpectrum(row.getAverageRT(), row.getAverageMZ(), rowCCS,
+              rowMassLists.get(i), ident);
           if (sim != null && (!needsIsotopePattern || checkForIsotopePattern(sim,
               mzToleranceSpectra, minMatchedIsoSignals)) && (best == null
-              || best.getSimilarity().getScore() < sim.getScore())) {
+                                                             || best.getSimilarity().getScore()
+                                                                < sim.getScore())) {
```
SteffenHeu, line 389: are you using the new formatter?

src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/PercentTolerance.java
```diff
@@ -38,54 +38,79 @@ public PercentTolerance(final double percentTolerance) {
...
+    if (base == null || value == null) {
+      return null;
+    }
+    return Math.abs(value - base) / base;
+  }
+
+  /**
+   * @return The error in per cent. 10 % = 0.1
+   */
+  public static Double getPercentError(@Nullable Double base, @Nullable Double value) {
```
SteffenHeu, line 82: i think returning different value types for different input types is unintuitive, although i must agree that it would be a precision loss. But i don't think that double precision is necessary here.

also i dont know if we should remove the primitives entirely, no use in boxing primitive values.

src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/RowsSpectralMatchTask.java
robinschmid: Nope. I want to switch over once we have most of the PR merged. Dont plan on having to much white space change (looking at you @SteffenHeu  :D)

src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/PercentTolerance.java
robinschmid: What do you mean?
I would keep all the methods and I would always return the same type. If someone is workign with floats you will get floats

SteffenHeu: Ok, but maybe we should use Double.POSITIVE_INFINITY then as return, to return a primitive? If someone does not check input types, he's probably not checking output types, too?

SteffenHeu: Maybe that will destroy project load/save though. Mh, i dont really know whats best.

robinschmid: You mean in the primitives version of this method?

PR #519, opened 2022-01-19
Filter 13C isotopes

robinschmid:
Added a more comprehensive 13C isotope filter to the rows filter. Previously, we only filter for minimum signals in an isotope pattern. For the Isotope finder, which is more inclusive, we need a better 13C filter.
- for each row's m/z we search for the +1 13C isotope peak in the isotope pattern and score the height. 
- there might be multiple signals in range so all are considered - only one needs to match the height requirements which derived from estimated min / max number of carbons.
- estimates the minimum and maximum number of 13 carbons based on mass defect and m/z 
- optionally looks for a -1 13C peak in the pattern (with matching height) to flag the row as an isotope and filter it
- optionally uses other elements to make sure we are not looking at a 18O or any other isotope signal (searching for the main signal at -X)
- added plenty of tests to check the performance with multiple charge states considered etc

image

### DIFF
--- a/src/main/java/io/github/mzmine/util/IntRange.java
+++ b/src/main/java/io/github/mzmine/util/IntRange.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2006-2021 The MZmine Development Team
+ *
+ * This file is part of MZmine.
+ *
+ * MZmine is free software; you can redistribute it and/or modify it under the terms of the GNU
+ * General Public License as published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * MZmine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with MZmine; if not,
+ * write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
+package io.github.mzmine.util;
+
+/**
+ * @author Robin Schmid (https://github.com/robinschmid)
+ */
+public record IntRange(int minInclusive, int maxExclusive) {
+
+  public static final IntRange EMPTY = new IntRange(-1, -1);
+
+  public boolean contains(int i) {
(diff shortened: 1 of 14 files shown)


robinschmid:
### Cummulative histogram of library matches that match isotope filters (13C = Isotope finder; 2iso = Isotope grouper)
image

image



### Cummulative spectral library matches matching the 13C filter (Isotope finder) or the 2 isotopes by isotope grouper
image

image

SteffenHeu:
What exactly are the graphs meant to mean? I dont see the colour of no match and match 2iso in the charts although theyre present in the legend

robinschmid:
Oh sorry forgot to say that.
So the no match was filtered. So I used 28 samples, each a mix of extracted compounds. I know what compounds to expect in which sample and filter for library matches to compounds that we expect in these samples. I removed unmatched from because I only wanted to look at those high quality matches.

match 2iso has no unique matches therefore its empty

robinschmid:
I updated the figures there was something wrong with my plotting script. Now it makes more sense
image

src/main/java/io/github/mzmine/util/IntRange.java
```diff
@@ -0,0 +1,39 @@
...
+ * write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
+package io.github.mzmine.util;
+
+/**
+ * @author Robin Schmid (https://github.com/robinschmid)
+ */
+public record IntRange(int minInclusive, int maxExclusive) {
```
SteffenHeu, line 24: Whats the benefit over the guava range?

SteffenHeu:
Maybe add it as a parameter to the spectral library matching?

robinschmid:
What do you mean by adding to as a parameter to spectral library matching?
This is a filter to not really a scoring method

src/main/java/io/github/mzmine/util/IntRange.java
robinschmid: what is the benefit of Range<Integer>.openClosed(min, max) over new IntRange(min, max) ?

robinschmid: primitives, record type, fixed behavior without all this open / closed stuff - the standard behavior of every index range

robinschmid:
Tested and ready from my side @SteffenHeu

PR #523, opened 2022-01-21
Split charge states in isotope patterns

robinschmid:
So far the isotope finder just throws all isotope signals into the same pattern. Now it uses a MultiChargeStateIsotopePattern and isotope pattern also carry information on the charge state.

relates to PR #519


SteffenHeu:
Also: No save/load test for MultiChargeIsotopePattern yet

SteffenHeu:
apparently not, so we need one too

robinschmid:
Added tests for simple and multi charge isotope pattern

robinschmid:
@SteffenHeu ready from my side

SteffenHeu:
thanks, robin!

PR #525, opened 2022-01-21
Rename isotope grouper

robinschmid:
Isotope grouper is a misleading name.
I would like to call it something like "13C isotope filter" or 13C feature filter


robinschmid:
With @SteffenHeu

tomas-pluskal:
But it’s not just for 13C, why limit the name to single isotope?

Tomas

PR #526, opened 2022-01-21
Add option to rows filter to keep all rows with MS2 no matter what

robinschmid:
Users might want to choose to keep features with MS2 and not apply as stringent filtering to those features.

Will also open one for the isotope grouper.


robinschmid:
with @SteffenHeu

PR #528, opened 2022-01-21
fix index out of bounds after resolving

SteffenHeu:
it's possible that there is no feature in the feature list prior to resolving (e.g. blanks in batch mode)


robinschmid:
What should be the response. Should we overall just not generate any new lists when its empty? Might be unintuitive maybe. At the same time its weird to run all methods on empty lists

SteffenHeu:
Mh good point. I think currently the new empty lists will be generated by default. So I don't feel like changing it

robinschmid:
Sounds good

PR #529, opened 2022-01-21
Refactor GNPS export

robinschmid:
The GNPS export used some unsafe methods to open / save files.
There are reports of mismatches between the number of exported features for mgf and csv export despite using the same rows filter for only those with MS2.

I will open a separate PR removing the best fragmentation DataType - we should only use the list of all fragment scans and sort it regarding the "best" scan = first in list.
Similar to Ion Identities and other List types


robinschmid:
included in #531 with more changes

PR #531, opened 2022-01-25
Fixes GNPS export in batch mode?

robinschmid:
Changing the iteration to a copy of featureList.rows fixes a problem where only in batch mode, the number of exported rows varies between MGF, CSV and modular CSV - all filtered to only contain those rows with MS2.

No idea why. Running the same FBMN export module manually on the same feature list results in all files with the same number of rows...
This sounds like concurrency problems in the batch mode. Some racing conditions.


SteffenHeu:
can merge after conflict is resolved

robinschmid:
With @SteffenHeu

PR #534, opened 2022-02-10
Export SIRIUS mgf harmonized

robinschmid:
@SteffenHeu I took your branch and added a few changes. I tested it with LC-MS data.

Replaces #351


src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
```diff
@@ -693,6 +693,22 @@ public IsotopePattern getBestIsotopePattern() {
     return null;
   }
 
+
+  @Override
+  public boolean hasIsotopePattern() {
```
SteffenHeu, line 698: I theory a row might now have an (invalid) isotope pattern but this function could return false.
I think the addIsotopePattern method should check if the added isotope pattern has more than one data point and not add it, if it doesnt.

SteffenHeu (commented):
Is it tested?

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
robinschmid: good idea

robinschmid: Removed the check here. It is checked in the isotope filter (grouper) and in the finder

robinschmid:
Tested and ready. It is more comparable to the GNPS export now. But we still need this export as SIRIUS needs other input (MS1 etc).

PR #537, opened 2022-02-15
MASST quickstart

robinschmid:
Directly submit MASST search from the feature table


src/main/java/io/github/mzmine/modules/io/export_features_gnps/GNPSUtils.java
```diff
@@ -166,24 +175,129 @@ public static String submitFbmnJob(@NotNull File mgf, @NotNull File quan, @Nulla
     return "";
   }
 
+
+  public static String submitMASSTJob(DataPoint[] data, double precursorMZ, MasstDatabase database,
```
SteffenHeu, line 179: does it have to be datapoint[]? can it be two double[]?

robinschmid:
Ready from my side

src/main/java/io/github/mzmine/modules/io/export_features_gnps/masst/GnpsMasstSubmitModule.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.logging.Logger;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Exports all files needed for GNPS feature based molecular networking (quant table (csv export)),
+ * MS2 mgf, additional edges (ion identity networks)
+ *
+ * @author Robin Schmid (robinschmid@uni-muenster.de)
+ */
+public class GnpsMasstSubmitModule implements MZmineProcessingModule {
```
SteffenHeu, line 55: Is it supposed to be executed in batch mode? as i understand it works on single spectra right? how would they be selected in batch?

src/main/java/io/github/mzmine/modules/io/export_features_gnps/masst/GnpsMasstSubmitModule.java
```diff
@@ -0,0 +1,108 @@
...
+import io.github.mzmine.util.ExitCode;
+import io.github.mzmine.util.scans.ScanUtils;
+import java.time.Instant;
+import java.util.Collection;
+import java.util.logging.Logger;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Exports all files needed for GNPS feature based molecular networking (quant table (csv export)),
+ * MS2 mgf, additional edges (ion identity networks)
```
SteffenHeu, line 51: update

src/main/java/io/github/mzmine/modules/io/export_features_gnps/masst/GnpsMasstSubmitModule.java
```diff
@@ -0,0 +1,108 @@
...
+  public String getName() {
+    return MODULE_NAME;
+  }
+
+  @Override
+  public Class<? extends ParameterSet> getParameterSetClass() {
+    return GnpsMasstSubmitParameters.class;
+  }
+
+}
```
SteffenHeu, line 108: shouldnt the formatter put a new line on the end of a file?

SteffenHeu (changes requested):

src/main/java/io/github/mzmine/modules/io/export_features_gnps/masst/GnpsMasstSubmitModule.java
robinschmid: done  -i thought I updated it everywhere

src/main/java/io/github/mzmine/modules/io/export_features_gnps/masst/GnpsMasstSubmitModule.java
robinschmid: Have to switch to the new formatter but there are a few things weird have to chekc with you

src/main/java/io/github/mzmine/modules/io/export_features_gnps/masst/GnpsMasstSubmitModule.java
robinschmid: Did not think about that. Changed it to runnable

src/main/java/io/github/mzmine/modules/io/export_features_gnps/masst/GnpsMasstSubmitModule.java
```diff
@@ -0,0 +1,93 @@
...
+  private static final Logger logger = Logger.getLogger(GnpsMasstSubmitModule.class.getName());
+
+  private static final String MODULE_NAME = "Submit GNPS MASST search";
+  private static final String MODULE_DESCRIPTION = "Submit an MS2 spectrum to GNPS MASST search, searching against public data.";
+
+  public static ExitCode submitSingleMASSTJob(double precursorMZ, MassSpectrum spectrum) {
+    return submitSingleMASSTJob(precursorMZ, ScanUtils.extractDataPoints(spectrum));
+  }
+
+  public static ExitCode submitSingleMASSTJob(double precursorMZ, DataPoint[] dataPoints) {
```
SteffenHeu, line 51: Are the results only presented on the webpage and not imported to mzmine?

robinschmid: This is the first step - this is to run a MASST job on GNPS. this takes a while - I could add a column with the MASST job link to the table and we can download the results later. MASST job is nice because its reproducible and you can share the link. 

However,  there is fastMASST now that has an API and is really fast. So next implementation step will be to use fastMASST to directly get the results and add them to the table

PR #541, opened 2022-02-15
Wizard MS2 pair

robinschmid:
The default settings for MS2 pairing were too wide - we used the width but it should actually be a tolerance


robinschmid:
Approved by @SteffenHeu

PR #544, opened 2022-02-16
Spectral library matching in batch mode

robinschmid:
The parameter to select spectral libraries was dependent on importing the libraries before starting the batch.
Now by default all imported libraries are used - otherwise users can specify what libraries to use.

Also added the import to the Advanced All Data Import. I guess that most users will import one or two big libraries (done with one task each). The module starts those tasks first and then adds the data import tasks. This should give better performance.

image


SteffenHeu:
awfully big window imo. can we make the buttons group in two columns?

robinschmid:
Actually fits easily on my screen with extra space.
I think it actually makes sense to have some height on both TextAreas. You can see more files / libraries then

robinschmid:
Same size like the Batch mode - seems fine to me. We could also change that in the future

image

SteffenHeu:
> Actually fits easily on my screen with extra space. I think it actually makes sense to have some height on both TextAreas. You can see more files / libraries then

Id say theres space for aroudn 15 libraries now, which is just a lot

robinschmid:
The GNPS libraries are fragmented into multiple parts and MassBankEU is one file per spectrum although you could convert folders to separate libraries:
https://gnps.ucsd.edu/ProteoSAFe/libraries.jsp

It think it does not really make a difference

robinschmid:
But if you like I can put the buttons on two columns

robinschmid:
Updated the layout 
image

PR #553, opened 2022-02-28
Fix zoom and spectrum select events in TIC / SpectraPlot

robinschmid:
Now we use the ChartGestureHandler - and click events are only used if the mouse was still during the event (no dragging)


SteffenHeu:
Works fine, thanks Robin!

PR #555, opened 2022-02-28
Final naming of DataType.uniqueID

robinschmid:
Those IDs should be as short as possible while still understood by most without much thinking. We will use them as column headers in the CSV export and in the project export. So they should not change after we release for good.

fwhm, ccs, ... should be okay to abbreviate in my view

### DIFF
--- a/src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonAdductType.java
+++ b/src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonAdductType.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2020 The MZmine Development Team
+ * Copyright 2006-2021 The MZmine Development Team
  *
  * This file is part of MZmine.
  *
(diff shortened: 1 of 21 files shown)


src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonAdductType.java
```diff
@@ -52,7 +53,7 @@ public String fromString(String s) {
   @Override
   public final String getUniqueID() {
     // Never change the ID for compatibility during saving/loading of type
-    return "ion_adduct";
+    return "adduct";
```
SteffenHeu, line 56: duplicate!

src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonTypeType.java
```diff
@@ -41,7 +41,7 @@
 
   @Override
   public @NotNull String getUniqueID() {
-    return "ion_type_type";
+    return "adduct";
```
SteffenHeu, line 44: duplicate!

SteffenHeu (changes requested):

src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonAdductType.java
robinschmid: is again ion_adduct

src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonTypeType.java
robinschmid: stayed "adduct" 
This is the one we are going to see most

src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonAdductType.java
SteffenHeu: ok 👍 Maybe annotate this with `@Deprecated` so it's not used anymore.

PR #562, opened 2022-03-02
fix invalid PaintScale range causing chart freeze

SteffenHeu:
it appears Double.MIN_VALUE caused an infinite loop. Increased to 0.1 now and added an assertion to check. If the assertion is triggered anyway, it should preventthe gui from crashing.

### DIFF
--- a/src/main/java/io/github/mzmine/gui/chartbasics/chartutils/paintscales/PaintScale.java
+++ b/src/main/java/io/github/mzmine/gui/chartbasics/chartutils/paintscales/PaintScale.java
@@ -91,11 +91,9 @@ public class PaintScale extends LookupPaintScale {
 
   private static double getValidUpperBound(double lower, double upper) {
     final double diff = upper - lower;
-    if (Double.compare(diff, 0d) != 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/mzmine/gui/chartbasics/chartutils/paintscales/PaintScale.java
```diff
@@ -86,4 +88,14 @@ public Paint getPaint(double value) {
 
     return super.getPaint(value);
   }
+
+  private static double getValidUpperBound(double lower, double upper) {
+    final double diff = upper - lower;
+    if (Double.compare(diff, 0d) != 0) {
```
robinschmid, line 94: Too complex for this simple thing and does not cover all cases.
You should only return upper if upper > lower
on Double.compare(lower, upper)>=0 -> return lower+0.1

SteffenHeu: This is ensured by the guava range, but ok ill change it.

PR #564, opened 2022-03-04
Tdf pseudo-profile import, allow re-merging of frame spectra

SteffenHeu:
- enabled pseudo-profile import for tdf raw data via command line "-tdfpseudoprofile" (no argument)
  - currently no reason to use this over the centroid import other than testing purposes -> disabled by default and not exposed to the users
- allow merging of mobility scans to frames even if frame data already exists

### DIFF
--- a/src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_tdf/TDFImportTask.java
+++ b/src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_tdf/TDFImportTask.java
@@ -290,7 +290,7 @@ public class TDFImportTask extends AbstractTask {
         setFinishedPercentage(0.1 * (loadedFrames) / numFrames);
         setDescription(
             "Importing " + rawDataFileName + ": Averaging Frame " + frameId + "/" + numFrames);
-        SimpleFrame frame = // profile frame import disabled
+        SimpleFrame frame =
             !importProfile ? tdfUtils.extractCentroidScanForTimsFrame(newMZmineFile,
                 frameId, metaDataTable, frameTable, framePrecursorTable, maldiFrameInfoTable,
                 ms1Detector, ms1DetectorParam, ms2Detector, ms2DetectorParam)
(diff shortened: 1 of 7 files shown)


src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_tdf/TDFImportTask.java
```diff
@@ -277,15 +281,22 @@ public void run() {
...
-        SimpleFrame frame = tdfUtils.extractCentroidScanForTimsFrame(newMZmineFile, frameId,
-            metaDataTable, frameTable, framePrecursorTable, maldiFrameInfoTable, ms1Detector,
-            ms1DetectorParam, ms2Detector, ms2DetectorParam);
+        SimpleFrame frame =
+            !importProfile ? tdfUtils.extractCentroidScanForTimsFrame(newMZmineFile,
+                frameId, metaDataTable, frameTable, framePrecursorTable, maldiFrameInfoTable,
+                ms1Detector, ms1DetectorParam, ms2Detector, ms2DetectorParam)
+                : tdfUtils.extractProfileScanForFrame(newMZmineFile, frameId, metaDataTable,
+                    frameTable, framePrecursorTable, maldiFrameInfoTable, ms1Detector,
+                    ms1DetectorParam, ms2Detector, ms2DetectorParam);
```
robinschmid, line 299: For those complex multiline arguments its sometimes better to just do an if else with a final SimpleFrame frame;

robinschmid:
Looks good to me. My comment is more a general one you can address it or leave it as it is

src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_tdf/TDFImportTask.java
SteffenHeu: true, will change it

robinschmid:
Thanks Steffen

PR #565, opened 2022-03-04
Apply sorting by RT and reset IDs 

robinschmid:
Chromatogram builder, feature resolver, and alignment creates new lists sorting should be applied and reset IDs.
@steffenheu maybe some other modules should also do this?


SteffenHeu:
They were ordered by mz (after chroms and resolving) by default, or? I thought this was quite intuitive.

PR #566, opened 2022-03-04
CSV legacy export add ion mobility

robinschmid:
Also added to GNPS FBMN / IIMN to support mobility in tables

### DIFF
--- a/src/main/java/io/github/mzmine/modules/io/export_features_csv_legacy/LegacyCSVExportTask.java
+++ b/src/main/java/io/github/mzmine/modules/io/export_features_csv_legacy/LegacyCSVExportTask.java
@@ -124,6 +124,15 @@ public class LegacyCSVExportTask extends AbstractTask implements ProcessedItemsC
   private void refineCommonElements() {
     List<LegacyExportRowCommonElement> list = Lists.newArrayList(commonElements);
 
+    boolean hasIonMobility = Arrays.stream(featureLists)
+        .anyMatch(flist -> flist.hasRowType(MobilityType.class));
+
+    if (!hasIonMobility) {
+      list.remove(LegacyExportRowCommonElement.ROW_ION_MOBILITY);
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/mzmine/modules/io/export_features_csv_legacy/LegacyCSVExportTask.java
```diff
@@ -122,6 +124,15 @@ public int getProcessedItems() {
   private void refineCommonElements() {
     List<LegacyExportRowCommonElement> list = Lists.newArrayList(commonElements);
 
+    boolean hasIonMobility = Arrays.stream(featureLists)
+        .anyMatch(flist -> flist.hasRowType(MobilityType.class));
+
+    if (!hasIonMobility) {
+      list.remove(LegacyExportRowCommonElement.ROW_ION_MOBILITY);
+      list.remove(LegacyExportRowCommonElement.ROW_ION_MOBILITY_UNIT);
+      list.remove(LegacyExportRowCommonElement.ROW_CCS);
```
SteffenHeu, line 133: In theory, one should not rely on the mobility unit of the row, but the feature. files from varying vendors might be aligned by ccs (todo for join aligner) and therefore have different mobility units.

SteffenHeu: i dont even know if this type is actually set

robinschmid: Most people will work with one type of instrument / mobility unit. So this should be fine.
I alos think people should work with each feature. But many want to use the average mobility. 

They should know that an average does not make sense when they use multiple units

robinschmid: Okay I will remove the unit which is not that important anyways and is also not set

Issue #567, opened 2022-03-06
MS/MS display


dwheeler23:
### Basic information

* My operating system: Windows 10.0.18363 Build 18363, 64 bit
* My MZmine version: MZMine 3.0.0beta

### What happened

Excellent new version! Unfortunately I can't get the MS/MS plot display to work (formerly called MS/MS visualizer). I've tried fiddling with all of the settings (axes, options, etc.), file selection, updating Java, etc. 

No error appears, it just doesn't fire up any visualizer once I click OK.

SteffenHeu:
Hey dwheeler, 

thank you for the bug reprot.
Can you please post the log output? You can find it via the "Show log file" button in the help menu.
Also: What kind of file raw data file are you using? Perhaps it might be good to upload the file somewhere so we can reproduce the error.

robinschmid:
Hey, thanks for the report. That's one of the modules I have never used or touched so far. 
There are plenty of options for MS/MS visualization but this one seems very different from the rest. 

During the change of the interfaces, we created the new visualizer but forgot to add it to the main window.  That's an easy fix. 
I suspect this will be in the next patch release 3.0.1-beta.

Keep up the good work and report what you experience.

Best Robin


![image](https://user-images.githubusercontent.com/10366914/157003955-e83bc85b-2596-47fe-9046-f579d1b296c6.png)

robinschmid:
Oh and updating java does not work because MZmine ships with its own java version. That's what modern java does.

dwheeler23:
Thank you so much! This was using Thermo RAW files. Again, excellent new version, I really appreciate the wonderful tool you have all made. 

Best,
Doug

PR #570, opened 2022-03-07
Fix ms/ms scatter plot

robinschmid:
@steffenheu why was this commented out in RawDataFilesSelection? Is it okay to do it this way?

fixes issue #567

### DIFF
--- a/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilesSelection.java
+++ b/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilesSelection.java
@@ -57,11 +57,15 @@ public class RawDataFilesSelection implements Cloneable {
   public RawDataFile[] getMatchingRawDataFiles() {
 
     if (evaluatedSelection != null) {
-      /*var value = Arrays.stream(evaluatedSelection)
-          .map(placeholder -> placeholder.getMatchingFile()).filter(Objects::nonNull)
+      var value = Arrays.stream(evaluatedSelection).map(RawDataFilePlaceholder::getMatchingFile)
           .toArray(RawDataFile[]::new);
-      return value;*/
-      throw new IllegalStateException("Raw data file selection has already been evaluated.");
+      for (var raw : value) {
+        if (raw == null) {
+          throw new IllegalStateException(
+              "Data file selection points to a missing file (maybe it was removed after first evaluation).");
+        }
+      }
+      return value;
     }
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilesSelection.java
```diff
@@ -57,11 +57,15 @@ public RawDataFilesSelection(RawDataFilesSelectionType selectionType) {
...
           .toArray(RawDataFile[]::new);
-      return value;*/
-      throw new IllegalStateException("Raw data file selection has already been evaluated.");
+      for (var raw : value) {
+        if (raw == null) {
+          throw new IllegalStateException(
+              "Data file selection points to a missing file (maybe it was removed after first evaluation).");
+        }
+      }
+      return value;
```
robinschmid, line 68: This is the part. Should be okay to use the placeholders this way right?
Some modules might pass the parameters on to multile objects that want to get the raw data files

SteffenHeu: These modules should pass the raw data files directly. This is a safety feature to ensure that the evaluation is done at the right time and only once.

robinschmid: Would this break anything?
I mean we are always using the correct files in this case right? 
I feel that it is not intuitive that you can use RawDataFileSelection only once

SteffenHeu: Before, it was always reevaluated. This broke the feature list summaries and made them not preproducible. while implementing project load/save, i changed it so that it would only be evaluated once and throw an error otherwise.

If it has already been evaluated, you can always use the getEvaluationResult method.

This way we have clear functionality and reproducible results. Furthermore, we see when the error happens and can clearly debug it. You can also clone the selection and evaluate the clone for parameter checking.
Furthermore, there's reset method, which clears the selection.

robinschmid: But there should be one method that just provides the matching files. It is hard to secure that no one will call this method twice as parameters are often stored somewhere in a visualization module or similar. 
There should be a method that either evaluates or just gives the evaluated results

PR #571, opened 2022-03-07
fix zip import

SteffenHeu:
Zip module created the wrong task

### DIFF
--- a/src/main/java/io/github/mzmine/modules/io/import_rawdata_zip/ZipImportModule.java
+++ b/src/main/java/io/github/mzmine/modules/io/import_rawdata_zip/ZipImportModule.java
@@ -20,24 +20,19 @@ package io.github.mzmine.modules.io.import_rawdata_zip;
 
 import com.google.common.base.Strings;
 import io.github.mzmine.datamodel.MZmineProject;
-import io.github.mzmine.datamodel.RawDataFile;
 import io.github.mzmine.main.MZmineCore;
 import io.github.mzmine.modules.MZmineModuleCategory;
 import io.github.mzmine.modules.MZmineProcessingModule;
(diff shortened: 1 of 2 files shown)


robinschmid:
So this will unzip to a temp file and then read. 
Will this end up with problems for saved projects?
The files will point to the temp files correct?

robinschmid:
Maybe this oculd be done in a similar way to the ThermoRawImport by passing the input stream from unzipping to the method

SteffenHeu:
Then we would have to refactor every import task. some dont even use an input stream. some also require multiple files.

But indeed this breaks project import, because the temp folder has a different name every time.

SteffenHeu:
Would it be sufficient if this works for zipped mzml files?

robinschmid:
Maybe - or we could at least fix it for mzML files and use the current solution for all othzer files?

src/main/java/io/github/mzmine/modules/io/import_rawdata_mzml/MSDKmzMLImportTask.java
```diff
@@ -127,26 +131,41 @@ public void run() {
...
+        newMZmineFile = new IMSRawDataFileImpl(this.file.getName(), file.getAbsolutePath(),
+            storage);
       } else {
-        buildLCMSFile(file);
+        newMZmineFile = new RawDataFileImpl(this.file.getName(), file.getAbsolutePath(), storage);
       }
 
+      if (newMZmineFile instanceof IMSRawDataFileImpl) {
+        ((IMSRawDataFileImpl) newMZmineFile).addSegment(
+            Range.closed(1, msdkFile.getScans().size()));
```
robinschmid, line 164: This should be resolved by next PR for segments

robinschmid:
Great solution. 
Interferes with #572

PR #572, opened 2022-03-07
use mobility segments

SteffenHeu:
instead of storing every frame's mobilities in a new double buffer, the mobilities are stored as an array in memory with the raw data file. a new segment is only added if the mobility values are not matching any other segment.


src/main/java/io/github/mzmine/datamodel/impl/SimpleFrame.java
```diff
@@ -129,23 +127,27 @@ public MobilityScan getMobilityScan(int num) {
...
+          getMobilities().length, originalMobilityScans.size()));
     }
     mobilityScanStorage = new MobilityScanStorage(getDataFile().getMemoryMapStorage(), this,
         originalMobilityScans, useAsMassList);
   }
 
   @Override
   public double getMobilityForMobilityScanNumber(int mobilityScanIndex) {
-    return mobilityBuffer.get(mobilityScanIndex);
+    return ((IMSRawDataFileImpl) (getDataFile())).getSegmentMobilities(
```
robinschmid, line 141: Should be interface IMSRawDataFile

SteffenHeu: i didnt want to expose this method by default, so it's just in the implementation.

robinschmid: Either use IMSRawDataFileImpl then in Frame to enforce this - casting is unsafe. 
or pull this method up to the interface

src/main/java/io/github/mzmine/datamodel/impl/SimpleFrame.java
```diff
@@ -172,17 +174,11 @@ public PasefMsMsInfo getImsMsMsInfoForMobilityScan(int mobilityScanNumber) {
...
-  public DoubleBuffer setMobilities(double[] mobilities) {
-    if (mobilityScanStorage != null && (getNumberOfMobilityScans() != mobilities.length)) {
-      throw new IllegalArgumentException(String.format(
-          "Number of mobility values (%d) does not match number of mobility scans (%d).",
-          mobilities.length, getNumberOfMobilityScans()));
-    }
-    mobilityBuffer = StorageUtils.storeValuesToDoubleBuffer(getDataFile().getMemoryMapStorage(),
-        mobilities);
+  public int setMobilities(double[] mobilities) {
+    mobilitySegment = ((IMSRawDataFileImpl) getDataFile()).addMobilityValues(mobilities);
```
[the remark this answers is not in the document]
robinschmid: still impl

robinschmid:
I checked everything again. Great work. 
Only the two comments left. And the workflow does not run - I will restart it and check later

PR #579, opened 2022-03-08
update IMSRawDataOverviewPane

SteffenHeu:
- add binning to selected and frame mobilogram
- update range field to selected mz range
- allow addition of mz from single field with mz tol

### DIFF
--- a/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/spectra/FrameSummedMobilogramProvider.java
+++ b/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/spectra/FrameSummedMobilogramProvider.java
@@ -92,7 +92,7 @@ public class FrameSummedMobilogramProvider implements PlotXYDataProvider {
 
   @Override
   public String getToolTipText(int itemIndex) {
-    MobilityScan scan = frame.getSortedMobilityScans().get(itemIndex);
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/spectra/FrameSummedMobilogramProvider.java
```diff
@@ -121,12 +127,13 @@ public double getComputationFinishedPercentage() {
   public void computeValues(SimpleObjectProperty<TaskStatus> status) {
     List<MobilityScan> scans = frame.getSortedMobilityScans();
 
-    int numScans = scans.size();
-    mobilites = new double[numScans];
-    intensities = new double[numScans];
+    final int numScans = ((int)(scans.size() / binWidth)) * binWidth;
```
robinschmid, line 130: Is that needed?

robinschmid (approved):
Looks good to me.

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/spectra/FrameSummedMobilogramProvider.java
SteffenHeu: yes, it's the fastest way to do the binning. A few scans more scans than expected might hit the last bin but it's fine for visualisation.

PR #583, opened 2022-03-08
Fix Waters raw import

robinschmid:


SteffenHeu:
looks good so far. why did you delete the source code of the dumper though?

robinschmid:
its all in the zip.

PR #594, opened 2022-03-11
Remove RT from images

robinschmid:
Untested. imzML reader is broken while reading OBO files

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,8 +70,8 @@ repositories {
     maven { url = "https://www.xypron.de/repository/" }
     // For cpdetector
     maven { url = "https://nexus.nuiton.org/nexus/content/groups/releases/" }
-    // For jimzml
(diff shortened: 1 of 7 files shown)


SteffenHeu:
Problems:

- RT range not set, used during alignment (and maybe other tools) - in ExpandingTrace fixed by:
 `final List<? extends Scan> spectra = f.getBestFeature().getFeatureData().getSpectra();
    rtRange = Range.closed(spectra.get(0).getRetentionTime(),
        spectra.get(spectra.size() - 1).getRetentionTime());`

- Quality parameters cannot be calculated without RT - in FeatureDataUtils:
`if (calcQuality && !(feature.getRawDataFile() instanceof ImagingRawDataFile)) {
      calculateQualityParameters(feature);
    }`
- Row sorter, Feature shape chart expect an rt to be present
- Range.all is used in join aligner, produces exception in RangeUtils.rangeLength

PR #597, opened 2022-03-11
dont show image in shape chart, remove space in smoothing suffix

SteffenHeu:


robinschmid:
This change removes the shape of images from aligned feautres
Looks good. Related to #594 that removes the RT completely from Feature of images.

Issue #592, opened 2022-03-09
Error running RANSAC aligner


mmzdouc:
### Basic information

* My operating system: Ubuntu 20.04
* My MZmine version: MzMine 3.0.0-beta

Dear developers,
while testing MzMine 3, I encountered a bug while trying to run the RANSAC algorithm on my data:

`2022-03-09 15:27:13 SEVERE io.github.mzmine.taskcontrol.impl.WorkerThread run Unhandled exception java.lang.NullPointerException: Cannot invoke "io.github.mzmine.datamodel.features.ModularFeatureList.setSelectedScans(io.github.mzmine.datamodel.RawDataFile, java.util.List)" because "this.alignedFeatureList" is null while processing task Ransac aligner, Aligned feature list (3 feature lists)
java.lang.NullPointerException: Cannot invoke "io.github.mzmine.datamodel.features.ModularFeatureList.setSelectedScans(io.github.mzmine.datamodel.RawDataFile, java.util.List)" because "this.alignedFeatureList" is null
	at io.github.mzmine.modules.dataprocessing.align_ransac.RansacAlignerTask.lambda$run$0(RansacAlignerTask.java:149)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at io.github.mzmine.modules.dataprocessing.align_ransac.RansacAlignerTask.run(RansacAlignerTask.java:148)
	at io.github.mzmine.taskcontrol.impl.WorkerThread.run(WorkerThread.java:57)`

For more information, please see the logfile attached.
[mzmine_0_0.log](https://github.com/mzmine/mzmine3/files/8216825/mzmine_0_0.log)

Please let me know if you need any more info from my side. 
Great job with the new version! 

Kind regards, 
Mitja Zdouc

robinschmid:
Should be fixed in the next release (or current development release)

SteffenHeu:
Here's the link to test it: https://github.com/mzmine/mzmine3/releases/tag/Development-release

please let us know if you have further issues.

PR #598, opened 2022-03-11
fix ransac aligner

SteffenHeu:
fixes issue #592


SteffenHeu:
The aligned feature list was null when the data files were added

robinschmid:
nice

PR #601, opened 2022-03-16
prevent tdf import from infinite loop on corrupt data

SteffenHeu:
- tdf went into infinite loop when a corrupt frame was in the data
- same method is also used to check for the adequate buffer size, so i just added a message check that throws an exception when corrupt data is found


SteffenHeu:
When the data file is corrupt as i described above.

PR #603, opened 2022-03-16
FeatureListSummaryController - use arrays.tostring for arrays

SteffenHeu:
FileNamesParameter was not correctly displayed


robinschmid:
So all files are now just concatenated? or what is the result?
Maybe they could be joined by a semi colon?

SteffenHeu:
it looks like this, which is fine imo
[file1, file2, file3]

robinschmid:
Thats great

Issue #602, opened 2022-03-16
GNPS FBMN export creates single edge file for multiple feature lists


SteffenHeu:
When exporting multiple Feature lists to GNPS via "Feature list methods -> export -> GNPS / FBMN" with {} in the file name, only a single file is created for the additional edges.

The GnpsFbmnExportAndSubmitTask creates multiple subtasks, such as the ExportCorrAnnotationTask, which is responsible for the edge annotations. The filename passed to this task is not properly updated for each feture list. A working example is the CSVExportModularTask, which is also launched by the GnpsFbmnExportAndSubmitTask.

Result: 

![grafik](https://user-images.githubusercontent.com/37407705/158575505-0a7058ea-4fa9-45e1-aba3-106a71d77d19.png)

Expected result: one {}_edges_msannotation.csv file for every feature list. E.g. 100_ms_aligned_corr_edges_msannotation.csv

SteffenHeu:
I think the only class that needs to be modified is the "ExportCorrAnnotationTask", because the file name is not set properly there. 
The CSVExportModularTask correctly replaces the filename, so that's where you can check how it should be done.

sanbit876:
@SteffenHeu since the name of the file should be replaced with the name of the feature list, I tried modifying the parameter of the writeToFile method as show in this code:

`if (added.get() > 0) {
        writeToFile(ann.toString(), filename, featureLists[0].getName()+"_edges_msannotation");
        return true;
      } else {
        return false;
      }

PR #605, opened 2022-03-17
Fix for the issue #602

sanbit876:
Bug Fix For the issue #602 titled: "GNPS FBMN export creates single edge file for multiple feature lists"

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/export/ExportCorrAnnotationTask.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/export/ExportCorrAnnotationTask.java
@@ -56,6 +56,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javafx.collections.ObservableList;
 import org.apache.commons.lang3.StringUtils;
@@ -171,8 +172,23 @@ public class ExportCorrAnnotationTask extends AbstractTask {
 
       // export ann edges
       // Filename
+
       if (added.get() > 0) {
-        writeToFile(ann.toString(), filename, "_edges_msannotation");
+        String CString="{}";
+        boolean check=filename.getPath().contains(CString);
+        if(check)
+        {
+          File curFile = filename;
+          for (FeatureList featureListN : featureLists)
+          {
+            String newFilename = filename.getPath().replaceAll(Pattern.quote(CString),featureListN.getName());
+            curFile = new File(newFilename);
+            writeToFile(ann.toString(),curFile, "_edges_msannotation");
+          }
+        }
+        else {
+          writeToFile(ann.toString(), filename, "_edges_msannotation");
+        }
         return true;
       } else {
         return false;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/export/ExportCorrAnnotationTask.java
```diff
@@ -228,7 +244,21 @@ public boolean exportIINRelationships(FeatureList pkl, File filename, Double pro
       // export ann edges
       // Filename
       if (added.get() > 0) {
-        writeToFile(ann.toString(), filename, "_edges_iin_relations");
+        String CString="{}";
+        boolean check=filename.getPath().contains(CString);
+        if(check)
+        {
+          File curFile = filename;
+          for (FeatureList featureList : featureLists)
```
SteffenHeu, line 252: You don't need to loop over the feature lists here. This method (exportIINRelationships) is called in a loop for every feature list. The current feature list is passed as a method parameter, just use that.
Also, it's better to replace all illegal characters in the feature list name. See the CSVExportModularTask.

src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/export/ExportCorrAnnotationTask.java
```diff
@@ -171,8 +172,23 @@ public boolean exportIonIdentityEdges(FeatureList featureList, File filename, Do
...
       // Filename
+
       if (added.get() > 0) {
-        writeToFile(ann.toString(), filename, "_edges_msannotation");
+        String CString="{}";
+        boolean check=filename.getPath().contains(CString);
+        if(check)
+        {
+          File curFile = filename;
+          for (FeatureList featureListN : featureLists)
```
SteffenHeu, line 182: same here

src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/export/ExportCorrAnnotationTask.java
```diff
@@ -548,7 +578,21 @@ public boolean exportAnnotationEdgesMerged(FeatureList[] featureLists, File file
...
+          {
+            String newFilename = filename.getPath().replaceAll(Pattern.quote(CString),featureList.getName());
+            curFile = new File(newFilename);
+            writeToFile(ann.toString(),curFile, "_edges_msannotation");
+          }
+        }
+        else {
+          writeToFile(ann.toString(), filename, "_edges_msannotation");
+        }
         return true;
```
SteffenHeu, line 596: This method loops over every feature list itself and creates a merged result. So this is creating a file with the same content, which does not really make sense. You can leave the original code here and maybe add a "_merged" suffix to the "_edges_msannotation". (@robinschmid, what do you think?)

SteffenHeu (changes requested):
Looks good so far, a few changes need to be made, though.

sanbit876:
@SteffenHeu  I have made changes in the code as required...please review the code

SteffenHeu:
Looks good, only a small comment: Please use the IntelliJ code formatter that is supplied with the project and reformat the code.

@robinschmid merge after?

robinschmid:
Thanks for the changes and it works like expected. 
Could you add two small things?
1. add the same mechanic to the `exportMap` method to replace {} by feature list name. You could add the feature list as a parameter or evaluate the filename before and pass the correct file name.
2. Change the descripton of filename parameter in `ExportCorrAnnotationParameters` to reflect the new behavior. It is the second argument of the constructor. Something like 
"Base file name of all edge files. Use {} to fill in the feature list name when exporting multple feature lists at once." 

Then it is perfect. 
Thank you for the contribution.

robinschmid:
Thanks for addressing everything.

Issue #599, opened 2022-03-15
Layout to big in IMS feature visualizer


robinschmid:
The IMS feature visualizer does not scale for smaller resolutions.

## Proposed fix:
Reduce minimum size / height for all components in the classes found in
```
io.github.mzmine.modules.visualization.ims_featurevisualizer
```

## Steps to reproduce
1. import IMS-MS data
2. Mass detection on MS1 level
3. ADAP Chromatogram Builder
4. Local minimum feature resolver
5. Ims expander
6. Double click on feature list and select "Show/Ion mobility trace"
![image](https://user-images.githubusercontent.com/10366914/158421305-d935fc10-f19c-4c43-81be-cd787fac9fba.png)

**Or use batch processing file in "Project/Batch mode**:
[Batch processing parameters](https://github.com/mzmine/mzmine3/files/8255121/ims_trace_batch_test.zip)

### Correct large screen
![image](https://user-images.githubusercontent.com/10366914/158419985-1b41a2ed-4720-46b1-a3b7-1b62eb58fceb.png)

### Small screen
![image](https://user-images.githubusercontent.com/10366914/158420062-e81f0b4c-7968-47a5-9fe4-a8af92420134.png)

PR #610, opened 2022-03-18
fix: ims visualizer layout for small resolutions

arteevraina:
Fixes: #599


arteevraina:
Hi Steffen . I have made the minimum width of mobilogram to match the minimum width of trace chart.

SteffenHeu:
Basically yes, but the maximum and minimum width of the mobilogram should be limited. i think the maximum can be kept as is, but the minimum should be around 150 pixels or so.

arteevraina:
Screenshot from 2022-03-23 21-15-48

Hi Steffen, just checked the code. We were not setting the maximum width of mobilogram earlier explicitly, we were setting up only minimum width of mobilogram. 

Now, restricted the maximum width of mobilogram to 250 pixels (previous minimum width) and minimum width to 150 pixels.

SteffenHeu:
Hi Arteev, i checked in with the latest changes. The mobilogram is not scaled between the minimum and maximum size when the window is resized. Can you take a look, please?

SteffenHeu:
Closed due to inactivity

Issue #575, opened 2022-03-07
[Bug] Ion Identity: Cannot import custom adducts/modifications list from csv


Adafede:
### Basic information

* My operating system: MacOS 12.3
* My MZmine version: 3.0.0-beta

### What happened

* Please describe what happened and how it differs from what you expected to happen.

I tried importing a custom adducts.csv file but it does not load desired adducts.

* Please include all parameter values for the MZmine modules you used. 

No parameters used, just did: `Feature list methods > Feature grouping > Add more ion identities`

You can try exporting the default list and re-importing while adding a single dummy line it does not work

* Please upload screenshots of any errors and your log/mzmine.log file.

```
2022-03-07 18:24:51 WARNING io.github.mzmine.parameters.parametertypes.MultiChoiceComponent loadChoices Line length is different to header length: 5 to 5
```

* Please upload an MZmine project file that demonstrates the problem you are reporting.

NA

robinschmid:
Will be fixed in the next release. Thanks for reporting.

PR #616, opened 2022-03-21
Add major negative ions to IIN refinement 

robinschmid:
@SteffenHeu @Adafede @Annexhc 
There is an optional refinement step in IIN to remove all IIN without major ion. Users can control the creation of IIN initially in the module so this is not super important but still a great hack to reduce the number of weird IIN with little parameter setup. 

I have added some missing negative ions and defined the maximum number of modifications that I would like to allow. For example, M+Na does usually not show abundant in source fragments, M+H does. The list is now:
- M+H (3 max modifications, e.g., M-3H2O+H)
- M+Na (0)
- M+NH4 (1)
- M-H (2)
- M+Cl (0)
- M+formic acid (0)

Any comments, anything to add?

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/id_ion_identity_networking/refinement/IonNetworkRefinementParameters.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/id_ion_identity_networking/refinement/IonNetworkRefinementParameters.java
@@ -1,19 +1,19 @@
 /*
- * Copyright 2006-2015 The MZmine 2 Development Team
- * 
- * This file is part of MZmine 2.
- * 
- * MZmine 2 is free software; you can redistribute it and/or modify it under the terms of the GNU
+ * Copyright 2006-2021 The MZmine Development Team
+ *
+ * This file is part of MZmine.
+ *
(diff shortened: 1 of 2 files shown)


SteffenHeu (approved):
looks good to me

robinschmid:
I was using HFA and FA... what would you use?

SteffenHeu:
Sorry i did not see your comment. Hard to say. If the usage is consistent throughout the IIN module its fine i think. Everyone familiar with massspec should know whats going on.

Adafede:
@robinschmid 

Sorry completely missed the tag on that one!

Hmmm....I am actually using n >> 3 modifications, this would imply doing in in n/3 steps, did I get it right?

Btw, could it be related to #575 or not at all?

robinschmid:
Oh this is only about the refinement of IIN. It is completely optional. It is just an option for people to easily filter out all ion identity networks that do not contain one of the main ions. the max 3 modifications for M+H just mean that we do accept that M+H fragments more than M+Na for example.

robinschmid:
#575 will be something else. have to look into this

PR #623, opened 2022-03-25
Fix baseline resolver

### LINKED ISSUE #612: Issue with BASELINE RESOLVER - 3.0.11-beta version
### Basic information

* My operating system: Windows 10
* My MZmine version: 3.0.11-beta

### What happened
Dear developers,
I am trying to reproduce a job I already performed on the old version MZmine 2.53, I started from Thermo raw data and I adopted the same parameters for mass detection and ADAP chromatogram builder, when it came to the deconvolution I tried to use the "baseline resolver" algorithm which I think is the equivalent of "baseline cut-off" in the 2.53 version, in fact, I put the same parameters, but this time I got this error for each feature list.

error ResolvedPeak java 105

Could you give an explanation for this?

Parameters
-mass detention: MS1 noise level: 10000, MS2 noise level 100
-ADAP chrom builder: Scans MSlevel 1, Min group size n of scans: 3, group intensity: 10000, min intensity:10000, scan accuracy 20 ppm
-Baseline resolver: MS/MS scan pairing (RT toll:0.3, MS1MS2 toll: 5ppm), Min peak height: 10000, peak duration 0-2 min, Baseline level:10000 ---  ERROR ---

Thank you in advance,
Giovanni

SteffenHeu:
Assertion was triggered when there was only one data point in the resolved feature. Don't know how it was handled in MZmine 2, the assertion was there sicne 2010 (accordingto git blame)

Also fixed FeatureConvertors#ResolvedPeakToMoularFeature

Edit: fixes #612

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/FeatureResolverTask.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/FeatureResolverTask.java
@@ -425,6 +425,7 @@ public class FeatureResolverTask extends AbstractTask {
     processedRows = 0;
     totalRows = originalFeatureList.getNumberOfRows();
     int peakId = 1;
+    final Integer minNumDp = parameters.getValue(GeneralResolverParameters.MIN_NUMBER_OF_DATAPOINTS);
 
     for (int i = 0; i < totalRows; i++) {
       final ModularFeatureListRow originalRow = (ModularFeatureListRow) originalFeatureList.getRow(
@@ -434,7 +435,6 @@ public class FeatureResolverTask extends AbstractTask {
       final ResolvedPeak[] peaks = resolver.resolvePeaks(originalFeature, parameters, rSession,
           mzCenterFunction, msmsRange, RTRangeMSMS);
 
-      final Integer minNumDp = parameters.getValue(GeneralResolverParameters.MIN_NUMBER_OF_DATAPOINTS);
       for (final ResolvedPeak peak : peaks) {
         if(peak.getScanNumbers().length < minNumDp) {
           continue;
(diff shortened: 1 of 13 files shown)


src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/baseline/BaselineFeatureResolver.java
```diff
@@ -110,7 +110,7 @@
         currentRegionEnd--;
 
         // Check current region, if it makes a good peak.
-        if (durationRange
+        if (currentRegionEnd - currentRegionStart > 0 && durationRange
```
robinschmid, line 113: so you basically added a minimum number of data points of 2 here

robinschmid (approved):
looks good to me. Only thing might be to introduce a minimum number of data points parameter (like we did for local min) instead of forcing 2 datapoints

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/baseline/BaselineFeatureResolver.java
SteffenHeu: Yes

SteffenHeu: But looking at the history of resolved peak, the assertion was there since 2010. So it must not have been triggered before.

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/FeatureResolverTask.java
```diff
@@ -434,7 +434,11 @@ private FeatureList resolvePeaks(final ModularFeatureList originalFeatureList,
       final ResolvedPeak[] peaks = resolver.resolvePeaks(originalFeature, parameters, rSession,
           mzCenterFunction, msmsRange, RTRangeMSMS);
 
+      final Integer minNumDp = parameters.getValue(GeneralResolverParameters.MIN_NUMBER_OF_DATAPOINTS);
```
robinschmid, line 437: Can you move this outside of the rows loop?
parameter.getValue is not really effective.

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/baseline/BaselineFeatureResolver.java
robinschmid, line 113: this check can now be removed. currentRegionEnd is always >= start. ResolvedPeak is now also allowing 1 dp

PR #629, opened 2022-03-28
add link to new docs

SteffenHeu:
Added a constructor parameter to the SimpleParameterSet to open the new documentation link (if available)

### DIFF
--- a/src/main/java/io/github/mzmine/parameters/impl/SimpleParameterSet.java
+++ b/src/main/java/io/github/mzmine/parameters/impl/SimpleParameterSet.java
@@ -43,6 +43,7 @@ import javafx.application.Platform;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.scene.control.ButtonType;
+import org.jetbrains.annotations.Nullable;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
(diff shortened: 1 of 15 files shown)


src/main/java/io/github/mzmine/parameters/impl/SimpleParameterSet.java
```diff
@@ -204,7 +210,7 @@ public ExitCode showSetupDialog(boolean valueCheckRequired) {
     if ((parameters == null) || (parameters.length == 0)) {
       return ExitCode.OK;
     }
-    ParameterSetupDialog dialog = new ParameterSetupDialog(valueCheckRequired, this);
+    ParameterSetupDialog dialog = new ParameterSetupDialog(valueCheckRequired, this, null);
```
robinschmid, line 213: This should actually copy the url as well - instead of null

robinschmid:
Is it safe to use the parameterset here?
Or is there any chance the url gets lost.

- [ ] clone parameterset is safe
- [ ] write and read of parameters is safe

src/main/java/io/github/mzmine/parameters/impl/SimpleParameterSet.java
SteffenHeu: What do you mean? That's the header message

SteffenHeu:
No clue what you mean, but i remembered the cloning

robinschmid:
Good like this. We are safe using it in the parameter classes as its only used in the dialogs.

PR #643, opened 2022-04-01
Fix compdb

### LINKED ISSUE #642: Local compound database search
### Basic information

* My operating system: Windows 10
* My MZmine version: MZmine 3.0.21 Beta

### What happened

Local compound database search is one way that we identify metabolites and have done with MZmine 2 for several years.  In the beta version of MZmine 3, performing this search results in no identifications.  I modified the .CSV file according to the updated parameters but still no results are reported out.  Checking m/z and retention time pairs in the .CSV and in the compiled list shows that there is good correspondence.  Wondering if I'm doing something incorrect or if the feature is broken.

SteffenHeu:
fix issue with RtRelativeErrorType (could not be instantiated due to protected constructor)
fix issue with CompoundAnnotationScoreType (duplicated)

- added test to check for duplicated unique ids
- added test to check for duplicated class names
- both tests check if data type can be instantiated

maybe fixes #642


robinschmid (approved):
Great tests and checks

PR #651, opened 2022-04-04
fix ransac, add logging for showSetupDialog

SteffenHeu:
- fix the ransac aligner dialog if not feature list is present
- surround showSetupDialog with a try/catch to get logging for the dialogs


robinschmid:
Great.

PR #654, opened 2022-04-05
add MissingMassListException to MobilityScanStorage

SteffenHeu:
During gap filling just a NPE was shown if no mass detection was run. Exception added to MobilityScanStorage to show a more descriptive error message.


robinschmid (approved):
Good that you added more error messages and logging

Issue #644, opened 2022-04-01
Restructure the metadata (sample parameters) 


robinschmid:
Simplify the metadata table with only three main data types.

```
class MetaDataColumn<T> for (String, Double, Date)

// column -> raw file -> value
Map<MetaDataColumn<T>, Map<RawDataFile, T/Object>>

// or raw file -> column -> value
Map<RawDataFile, Map<MetaDataColumn<T>, T/Object>> 
```

- [x] three data types: String, double, Date
- [x] replace the data structure in MZmineProject (everything with UserParameter)
- [x] use the dialog to add new columns and set values in the TableView
- [ ] Import/Export to csv / tsv (tab-separated) (header, types, values)

InspiritedCoder:
I've started working on this issue

Tarush-Singh35:
@InspiritedCoder i would like to help you out tell me how can I get started

InspiritedCoder:
> @InspiritedCoder i would like to help you out tell me how can I get started

@Tarush-Singh35 Thank you for your willingness to help. As far as I understand, it is necessary to implement some "generalized" class through which we will perform basic operations with the project metadata (parameters) table. If you want, we can discuss a possible solution as soon as I come up with one. :)

robinschmid:
Its good if you propose a structure and create a PR once you are ready to show. 
Then we can discuss more together. @Tarush-Singh35, best way would be to then review the code and we can improve it together.

PR #655, opened 2022-04-07
Restructure the metadata (sample parameters) #644

InspiritedCoder:
New project metadata structure was added (MetadataTable).


src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Return parameter value of the corresponding RawData file.
+   *
+   * @param parameter   project parameter
+   * @param rawDataFile RawData file
+   * @param <T>         type of the project parameter
+   * @return parameter value
+   */
+  public <T> T getParameterValue(UserParameter<T, ?> parameter, RawDataFile rawDataFile) {
```
robinschmid, line 88: that function will just use the MetadataColumn directly this way you have both keys for the column and row and the query is effective.
The getColumnByName will not be used inside this method, as it needs to loop over the set to retrieve the column key. Once the programmer has that column they can start adding information.

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -0,0 +1,111 @@
...
+    if (hasParameter(parameter)) {
+      return (T) data.get(getParameterColumnByName(parameter.getName())).get(rawDataFile);
```
robinschmid, line 90: Instead of checking if the column exists I would just do this:'
```java
var row = data.get(columnKey);
if(row!=null) return (T) row.get(dataFile);
else return null;
```
for the (T) to work you will need to change the input parameter to MetadataColumn<T>

PR #659, opened 2022-04-08
Metadata structures fixes #644

### LINKED ISSUE #644: Restructure the metadata (sample parameters)

InspiritedCoder:
MetadataColumn basic types implementations were added. 

I decided to get rid of the generic parameter in MetadataColumn class. In my opinion, it's redundant because we will be able to define the type of the parameter by pattern matching it to the sealed classes of MetadataColumn.


robinschmid:
based on PR #655

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
```diff
@@ -46,240 +46,243 @@
...
-                                rawDataFile = file;
-                                break;
-                            }
-                        }
-                        currentProject.setParameterValue(parameter,rawDataFile,newParaVal);
-                        updateParametersToTable();
-                    }
-            );
+  @FXML
+  public void addPara(ActionEvent actionEvent) {
```
robinschmid, line 151: This whole method could be extracted in to a separate dialog class. 
But thinking more about the simplicity of this, we could just create a new ParameterSet, show the setup dialog and get the results if the user clicks on ok button. So no need to write the dialog yourself. 
1. Create a new parameters class (any module has one: see GridMassParameters for an example)
2. add all parameters that you need: StringParameter title; ComboParameter<String> valueType;
3. check for exit code from dialog and create new column

```java
ExitCode exitCode = moduleParameters.showSetupDialog(true);
    if (exitCode != ExitCode.OK){
    ...
    }
 ```

**Comment by robinschmid** [comment]
Data structures look good already. Just a few comments to improve it.
The next steps should be to setup some small tests for you to see that it actually does what you want. For tests we are using JUnit 5 tests. You can find some examples in main/test/ directory. Just create a new test class there and add a few test methods. Usually they should test very simple things like adding columns, adding rows, checking that the data is available by calling the get methods. 

If you are really interested you could also try out Mockito to @Mock the RawDataFiles - this just means that mockito will generate an empty object that you can use in the table to generate rows and add values. You can see how its used in the `IonIdentityTest`

**Comment by robinschmid** [comment]
Then the next step is the dialog. You can either use the one that is already there or completely redo it. Sometimes its easier to redo in case you think the old one is messy. Especially when the concept that we are applying now is much simpler than the original

File: build.gradle [hunk 2]
```diff
@@ -309,7 +316,7 @@ runtime {
         if (OperatingSystem.current().isLinux()) {
             // leave installer type empty to generate all that match
             // for rpm, rpmbuild and rpm tools need to be installed, installerType needs to be set to "rpm"
-            // installerType = "deb"
+            installerType = "deb"
```
robinschmid, line 319: comment out

robinschmid, line 319: comment out (or should be in master - merge master)

src/main/java/io/github/mzmine/project/parameterssetup/ProjectMetadataParameters.java
```diff
@@ -0,0 +1,38 @@
...
+import io.github.mzmine.parameters.impl.SimpleParameterSet;
+import io.github.mzmine.parameters.parametertypes.ComboParameter;
+import io.github.mzmine.parameters.parametertypes.StringParameter;
+
+public class ProjectMetadataParameters extends SimpleParameterSet {
+
+  public static final StringParameter title = new StringParameter("Title",
+      "Title of the new parameter", "", true, true);
+
+  public static final String[] availableTypes = new String[]{"String", "Double", "Datetime"};
```
robinschmid, line 31: Best create an enum with the three types. Its always easier than having stings. 
In case the string shoudl be different than the enums value name, you can override the toString method.

Lets do Text instead of String. Might be more straightforward for non coders.

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
```diff
@@ -23,263 +23,279 @@
...
+    // display each row of the table
+    ObservableList<ObservableList<StringProperty>> tableRows = FXCollections.observableArrayList();
+    for (RawDataFile rawFile : fileList) {
+      ObservableList<StringProperty> fileParametersValue = FXCollections.observableArrayList();
+      fileParametersValue.add(new SimpleStringProperty(rawFile.getName()));
+      for (MetadataColumn<?> column : columns) {
+        // either convert parameter value to string or display an empty string in case if it's unset
+        fileParametersValue.add(new SimpleStringProperty(
+            metadataTable.getValue(column, rawFile) == null ? ""
+                : metadataTable.getValue(column, rawFile).toString()));
```
robinschmid, line 105: introduce a variable to not call the method twice. 
CTRL ALT V with the cursor somewhere in the expression lets you extract a variable quiickly.

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
```diff
@@ -23,263 +23,279 @@
...
+              Alert alert = new Alert(Alert.AlertType.INFORMATION);
+              alert.setTitle("Wrong parameter value format");
+              alert.setHeaderText(null);
+              alert.setContentText(
+                  "Please respect the String parameter value format, e.g. \"String\"");
+              alert.showAndWait();
             }
-            if(currentProject.getParameterByName(parameterName)!=null){
-                return;
+          }
```
robinschmid, line 169: in switch case you should try to extract everything that is duplciated and only use it where you diverge.
so checkInput, convert, set value, and showing the dialog can all be done outside. The only thing that is different is the format hint in the message dialog. 
Use 
`String format = switch(parameter) { case StringMetadataColumn -> "String"` ....

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
```diff
@@ -23,263 +23,279 @@
...
+        alert.setTitle("Parameter already present");
+        alert.setHeaderText(null);
+        alert.setContentText("Please enter unique parameter name.");
+        alert.showAndWait();
+        return;
+      }
+
+      // add the new column to the parameters table
+      switch (parameterType.getValue()) {
+        case "String" -> {
```
robinschmid, line 230: use the enum that I mentioned in another comment to handle those cases

src/test/java/parameterssetup/MetadataTableTest.java
```diff
@@ -0,0 +1,161 @@
...
+  Double columnValue2 = 0.14d;
+  LocalDateTime columnValue3 = LocalDateTime.now();
+  String columnValueNew = "stringValueNew";
+  MetadataColumn<String> column1 = new StringMetadataColumn(columnName1);
+  MetadataColumn<Double> column2 = new DoubleMetadataColumn(columnName2);
+  MetadataColumn<LocalDateTime> column3 = new DateMetadataColumn(columnName3);
+  MetadataColumn<String> columnNew = new StringMetadataColumn(columnNameNew, columnValueNew);
+
+  @Mock
+  RawDataFile rawDataFile1;
```
robinschmid, line 58: Always good to Mock those instead of creating complex structures that are only used as keys here. 
see my PR to your bnranch

robinschmid (commented):
Looks really good. I think we should merge it as soon as you have implemented these few changes and then extend it

src/test/java/parameterssetup/MetadataTableTest.java
robinschmid: Sorry I directly pushed to your remote branch instead of creating a PR.
But you can look at the changes in the commit - which are very small but effective:
https://github.com/mzmine/mzmine3/pull/659/commits/64ae2314168d60deb6d81afa86397f9b32f7be54

build.gradle
robinschmid: still needs to be commented out

src/main/java/io/github/mzmine/project/parameterssetup/columns/DateMetadataColumn.java
```diff
@@ -0,0 +1,54 @@
...
+  public boolean checkInput(Object value) {
+    return value instanceof LocalDateTime;
+  }
+
+  @Override
+  public LocalDateTime convert(String input) {
+    try {
+      //ISO-8601 format
+      return LocalDateTime.parse(input.trim());
+//      return LocalDateTime.parse(input.trim(), DateTimeFormatter.ofPattern("dd-MM-yyyy HH:mm:ss"));
```
robinschmid, line 48: in the future it might be good to be more flexible with the date input to support multiple. No idea if something like this already exists in the LocalDateTime or in some utility (apache commons or similar)

src/main/java/io/github/mzmine/project/parameterssetup/columns/StringMetadataColumn.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public StringMetadataColumn(String title) {
+    super(title);
+  }
+
+  public StringMetadataColumn(String title, String description) {
+    super(title, description);
+  }
+
+  @Override
```
robinschmid, line 34: Maybe we should add a default value here
convert(String input, String defaultValue)
return input == null? defaultValue : input;

Same for the rest (following the T defaultValue)

robinschmid:
Looks good to me. 
Maybe only the null handling in the StringMetadataColumn. You could add a default value to the method signature or just create both methods.

build.gradle
InspiritedCoder: Isn't it already commented out (in the commit № 4d76b44)?

src/main/java/io/github/mzmine/project/parameterssetup/columns/StringMetadataColumn.java
InspiritedCoder: Yeah, sounds good, I will add it

robinschmid:
Great progress. Thanks for improving metadata support

PR #676, opened 2022-04-21
LC-Image Aligner

SteffenHeu:
added a module to properly allow alignment of an (aligned) LC feature list and an image feature list. 

Compared to the Join aligner, RT tolerance is removed and one image can be aligned to multiple LC features (since we miss one dimension of separation)


SteffenHeu:
Wrong commits snuck into this branch

PR #679, opened 2022-04-22
Romove id_sirius module

InspiritedCoder:
id_SIRIUS module is deleted; gradle was updated and some other files previously using id_SIRIUS were updated


build.gradle
```diff
@@ -238,15 +245,15 @@ javafx {
 
 tasks.withType(JavaCompile) {
     options.encoding = "UTF-8"
-    /*options.compilerArgs += ['--enable-preview']*/
+    options.compilerArgs += ['--enable-preview']
```
SteffenHeu, line 248: Is this needed?

build.gradle
```diff
@@ -238,15 +245,15 @@ javafx {
...
 }
 
 tasks.withType(Test) {
-    /*jvmArgs += "--enable-preview"*/
+    jvmArgs += "--enable-preview"
```
SteffenHeu, line 252: Is this needed?

build.gradle
```diff
@@ -238,15 +245,15 @@ javafx {
...
 }
 
 tasks.withType(JavaExec) {
-    /*jvmArgs += '--enable-preview'*/
+    jvmArgs += '--enable-preview'
```
SteffenHeu, line 256: Is this needed?

build.gradle
```diff
@@ -309,7 +316,7 @@ runtime {
         if (OperatingSystem.current().isLinux()) {
             // leave installer type empty to generate all that match
             // for rpm, rpmbuild and rpm tools need to be installed, installerType needs to be set to "rpm"
-            // installerType = "deb"
+            installerType = "deb"
```
SteffenHeu, line 319: why uncomment?

build.gradle
InspiritedCoder: My fault, I uncommented this line because it's needed for the other task I'm implementing (Pattern Matching is still a preview feature)

build.gradle
InspiritedCoder: It's not necessary neither

robinschmid: if not defined both installers are created rpm and deb

PR #689, opened 2022-05-09
sort spectral data by mz in mzXML

robinschmid:
- old converters create unsorted spectral data. 
- This might also apply to mzML

### DIFF
--- a/src/main/java/io/github/mzmine/modules/io/import_rawdata_mzxml/MzXMLImportTask.java
+++ b/src/main/java/io/github/mzmine/modules/io/import_rawdata_mzxml/MzXMLImportTask.java
@@ -19,12 +19,14 @@
 package io.github.mzmine.modules.io.import_rawdata_mzxml;
 
 import com.google.common.base.Strings;
+import io.github.mzmine.datamodel.DataPoint;
 import io.github.mzmine.datamodel.MZmineProject;
 import io.github.mzmine.datamodel.MassSpectrumType;
 import io.github.mzmine.datamodel.PolarityType;
 import io.github.mzmine.datamodel.RawDataFile;
 import io.github.mzmine.datamodel.features.SimpleFeatureListAppliedMethod;
 import io.github.mzmine.datamodel.impl.DDAMsMsInfoImpl;
+import io.github.mzmine.datamodel.impl.SimpleDataPoint;
 import io.github.mzmine.datamodel.impl.SimpleMassSpectrum;
 import io.github.mzmine.datamodel.impl.SimpleScan;
 import io.github.mzmine.datamodel.impl.masslist.ScanPointerMassList;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/mzmine/modules/io/import_rawdata_mzxml/MzXMLImportTask.java
```diff
@@ -420,10 +425,16 @@ public void endElement(String namespaceURI, String sName, // simple name
...
+            dps[i] = new SimpleDataPoint(mz, intensity);
+          }
+          // sort because old converters might create unsorted spectral data
+          Arrays.sort(dps, DataPointSorter.DEFAULT_MZ_ASCENDING);
 
+          for (int i = 0; i < dps.length; i++) {
+            mzValues[i] = dps[i].getMZ();
+            intensityValues[i] = dps[i].getIntensity();
           }
+
```
SteffenHeu, line 437: I think we have this issue more often than here and creating datapoints from an array, sorting them and then creating arrays from there, is a bit pointless. 

I think at this point we should c+p the Arrays.sort function to our array utils and make it apply the sorting of one array to another.

robinschmid: I think the java way of sorting two arrays based on one is to create an object / like we do here. Memory allocation is cheap in java and the sorting algorithms are optimized / so we do not want to reimplement those to reflect the sorting of one to another array, which is not trivial

SteffenHeu: ok, so what do we do about mzml? add it there, too or no? i think we should

robinschmid: I added it for non-IMS mzml

robinschmid:
We now apply sorting in the mzXML and mzML import.
mzML import sorting is done either after mass detection (assuming that only centroid mzML might be unsorted) or before converting to the SimpleScan.
It is not applied for IMS data also assuming that the conversion should be fine.
I think its only a problem for older files converted with previous version of msconvert or others

src/main/java/io/github/mzmine/util/DataPointUtils.java
```diff
@@ -216,4 +201,30 @@ public static double[][] getDataPointsInMzRange(double[] rawMzs, double[] rawInt
     return data;
   }
 
+  /**
+   * @return array of data points
+   */
+  public static DataPoint[] getDataPoints(double[] mzs, double[] intensities) {
```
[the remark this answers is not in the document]
robinschmid: The thing is that every time that we want to sort spectra - we need to get data points. We also get them from the byte buffers. It is discouraged to store them in memory but their use, in general, is unavoidable. I think its better to have this utility function here but to keep on being strict within structures that store data

PR #694, opened 2022-05-13
633_metadata_export_to_tsv

InspiritedCoder:
The project metadata can be exported and imported using a .tsv file;
If everything is OK, then the next step is to add manual import to metadata dialog;


src/main/java/io/github/mzmine/project/parameterssetup/columns/MetadataColumn.java
```diff
@@ -81,4 +81,27 @@ public String getDescription() {
    */
   public abstract T convert(String input, T defaultValue);
 
+  @Override
+  public boolean equals(Object o) {
```
robinschmid, line 85: Just a general thing. Maybe you already used that.
In IntelliJ you can generate equals and hashCode functions by ALT+insert

src/main/java/io/github/mzmine/project/parameterssetup/columns/MetadataColumn.java
```diff
@@ -81,4 +81,27 @@ public String getDescription() {
...
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof MetadataColumn)) {
+      return false;
+    }
+
+    MetadataColumn<?> that = (MetadataColumn<?>) o;
+
+    if (!title.equals(that.title)) {
```
robinschmid, line 95: Lets add @NotNull annotation from jetbrains to title and description. Then force a non null value in the constructors. Lets just use those defaults: description = "" and !title.isBlank()

src/main/java/io/github/mzmine/project/parameterssetup/columns/MetadataColumn.java
```diff
@@ -81,4 +81,27 @@ public String getDescription() {
...
+      return false;
+    }
+    return description != null ? description.equals(that.description) : that.description == null;
```
robinschmid, line 98: After adding @NotNull, all these checks can be simplified. But also check out the generate code funciton in intellij. 
Generally its good to use something like Objects.equals(a,b) as this is null safe by design

src/main/java/io/github/mzmine/project/parameterssetup/columns/MetadataColumn.java
```diff
@@ -81,4 +81,27 @@ public String getDescription() {
...
+  }
+
+  @Override
+  public int hashCode() {
+    int result = title.hashCode();
+    result = 31 * result + (description != null ? description.hashCode() : 0);
```
robinschmid, line 104: simplify after NotNull. There are also functions to build hashCodes with something like Objects.hashCode(Object... objs)  something like this

robinschmid:
Always use a private static final Logger logger ... 
You can do this by adding a live template that shows up in the auto complete
image

`private static final java.util.logging.Logger logger = java.util.logging.Logger.getLogger($CLASSNAME$.class.getName());`

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -31,6 +52,17 @@ public class MetadataTable {
...
+  // define the header fields names of the file with imported metadata
+  private enum HeaderFields {
+    NAME, DESC, TYPE, FILE, VALUE
+  }
+
+  // we will need HeaderFields enum converted into array
+  private final String[] HeaderFieldsArr = Stream.of(HeaderFields.values()).map(Enum::toString)
+      .toArray(String[]::new);
+
+  private final Logger logger = Logger.getLogger(this.getClass().getName());
```
robinschmid, line 64: change to private static final. 
See comment on how to use live template

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -31,6 +52,17 @@ public class MetadataTable {
...
+
   public MetadataTable() {
     this.data = new HashMap<>();
```
robinschmid, line 67: we should also start thinking about sychroniyation. 
I would say that this hashmap that holds the columns can stay unsynchronized. But the maps that we add for each column should be a ConcurrentHashMap because multiple threads might add raw data files or values

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -137,4 +182,178 @@ public <T> void setValue(MetadataColumn<T> column, RawDataFile file, T value) {
...
+  public boolean exportMetadata(File file) {
+    try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, false))) {
+      // in case if there's no metadata to export
+      if (data.isEmpty()) {
+        logger.info("There's no metadata to export");
+        return false;
+      }
+
+      // create the .tsv file header
+      String header = String.join("\t", HeaderFieldsArr);
```
robinschmid, line 211: Just stream over the enum.values() and join the strings in the stream

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -137,4 +182,178 @@ public <T> void setValue(MetadataColumn<T> column, RawDataFile file, T value) {
...
+   * TYPE  - type of the parameter
+   * DESC  - description of the parameter
+   * FILE  - name of the file to which the parameter belong to
+   * VALUE - value of the parameter
+   *
+   * @param file the file in which exported metadata will be stored
+   * @return true if the export was successful, false otherwise
+   */
+  public boolean exportMetadata(File file) {
+    try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, false))) {
```
robinschmid, line 203: Not sure if this is still the case but at one point you ahde to put both on separate lines to auto close the file input in all cases.

```java
try (FileWriter fw = new FileWriter(new File(fileName), StandardCharsets.UTF_8);
             BufferedWriter writer = new BufferedWriter(fw)) {
```

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -137,4 +182,178 @@ public <T> void setValue(MetadataColumn<T> column, RawDataFile file, T value) {
...
+      if (data.isEmpty()) {
+        logger.info("There's no metadata to export");
+        return false;
+      }
+
+      // create the .tsv file header
+      String header = String.join("\t", HeaderFieldsArr);
+      header += System.lineSeparator();
+      // write the header down
+      bufferedWriter.write(header);
```
robinschmid, line 214: you can do `bufferedWriter.append(header).newLine():`

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -137,4 +182,178 @@ public <T> void setValue(MetadataColumn<T> column, RawDataFile file, T value) {
...
+      logger.info("Header was successfully written down");
+
+      // write the parameters value down
+      for (var param : data.keySet()) {
+        var record = data.get(param);
```
robinschmid, line 219: Loop over data.entrySet() instead in both loops. And you could use names that signal what we are looking at: 
record -> column
paramVal could be cellValue

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -137,4 +182,178 @@ public <T> void setValue(MetadataColumn<T> column, RawDataFile file, T value) {
...
+        for (var rawDataFile : record.keySet()) {
+          var paramVal = record.get(rawDataFile);
+          // pattern match the parameter type
+          AvailableTypes paramType = switch (param) {
+            case (StringMetadataColumn stringMetadataColumn) -> TEXT;
+            case (DoubleMetadataColumn doubleMetadataColumn) -> DOUBLE;
+            case (DateMetadataColumn dateMetadataColumn) -> DATETIME;
+          };
```
robinschmid, line 227: Push this code into the MetadataColumn class as a function getType() or similar.

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -137,4 +182,178 @@ public <T> void setValue(MetadataColumn<T> column, RawDataFile file, T value) {
...
+
+          // create the record line
+          List<String> lineFieldsValues = new ArrayList<>();
+          lineFieldsValues.add(param.getTitle());
+          lineFieldsValues.add(param.getDescription());
+          lineFieldsValues.add(paramType.toString());
+          lineFieldsValues.add(rawDataFile.getName());
+          lineFieldsValues.add(paramVal.toString());
```
robinschmid, line 235: You are currently writing a long-table I think we should add an option to write either long or wide table, with wide being the standard. 

this would mean we use multiple lines as headers and then a block of all values just like a standard table.

robinschmid: image

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -137,4 +182,178 @@ public <T> void setValue(MetadataColumn<T> column, RawDataFile file, T value) {
...
+
+          String line = String.join("\t", lineFieldsValues);
+          line += System.lineSeparator();
+          bufferedWriter.write(line);
+        }
+      }
+
+      logger.info("The metadata table was successfully exported");
```
robinschmid, line 243: change to logger.warning if not severe and to logger.severe if its a real issue.

this case severe might be used as we did not produce any results at this point.

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -137,4 +182,178 @@ public <T> void setValue(MetadataColumn<T> column, RawDataFile file, T value) {
...
+   * Import the metadata to the metadata table.
+   * todo: add extra argument defining the format of the imported data (e.g. GNPS or .tsv)
+   *
+   * @param file       source of the metadata
+   * @param appendMode whether the new metadata should be appended or they should replace the old
+   *                   metadata
+   * @return true if the metadata were successfully imported, false otherwise
+   */
+  public boolean importMetadata(File file, boolean appendMode) {
+    try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
```
robinschmid, line 265: change similar to the try with pattern for writing

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -137,4 +182,178 @@ public <T> void setValue(MetadataColumn<T> column, RawDataFile file, T value) {
...
+            case TEXT -> {
+              yield new StringMetadataColumn(splitLine[namePos]);
+            }
+            case DOUBLE -> {
+              yield new DoubleMetadataColumn(splitLine[namePos]);
+            }
+            case DATETIME -> {
+              yield new DateMetadataColumn(splitLine[namePos]);
+            }
+          };
```
robinschmid, line 329: add this part also to metadatacolumn as a factory method. 
MetadataColumn.forType(AvailableTypes type).
Also remove the {} blocks from each statement and the yield. case DOUBLE -> new ....

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
```diff
@@ -62,6 +63,8 @@ private void initialize() {
     parameterTable.setEditable(true);
     parameterTable.getSelectionModel().setCellSelectionEnabled(true);
     fileList = currentProject.getDataFiles();
+    metadataTable.importMetadata(new File("metadata_exp.tsv"), false);
```
robinschmid, line 66: Why do you import here?
I guess its only a test?

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
```diff
@@ -287,6 +290,9 @@ public void removePara(ActionEvent actionEvent) {
 
   @FXML
   public void onClickOK(ActionEvent actionEvent) {
+    // testing export
+    metadataTable.exportMetadata(new File("metadata_exp.tsv"));
```
robinschmid, line 294: We should add buttons for import, export, last files. Instead of automatically saving it.
So that people can import export manually. 
Later we add support to the project load save.

robinschmid: Load last- you can see an example in the batch mode
image

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -31,6 +52,17 @@ public class MetadataTable {
 
   private final Map<MetadataColumn<?>, Map<RawDataFile, Object>> data;
 
+  // define the header fields names of the file with imported metadata
+  private enum HeaderFields {
+    NAME, DESC, TYPE, FILE, VALUE
+  }
+
+  // we will need HeaderFields enum converted into array
+  private final String[] HeaderFieldsArr = Stream.of(HeaderFields.values()).map(Enum::toString)
```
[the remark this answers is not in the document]
InspiritedCoder: You're absolutely right, but the issue is that I would still need the array of strings created from Enum in importMetadata() method, so wouldn't it be better not to use streams repeatedly and just to move this array of strings into a separate variable (moreover, if this header values wouldn't be changed during the run of our application)?

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
InspiritedCoder: OK, but why it might be severe in this case? I'm just logging that the metadata were successfully exported, thus this is a log with some good news, not an issue report :)

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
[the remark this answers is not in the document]
robinschmid: Only if this code is run often... but it's not a performance-critical part so it's better to keep everything local. In general, don't apply optimizations if not needed. JVM does the memory management for you and short-lived objects are no problem

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
robinschmid: I guess the comment slip in a different line:D

src/main/java/io/github/mzmine/project/parameterssetup/MetadataTable.java
```diff
@@ -18,28 +18,62 @@
...
 
-  public MetadataTable(Map<MetadataColumn<?>, Map<RawDataFile, Object>> data) {
+  public MetadataTable(Map<MetadataColumn<?>, ConcurrentMap<RawDataFile, Object>> data) {
     this.data = data;
   }
 
-  public Map<MetadataColumn<?>, Map<RawDataFile, Object>> getData() {
+  public Map<MetadataColumn<?>, ConcurrentMap<RawDataFile, Object>> getData() {
     return data;
   }
```
robinschmid, line 79: Generally you would always try to use the general types/interfaces. SO instead of ConcurrentMap always use Map in the declaration. But its fine here for now as there will only be one implementation

robinschmid:
Its getting there

robinschmid:
The table looks great. Lets move the title column to the bottom just above the data. This is a bit more convenient for other people to parse
image

src/main/java/io/github/mzmine/project/parameterssetup/table/WideTableExportUtility.java
```diff
@@ -0,0 +1,223 @@
...
+
+  @Override
+  public boolean importFrom(File file, boolean appendMode) {
+    try (FileReader fr = new FileReader(file); BufferedReader bufferedReader = new BufferedReader(
+        fr)) {
+      // read the header
+      List<String> parametersTitles = new ArrayList<>(
+          List.of(bufferedReader.readLine().split("\t", -1)));
+      List<String> parametersDescriptions = new ArrayList<>(
+          List.of(bufferedReader.readLine().split("\t", -1)));
```
robinschmid, line 142: Why -1. You want to split all of them so just use split("\t")

src/main/java/io/github/mzmine/project/parameterssetup/table/WideTableExportUtility.java
```diff
@@ -0,0 +1,223 @@
...
+      List<String> parametersTypes = new ArrayList<>(
+          List.of(bufferedReader.readLine().split("\t", -1)));
+
+      // represents the names of the RawDataFiles
+      StringMetadataColumn dataFileCol = new StringMetadataColumn("Datafile", "File name");
+
+      // compare the headers
+      if (!parametersTitles.get(0).equals(HeaderFields.TITLE.toString())
```
robinschmid, line 150: nice sanity checks

src/main/java/io/github/mzmine/project/parameterssetup/ProjectMetadataParameters.java
```diff
@@ -22,13 +22,17 @@
...
+import io.github.mzmine.parameters.parametertypes.TextParameter;
 import java.util.stream.Stream;
 
 public class ProjectMetadataParameters extends SimpleParameterSet {
 
   public static final StringParameter title = new StringParameter("Title",
       "Title of the new parameter", "", true, true);
 
+  public static final TextParameter description = new TextParameter("Description",
+      "Description of the new parameter", "");
```
robinschmid, line 34: add a boolean valueRequired state to this parameter - by default set to true as this was the default before.  
Here for the descriptions set it to false.
You need to adjust the checkValue method then

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialog.fxml
```diff
@@ -32,9 +32,10 @@
         </TableView>
         <ButtonBar buttonMinWidth="50.0" prefHeight="40.0" prefWidth="200.0">
```
robinschmid, line 33: - Wrap the VBox in a BorderPane with the tableView as its center and the VBox with buttons as bottom.
- Remove maxHeight from VBox

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
```diff
@@ -153,43 +162,25 @@ private TableColumn<ObservableList<StringProperty>, String> createColumn(final i
...
-          case DateMetadataColumn dateMetadataColumn -> {
-            parameterMatchedType = "Datetime";
-            parameterMatchedExample = "\"2022-12-24T10:11:36\"";
-            parameterMatchedDefaultValue = LocalDateTime.now();
-            yield dateMetadataColumn;
-          }
+          case StringMetadataColumn stringMetadataColumn -> stringMetadataColumn;
+          case DoubleMetadataColumn doubleMetadataColumn -> doubleMetadataColumn;
+          case DateMetadataColumn dateMetadataColumn -> dateMetadataColumn;
         };
```
robinschmid, line 169: This can be deleted. it only casts the parameter to parameterMatched but still to the generic MetadataColumn.  just use parameter

robinschmid:
Merge the latest master

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
```diff
@@ -223,16 +216,22 @@ public void addPara(ActionEvent actionEvent) {
         return;
       }
 
+      // it's important to replace tabs due to the tsv file format
+      String parameterDescriptionVal = parameterDescription.getValue().replace("\t", " ");
+
       // add the new column to the parameters table
       switch (AvailableTypes.valueOf(parameterType.getValue())) {
```
robinschmid, line 223: This one slipped through last time. 
this parameter should be ComboParameter<AvailableTypes> instead of String type. Java automatically uses toString to show an enum value on a combobox. 
This way you do not have to cast to String inside the parameter and here back to the enum.

src/main/java/io/github/mzmine/project/parameterssetup/table/WideTableExportUtility.java
InspiritedCoder: AFAIK, we should add some negative number as the second argument of .split(), otherwise all empty substrings would be discarded which can cause import failure (e.g. if we would have an empty parameter's description value)

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
InspiritedCoder: It doesn't work for me without this pattern matching. I guess that it's necessary for inferring of the generic parameter's T (in MetadataColumn<T>) type

src/main/java/io/github/mzmine/project/parameterssetup/table/WideTableExportUtility.java
robinschmid: Man that is such a weird standard behavior. it is retaining empty strings but discarding trailing empty strings. 
I guess you are right

src/main/java/io/github/mzmine/project/parameterssetup/ProjectParametersSetupDialogController.java
robinschmid: you are casting the wildtype MetadataColumn<?> to MetadataColumn
Just change parameter to MetadataColumn without the <?>

InspiritedCoder: Yeah, now it makes sense, thanks, I fixed it and pushed the changes

robinschmid:
Great thanks for the progress

PR #695, opened 2022-05-13
Aird2.0 support

lumiaoshan:
Aird format is update to 2.0.0. The new compression algorithm ComboComp is enbaled in 2.0.0, which can achieve 20% more compression ratio than Aird 1.0 
Aird2.0 also supports the conversion of TIMSTOF PASEF files, which also has a compression improvement of about 20%-25% compared to vendor files.


SteffenHeu:
hey lumiaoshan,

it's great to see the timstof data implemented to aird! In MZmine we have the 
´´´
io.github.mzmine.datamodel.Frame
io.github.mzmine.datamodel.MobilityScan
io.github.mzmine.datamodel.IMSRawDataFile
´´´
data types to represent ion mobility experiments.

In the PasefScan you created, I only see one set of mz + intensity values. Are these the accumulated frame values? It's better to use the existing data structures than to create a new one, so all downstream processing works as expected. 

If you have any questions about the IMS implementation in mzmine, feel free to reach out to me via the MZmine slack!

lumiaoshan:
This is indeed a version code for accumulated frame of TIMS. Aird can only know whether it is PASEF mode or other acquisition mode after parsing the metadata part(determined whether using RawDataFile or IMSRawDataFile). According to the current architecture of MZMine, I need to initialize the metadata JSON twice, and I am thinking about how to implement it more elegantly. The main point of this submission is the latest combined compression algorithm-ComboComp(Relevant papers have just been written and are being submitted). 
It will take some time to make Aird better compatible with MZmine's PASEF mode. I will study the new architecture of MZmine again. It will take some time to make Aird more compatible with MZmine's PASEF mode. I will further study MZmine's new architecture to determine how the integration would be more appropriate. Thanks @SteffenHeu for the reply.

MauryanKansara:
hey @SteffenHeu can you send me the Slack link please??

SteffenHeu:
Hi @lumiaoshan, unfortunately we have to remove aird for the next release because we found it contains a library that is licensed under gpl license, which is incompatible with the MIT license of mzmine.  

It would be great if that coud be adressed in aird 2.0, so we can implement it again

lumiaoshan:
Hi @SteffenHeu，Thank you for informing me of this problem, I will solve this problem as soon as possible.

Issue #630, opened 2022-03-28
Add date and time of data collection to RawDataFile


robinschmid:
Read date and time from raw data files and provide this information through the metadata
Maybe use LocalDateTime?
https://www.baeldung.com/java-8-date-time-intro

- [ ] mzML (xml element is maybe startTimeStamp)
- [ ] mzXML
- [ ] .d
- [ ] .tdf
- [ ] thermo raw --> uses mzML

Needs #633 to fix project save

Project/Set sample parameters
![image](https://user-images.githubusercontent.com/10366914/160572762-e42467b6-3e96-4f98-bd96-a7ff0f7c91c1.png)

nikhilgotpagar:
I want to work on this issue  ! Please assign me this

Mohamadi98:
Hello, if no one is working on this issue currently, I would like to try and work on it. I've already set up MZmine locally using this guide https://mzmine.github.io/mzmine_documentation/contribute_intellij.html, and the project is working fine. 
It seems like date-time was already added to `RawDataFile` in this issue #696. 
If I'm understanding the codebase correctly, to complete this issue a method should be implemented in src/main/java/io/github/mzmine/modules/io/projectsave/RawDataFileSaveHandler.java that gets the date-time information, please correct me if I'm wrong.

PR #696, opened 2022-05-13
#630: extracting run_date parameter from a RawDataFile

InspiritedCoder:
I've added a **startTimeStamp** attribute to MzMLRawDataFile object, so the value of the run_date parameter would be saved in it while parsing a RawDataFile.

However, I've encountered a problem with using this attribute to add the parameter value into our metadata table. Ideally, I'd like to cast "general" RawDataFile instance to MzMLRawDataFile, thus I'd be capable to use that startTimeStamp attribute, but for some reason io.github.msdk.datamodel.RawDataFile class is used in MSDKmzMLImportTask.java (147 line) instead of io.github.mzmine.datamodel.RawDataFile, so the cast is unreal...


src/main/java/io/github/mzmine/modules/io/import_rawdata_mzml/msdk/data/MzMLRawDataFile.java
```diff
@@ -42,6 +42,7 @@ public class MzMLRawDataFile implements RawDataFile {
   private @NotNull String defaultInstrumentConfiguration;
   private @NotNull String defaultDataProcessingScan;
   private @NotNull String defaultDataProcessingChromatogram;
+  private String startTimeStamp;
```
robinschmid, line 45: Try adding null annotations here and in the methods @NotNull should be fine. and Then use empty string instead.

src/main/java/io/github/mzmine/modules/io/import_rawdata_mzml/msdk/data/MzMLParser.java
```diff
@@ -77,20 +78,32 @@ public void processOpeningTag(XMLStreamReaderImpl xmlStreamReader, InputStream i
...
+      // startTimeStamp may be optional, so it makes no sense to stop import of a RawDataFile
+      // if this tag is skipped
+      try {
+        final CharArray startTimeStamp = getRequiredAttribute(xmlStreamReader,
+            MzMLTags.ATTR_START_TIME_STAMP);
+
+        newRawFile.setStartTimeStamp(startTimeStamp.toString());
+        System.out.println("startTimeStamp value is: " + startTimeStamp);
+      } catch (IllegalStateException illegalStateException) {
+        System.out.println("startTimeStamp wasn't set");
```
robinschmid, line 94: use logger.info

src/main/java/io/github/mzmine/modules/io/import_rawdata_mzml/msdk/data/MzMLParser.java
```diff
@@ -77,20 +78,32 @@ public void processOpeningTag(XMLStreamReaderImpl xmlStreamReader, InputStream i
...
-          getRequiredAttribute(xmlStreamReader, MzMLTags.ATTR_DEFAULT_INSTRUMENT_CONFIGURATION_REF);
+      final CharArray defaultInstrumentConfigurationRef = getRequiredAttribute(xmlStreamReader,
+          MzMLTags.ATTR_DEFAULT_INSTRUMENT_CONFIGURATION_REF);
       newRawFile.setDefaultInstrumentConfiguration(defaultInstrumentConfigurationRef.toString());
+
+      // startTimeStamp may be optional, so it makes no sense to stop import of a RawDataFile
+      // if this tag is skipped
+      try {
+        final CharArray startTimeStamp = getRequiredAttribute(xmlStreamReader,
+            MzMLTags.ATTR_START_TIME_STAMP);
```
robinschmid, line 89: Just use the method to get the attribute directly or create a simple method to getAttributeUnchecked 
`xmlStreamReader.getAttributeValue(null, attr);`

robinschmid:
The MSDK data model is legacy but we are still using it here. 
The best way would be one of these two options:
1. after adding the file to the project (MSDKmzMLImportTask.java; line 185) add the date to the metadata
2. add the date to the RawDataFile interface and its implementations before calling project.addFile() and then while adding a raw data file, also set its acquisition date

I think I prefer option 1 because that eliminates doubling the data. And in any case people should use the date from the metadata because people can define it there.

src/main/java/io/github/mzmine/modules/visualization/projectmetadata/ProjectParametersSetupPaneController.java
```diff
@@ -63,6 +64,15 @@ private void initialize() {
     parameterTable.setEditable(true);
     parameterTable.getSelectionModel().setCellSelectionEnabled(true);
     fileList = currentProject.getDataFiles();
+
+    // set the run start timestamp of the sample for each file
+    DateMetadataColumn startTimeStamp = new DateMetadataColumn("Run start time stamp",
```
robinschmid, line 69: title should be shorter. Maybe 
run_date

src/main/java/io/github/mzmine/modules/visualization/projectmetadata/ProjectParametersSetupPaneController.java
```diff
@@ -63,6 +64,15 @@ private void initialize() {
...
+        "Run start time stamp of the sample");
+    for (RawDataFile rawFile : fileList) {
+      metadataTable.setValue(startTimeStamp, rawFile, rawFile.getStartTimeStamp());
+      logger.info("Run start time stamp: " + rawFile.getStartTimeStamp());
+    }
```
robinschmid, line 74: this only runs once the user opens the dialog. Very often people just run the whole procedure without the GUI. So this should all happen once a new file is added

PR #703, opened 2022-05-31
Options for TaskManager position

robinschmid:
Options to show task manager in
- hide
- main
- tab
- external window

That fixes two things:
- you can actually see more tasks at the same time
- You can better see the raw data overview

See examples of the task manager in a tab or window: 
image

image


SteffenHeu:
"external" is a weird window name. just call it external window or so. 
Also i'd put it to the windows menu, not the help.

robinschmid:
Like you wish @SteffenHeu 

image

PR #705, opened 2022-05-31
Clean Raw data overview

robinschmid:
put all parameters in an accordion


robinschmid:
image


image

robinschmid:
This is without all the other changes

robinschmid:
How about this @SteffenHeu 

image

image

PR #738, opened 2022-06-15
Spectral library search in wizard

robinschmid:
- Adds import of libraries
- spectral library search for LC-MS2 
- improves layout

Only need to specify some libraries and the spectral library matching is added - without additional parameter.

Creates this batch
image

On large display
image

On smaller display
image


robinschmid:
Creates this batch
image

On large display
image

On smaller display
image

PR #740, opened 2022-06-15
Updated MSn Feature Detection

Kevin-Murray:
Updated MSn feature detection module to reflect changes to MSn spectra implementations. 

Brief description of changes:

- MS Level set directly in Scan Selection field. Produces error if  set to 1 or left empty.
- Scan array is restricted to only specified MSn scans in range.
- Precursor MZ is obtained directly from scan (MS2 only) or through MSnInfoImpl (getMS2PrecursorMZ) for MSn scans. 
- Simple modular feature is built for each precursor using specified MZ and RT tolerances. 
- Help documents reflect changes to algorithm, input parameter changes, and issues arising from using Thermo Raw files converted using older versions of MSConvert or ThermoRawFileParser.

I've tested the module using data generated from Thermo Tribrid instruments only. Works for MS2 and MS3 feature detection. I have not tested for MSn>3 or other vendor platforms.


Kevin-Murray:
I need to add a commit. I realize I didn't add the scans to the created ModularFeatureList. Presently, you could not use the resulting feature list with some of the chromatogram resolving tools. 

Additionally, I could update the module to use EffecientDataAccess for scan looping.

Issue #730, opened 2022-06-13
Compute the network layout in other thread (not JavaFX) 


robinschmid:
Currently, the layout is computed on the JavaFX thread, explore options to run on other thread for performance.

https://github.com/mzmine/mzmine3/blob/6fc516e958a399c6f3a560167f559a806b5e27b3/src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java#L239-L245

PR #742, opened 2022-06-16
Fix for the issue #730

sanbit876:
Issue #730 
 Replaced the existing threading model with GRAPH_IN_ANOTHER_THREAD model


robinschmid:
The conflict is resolved when you merge the latest master into your branch. 
Did you check this option?
Its best if you have a small example dialog where you just visualize synthetic data.

PR #750, opened 2022-06-27
682 metadata batch mode

InspiritedCoder:


src/main/java/io/github/mzmine/modules/visualization/projectmetadata/ProjectMetadataImportModule.java
```diff
@@ -0,0 +1,99 @@
...
+  public @NotNull MZmineModuleCategory getModuleCategory() {
+    return MZmineModuleCategory.PROJECTMETADATA;
+  }
+
+  @Override
+  public @NotNull ExitCode runModule(@NotNull MZmineProject project,
+      @NotNull ParameterSet parameters, @NotNull Collection<Task> tasks,
+      @NotNull Instant moduleCallDate) {
+    // get the all selected files
+    File[] fileNames = parameters.getParameter(MzMLImportParameters.fileNames).getValue();
```
robinschmid, line 71: use the correct parameters class here

src/main/java/io/github/mzmine/modules/visualization/projectmetadata/ProjectMetadataImportModule.java
```diff
@@ -0,0 +1,99 @@
...
+      }
+
+      if (metadataTable.importMetadata(fileName, false)) {
+        logger.info("Successfully imported parameters from " + fileName);
+      } else {
+        MZmineCore.getDesktop().displayErrorMessage("Couldn't import metadata from " + fileName);
+        logger.warning("Error while processing " + fileName);
+        return ExitCode.ERROR;
+      }
+    }
```
robinschmid, line 95: You should put this into a task similar to most other modules. This way you can easily control when it fails etc
One task that loads all files should be fine.
you just have to add the task to the list of tasks

robinschmid (approved):
Great thanks Kirill

PR #751, opened 2022-06-27
Create .mailmap to map redundant committers

robinschmid:
This way git will know that those committers are the same person


tomas-pluskal:
Great idea! I wonder if this is also going to apply to all the Insights plots on GitHub.

robinschmid:
I think it does but there were multiple people with different experiences. Lets see

PR #752, opened 2022-06-27
Gource analysis of the repository

robinschmid:
Added pictures of contributors that can be used by gource to visualize the codebase graph


SteffenHeu:
Not sure if all contributors are ok with adding their picture to the repository.

robinschmid:
I mean we should just ask them. All pictures are from GitHub. Also we might add some of those pictures to the website if people are okay with it.

PR #760, opened 2022-07-01
Use scan selection for mobility scans in ScanAccess

robinschmid:
Mobility filter was not applied. Together with @SteffenHeu


SteffenHeu:
Did you push the latest changes? The test is failing

PR #769, opened 2022-07-07
fix isotope finder adding duplicate data points

SteffenHeu:
isotope finder module created a new data point with feature mz + intensity as a starting point. on some occasions, mostly more complex isotope patterns than just carbon, the actual data point in the scan was also added and lead to duplicate signals in the isotope pattern.

### DIFF
--- a/src/main/java/io/github/mzmine/util/IsotopesUtils.java
+++ b/src/main/java/io/github/mzmine/util/IsotopesUtils.java
@@ -240,7 +240,7 @@ public class IsotopesUtils {
           for (int j = i + 1; j < abundantIsotopes.size(); j++) {
             final double deltaMZ =
                 (abundantIsotopes.get(j).getExactMass() - abundantIsotopes.get(i).getExactMass())
-                / charge;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/mzmine/util/IsotopesUtils.java
```diff
@@ -454,7 +458,7 @@ public static List<DataPoint> findIsotopesInScan(List<Double> isoMzDiffs, double
         final var dataPoint = new SimpleDataPoint(mz, spectrum.getIntensityValue(dp));
         if (mz > maxMZ) {
           candidates.add(dataPoint);
-          maxMZ = mz;
+          maxMZ = Math.max(mz, maxMZ);
         } else {
```
robinschmid, line 462: why Math.max if we are already checking mz>maxMZ

robinschmid: Seems like there was an issue without it. Will merge for now.

PR #770, opened 2022-07-07
Isofinder remove redundant rows

SteffenHeu:
includes #769 

adds the option to remove rows that are not the most intense or the first peak in an isotope pattern (useful for deisotoping a feature list with more than 13C)


robinschmid:
Have to check it but I am not sure if this should be done here. 
We do have the 13C filter in the rows filter... The finder is supposed to find all possible signals but not to score or filter. This should be done by other steps

robinschmid:
Filters will go into RowsFilterTask

PR #771, opened 2022-07-08
Updated ThermoRawFileParser to v1.4.0

tomas-pluskal:
This should fix the issue with reading scans of higher MS levels, as discussed here: 
https://github.com/compomics/ThermoRawFileParser/issues/126


SteffenHeu:
thanks, Tomas!

PR #790, opened 2022-07-14
Improved error reporting of the Thermo raw parser

tomas-pluskal:


robinschmid:
Adapted and optimized to ship the unzipped tool in an external_tools directory in #811

PR #793, opened 2022-07-15
Adapting mzTab-M exporter to new feature handling.

nilshoffmann:
This is just a first step and not the complete implementation yet. 
**Please do not merge**!


src/main/java/io/github/mzmine/modules/io/export_features_mztabm/MZTabmExportTask.java
```diff
@@ -34,277 +35,374 @@
...
+
+                    //Cancelled?
+                    if (isCanceled()) {
+                        return;
+                    }
+                    sm.setReliability("2");
+                    FeatureIdentity featureIdentity = featureListRow.getPreferredFeatureIdentity();
+                    List<IonIdentity> ionIdentities = featureListRow.getIonIdentities();
+                    // set default identification method
+                    sme.setIdentificationMethod(new Parameter().name("Unknown"));
```
SteffenHeu, line 243: How about we add a getIdentificationMethodName() to the AnnotationType, that you can use here. Then we can loop over all annotations similar to the ModularFeatureListRow.isIdentified() method. Since all of the new annotations are a ListWithSubsType you can then loop over the subtypes (CompoundNameType, FormulaType... are reused) and fill in the fields of the mztab file.

nilshoffmann: Sounds good!

robinschmid: So the ListWithSubsType should extend something like AnnotationsContainer and this would have the method name?

src/main/java/io/github/mzmine/modules/io/export_features_mztabm/MZTabmExportParameters.java
```diff
@@ -44,8 +44,15 @@ public class MZTabmExportParameters extends SimpleParameterSet {
 
   public static final BooleanParameter exportAll = new BooleanParameter("Include all peaks"//
       , "Include peaks with unknown identity");
+  
+  public static final BooleanParameter validateOnWrite = new BooleanParameter(
+      "Validate mzTab-M on write ",
```
robinschmid, line 49: remove trailing white space after on write

src/main/java/io/github/mzmine/modules/io/export_features_mztabm/MZTabmExportParameters.java
```diff
@@ -44,8 +44,15 @@ public class MZTabmExportParameters extends SimpleParameterSet {
 
   public static final BooleanParameter exportAll = new BooleanParameter("Include all peaks"//
       , "Include peaks with unknown identity");
```
robinschmid, line 46: please add default value

src/main/java/io/github/mzmine/modules/io/export_features_mztabm/MZTabmExportParameters.java
```diff
@@ -44,8 +44,15 @@ public class MZTabmExportParameters extends SimpleParameterSet {
...
+  
+  public static final BooleanParameter validateOnWrite = new BooleanParameter(
+      "Validate mzTab-M on write ",
+      "If \"true\", disable validation of mzTab-M file during writing. "
+      + "This may lead to inconsistent broken files that may not be "
+      + "recoverable by other tools. Set to \"false\" to debug issues with "
+      + "feature export only!", true);
```
robinschmid, line 53: wait true disables the validation?
shouldnt it be the other way around

nilshoffmann:
Obsolete due to #1043

PR #802, opened 2022-07-19
MassDetector that uses the lowest intensity * user factor as noise level

robinschmid:
Each individual scan will have a noise level defined by the lowest intensity and a user factor. Only works with centroid data for now.


SteffenHeu (commented):
Why this and not the injection time thing?

robinschmid:
I think this option should have been there long ago because it's the simplest thing after just using a static noise level. 
In case people use non-trap data or converted with an older version of MS convert and do not have an inject time information this mass detector works and is easy to understand

SteffenHeu:
Is the spectrum type correctly set to thresholded if prefiltered data is imported? if yes, do we have a warning message with this mass detector for that?

robinschmid:
Merged with @SteffenHeu

PR #806, opened 2022-07-20
Add specturm to inject time analysis

robinschmid:
When you click on a circle in the plot you open the spectrum in the bottom plot. For this I added a sealed class that either gives features or Scan for an XYDataset like ColoredXYDataset

### DIFF
--- a/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/XYItemScanProvider.java
+++ b/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/XYItemScanProvider.java
@@ -18,18 +18,23 @@
 
 package io.github.mzmine.gui.chartbasics.simplechart.providers;
 
-import org.jetbrains.annotations.Nullable;
-import org.jfree.chart.renderer.PaintScale;
+import io.github.mzmine.datamodel.Scan;
 
 /**
- * Used to color XYZ plots based on their z values.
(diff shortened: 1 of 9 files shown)


src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/XYItemScanProvider.java
```diff
@@ -0,0 +1,40 @@
...
+import io.github.mzmine.datamodel.Scan;
+
+/**
+ * Scans connected to each XYItem in an XY dataset and chart
+ *
+ * @author Robin Schmid <a href="https://github.com/robinschmid">https://github.com/robinschmid</a>
+ */
+public non-sealed interface XYItemScanProvider extends XYItemObjectProvider<Scan> {
+
+  Scan getScan(int series, int item);
```
SteffenHeu, line 30: We never use the "series" in the coloredxydataset and the providers, so i don't think it makes sense to introduce it here.

robinschmid: I made it so that it is generally applicable to jfrerchart charts.... I think that is reasonable.

robinschmid: Removed the usage of series

robinschmid:
Merged together with @SteffenHeu

PR #807, opened 2022-07-20
Update SiriusExportTask.java

Adafede:
Hi to the team!

Just ran into this issue while exporting both "classical" and "SIRIUS" MGF's.

It is a dirty fix and having a small helper called everywhere in the exports could be nicer but for the moment, it does the work. (I did not check other exports)

Best,


robinschmid:
Hey Adriano, 
Never tried to export to a nonexistent folder but you are right. Thanks for the fix. And great first contribution :)

Maybe you could do the same for the project?

robinschmid:
I pushed a small fix to your branch. The File curFile was not declared. Rest is good. 

If you really want to start and work in an IDE that helps you spot those issues, navigate the code better, and build your own versions - you could get the student/teacher version of IntelliJ and follow this part of our documentation:
https://mzmine.github.io/mzmine_documentation/contribute_intellij.html

Adafede:
Thank you for the tip!

Yup, overlooked it, as the names are different between gnps MGF export and sirius MGF export... probably part of the upcoming housekeeping 😜

PR #811, opened 2022-07-21
Thermo raw file parser as external tool

robinschmid:
I pushed the parser into an external_tools folder which will be copied into the MZmine directory. We access the tool by path with the ExternalTool enum. 

Added the error loggings from tomas


robinschmid:
@tomas-pluskal @Annexhc @roman-bushuiev can someone try the version on mac
Windows seems to work
@Annexhc can you try linux?

tomas-pluskal:
This messes up the macOS build. The final .dmg only contains the external_tools folder and no MZmine.app. The portable zip build contains the external_tools folder and the MZmine.app, but the external_tools should be included inside MZmine.app (on macOS apps are folders). Maybe @roman-bushuiev could try to fix this?

roman-bushuiev:
> This messes up the macOS build. The final .dmg only contains the external_tools folder and no MZmine.app. The portable zip build contains the external_tools folder and the MZmine.app, but the external_tools should be included inside MZmine.app (on macOS apps are folders). Maybe @roman-bushuiev could try to fix this?

I can try.

roman-bushuiev:
@robinschmid, did you implement it by some tutorial? I found a [related Stack Overflow thread](https://stackoverflow.com/questions/56778477/how-to-bundle-extra-files-with-jlink-image), and the answer there proposes to do the copying in a slightly different way.

build.gradle
```diff
@@ -337,6 +337,14 @@ runtime {
...
     }
 }
+def JLINK_DIR = "$buildDir/jpackage/MZmine"
+
+tasks.jpackage.doFirst {
+    copy {
+        from "${projectDir}/external_tools"
+        into JLINK_DIR + "/external_tools"
+    }
+}
```
robinschmid, line 347: Its very similar to their approach. We always run jpackage, which includes jlink as a previous step.
So the doFirst here should be correct. At least it works on windows. we just have to adjust the path in case its macOS

roman-bushuiev:
I will work on macOS signing after Dec 15. I hope the updated macOS build will implicitly fix it. 😆

robinschmid:
@ansgarkorf

SteffenHeu:
need to check this again @robinschmid @ansgarkorf

PR #815, opened 2022-07-22
NIST MS Search Update - Add SpectralDBAnnotation

Kevin-Murray:
Update to the NIST MS Search Module to reflect addition of SpectralDBAnnotation methods for recording Spectral Library Search Results.

Brief Summary of Changes:
- Add import option (Overwrite or Append). This a quality of life update.
- Add SpectralDBAnnotation to record results from NIST MS Search.
- Improve result REGEX patterns for more flexible parsing of incomplete database results.


src/main/java/io/github/mzmine/modules/dataprocessing/id_nist/ImportOption.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along with MZmine; if not,
+ * write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+package io.github.mzmine.modules.dataprocessing.id_nist;
+
+public enum ImportOption {
+
+  PREPEND("Append"),
```
robinschmid, line 22: I guess you should decide for one :D
Pre or append

src/main/java/io/github/mzmine/modules/dataprocessing/id_nist/NistMsSearchParameters.java
```diff
@@ -54,24 +54,17 @@ public class NistMsSearchParameters extends SimpleParameterSet {
    * MS Level for search.
    */
   public static final IntegerParameter MS_LEVEL = new IntegerParameter("MS level",
-      "Choose MS level for spectal matching. Enter \"1\" for MS1 precursors or ADAP Cluster Spectra.",
+      "Choose MS level for spectal matching. Enter \"1\" for MS1 precursors or ADAP-GC Cluster Spectra.",
```
robinschmid, line 57: maybe to make it clearer say 
for MS1 spectra or ADAP-GC-MS clustered spectra

src/main/java/io/github/mzmine/modules/dataprocessing/id_nist/NistMsSearchParameters.java
```diff
@@ -54,24 +54,17 @@ public class NistMsSearchParameters extends SimpleParameterSet {
...
       2, 1, 1000);
 
   /**
    * Match factor cut-off.
    */
-  public static final IntegerParameter MIN_MATCH_FACTOR = new IntegerParameter("Min. match factor",
-      "The minimum match factor (0 .. 1000) that search hits must have", 700, 0, 1000);
+  public static final IntegerParameter DOT_PRODUCT = new IntegerParameter("Min. Dot Product",
+      "The minimum cosine dot product (0 .. 1000) that search hits must have", 700, 0, 1000);
```
robinschmid, line 64: lets do title: **Min cosine similarity**
Also a question would be if we should use a DoubleParameter with the MZmineCore.getPreferences().getScoreFormatter() then its more similar to other matching modules in MZmine.

In the task you can still convert it to the NIST MS search output either way

robinschmid (commented):
Thanks for the changes. They will improve the usage of this module. 
One question, did you try opening feature list/right click/show/Spectral DB search results
I guess it will throw some issues with the spectral data missing right?

src/main/java/io/github/mzmine/modules/dataprocessing/id_nist/ImportOption.java
Kevin-Murray: Haha good catch. In my defense, the NIST MS Search tool is inconsistent with its naming too. The import option is called 'Prepend' but the command line option is 'Append'.

I will just stick with Append. =)

Kevin-Murray:
>

src/main/java/io/github/mzmine/modules/dataprocessing/id_nist/NistMsSearchParameters.java
Kevin-Murray: > I guess it will throw some issues with the spectral data missing right?

It doesn't produce an error, but it never leaves the processing step. It just says "I'm working on it" and never progresses. 

The SpectralDBAnnotations software field is explicitly 'NIST MS Search'. We could add a message saying Spectral data visualization is not supported for this software. 

What do you think?

src/main/java/io/github/mzmine/modules/dataprocessing/id_nist/NistMsSearchParameters.java
Kevin-Murray: Easy change. Shouldn't have any impact of the results.

src/main/java/io/github/mzmine/modules/dataprocessing/id_nist/NistMsSearchParameters.java
robinschmid: Those could be mixed with the spectral db matches from our own module.... so I guess they should just not show up in the dialog. or there should be a placeholder in the dialog saying (NIST Search match without spectral preview)

Kevin-Murray:
Presently, it works for me. I've tested it with data generated on a Thermo Orbitrap instrument and Agilent GC-MS. 

I can be responsible for maintaining and updating the module in the future.

Kevin-Murray:
I think I found a solution for the view Spectral DB Search Results issue. 

The visualization module will now skip results with empty SpectralDBEntry data points field. If the row has other entries with valid SpectralDBEntries (ie, those with library spectra data points), they will be displayed. If no valid match is found, the initial no match entry will update and inform users how to visualize NIST MS Search results.

robinschmid:
This way, the NIST library matches will not appear in the dialog. 

We could also modify `SpectralMatchPanelFX` in the constructor where the mirror chart is created. You could check if hit contains spectral data. If not you create a blank pane with a label instead saying "no spectral data available", this would then work for all cases, the user would still see all matches and also all metadata. 

Then you do not need to change things in `SpectralIdentificationResultsWindowFX`

Here is the relevant code in the constructor:
```java
    mirrorChart = MirrorChartFactory.createMirrorPlotFromSpectralDBPeakIdentity(hit);
    MZmineCore.getConfiguration().getDefaultChartTheme().apply(mirrorChart.getChart());
    mirrorChartWrapper = new BorderPane();
    mirrorChartWrapper.setCenter(mirrorChart);
```

robinschmid:
If you want to give a specific message to the NIST search module, you could set an attribute for the identification method/module in the `SpectralDBAnnotation`, check it when creating the mirror chart, and display a customized label instead

Kevin-Murray:
I made the changes suggested. However, creating a customized label in the plot is beyond my current skill set. Instead, I added a comment to the SpectralDBAnnotation results noting the results can only be viewed in the NIST MS Search program. I admit this is a bit lazy but it should get the job done.

empty_result

robinschmid:
Great. Thanks @Kevin-Murray

PR #816, opened 2022-07-22
GAT fix analytics for MZmine 3

robinschmid:
Bad news the GAT was not working for MZmine 3. Somehow Tasks were not tracked and module calls were tracked with name and they broke when there were spaces because they were not escaped.

TODO remove superfluous loggings


tomas-pluskal:
It looks good to me. I assume you checked that the tracking actually works in the GAT website?

tomas-pluskal (dismissed):

robinschmid:
Yes. It works. But each tracking call is made from a different user. Would be nice to keep it session wise but no idea how to accomplish that.

robinschmid:
Merged with @SteffenHeu

PR #825, opened 2022-08-01
[GSOC'22 Project] Latest version of FeatureList Networking Visualization Module

sanbit876:


src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -269,19 +282,12 @@ public void collapseIonNodes(boolean collapse) {
...
-            setVisible(edge, showNetRelationsEdges);
-            break;
-          default:
-            break;
+          case ION_IDENTITY -> setVisible(edge, !collapse && showIonEdges);
+          case MS2_SIMILARITY_NEUTRAL_M_TO_FEATURE, MS2_SIMILARITY_NEUTRAL_M, MS2_SIMILARITY ->
+              setVisible(edge, showMs2SimEdges);
+          case NETWORK_RELATIONS -> setVisible(edge, showNetRelationsEdges);
+          default -> {
+          }
```
robinschmid, line 290: we should make the switch exhaustive so handle all cases and then get rid of the default. This way we will se that we need to handle new cases here

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
```diff
@@ -185,6 +185,7 @@ public class NetworkPane extends BorderPane {
   protected Graph graph;
   protected Viewer viewer;
   protected FxViewPanel view;
+  protected Node selectedNode;
```
robinschmid, line 188: selected node should be an array or a list. 
array might be nice considering that we could have a method 
`setSelectedNodes(@Nullable Node... selectedNodes)`

this way we can set one or multiple nodes

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
```diff
@@ -241,13 +242,14 @@ public NetworkPane(String title, String styleSheet2, boolean showTitle) {
     graph.setAutoCreate(true);
     graph.setStrict(false);
 
-    viewer = new FxViewer(graph, Viewer.ThreadingModel.GRAPH_IN_GUI_THREAD);
+    viewer = new FxViewer(graph, Viewer.ThreadingModel.GRAPH_IN_ANOTHER_THREAD);
     viewer.enableAutoLayout();
-
+    graph.setAttribute("Layout.frozen"); //Block the layout algorithm
```
robinschmid, line 247: Can we just do this here without checking if the layout is already finished?
The layout runs multiple iterations to stabilize. So will this always work or does it work sometimes?

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
```diff
@@ -241,13 +242,14 @@ public NetworkPane(String title, String styleSheet2, boolean showTitle) {
...
     view = (FxViewPanel) viewer.addDefaultView(false);
     // wrap in stackpane to make sure coordinates work properly.
     // Might be confused by other components in the same pane
     StackPane graphpane = new StackPane(view);
+    view.setStyle("-fx-effect: innershadow(three-pass-box, gray, 12 , 0.5, 1, 1);");
```
robinschmid, line 252: why the inner shadow? is it to plain otherwise?

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
```diff
@@ -285,6 +287,8 @@ public void mouseLeft(String id) {
     view.setOnScroll(event -> zoom(event.getDeltaY() > 0));
 
     view.setOnMouseClicked(e -> {
+      selectedNode = (Node) view.findGraphicElementAt(EnumSet.of(InteractiveElement.NODE), e.getX(),
```
robinschmid, line 290: create method like discussed above `setSelectedNode` maybe with a flag `boolean notifyListeners`
its gonna be called from here and from the listener itself.
If the listener sets the selected nodes we of course do not call the listener again

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
```diff
@@ -534,6 +536,10 @@ public Pane getPnSettings() {
     return pnSettings;
   }
 
+  public Node getSelectedNode() {
```
robinschmid, line 539: array or list

src/test/java/modules/TestNodeNeighbourMethod.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.stream.Stream;
+import org.graphstream.algorithm.generator.FullGenerator;
+import org.graphstream.algorithm.generator.Generator;
+import org.graphstream.graph.Graph;
+import org.graphstream.graph.Node;
+import org.graphstream.graph.implementations.MultiGraph;
+import org.junit.Test;
+public class TestNodeNeighbourMethod {
+
+  public Graph CreateGraph()
```
robinschmid, line 12: lower case createGraph

src/test/java/modules/TestNodeNeighbourMethod.java
```diff
@@ -0,0 +1,48 @@
...
+  {
```
robinschmid, line 13: apply default style format

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
sanbit876: ok

src/test/java/modules/TestNodeNeighbourMethod.java
```diff
@@ -0,0 +1,48 @@
...
+    Graph graph = new MultiGraph("Test-Graph");
```
robinschmid, line 14: Where is the graph that you manually entered?

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
sanbit876: As per my observations, the "layout.frozen" attribute sets after the graph is stabilized

src/test/java/modules/TestNodeNeighbourMethod.java
```diff
@@ -0,0 +1,48 @@
...
+    gen.end();
+    int i=0;
+    for (Node node : graph) {
+      node.setAttribute("ui.label",(i+1)+"");
+      i++;
+    }
+    return graph;
+  }
+  @Test
+  public Stream<Node> streamNodeNeighborsBreadthFirst(Node node, int edgeDistance) {
```
robinschmid, line 28: You should push this code into the main code base and only have tests here that actually check some numbers.
Maybe `NetworkUtils` or `GraphStreamUtils`

src/test/java/modules/TestNodeNeighbourMethod.java
```diff
@@ -0,0 +1,48 @@
...
+        Stream<Node> nextLevel = node.neighborNodes().flatMap(
+            neighbor -> streamNodeNeighborsBreadthFirst(neighbor, edgeDistance - 1).distinct());
+        yield Stream.concat(stream, nextLevel);
+      }
+    };
+  }
+  @Test
+  public void printNodes()
+  {
+    streamNodeNeighborsBreadthFirst(CreateGraph().getNode("1"),3).forEach(s -> System.out.println(s.getId()));
```
robinschmid, line 46: this should be the actual test code so do something like counting the number of unique nodes returned
Assertions.assertEquals(expected, actual);

with `graph.nodes().distinct().count()`

src/test/java/modules/TestNodeNeighbourMethod.java
```diff
@@ -0,0 +1,48 @@
...
+    if (edgeDistance < 0) {
+      throw new IllegalArgumentException("Distance cannot be negative, value=" + edgeDistance);
+    }
+    return switch (edgeDistance) {
+      case 0 -> Stream.of();
+      case 1 -> node.neighborNodes();
+      default -> {
+        Stream<Node> stream = node.neighborNodes();
+        Stream<Node> nextLevel = node.neighborNodes().flatMap(
+            neighbor -> streamNodeNeighborsBreadthFirst(neighbor, edgeDistance - 1).distinct());
```
robinschmid, line 38: the distinct here actually does not work as intended because we still end up with duplicates.

I think the best way would actually be to have two methods and use a map to track visited nodes with their remaining edge distance
Both methods should go into the NetworkUtils or something similar

```java

/**
 * Unique list of node neighbors within edge distance
 *
 * @param node         node to visit and all its neighbors
 * @param edgeDistance number of consecutive edges connecting neighbors
 * @return list of all neighbors + the initial node
 */
public List<Node> getNodeNeighbors(Node node, int edgeDistance) {
    Object2IntOpenHashMap<Node> visited = new Object2IntOpenHashMap<>();
    visited.put(node, edgeDistance);
    addNodeNeighbors(visited, node, edgeDistance);
    return new ArrayList<>(visited.keySet());
    }

/**
 * Add all neighbors to the visited list and check for higher edgeDistance to really capture all
 * neighbors
 *
 * @param visited map that tracks visited nodes and their edgeDistance
 * @param node
 * @param edgeDistance
 */
private void addNodeNeighbors(Object2IntOpenHashMap<Node> visited, Node node, int edgeDistance) {
final int nextDistance = edgeDistance-1;
    node.neighborNodes().forEach(neighbor -> {
    if (visited.getOrDefault(neighbor, -1) < edgeDistance) {
    // was never visited or was visited with lower edgeDistance - visit this time 
    visited.put(neighbor, nextDistance);
    if (edgeDistance > 1) {
    addNodeNeighbors(visited, neighbor, nextDistance);
    }
    }
    });
    }
```

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
robinschmid: Okay so lets try it this way

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
robinschmid: Oh we can keep it for now. I will look at it once we are further with the implemented methods

src/test/java/modules/GraphStreamUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import org.graphstream.graph.Graph;
+import org.graphstream.graph.Node;
+import org.graphstream.graph.implementations.MultiGraph;
+import org.junit.Test;
+import org.junit.jupiter.api.Assertions;
+
+public class GraphStreamUtils {
+
+  public Graph createGraph()
+  {
```
robinschmid, line 15: set the mzmine code formatter in IntelliJ and you can have it automatically run before commiting changes

src/test/java/modules/GraphStreamUtils.java
```diff
@@ -0,0 +1,83 @@
...
+        // was never visited or was visited with lower edgeDistance - visit this time
+        visited.put(neighbor, nextDistance);
+        if (edgeDistance > 1) {
+          addNodeNeighbors(visited, neighbor, nextDistance);
+        }
+      }
+    });
+  }
+  @Test
+  public void testmethod()
```
robinschmid, line 78: Your approach to testing was correct before. See this as a reference, they have more info on unit testing.
https://www.baeldung.com/junit-before-beforeclass-beforeeach-beforeall#before
1. The class name should be something like GraphStreamTests or GraphStreamUtilsTests (ending with Tests) and located in the test directory. 
2. only code that is necessary for the testing. So initializing stuff like the createGraph should be done in a method annotated with @Before or @BeforeClass (see link)
3. the actual logic should go in to the main code base, maybe in main/java/....mzmine/util/GraphStreamUtils
4. Make all methods in the utils static - as they do not have a state. So you should access them by GraphStreamUtils.getNodeNeighbors

src/test/java/modules/GraphStreamUtils.java
```diff
@@ -0,0 +1,83 @@
...
+  {
+    String traversingSequence=getNodeNeighbors(createGraph().getNode("A"),2).toString();
+    Assertions.assertEquals("FABJCED",traversingSequence);
```
robinschmid, line 81: Make sure the output actually matches. I think list.toString will not join the strings like this. You could use String.join("", myList)

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -146,48 +153,67 @@ private void addMenu() {
...
+    showGNPSMatches.setMaxWidth(Double.MAX_VALUE);
     showGNPSMatches.setOnAction(e -> showGNPSMatches());
 
     Button showLibraryMatches = new Button("Library matches");
+    showLibraryMatches.setMaxWidth(Double.MAX_VALUE);
     showLibraryMatches.setOnAction(e -> showLibraryMatches());
 
+    Spinner nodeNeighbours = new Spinner(3, 50, 3, 1);
+    Label l = new Label("No. of node neighbours:");
+    neighbours = Integer.parseInt(nodeNeighbours.getValue().toString());
```
robinschmid, line 206: you can make an Intger spinner. Also minimum should be 1, maximum should be huge. can also be Integer.MAX_VALUE or similar
```java
    Spinner<Integer> spinner = new Spinner<>(1, 10000000,3,1);
    int neighbors = spinner.getValue();
```

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -146,48 +153,67 @@ private void addMenu() {
...
+
+    Button updateGraphButton = new Button("Update graph");
+    updateGraphButton.setMaxWidth(Double.MAX_VALUE);
```
robinschmid, line 209: add an action to update the graph

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -424,6 +450,50 @@ private void applyNodeColorStyle() {
...
+        if (edgeDistance > 1) {
+          addNodeNeighbors(visited, neighbor, nextDistance);
+        }
+      }
+    });
+  }
+  public void updateGraph() {
+    graph.removeAttribute("Layout.frozen");
+    graph.nodes().forEach(node -> {
+      if(!(mouseSelectedNodes.contains(node))) {
```
robinschmid, line 491: this variable is not declared. 
do 
Node[] selectedNodes
and use all selected Nodes and their neighbors here

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
```diff
@@ -409,6 +408,11 @@ public void showEdgeLabels(boolean show) {
     });
   }
 
+  public void setSelectedNodes(@Nullable Node mouseSelectedNodes)
+  {
```
robinschmid, line 412: apply style. and call them just selected nodes.

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
```diff
@@ -534,6 +538,10 @@ public Pane getPnSettings() {
     return pnSettings;
   }
 
+  public List<Node> getmouseSelectedNode() {
```
robinschmid, line 541: you can auto generate those methods in intellij and then it is written the correct way. 
getSelectedNodes

you got two typos in there get**m**ouseSelectedNod**e(**)  (should be M and Node**s**)

src/test/java/modules/GraphStreamUtilsTest.java
```diff
@@ -0,0 +1,48 @@
...
+    graph.addEdge("KL", "K", "L");
+    graph.addEdge("JL", "J", "L");
+    graph.addEdge("IJ", "I", "J");
+    graph.addEdge("FJ", "F", "J");
+    graph.addEdge("CJ", "C", "J");
+    graph.addEdge("KJ", "K", "J");
+    for (Node node : graph) {
+      node.setAttribute("ui.label", node.getId());
+    }
+    return graph;
```
robinschmid, line 39: This needs to be a global variable. 
The idea is that @Before lets this run before the tests so you initialize all variables.
Then in the test yuo can just use graph

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkGenerator.java
```diff
@@ -111,6 +110,24 @@ public void createNewGraph(FeatureListRow[] rows, Graph graph, boolean onlyBestN
...
+    graph.removeAttribute("Layout.frozen");
+    List<String> updatedList= new ArrayList<>();
+    for(int k=0;k<neighboringNodes.size();k++)
+    {
+      updatedList.add(k,neighboringNodes.get(k).getId());
+    }
+    int j = 0;
+    for (int i = 0; i < graph.getNodeCount(); i++) {
+      if (!(updatedList.contains(graph.getNode(i).getId()))) {
+        graph.removeNode(i);
```
robinschmid, line 123: you cannot loop over some collection and remove at the same time. for this use an iterator. 
Or collect all elements that you want to delete and then delete them in a second loop.

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -146,48 +153,68 @@ private void addMenu() {
...
     showGNPSMatches.setOnAction(e -> showGNPSMatches());
 
     Button showLibraryMatches = new Button("Library matches");
+    showLibraryMatches.setMaxWidth(Double.MAX_VALUE);
     showLibraryMatches.setOnAction(e -> showLibraryMatches());
 
+    Spinner<Integer> nodeNeighbours = new Spinner<>(1, Integer.MAX_VALUE, 3, 1);
+    Label l = new Label("No. of node neighbours:");
+    neighbours = nodeNeighbours.getValue();
+
```
robinschmid, line 207: You have to update the neighbors value somewhere...
Best to just use an IntegerProperty for everythin JavaFX and bind the valueProperty of the spinner to your neighbors value

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -209,7 +236,19 @@ private void setAttributeForAllEdges(GraphStyleAttribute attribute, EdgeAtt prop
...
+  private void updateGraph()
+  {
+    if(getMouseClickedNode()==null)
+    {
+      Alert alert = new Alert(AlertType.INFORMATION);
+      alert.setContentText("Please click on any node First!!");
+      alert.showAndWait();
+    }
+    else {
+      generator.createGraphwithNeighboringNodes(graph,
```
robinschmid, line 248: adhere to standards. createGraph**W**ith

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -209,7 +236,19 @@ private void setAttributeForAllEdges(GraphStyleAttribute attribute, EdgeAtt prop
...
+          getNodeNeighbors(getMouseClickedNode(), neighbours));
```
robinschmid, line 249: we have to be kind of careful here. the clicked node will be a GraphicsNode and the graph object contains some other class from AbstractNodes

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
robinschmid, line 252: i do not like the shadow. it looks way different then all teh rest of mzmine

robinschmid:
Look at this
https://github.com/sanbit876/mzmine3/pull/4

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FilteredGraph.java
```diff
@@ -0,0 +1,19 @@
+package io.github.mzmine.modules.visualization.networking.visual;
+
+import org.graphstream.graph.implementations.MultiGraph;
+
+
+public class FilteredGraph extends MultiGraph {
+
+  private MultiGraph filteredGraph;
```
robinschmid, line 8: FilteredGraph class is already a MultiGraph. 
So the main field that you have here should be the full graph. You can have a set method or you can pass the full graph in the constructor.

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -69,12 +79,15 @@ public class FeatureNetworkPane extends NetworkPane {
...
   private boolean onlyBest;
+
+  private final IntegerProperty bNeighbors = new SimpleIntegerProperty(1);
   private boolean showNetRelationsEdges;
   private boolean collapse = true;
   private boolean showIonEdges = true;
   private boolean showMs2SimEdges;
   private boolean ms1FeatureShapeEdges = false;
 
+  private Graph fullGraph;
```
robinschmid, line 90: FeatureNetworkPane should only contain things that are specific to the feature network.
the graph is a general field that should be in NetworkPane.
We only have one the FilteredGraph graph
Inside the filtered graph is a field fullGraphj

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -209,13 +247,25 @@ private void setAttributeForAllEdges(GraphStyleAttribute attribute, EdgeAtt prop
...
 
+  private void updateGraph() {
+    if (getMouseClickedNode() == null) {
+      Alert alert = new Alert(AlertType.INFORMATION);
+      alert.setContentText("Please click on any node First!!");
+      alert.showAndWait();
+    }
+    else
+    {
+      generator.createGraphWithNeighboringNodes(graph, GraphStreamUtils.getNodeNeighbors(getMouseClickedNode(),bNeighbors.get()));
```
robinschmid, line 258: here you need to call something like graph.setNodeFilter(n -> //check if node is in neighbor set)
this method should be inside FilteredGraph to set a filter onto the nodes

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -586,4 +601,14 @@ public void setFeatureList(FeatureList featureList) {
   public void setUseMs1FeatureShapeEdges(boolean ms1FeatureShapeEdges) {
     this.ms1FeatureShapeEdges = ms1FeatureShapeEdges;
   }
+ public void setFilteredGraph()
+ {
+   FilteredGraphUtils fgu = new FilteredGraphUtils();
```
robinschmid, line 606: Utils usually contain static methods but no state - so no non static fields.
so you never do new ...

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -586,4 +601,14 @@ public void setFeatureList(FeatureList featureList) {
...
+   fgu.addContentsOfFullGraph(fullGraph,graph);
```
robinschmid, line 607: just do something like graph (which is a FilteredGraph) graph.setFullGraph(fullGraph)

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
```diff
@@ -236,14 +236,15 @@ public NetworkPane(String title, String styleSheet2, boolean showTitle) {
 
     selectedNodes = new ArrayList<>();
 
-    graph = new MultiGraph(title);
+    FilteredGraph fg = new FilteredGraph("Filtered-Graph");
+    graph = fg.getFilteredGraph();
```
robinschmid, line 240: dont understand what you do here.
You create a FilteredGraph and extract a filteredgraph from it... that does not make sense.
FilteredGraph is a graph

src/main/java/io/github/mzmine/modules/visualization/networking/visual/NetworkPane.java
```diff
@@ -285,27 +286,16 @@ public void mouseLeft(String id) {
...
-      if (e.getButton() == MouseButton.PRIMARY) {
-        if (e.getClickCount() == 2) {
-          resetZoom();
-          e.consume();
-        } else if (e.getClickCount() == 1) {
-          setCenter(e.getX(), e.getY());
-        }
-      } else if (e.getButton() == MouseButton.SECONDARY) {
-        openSaveDialog();
-        e.consume();
```
robinschmid, line 297: why did you remove all of this. this is supposed to give mouse control

src/main/java/io/github/mzmine/util/FilteredGraphUtils.java
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Copies the edges and nodes of the original_graph into the filtered_graph
+   *
+   * @param originalGraph the original Graph
+   * @param filteredGraph the filtered Graph
+   */
+
+  public void addContentsOfFullGraph(Graph originalGraph, Graph filteredGraph) {
+    originalGraph.nodes().forEach(aNode -> {
```
robinschmid, line 17: add a filteredGraph clear before that to be sure that it was empty

src/main/java/io/github/mzmine/util/FilteredGraphUtils.java
```diff
@@ -0,0 +1,33 @@
...
+public class FilteredGraphUtils {
+
+  /**
+   * Copies the edges and nodes of the original_graph into the filtered_graph
+   *
+   * @param originalGraph the original Graph
+   * @param filteredGraph the filtered Graph
+   */
+
+  public void addContentsOfFullGraph(Graph originalGraph, Graph filteredGraph) {
```
robinschmid, line 16: this can be logic inside the filteredgraph

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -593,22 +587,31 @@ public void setFeatureList(FeatureList featureList) {
...
+      });
+    });
+    fullGraph.edges().forEach(anEdge -> {
+      Edge e;
+      e = filteredGraph.addEdge(anEdge.getId(), anEdge.getSourceNode().getId(),
+          anEdge.getTargetNode().getId(), anEdge.isDirected());
+      anEdge.attributeKeys().forEach(attribute -> {
+        e.setAttribute(attribute, anEdge.getAttribute(attribute));
+      });
+    });
```
robinschmid, line 615: I tell you to put this part of code into the FilteredGraph class and you push it from the FilteredGraphUtils to the FeatureNetworkPane which is obviously the wrong location because its supposed to be a specialized Pane. Whereas this code performs something that can be done with any Graph

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -327,6 +355,7 @@ public void createNewGraph(FeatureListRow[] rows) {
 
     // apply dynamic style
     applyDynamicStyles();
+    graph.setInitialGraph(graph,graph.getFullGraph());
```
robinschmid, line 358: this does not make sense. passing graph and a field of graph to graph

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FilteredGraph.java
```diff
@@ -0,0 +1,31 @@
...
+public class FilteredGraph extends MultiGraph {
+
+  private final MultiGraph fullGraph;
+
+  public FilteredGraph(String id) {
+    super(id);
+    this.fullGraph = new MultiGraph("Full-Graph");
+  }
+
+  public void setInitialGraph(MultiGraph graph1,MultiGraph graph2) {
```
robinschmid, line 16: this does not make any sense. 
First there should be a way to set the fullGraph in the constructor or via a setter.
Then add all the methods for filtering to this class.
The update methode should use a filter and use the fullGraph, filter it and add all nodes and eged to MutiGraph fg = this;

src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableContextMenu.java
```diff
@@ -248,6 +250,9 @@ private void initSearchMenu() {
   }
 
   private void initShowMenu() {
+
+    final MenuItem showInNetworkVisualizerItem = new MenuItem("Show in Network Visualizer");
```
robinschmid, line 254: Please conform to the other menu item titles.
just use "Feature network"

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkGenerator.java
```diff
@@ -111,6 +113,23 @@ public void createNewGraph(FeatureListRow[] rows, Graph graph, boolean onlyBestN
     }
   }
 
+  /**
+   * only show neighboring nodes
+   *
+   * @param graph
+   * @param neighboringNodes a set for fast contains query
+   */
+  public void createGraphWithNeighboringNodes(Graph graph, Set<Node> neighboringNodes) {
```
robinschmid, line 122: this is still in the wrong class. This should happen in the filtered graph

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -50,6 +57,8 @@ public class FeatureNetworkPane extends NetworkPane {
   public static final float MIN_NODE_WIDTH_GU = 0.02f;
   private static final Logger logger = Logger.getLogger(FeatureNetworkPane.class.getName());
 
+  private final FilteredGraph fg = new FilteredGraph("Filtered-Graph");
```
robinschmid, line 60: Please read our conversations again. We are cycling through the same mistakes.
NetworkPane has a FilteredGraph graph that should be protected or accessible by getter

Why should there be two graphs in FeatureNetworkPane?

sanbit876: I have created object of the FilteredGraph Class here for calling the setFullGraph function (inside the createnewgraph function) of the filteredgraph class in order to set the intial state of the graph (with all edges and nodes) to the full graph

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkGenerator.java
sanbit876: okay... i will move it into the filteredgraph class

src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableContextMenu.java
sanbit876: okay

robinschmid (changes requested):

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -209,6 +239,17 @@ private void setAttributeForAllEdges(GraphStyleAttribute attribute, EdgeAtt prop
     }
   }
 
+  private void updateGraph() {
+    if (getMouseClickedNode() == null) {
+      Alert alert = new Alert(AlertType.INFORMATION);
+      alert.setContentText("Please click on any node First!!");
+      alert.showAndWait();
+    } else {
+      generator.createGraphWithNeighboringNodes(graph,
```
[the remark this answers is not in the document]
sanbit876: okay

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkPane.java
```diff
@@ -318,6 +344,7 @@ public void createNewGraph(FeatureListRow[] rows) {
 
     clear();
     generator.createNewGraph(rows, graph, onlyBest, relationMaps, ms1FeatureShapeEdges);
+    fg.setFullGraph(graph); //setting the full-graph generated to the filtered-graph
```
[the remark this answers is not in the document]
sanbit876: I have update the code:
graph.setFullGraph(graph)

src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableContextMenu.java
robinschmid: you seem to not read my comments.
Please change it to "Feature network" to the letter.  All other items are also called this way without Show in because the menu is named Show

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FilteredGraph.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.graphstream.graph.Element;
+import org.graphstream.graph.Node;
+import org.graphstream.graph.implementations.MultiGraph;
+
+public class FilteredGraph extends MultiGraph {
+
+  MultiGraph fullGraph;
```
robinschmid, line 12: always private if possible. Limit the visibility as much as possible and only write getter and setters where needed

sanbit876: okay....i will make it private

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkGenerator.java
robinschmid: why is it still here

src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableContextMenu.java
sanbit876: okay i will change it..... i thought u were telling to change only the network-visualizer part

src/main/java/io/github/mzmine/modules/visualization/networking/visual/FilteredGraph.java
```diff
@@ -0,0 +1,36 @@
...
+  public FilteredGraph(String id) {
+    super(id);
+    this.fullGraph= new MultiGraph("Full-Graph");
+  }
+
+  public void setFullGraph(MultiGraph graph) {
+    this.fullGraph = graph;
+  }
+
+  public void setNodeFilter(Set<Node> neighboringNodes) {
```
robinschmid, line 23: This is okay for the start but is not what I asked for. 
You should be able to write a real filter something like
setNodeFilter(Predicate<Node> filter)

PR #827, opened 2022-08-02
Discussion: Best Approach to Get Most Intense MSn Scan of FeatureList Row

Kevin-Murray:
I would like to discuss what other developers think is the best way to get the most intense MSn scan associated with a feature list row entry. I am not necessarily aiming to have this code merged unless we decide its an appropriate approach. 

### Background
Our lab utilizes a constant neutral loss MS3 approach in our research (LC-MS/MS-CNL-MS3). In short, if the acquisition software observes a diagnostic neutral loss in the MS2 it triggers a singular MS3 on the associated fragment ion. We would like to use these spectra with various annotative tools, such as, spectral library searching, formula assignment, etc. My aim was to create a function that would return the "most intense" MSn scan associated with feature list row for data generated in a singular DDA-MSn fashion, ie. assuming only one data dependent MSn scan per precursor. I don't think this function would be particularly useful for a fragmentation tree acquisition design. 

### Approach
The Util functions return the "most intense" MSn scan of a feature list row object. MSn scans are found through searching the PrecursorIonTree of the raw data file associated with the most intense fragment scan of the row. If the precursor mass is found in the tree and the node possess an appropriate max MS depth, the most intense MSn scan (determined by TIC) in the feature scan region is returned. Currently the search region is limited by the scan number range of the row feature but this could be easily changed to retention time. For iterative use, if you create a HashMap object to store the PrecursorIonTrees for each raw datafile of an aligned feature list, it saves quite a bit of processing time. 

This approach works with my test data. I've tried this with data generated on the Thermo Fusion and Lumos with MS3, MS4, MS5, and MS6 data. I have not tested this with other vendor data types. Additionally, this approach does not have any ion mobility components yet (I only have FAIMS ion mobility data to test). 

### Discussion
I'm curious other developers thoughts are on this approach? I feel like its a bit clunky and inefficient but there are not many other ways to link MSn scans with feature list rows currently implemented.

### DIFF
--- a/src/main/java/io/github/mzmine/util/scans/ScanUtils.java
+++ b/src/main/java/io/github/mzmine/util/scans/ScanUtils.java
@@ -1635,33 +1635,36 @@ public class ScanUtils {
   /**
    * Return the closest MSn scan to the most intense MS2 fragment scan.
    *
-   * I would recommend using PrecursorIonTree caller for iterative uses to avoid repeated
-   * MSn Fragment Tree building.
+   * Use PrecursorIonTree caller for iterative uses to avoid repeated MSn Fragment Tree building.
    *
    * @param row Feature list row
    * @param msLevel MSn level
    * @return MSn scan
    */
-  public static Scan getClosestMSnScan(FeatureListRow row, int msLevel){
+  public static Scan getMostIntenseMSnScan(FeatureListRow row, int msLevel){
 
     Scan ms2Scan = row.getMostIntenseFragmentScan();
     List<PrecursorIonTree> tree = getMSnFragmentTrees(ms2Scan.getDataFile());
 
-    return getClosestMSnScan(tree, msLevel, ms2Scan);
+    return getMostIntenseMSnScan(row, msLevel, tree);
   }
 
   /**
-   * Return the closest MSn scan to the most intense MS2 fragment scan.
+   * Get most intense MSn scan for feature list row
    *
-   * @param msNFragmentTrees PrecursorIonTree of raw data file
-   * @param msLevel MSn level
-   * @param ms2Scan Most intense fragment scan
-   * @return MSn scan
+   * @param row Feature List row
+   * @param msLevel MSn scan level
+   * @param msNFragmentTrees PrecursorIonTree for raw data file of most intense fragment scan.
+   * @return
    */
-  public static Scan getClosestMSnScan(List<PrecursorIonTree> msNFragmentTrees, int msLevel,
-      Scan ms2Scan) {
+  public static Scan getMostIntenseMSnScan(FeatureListRow row, int msLevel,
+      List<PrecursorIonTree> msNFragmentTrees) {
 
-    int ms2ScanNum = ms2Scan.getScanNumber();
+    Scan ms2Scan = row.getMostIntenseFragmentScan();
+
+    List<Scan> scanRange = row.getFeature(ms2Scan.getDataFile()).getScanNumbers();
+    int firstScanNum = scanRange.get(0).getScanNumber();
+    int lastScanNum = scanRange.get(scanRange.size() - 1).getScanNumber();
 
     // Loop through each tree
     for (PrecursorIonTree tree : msNFragmentTrees) {
@@ -1671,31 +1674,32 @@ public class ScanUtils {
       // check if tree has MSLevel and root m/z matches MS2 Scan
       if (root.getMaxMSLevel() >= msLevel && root.matches(ms2Scan.getPrecursorMz())) {
 
-        List<PrecursorIonTreeNode> child = root.getChildPrecursors();
-
         // Get scans associated with MS Level
         List<Scan> fragmentScans = root.getFragmentScans(msLevel - 2);
 
         if (fragmentScans != null) {
 
-          Scan closestScan = null;
-          int closest = ms2ScanNum;
+          Scan msnScan = null;
+          double intensity = 0;
 
-          // Get scan closest (by scan number) to origin scan
           for(Scan msn : fragmentScans){
 
-            int dist = msn.getScanNumber() - ms2ScanNum;
+            Double value = msn.getTIC();
 
-            if(dist > 0 && dist < closest){
-              closest = dist;
-              closestScan = msn;
+            // Get most intense MSn scan in feature scan range
+            if(msn.getScanNumber() > firstScanNum &&
+                msn.getScanNumber() < lastScanNum &&
+                value > intensity){
+              msnScan = msn;
+              intensity = value;
             }
           }
-
-          return closestScan;
+          return msnScan;
         }
       }
     }
+
+    // No MSn scans found
     return null;
   }
 }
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/mzmine/util/scans/ScanUtils.java
```diff
@@ -1630,4 +1630,76 @@ public String toString() {
...
+
+  /**
+   * Get most intense MSn scan for feature list row
+   *
+   * @param row Feature List row
+   * @param msLevel MSn scan level
+   * @param msNFragmentTrees PrecursorIonTree for raw data file of most intense fragment scan.
+   * @return
+   */
+  public static Scan getMostIntenseMSnScan(FeatureListRow row, int msLevel,
```
SteffenHeu, line 1660: I think this method should run on features instead of rows. You can still add a method that runs on rows, but it will be more universal if it works for features.

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
```diff
@@ -1630,4 +1630,76 @@ public String toString() {
...
+    // Loop through each tree
+    for (PrecursorIonTree tree : msNFragmentTrees) {
+
+      PrecursorIonTreeNode root = tree.getRoot();
+
+      // check if tree has MSLevel and root m/z matches MS2 Scan
+      if (root.getMaxMSLevel() >= msLevel && root.matches(ms2Scan.getPrecursorMz())) {
+
+        // Get scans associated with MS Level
+        List<Scan> fragmentScans = root.getFragmentScans(msLevel - 2);
```
SteffenHeu, line 1678: is it always -2 here?

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
```diff
@@ -1630,4 +1630,76 @@ public String toString() {
...
+
+          Scan msnScan = null;
+          double intensity = 0;
+
+          for(Scan msn : fragmentScans){
+
+            Double value = msn.getTIC();
+
+            // Get most intense MSn scan in feature scan range
+            if(msn.getScanNumber() > firstScanNum &&
```
SteffenHeu, line 1690: the cleaner version would be to use feature.getRtRange() and then just use rtRange.contains

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
```diff
@@ -1630,4 +1630,76 @@ public String toString() {
...
+   * Use PrecursorIonTree caller for iterative uses to avoid repeated MSn Fragment Tree building.
+   *
+   * @param row Feature list row
+   * @param msLevel MSn level
+   * @return MSn scan
+   */
+  public static Scan getMostIntenseMSnScan(FeatureListRow row, int msLevel){
+
+    Scan ms2Scan = row.getMostIntenseFragmentScan();
+    List<PrecursorIonTree> tree = getMSnFragmentTrees(ms2Scan.getDataFile());
```
SteffenHeu, line 1647: not querying all trees here but only the ones in a specific rt range of the feature would also save processing time.

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
Kevin-Murray: That was a fairly easy change. I suspect these methods will change considerably in the future. I've been reviewing MSn fragment trees lately and I think these methods will be extend some `getFeatureFragmentTree` method in the future. I will finish on these comments and look into building such a function.

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
Kevin-Murray: Yes, to my understanding it will always be -2 in the current implementation. PrecursorIonTreeNode `getFragmentScans` argument is levelFromRoot. levelFromRoot == 0 represents the MS2 level of the tree. As such if you wanted to get the MS3 of the PrecursorIonTreeNode, you would need to go 1 level from the root. 

@robinschmid can you confirm this?

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
Kevin-Murray: I knew I would need to do that eventually... Thank you for the reminder. 

I made the changes using your suggestion. One question - I could not find a function named `feature.getRtRange()`, however I did find `feature.getRawDataPointsRTRange()`. Are these functions the same?

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
Kevin-Murray: Yes, I agree this would be the better approach. I think if we made a method overload of `getMSnFragmentTrees` that utilized a Feature rather than RawDataFile we could build just a Feature-specific fragment tree. Alternatively, we could build a method that extracts the subtree(s) associated with a feature from the master PrecursorIonTree of the entirety of the raw data file. Maybe there is value to both methods...?

@SteffenHeu  @robinschmid What are your thoughts on this topic? I know there are multiple parties interested in MSn functionality, should we have a larger discussion about the end goals for the fragmentation trees in MZmine3?

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
SteffenHeu: yes that's what i meant

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
SteffenHeu: I can only mark @robinschmid again, i don't do much with MSn

robinschmid:
Sorry for the delay.
This is actually really important. So far we only have one list of fragmentation scans per feature and then per row.
I think we should find a good way to achieve multiple things here:

1. Have a list of all MSn scans per feature/row
2. Have a list of all MSn trees
3. Have merged MSn spectra: for each sample, row, and ms level precursor.

We have two options.
a. We could use the PrecursorIonTree the the FragmentScanList and add merged spectra to each precursor ion. 
b. just keep all MSn spectra in a simple list for each feature and build trees only on demand. Thats super fast as we only convert the structure.

Kevin-Murray:
@robinschmid No worries on delay - busy time of year. I am already delayed in responding to you.

My initial thought is Option B would be the easier approach. I could try putting together a ScanUtils function that would add all MSn spectra to each feature, similar to the `streamAllMS2FragmentScans` method currently utilized in some feature building methods. A 'getAllMSnScans' and 'getMostIntenseMSnScan' could easily retrieve those scans by feature/row to build a PrecursorIonTree or directly use the spectra in a different operation, eg. spectral library searching, etc. 

I think a big question is: how will researchers be using MSn spectra? It might be worthwhile outlining the most obvious use cases for MSn spectra and ensuring our implementation is versatile enough to accommodate the immediate applications.

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
robinschmid: Yeah we definitely need some functions to create trees for features and from feature list rows. How about just passing a row / feature or the list of fragmentation scans from them?

robinschmid:
Thats a good question. Lets wait for our meeting in a few. But I am currently putting together some ways to merge spectra on different levels and energies

SteffenHeu:
@robinschmid / @Kevin-Murray did #973 introduce the functionality this thread was about?

robinschmid:
Yes I would say this PR mostly delivered on this.
Selection of MSn spectra for the total number of spectra and then the merged spectra from this

robinschmid:
We included MSn datastructures and scan selection and merging

PR #834, opened 2022-08-05
ims expander update

SteffenHeu:
- restructured ims expander
- before, all features were sorted by mz and frames were split into threads to expand all traces at once
  - every data point was kept in ram until all threads were finished

- now the traces are sorted by rt range and only frames in the selected rt ranges must be queried per trace set
  - one package is processed per thread, making it possible to write finished, expanded traces to the disc
- the number of traces per thread can also be limited, making it possible to process large imaging datasets without too much ram consumption


src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderModule.java
```diff
@@ -37,7 +37,7 @@ public class ImsExpanderModule implements MZmineProcessingModule {
 
   @Override
   public @NotNull String getName() {
-    return "Ims expander";
+    return "Ims expander2";
```
robinschmid, line 40: needed?

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderParameters.java
```diff
@@ -63,9 +63,16 @@ public class ImsExpanderParameters extends SimpleParameterSet {
   public static final OriginalFeatureListHandlingParameter handleOriginal = //
       new OriginalFeatureListHandlingParameter(false);
 
+  public static final OptionalParameter<IntegerParameter> maxNumTraces = new OptionalParameter<>(
+      new IntegerParameter("Maximum traces per thread", """
+          Sets the maximum number of traces to be processed per thread.
+          For LC-IMS-MS measurements, this is typically not required.
+          However, it can be beneficial for imaging experiments to reduce the memory consumption during this step.""",
+          2_000), false);
```
robinschmid, line 71: should the user define the traces or rather the number of threads to use and then divide the traces on the threads? 
This could reduce this parameter to the NumOfThreadsParameter which can be set to auto

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderSubTask.java
```diff
@@ -137,14 +160,54 @@ public void run() {
           }
         }
         processedFrames.getAndIncrement();
+
+        desc = flist.getName() + ": expanding traces for frame " + processedFrames.get() + "/"
+            + totalFrames + " m/z range: " + RangeUtils.formatRange(traceMzRange,
+            MZmineCore.getConfiguration().getMZFormat());
```
robinschmid, line 166: when in a loop we should convert this to a local var outside the loop.

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderSubTask.java
```diff
@@ -137,14 +160,54 @@ public void run() {
...
+        return;
+      }
+    }
+
+    // allow traces to be released
+    expandingTraces = null;
+
+    synchronized (newFlist) {
+      newRows.forEach(newFlist::addRow);
+    }
```
robinschmid, line 209: Maybe instead of adding all rows here in a synchronized block  - can we add the rows from the main task? So we just keep a list of results in this task and once we set to finished the main task adds the rows  - so its single threaded.
I think we do not need to make the whole feature list thread safe on add / remove rows right?

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderParameters.java
robinschmid: Maximum features per thread

PR #835, opened 2022-08-05
Small fixes in mass calibration 

omokshyna:
Previously mass calibration dialog was not launching with no loaded raw files. Now, the dialog launches and gives a warning.
Additionally, slight changes were added to the gui removing an extra column in one of the panes.


SteffenHeu (approved):
Thanks!

PR #839, opened 2022-08-10
fix slow file type detection in all data import

robinschmid:
By default we were using in input stream two times to determine the correct mzML import: IMS or no IMS
Now we are handling IMS and mzML files the same way with the advanced MSDK import.

This actually took a long time sometimes for ~500 files.

Also, I am checking if the file name ends with .mzml case insensitive.

### DIFF
--- a/src/main/java/io/github/mzmine/util/RawDataFileTypeDetector.java
+++ b/src/main/java/io/github/mzmine/util/RawDataFileTypeDetector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2021 The MZmine Development Team
+ * Copyright 2006-2022 The MZmine Development Team
  *
  * This file is part of MZmine.
  *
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/mzmine/util/RawDataFileTypeDetector.java
```diff
@@ -99,31 +94,38 @@ public static RawDataFileType detectDataFileType(File fileName) {
...
+              return RawDataFileType.AIRD;
+            }
           }
+          logger.info("It's not an aird format file or the aird index file not exist");
+        }
+        if (fileName.getName().contains(TDF_SUFFIX) || fileName.getName()
+            .contains(TDF_BIN_SUFFIX)) {
+          return RawDataFileType.BRUKER_TDF;
+        }
+        if (fileName.getName().contains(TDF_SUFFIX) || fileName.getName()
```
SteffenHeu, line 120: TSF

src/main/java/io/github/mzmine/util/RawDataFileTypeDetector.java
```diff
@@ -160,7 +162,7 @@ public static RawDataFileType detectDataFileType(File fileName) {
         }
 
         if (fileHeader.contains(MZML_HEADER)) {
-          return getMzmlType(fileName);
+          return RawDataFileType.MZML;
```
SteffenHeu, line 165: I think im missing it, but where is the discrimination between ims and non ims?

src/main/java/io/github/mzmine/util/RawDataFileTypeDetector.java
robinschmid: fixed. must have been like this for a while?

src/main/java/io/github/mzmine/util/RawDataFileTypeDetector.java
robinschmid: no discrimination needed at this point as we are doing this already in the mzml import task

robinschmid:
ready from my side

src/main/java/io/github/mzmine/util/RawDataFileTypeDetector.java
SteffenHeu: yeah

PR #844, opened 2022-08-12
Speedup advanced mzML import 6 times 

robinschmid:
Data access on MzMLScan and the MSDKScanWrapper was very inefficient as used in the advanced import where mass detection is applied directly, to safe memory and need for memory mapping.

for 250 mzML files it went down from 12 min --> 2 min


robinschmid:
## Before
image


image


## After
image
image

robinschmid:
disk is finally showing some activity

PR #847, opened 2022-08-14
Add warning log and icon for incorrectly centroided files 

robinschmid:
MSconvert needs the peak picking to be the first step to apply the vendor algorithm.

image


robinschmid:
Reviewed by @tomas-pluskal

PR #848, opened 2022-08-15
Update .mailmap

tomas-pluskal:
I updated the mailmap according to documentation (https://git-scm.com/docs/gitmailmap). The name of the person doesn't need to be specified in the 3rd column, it will match any name with the given email address.


.mailmap
```diff
@@ -1,55 +1,58 @@
+Robin Schmid <rschmid1789@gmail.com> <rschmid1789@gmail.com>
```
robinschmid, line 1: Is this why you can leave out the name  - because email address might be the same but name is different

tomas-pluskal: When you leave out the name, it means "any commit with this EMAIL address should be updated with the correct NAME and EMAIL address". So it's like a wildcard for names.

PR #851, opened 2022-08-18
Patch out of bound in IsotopeFinder

robinschmid:
Empty mobility scans were throwing out of bounds exceptions


SteffenHeu:
Looks good to me. But how did this happen? It should be at the apex of the peak.

PR #852, opened 2022-08-18
Simple statistics on RawDataFIle, scans, data points

robinschmid:
Need this to get more insight into the number of files, scans, datapoints


SteffenHeu:
I think this is a bit misleading, becaus the number of data points will be very high for profile scans. 

Also it might be interesting to also display the total number of points to check if the file was centroided correctly

robinschmid:
We can change some stuff

robinschmid:
Anything interesting to know about raw data files? Spectra? ...

PR #854, opened 2022-08-19
Feature list methods help fix

omokshyna:
Removed offline help from several more methods in Feature list methods, namely 

- MS2 scan pairing,
- Isotope pattern finder,
- Isotope peak scanner,
- Both Spectral deconvolution (GC) modules


src/main/java/io/github/mzmine/modules/dataprocessing/filter_groupms2/GroupMS2Parameters.java
```diff
@@ -56,10 +56,7 @@ public class GroupMS2Parameters extends SimpleParameterSet {
 
   public static final BooleanParameter lockMS2ToFeatureMobilityRange = new BooleanParameter(
       "Lock to feature mobility range",
-      "If checked, only mobility scans from the mobility range of the feature will be merged.\n"
-          + "This is usually not needed. However, if isomeres/isobares elute at the same retention time and are close in mobility, "
-          + "the MS/MS window might be larger than the peak in mobility dimension and thus cause chimeric MS/MS spectra.\n"
-          + "This can be investigated in hte \"All MS MS\" window", false);
+      "If checked, only mobility scans from the mobility range of the feature will be merged.", false);
```
SteffenHeu, line 59: why did you remove the additional explaination?

omokshyna: Because it is available in the documentation and might not be very useful as a pop-up hint. However, if you think it's necessary there, I can put it back

SteffenHeu: I think this is good to have in there, so we can explain the average user what the parameter can be used for.

omokshyna: Sure :) I've added it back

PR #863, opened 2022-08-25
update default ms2 grouping selection (true = default)

SteffenHeu:
Change the selection of the group ms2 parameter in the resolver to true, so MS2s are grouped by default.
Also select the limit by rt edges by default.


robinschmid (approved):
Better

PR #867, opened 2022-08-26
Update datatype loadsafe

SteffenHeu:
For some stuff i want to do, it is possible that a feature is grouped with MS2s from different raw files. Therefore, I adapted the load/save to accomodate this. This change is not breaking and existing projects still work. If no raw file information is given for MS2 scans or MsMsInfos, we still default to the file of the MS1 feature.


robinschmid (approved):

robinschmid (approved):
Looks good to me

PR #865, opened 2022-08-25
Fix cell factory

SteffenHeu:
fix the cell factory from going crazy in the feature table. Also update java fx to 17

related to https://bugs.openjdk.org/browse/JDK-8280442

PR #868, opened 2022-08-26
Java 18

SteffenHeu:
includes #865


SteffenHeu:
When scrolling through the feature table it jumps to the bottom at some point, making it super annoying. Could not find a fix for that yet.

robinschmid:
I have the same issue and its annoying. Even trying to scroll up it jumps down sometimes. More often for the aligned list.
Maybe its a bug of the new JFX. I was using JDK18 for a while

tomas-pluskal:
Have you guys tried using JavaFX 19 with Java 18? Does it still cause this problem?

robinschmid:
Maybe we should try this

Issue #873, opened 2022-08-29
Can not open the ion mobility data file in MZmine 3


luluwang0702:
### Basic information

* My operating system: Bruker trapped ion mobility (TIMS)-TOF Flex
* My MZmine version: MZmine 3,2,3

### What happened

* Please describe what happened and how it differs from what you expected to happen. I would like to process the ion mobility data in MZmine to generate the feature list for the following GNPS analysis. But the data file could not be imported into MZmine.

* Please include all parameter values for the MZmine modules you used. 
* Please upload screenshots of any errors and your log/mzmine.log file.
![Picture1](https://user-images.githubusercontent.com/62401677/187155424-444dfcd9-29ae-4426-8d95-6e8b3ce1f694.png)

* Please upload an MZmine project file that demonstrates the problem you are [reporting.]
[analysis.zip](https://github.com/mzmine/mzmine3/files/9442905/analysis.zip)

SteffenHeu:
Hi,

I've never seen this occur before. What kind of data are you importing? Which acquisition mode did you use? Could you upload an example file and the full log file please so we can reproduce this?

Best 
Steffen

luluwang0702:
Hi Steffen,

Here is the link for the data file. I tried to import the .d folder and the BDal DataAnalysis Application class inside the folder. The data is in the data dependent mode.
https://drive.google.com/file/d/1agNDlEfK0awuFoV0bNsPc84ZvJ8ivVha/view?usp=sharing

Thanks in advance.

Best regards,
Xinhui

SteffenHeu:
Hey, 

I looked into the file. I know what the issue is and it can be fixed, but It will take some time.

The issue is that no data is stored in the first acquisition segment, which I haven't seen before. How did you acquire the dataset? What were the acquisition parameters in the first segment in order to not not store any ms data?

PR #886, opened 2022-09-07
Fix no data in tdf

SteffenHeu:
should fix issues with tdf files that don't have any data in the first segment as shown in #873

We add a single mobility scan to these frames, so we don't introduce issues with later processing modules.


src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_tdf/TDFImportTask.java
```diff
@@ -433,6 +433,9 @@ private void appendScansFromTimsSegment(@Nonnull final TDFUtils tdfUtils,
       final ParameterSet param = msLevel == 1 ? ms1DetectorParam : ms2DetectorParam;
       final List<BuildingMobilityScan> spectra = tdfUtils.loadSpectraForTIMSFrame(
           frame.getFrameId(), frameTable, detector, param);
+      if (spectra.isEmpty()) {
+        spectra.add(new BuildingMobilityScan(0, new double[]{0}, new double[]{0}));
```
SteffenHeu, line 437: check with empty array

src/main/java/io/github/mzmine/datamodel/impl/SimpleFrame.java
```diff
@@ -46,6 +47,8 @@
  */
 public class SimpleFrame extends SimpleScan implements Frame {
 
+  private static Logger logger = Logger.getLogger(SimpleFrame.class.getName());
```
robinschmid, line 50: you can change your live template to private static **final**

robinschmid (approved):

robinschmid (approved):
Great thanks for fixing

Issue #859, opened 2022-08-22
RANSAC Aligner preview crashes


omokshyna:
### Basic information

* My operating system: Windows 10 Pro
* My MZmine version: 3.2.4

### What happened

When trying to launch preview in RANSAC aligner, the module first freezes and then crashes.

SteffenHeu:
i think its because the dialog runs the preview on the java FX (=the gui) thread

omokshyna:
@SteffenHeu I checked and indeed preview seems to be running in thread called "JavaFX Application Thread". Should it be otherwise? At the moment, I introduced some fixes and the preview is displayed (but horrifyingly slow). Please, take a look if you have some time before I create a PR - https://github.com/omokshyna/mzmine3/tree/ransac-align-preview-fix

SteffenHeu:
Hi olena, I don't have to time to go through the branch before next week, but thanks for looking into it. 
It's favourable of the data for the preview was calculated in a separate thread which then updates the GUI when it's done. Basically In the same way we use the processing tasks in Mzmine.

omokshyna:
Thanks, next week would be also great. Meanwhile, I'll try to solve the problem with the threads.

PR #900, opened 2022-09-12
Ransac aligner preview fix

omokshyna:
So, the RANSAC aligner preview is now working and has been moved into a separate task. However, some issues still remain - such as speed of update, which might require optimizing the underlying calculations. 
Addresses #859


src/main/java/io/github/mzmine/modules/dataprocessing/align_ransac/RansacAlignerSetupDialog.java
```diff
@@ -18,187 +18,143 @@
...
+  @Override
+  protected void showPreview(boolean show) {
+    super.showPreview(show);
+    if(show) {
+      updatePreview();
+    }
   }
 
+  //using this method makes module too slow
+//  protected void parametersChanged() {
```
SteffenHeu, line 148: I think this should be used when some parameters are changed. otherwise it will only update once a feature list is changed, or? 
You can check the IsotopePatternPreviewDialog or so, it has a FutureTask withg a CountdownLatch or so that will only run a new task if there wasnt a new entry for 500 ms or so.

src/main/java/io/github/mzmine/modules/dataprocessing/align_ransac/RansacPreviewTask.java
```diff
@@ -0,0 +1,135 @@
...
+    this.parameters = parameterSet;
+
+  }
+
+  @Override
+  public String getTaskDescription() {
+    return message;
+  }
+
+  //TODO Write a working percentage calculation
```
SteffenHeu, line 47: this could the same method as the main task for the aligner, right?

omokshyna: Thanks for the idea, with great help from @robinschmid it was implemented alongside some minor performance optimization

src/main/java/io/github/mzmine/modules/dataprocessing/align_ransac/RansacPreviewTask.java
```diff
@@ -0,0 +1,131 @@
...
+    setStatus(TaskStatus.FINISHED);
+  }
+
+  //TODO This method at the moment is quite slow. Optimize it (or underlying methods) if possible
+  private void updateRansacPlot() {
+//    this.plot.removeSeries();
+
+    // Select the rawDataFile which has more peaks in each feature list
+    RawDataFile file = featureListX.getRawDataFiles().stream().max(Comparator.comparingInt(raw -> featureListX.getFeatures(raw).size())).get();
+    RawDataFile file2 = featureListY.getRawDataFiles().stream().max(Comparator.comparingInt(raw -> featureListY.getFeatures(raw).size())).get();
```
robinschmid, line 79: apply the code style. You can do that after each save action or before commiting automatically.

robinschmid (approved):
Thanks for fixing it. We can merge it in the current state after the code formatter ran. Any further optimization can be done later

PR #905, opened 2022-09-15
Batch library generation

robinschmid:
Started batch library generation. Based on the local compound DB annotations.

I needed a real MSn tree builder for this. It generates the MSn trees first, then creates all chromatograms on MS1 and links the MSn tree to it.

Reuses the SpectralDBEntry to write json or msp format libraries.

image

## TODO
Add metadata 

- [x] like instrument and instrument type trough parameters: instrument, type, sample type, PI, data collector


src/main/java/io/github/mzmine/modules/dataprocessing/featdet_msn_tree/MsnTreeFeatureDetectionModule.java
```diff
@@ -0,0 +1,80 @@
...
+import java.time.Instant;
+import java.util.Collection;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Search for MSn trees in the whole retention time range and extract chromatograms for them
+ *
+ * @author Robin Schmid <a href="https://github.com/robinschmid">https://github.com/robinschmid</a>
+ */
+public class MsnTreeFeatureDetectionModule implements MZmineProcessingModule {
```
SteffenHeu, line 38: this should be in da separate PR. Merge that one first?

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_msn_tree/SimpleFullChromatogram.java
```diff
@@ -0,0 +1,122 @@
...
+   * @param scans the original full list of scans that was used to create this chromatogram
+   * @return an ion time series with only data points > 0
+   */
+  public IonTimeSeries<? extends Scan> toIonTimeSeries(@Nullable MemoryMapStorage storage,
+      final List<Scan> scans) {
+    DoubleArrayList fmzs = new DoubleArrayList();
+    DoubleArrayList fintensities = new DoubleArrayList();
+    List<Scan> fscans = new ArrayList<>();
+    for (int i = 0; i < scans.size(); i++) {
+      if (intensities[i] > 0) {
```
SteffenHeu, line 70: this leads to ugly chromatograms. zeros should only be excluded if the previous and next value are zero, too.

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/formats/MSPEntryGenerator.java
```diff
@@ -18,86 +18,101 @@
...
 
 public class MSPEntryGenerator {
 
+  public static final List<DBEntryField> EXPORT_FIELDS = List.of(DBEntryField.NAME, DBEntryField.MZ,
+      DBEntryField.ION_TYPE, DBEntryField.EXACT_MASS, DBEntryField.MS_LEVEL, DBEntryField.CHARGE,
+      DBEntryField.FORMULA, DBEntryField.SMILES, DBEntryField.INCHI, DBEntryField.INCHIKEY,
+      DBEntryField.FRAGMENTATION_METHOD, DBEntryField.COLLISION_ENERGY,
+      DBEntryField.ISOLATION_WINDOW, DBEntryField.MSN_FRAGMENTATION_METHODS,
+      DBEntryField.MSN_COLLISION_ENERGIES, DBEntryField.MSN_PRECURSOR_MZS,
+      DBEntryField.MSN_ISOLATION_WINDOWS);
```
SteffenHeu, line 52: is  there no CCS field?

src/main/java/io/github/mzmine/datamodel/features/types/annotations/InChIKeyStructureType.java
```diff
@@ -0,0 +1,62 @@
...
+import io.github.mzmine.datamodel.features.types.modifiers.AnnotationType;
+import io.github.mzmine.datamodel.features.types.modifiers.EditableColumnType;
+import io.github.mzmine.datamodel.features.types.modifiers.StringParser;
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.property.StringProperty;
+import javafx.util.StringConverter;
+import javafx.util.converter.DefaultStringConverter;
+import org.jetbrains.annotations.NotNull;
+
+public class InChIKeyStructureType extends StringType implements EditableColumnType,
```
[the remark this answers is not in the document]
robinschmid: not for inchikey aparently

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/formats/MSPEntryGenerator.java
robinschmid: added rt and CCS. MSP is a weird format

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/ScanSelector.java
```diff
@@ -0,0 +1,34 @@
...
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with MZmine; if not,
+ * write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
+package io.github.mzmine.modules.io.spectraldbsubmit.batch;
+
+public enum ScanSelector {
```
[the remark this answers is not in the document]
robinschmid: added doc. Its for selecting what spectra to export. Based on what criteria.... Could be done differently

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_msn_tree/MsnTreeFeatureDetectionModule.java
robinschmid: Maybe we could merge the whole thing as it is and we add more changes in the future this would make it easier to review for additions

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_msn_tree/SimpleFullChromatogram.java
robinschmid: added a trailing and leading 0

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchTask.java
```diff
@@ -266,8 +300,20 @@ private CompoundDBAnnotation getCompoundFromLine(String[] values,
     final Double neutralMass =
         entry.get(neutralMassType) != null ? Double.parseDouble(entry.get(neutralMassType)) : null;
     final String smiles = entry.get(smilesType);
+    final String inchi = entry.get(inchiType);
+    final String inchiKey = entry.get(inchiKeyType);
     final String pubchemId = entry.get(pubchemIdType);
 
+    final String lineComment;
+    if (!commentFields.isEmpty()) {
```
SteffenHeu, line 308: commentFields is nullable here

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchTask.java
```diff
@@ -152,19 +155,43 @@ public void run() {
...
       List<ImportType> lineIds = findLineIds(importTypes, databaseValues[0]);
 
-//      peakList.addRowType(new CompoundDatabaseMatchesType());
+      // option to read more fields and append to comment as json
+      final DataType<String> type = DataTypes.get(CommentType.class);
+      List<ImportType> commentFields = Arrays.stream(
+              parameters.getValue(LocalCSVDatabaseSearchParameters.commentFields).split(","))
+          .map(s -> s.trim().toLowerCase()).map(s -> new ImportType(true, s, type)).toList();
+
+      commentFields = findLineIds(commentFields, databaseValues[0]);
```
SteffenHeu, line 169: findLineIds is nullable due to your change

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchTask.java
```diff
@@ -296,18 +344,11 @@ private List<ImportType> findLineIds(List<ImportType> importTypes, String[] firs
...
-            setErrorMessage(
-                "Library file " + dataBaseFile.getAbsolutePath() + " contains two columns called \""
-                    + columnName + "\".");
-            setStatus(TaskStatus.ERROR);
-          }
-          importType.setColumnIndex(i);
-        }
+      int columnIndex = getHeaderColumnIndex(firstLine, importType.getCsvColumnName());
+      if (columnIndex == -1) {
+        return null;
```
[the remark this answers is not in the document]
SteffenHeu: Ah it's in the getHeaderColumIndex method. Nevertheless, by always returning -1 as default and exiting here, it is also not clear if a column was not found (indicated by the error below).
This removes a lot of user friendlieness from the module.

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchParameters.java
```diff
@@ -95,7 +110,7 @@ public class LocalCSVDatabaseSearchParameters extends SimpleParameterSet {
   public LocalCSVDatabaseSearchParameters() {
     super(
         new Parameter[]{peakLists, dataBaseFile, fieldSeparator, columns, mzTolerance, rtTolerance,
-            mobTolerance, ccsTolerance, ionLibrary},
+            mobTolerance, ccsTolerance, filterSamples, ionLibrary, commentFields},
```
SteffenHeu, line 113: i'd put filterSamples below the ion library

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationTask.java
```diff
@@ -0,0 +1,276 @@
...
+    return chimericMap;
+  }
+
+  private ChimericPrecursorResult scoreChimericIsolation(final FeatureListRow row,
+      final RawDataFile raw) {
+    Feature feature = row.getFeature(raw);
+    if (feature == null) {
+      return ChimericPrecursorResult.PASSED;
+    }
+    MassList massList = feature.getRepresentativeScan().getMassList();
```
SteffenHeu, line 241: taking the representative scan does not mean it's close to when the MS2 was actually acquired

SteffenHeu: we have ScanUtils.findSucceedingPrecursorScan, findSucceedingPrecursorScanForMerged (for IMS) and findPrecursorScan, findPrecursorScanForMerged (for IMS). 
this also makes the module IMS aware. you can use the default tolerances in the SpectraMerging class for the IMS methods

SteffenHeu:
sorry, even more findings:

the ranges in the msp and json are written like this: [271.2795373328638..272.97661129192215] can this be parsed in the end?
in the json: two fields called "FRAGMENTATION_METHOD"
one says "FRAGMENTATION_METHOD":20.0, the other one says "Fragmenation_method":"CID"
in the json: "IONMODE":"false", i guess the same error as in the msp parsing

where are you currently taking the ccs information from? also from the comp db file? it should be possible to also use the measured one (imo)

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchTask.java
robinschmid: I added NotNull annotations and also throw an exception now if one of the fields is not in the metadata file. 
So if the user asks for inchi but its not there it will directly fail.

robinschmid: the fields are now at least empty list

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchTask.java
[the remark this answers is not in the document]
robinschmid: I extracted the code as a method so the return value was the default I guess.
This time I changed to throwing an exception, which will show the whole call stack and error out the task and log all the missing headers (instead of only the first like it used to be)

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchTask.java
robinschmid: I think it was that before because it was only set to error but did not end the task. Changed that. Good catch

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationParameters.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public static final ComboParameter<ScanSelector> scanExport = new ComboParameter<>("Export scans",
+      "Select scans to export", ScanSelector.values(), ScanSelector.ALL);
+
+  public static final FileNameParameter file = new FileNameParameter("Export file",
+      "Local library file", FileSelectionType.SAVE);
+
+  public static final ComboParameter<SpectralLibraryExportFormats> exportFormat = new ComboParameter<>(
+      "Export format", "format to export", SpectralLibraryExportFormats.values(),
+      SpectralLibraryExportFormats.json);
```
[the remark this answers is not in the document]
robinschmid: I thought about this... but it just makes the whole export more complex. People can just run the export twice... 
And there are more formats.
We should add a convert function - or basically an export of spectral libs to different formats.

robinschmid:
This is the error that pops up when columns were selected in CsvMatch that are not in the table.
The problem is that we are importing the library for each RawDataFile (each task).

How about I change the task 
1. to run 1 task on all data files
2. import the lib
3. parallel stream over all rows in all RawDataFile

image

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchParameters.java
robinschmid: ok done

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/HandleChimericMsMsParameters.java
```diff
@@ -0,0 +1,78 @@
...
+ * How to handle chimeric MS/MS spectra, where the isolation selected multiple precursor
+ *
+ * @author Robin Schmid <a href="https://github.com/robinschmid">https://github.com/robinschmid</a>
+ */
+public class HandleChimericMsMsParameters extends SimpleParameterSet {
+
+  public static final MZToleranceParameter mainMassWindow = new MZToleranceParameter(
+      "Main mass tolerance",
+      "Used to determine the main signal. Usually smaller than the precursor isolation "
+          + "window and dependent on the accuracy and resolution of the instrument.", 0.005, 2);
```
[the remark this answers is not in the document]
robinschmid: I changed the name and description a bit. Hope this makes it clearer. Added that we search in MS1 data for
1. the main precursor signal. This signal triggered the MS2 acquisition and is the main signal
2. the range for contamination. This range is very often different from the isolation range set for the instrument. If you set unit resolution on TOF-MS you typically still isolate the isotope

```java
"Precursor m/z isolation tolerance", """
      Real isolation window to check MS1 scan for co-isolated signals. This might be greater than
      the actually set m/z window. Especially some TOF-MS instruments isolate more than a unit
      resolution, resulting in co-isolated isotopes"""
```

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/HandleChimericMsMsParameters.java
```diff
@@ -0,0 +1,78 @@
...
+  public static final MZToleranceParameter isolationWindow = new MZToleranceParameter(
+      "Precursor isolation tolerance",
+      "Real isolation window to check for co-isolated signals. This might be greater than "
+          + "the actually set value. Especially some TOF-MS instruments isolate more than a unit "
+          + "resolution, resulting in co isolated isotopes.", 0.4, 0);
```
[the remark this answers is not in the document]
robinschmid: Not sure. There are multiple problems with the isolation window imported from the files:
1. many files do not have it (dependent on MS convert version and vendor)
2. even if you set it to a unit resolution - TOF-MS and other instruments still isolate the whole isotope pattern. So this would be something like the actual isolation window.

Would be interesting to actually compute something for each raw data file to get this automatically. Should be easy just look at the isotope pattern in the MS2 for residual precursor m/z for all MS2.

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibaryMetadataParameters.java
```diff
@@ -0,0 +1,105 @@
...
+import io.github.mzmine.util.spectraldb.entry.DBEntryField;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Metadata for batch library submission
+ *
+ * @author Robin Schmid <a href="https://github.com/robinschmid">https://github.com/robinschmid</a>
+ */
+public class LibaryMetadataParameters extends SimpleParameterSet {
```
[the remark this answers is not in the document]
robinschmid: The initial design was simple. I made it for GNPS and it was following strictly the naming of GNPS library submission :D

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibaryMetadataParameters.java
```diff
@@ -0,0 +1,105 @@
...
+
+  public static final ComboParameter<CompoundSource> ACQUISITION = new ComboParameter<>(
+      "ACQUISITION", "", CompoundSource.values(), CompoundSource.Crude);
+  public static final OptionalParameter<ComboParameter<Polarity>> IONMODE = new OptionalParameter<>(
+      new ComboParameter<>("IONMODE",
+          "Exchange all polarities with this value in case its not provided.", Polarity.values(),
+          Polarity.Positive), false);
+  public static final ComboParameter<Instrument> INSTRUMENT = new ComboParameter<>("INSTRUMENT", "",
```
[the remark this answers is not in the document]
robinschmid: Again the GNPS naming... we could change it internally and rework the export. But I guess we should do that after merging this one

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationTask.java
robinschmid: True that will be better. 
I was taking a short cut and only evaluated it once per raw data file... Will change it to check for each individual MS2 scan.

robinschmid: Applied what you said.
MergedMsMsSpectrum will use the merged function all other the getPrecursorScan

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/formats/MSPEntryGenerator.java
```diff
@@ -25,86 +25,101 @@
...
   public static String createMSPEntry(LibrarySubmitIonParameters param, DataPoint[] dps) {
 
-    LibraryMetaDataParameters meta = (LibraryMetaDataParameters) param
-        .getParameter(LibrarySubmitIonParameters.META_PARAM).getValue();
+    LibraryMetaDataParameters meta = (LibraryMetaDataParameters) param.getParameter(
+        LibrarySubmitIonParameters.META_PARAM).getValue();
 
     boolean exportRT = meta.getParameter(LibraryMetaDataParameters.EXPORT_RT).getValue();
     String ionMode =
         meta.getParameter(LibraryMetaDataParameters.IONMODE).getValue().equals(Polarity.Positive)
```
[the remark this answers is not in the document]
robinschmid: So this one is when you run it on a single spectrum. And people are supposed to just supply that.
The batch export uses the polarity from the scan

robinschmid:
For CCS we still need to add this information. I guess we would use the feature information for that right?

robinschmid:
I now renamed everything to MZmine json. 
GNPS json has changed and we also add way more details. Also I want it more harmonized from the names. 

In a separate PR I will create a new GNPS json parser.

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/HandleChimericMsMsParameters.java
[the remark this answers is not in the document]
SteffenHeu: ok
i'd call it precursor isolation window though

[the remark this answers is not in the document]
robinschmid: You think a window is better than mztol parameter?

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/formats/MSPEntryGenerator.java
```diff
@@ -25,86 +25,101 @@
...
-        .getParameter(LibrarySubmitIonParameters.META_PARAM).getValue();
+    LibraryMetaDataParameters meta = (LibraryMetaDataParameters) param.getParameter(
+        LibrarySubmitIonParameters.META_PARAM).getValue();
 
     boolean exportRT = meta.getParameter(LibraryMetaDataParameters.EXPORT_RT).getValue();
     String ionMode =
-        meta.getParameter(LibraryMetaDataParameters.IONMODE).getValue().equals(Polarity.Positive)
-            ? "P"
-            : "N";
+        meta.getParameter(LibraryMetaDataParameters.IONMODE).getValue().equals(Polarity.Negative)
```
SteffenHeu, line 76: you are still checking the OptionalParameter here! getValue for an optional parameter is true or false, comparing it to the polarity does not make sense.

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/formats/MSPEntryGenerator.java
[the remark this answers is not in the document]
SteffenHeu: the check still doesnt make sense though

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/HandleChimericMsMsParameters.java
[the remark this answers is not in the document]
SteffenHeu: yeah, because it is set as a window in most instruments

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/formats/MSPEntryGenerator.java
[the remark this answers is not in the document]
robinschmid: IONMODE is a combo parameter<Polarity>

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/formats/MSPEntryGenerator.java
robinschmid: IONMODE is a combo parameter for <Polarity>

robinschmid: The names of the two parameter sets was confusing similar.
This is for the single entry submission and it contains way more metadata to set. 
For the batch submission, most of the compound specific metadata is in the CompoundDBMatch - so I renamed to the specific class for instrument and analysis metadata to LibraryBatchMetadataParameters

SteffenHeu:
fine with me when the ccs is added either from the database or the measured ccs for the library entry

SteffenHeu:
Thanks!

PR #914, opened 2022-09-23
fix MobilityScanMering always summing

SteffenHeu:


robinschmid (approved):
Very nice. Thanks Steffen

PR #930, opened 2022-10-04
fix concurrent mod exception in ImsExpanderTask

SteffenHeu:


build.gradle
```diff
@@ -248,15 +248,15 @@ javafx {
 
 tasks.withType(JavaCompile) {
     options.encoding = "UTF-8"
-//    options.compilerArgs += ['--enable-preview']
+    options.compilerArgs += ['--enable-preview']
```
tomas-pluskal, line 251: I think this depends on the JDK version used

tomas-pluskal: Ok

PR #932, opened 2022-10-05
Bug fix - getScanNumbers

Kevin-Murray:
Javadoc for RawDataFile method `getScanNumbers(int msLevel)` indicates that msLevel=0 should return all scans, current implementation returns no scans. 

I corrected this method to return all scans if msLevel=0. There are other implementations that could be updated to reflect this functionality. 

Alternatively, if the Javadoc is out-of-date and this functionality is not desired, this pull request can be disregarded and I will update the Javadoc in question.


SteffenHeu:
Hi Kevin, 

the javadoc is outdated indeed. In fact, even the function name is outdated, but we haven't renamed it due to the probably 1000s of merge conflicts.

If you want to access all scans, you should use RawDataFile.getScans() instead. So updating the javadoc is the better solution. Thanks in advance!

Kevin-Murray:
@SteffenHeu Sounds good. I thought it might be outdated but I wanted to check to be sure. 

I updated the javadoc to reflect the intended usage. I will use your suggestion in my work.

SteffenHeu:
thanks!

Issue #925, opened 2022-09-30
GCMS spectral deconvolution


NthlLcrmp:
### Basic information

* My operating system: Windows 10 version 21H1
* My MZmine version: 3.2.8

### What happened

I try to perform spectral deconvolution on GCMS data.

Whether I use Hierarchical Clustering or Multivariate curve resolution algorithm, the preview doesn't work.

With Hierarchical clustering algorithm, the cluster list is empty :
![image](https://user-images.githubusercontent.com/114736047/193242325-a7adbcf6-4753-4390-be28-33cf1d51a84a.png)

With Multivariate curve resolution algorithm, I can choose the cluster but nothing happened :
![image](https://user-images.githubusercontent.com/114736047/193247230-0d297f96-2c80-4daa-a3fc-03ec8646a5ca.png)

I can go through spectral deconvolution with Multivariate curve resolution algorithm but not with Hierarchical Clustering algorithm where the following error appears (which was already reported in issue #744) :
![image](https://user-images.githubusercontent.com/114736047/193243428-bc490d63-9b01-4894-af5f-796e17776239.png)

asmirn1:
@NthlLcrmp Could you upload your .mzmine project file or raw data files, so I could reproduce the error?

NthlLcrmp:
Here is a link to download the project and the raw file :
https://filesender.renater.fr/?s=download&token=918d091b-95ce-44b9-b0ee-620ad1377786

asmirn1:
@NthlLcrmp We have fixed a few issues with the Spectral Deconvolution in the version of MZmine. I have tested it with your project file, and it is working for me now.

Could you try the latest development build of MZmine from here: https://github.com/mzmine/mzmine3/releases/tag/Development-release

Let us know if it's working for you or not.

Alex

NthlLcrmp:
Thanks Alex for your help.

This is better with this version. Spectral deconvolution with the Multivariate curve resolution algorithm is now working. However, I still have issue with the Hierarchical clustering algorithm. I can select feature and cluster, and I can preview the first clustering but not the second : 
![image](https://user-images.githubusercontent.com/114736047/193753565-fe80168b-a445-4236-af06-0eb3feae44ac.png)


And when I try to run it anyway  I still have an 'Unknown error' : 
![image](https://user-images.githubusercontent.com/114736047/193753900-cb025924-7bd1-4f8c-8f59-41d89a660226.png)

Could this be due to the chosen parameters ?

Nathalie

[...]

PR #933, opened 2022-10-05
Fix some issues in Spectral Deconvolution / Hierarchical Clusters #925

asmirn1:
- Fix the plot update when the parameters are changed
- Fix the IndexOutOfBoundsException in EICPlot


src/main/java/io/github/mzmine/modules/dataprocessing/adap_hierarchicalclustering/EICPlot.java
```diff
@@ -159,14 +159,15 @@ public void updateData(List<List<NavigableMap<Double, Double>>> clusters, List<D
           series.add(e.getKey(), e.getValue());
 
         try{
+          colorDataset.add(color);
```
SteffenHeu, line 162: why does this have to be run in a try-catch environment?

asmirn1: @SteffenHeu It doesn't. it just needs to be called before `xyDataset.addSeries(series)`. I'll move it out the try-catch.

SteffenHeu: but why is the try catch there in the first place? what can cause an exception here?

asmirn1: xyDataset.addSeries() fires an event to redraw the plot. So if the plot draw fails, we'll catch it. I don't remember exactly why we had this try-catch here. I assume, it couldn't redraw the plot for some reason.

SteffenHeu: That's weird. From where is the dataset added? Changes to the gui need to be run on the java FX thread, otherwise they fire an exception. maybe it was that?

 We have the MZmineCore.runLater() method for that. If that's not the issue, we can keep the try-catch.

src/main/java/io/github/mzmine/modules/dataprocessing/adap_hierarchicalclustering/ADAP3DecompositionV1_5Task.java
```diff
@@ -224,15 +224,15 @@ private FeatureList decomposePeaks(FeatureList peakList)
...
-            new HashMap<>(refPeakRow.getFeatureInformation().getAllProperties()));
+            new HashMap<>(originalPeakRow.getFeatureInformation().getAllProperties()));
         row.setFeatureInformation(information);
       }
 
       // Set row properties
-      row.setAverageMZ(refPeakRow.getAverageMZ());
-      row.setAverageRT(refPeakRow.getAverageRT());
+      row.setAverageMZ(originalPeakRow.getAverageMZ());
+      row.setAverageRT(originalPeakRow.getAverageRT());
```
robinschmid, line 235: All the row types you actually do not have to set. The average values are computed automatically once a new feature is added or the values are changed.

robinschmid: this means that you just set the values in the samples.
Also for example for fragmentation spectra, the list is set in the feature and then reflected to the row

robinschmid (changes requested):
Looks good. Thanks for the changes. Only the one comment remaining about the average row values

src/main/java/io/github/mzmine/modules/dataprocessing/adap_hierarchicalclustering/ADAP3DecompositionV1_5Task.java
asmirn1: @robinschmid So you mean that I don't need to call `row.setAverageMZ()` and `row.setAverageRT()` here. Correct?

robinschmid: Yes. All the average row values are calculated based on bindings

robinschmid (approved):
Great thanks for the fix

PR #942, opened 2022-10-12
Option to skip files and libraries that are already present in project

robinschmid:
1. optional - so you need to select
2. made ALL_FILES the default for all steps that use raw data files parameter in batch wizard. This way people can just rerun the batch after changing few parameters

### DIFF
--- a/src/main/java/io/github/mzmine/modules/tools/batchwizard/BatchWizardController.java
+++ b/src/main/java/io/github/mzmine/modules/tools/batchwizard/BatchWizardController.java
@@ -506,7 +506,7 @@ public class BatchWizardController {
     final ParameterSet param = MZmineCore.getConfiguration()
         .getModuleParameters(MassDetectionModule.class).cloneParameterSet();
     param.getParameter(MassDetectionParameters.dataFiles)
-        .setValue(new RawDataFilesSelection(RawDataFilesSelectionType.BATCH_LAST_FILES));
+        .setValue(new RawDataFilesSelection(RawDataFilesSelectionType.ALL_FILES));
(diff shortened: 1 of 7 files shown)


robinschmid:
image

This also handles duplicates in the import. If detected it shows this error:
image

robinschmid:
Oh so I actually added that it detects duplicates - and then fails before even trying to import. Otherwise, it was always importing and then failed once it got duplicates.

The option only checks if people already imported the data.

The problem is that people usually 
1. create their batch
2. run it once
3. optimize it
4. want to run again - but it fails because it imports the files again. Therefore, you have to either clear the project or remove the import step and then change the mass detection steps and chromatogram builder etc.

With this change, it just skips the library and data import if files are already loaded.

src/main/java/io/github/mzmine/modules/tools/batchwizard/BatchWizardController.java
```diff
@@ -544,7 +544,7 @@ private MZmineProcessingStep<MZmineProcessingModule> makeAdapStep(
     final ParameterSet param = MZmineCore.getConfiguration()
         .getModuleParameters(ModularADAPChromatogramBuilderModule.class).cloneParameterSet();
     param.setParameter(ADAPChromatogramBuilderParameters.dataFiles,
-        new RawDataFilesSelection(RawDataFilesSelectionType.BATCH_LAST_FILES));
+        new RawDataFilesSelection(RawDataFilesSelectionType.ALL_FILES));
```
SteffenHeu, line 547: don't know if this is the best solution. again, this will affect reproducibility.

Can we edit the batchlastfiles list?

robinschmid:
Something similar is in the main branch now

PR #943, opened 2022-10-12
Update normalization

SteffenHeu:
- Different normalization was used in the feature table and the image chart opened from the feature table
- use all scans instead of scans where the image feature was detected in for normalization


src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
```diff
@@ -1,6 +1,5 @@
 /*
  * Copyright (c) 2004-2022 The MZmine Development Team
- *
```
robinschmid, line 3: Are you using a different Copyright

robinschmid (approved):
Looks good to me.

SteffenHeu:
fixed

robinschmid:
Thanks steffen

Issue #911, opened 2022-09-22
Manual removal of false DB:annotations


tsufz:
**Is your feature request related to a problem? Please describe.**
Using the DB:annotation, in some cases, multiple peaks could be annotated with the same compound. Unfortunately, it is not possible to remove false annotations manually. This was possible in `MZmine 2`.

**Describe the solution you'd like**
I would like to get a simple method to remove the false DB:annotation via the context menu.

**Describe alternatives you've considered**
No alternatives.

**Additional context**
![2022-09-22 15_27_10-datascience2 - Remote Desktop Connection](https://user-images.githubusercontent.com/2526819/191761033-fb3914fe-d64e-4710-8b59-a06db39d19bb.png)

PR #951, opened 2022-10-17
add option to clear specific annotations

SteffenHeu:
as requested in #911, adds the option to remove not just compound but all other identifications from the specific rows from the feature table. 
@robinschmid is there anything special about the IIN? Can we just remove an annotation from there or should that type be blocked?


SteffenHeu:
grafik

SteffenHeu:
Ok, i will disable IIN removal for now and you can look after the IIN implementation when you have the time, ok?

robinschmid (approved):
Very nice solution

PR #955, opened 2022-10-19
Merge minrelative

SteffenHeu:
Added minimum relative intensity for ims ms2 spectra after merging. also added it to the batch wizard. minimum intensity is 1% by default and enabled.


robinschmid (approved):
Looks good to me. Thanks Steffen

PR #960, opened 2022-10-24
add DataType#getDefaultVisibility

SteffenHeu:


src/main/java/io/github/mzmine/datamodel/features/types/DataType.java
```diff
@@ -245,6 +245,14 @@ public TreeTableColumn<ModularFeatureListRow, Object> createColumn(
     }
   }
 
+  /**
+   * Defines if this data type is visible in the feature table by default.
+   * @return true or false.
+   */
+  public boolean getDefaultVisibility() {
+    return false;
+  }
```
robinschmid, line 254: Why not do true as the default and then only define those as false that we do not want. 
Would make sense to me

robinschmid (approved):
Looks good. Maybe true could be the default and we do not have to override the method then

SteffenHeu:
No, we have so many types, most of which are not interesting most of the time. I always disable most of the columns when i open the feature table on a new pc.

robinschmid:
Okay

PR #961, opened 2022-10-24
add ims type supported to Isofinder and rowsfilter

SteffenHeu:


robinschmid (approved):
Thanks for tagging those

PR #962, opened 2022-10-25
Minor fixes

### LINKED ISSUE #781: 2D plot log scale palette creates cyan background
* Windows 10 Pro
* MZmine 3.1.0-beta 

In 2D plot, choosing logarithmic scale palette gives the following result:

image

omokshyna:
- Fixed an error in the Histogram plot that occurred due to the empty "Plotted data" field
- Added module names to the parameter dialogues
- Modified log color scale in 2D plot (closes #781)


SteffenHeu:
Nice to finally have the module name in the dialog. 
Currently thinking if this is the best place to set it. Usually we retrieve the parameters via the `public @Nullable ParameterSet getModuleParameters(Class<? extends MZmineModule> moduleClass)` mthod in the MZmineConfiguration. Effectively every parameter set is a singleton and is usually cloned for the batch mode etc.. So we could set it there and also put it in the clone methods, so we always have it.

src/main/java/io/github/mzmine/parameters/impl/SimpleParameterSet.java
```diff
@@ -83,6 +85,12 @@ public SimpleParameterSet(Parameter<?> parameters[], String onlineHelpUrl) {
     this.helpUrl = onlineHelpUrl;
   }
 
+  public SimpleParameterSet(Parameter<?> parameters[], String onlineHelpUrl, String moduleNameAttribrute) {
+    this.parameters = parameters;
+    this.helpUrl = onlineHelpUrl;
+//    this.moduleNameAttribrute = moduleNameAttribrute;
```
robinschmid, line 91: Why is this commented out. the argument is not used

omokshyna: will fix that

omokshyna:
@SteffenHeu, thanks, just a couple of questions to be sure. So, do you think it's better to move module name functionality from ParameterSet to MZmineConfigurationImpl to the setModuleParameters()? But would it require then any changes in clone() methods if a module name is already in ParameterSet?

src/main/java/io/github/mzmine/parameters/impl/SimpleParameterSet.java
omokshyna: I removed this constructor altogether

omokshyna:
Updated the PR, now the module name in the dialogues is accessible from batch queue and also added to clone() method of SimpleParameterSet

PR #965, opened 2022-10-25
fix adap error spam

### LINKED ISSUE #938: Bug in ADAP Chromatogram builder with Scans "All"
### Basic information

* My operating system: Windows 11
* My MZmine version:  3.2.28

### What happened

Launching the ADAP Chromarogram builder without scan filters leads to a cascade of errors, and MZmine crashes. Error reproduces in multiple different data files. This behavior is not reproduced in case if MS level is set to 2 (though it is not intended method usage, the user can do that for now without warning). 

image

image

The console output shows the following:

`Exception in thread "JavaFX Application Thread" java.lang.StackOverflowError`

`Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Task controller thread"`

SteffenHeu:
fix #938 

#git shame:
grafik


robinschmid:
I copied the module in 2019 when we moved to modular data model.
So this is not mine :D I just did not remove it.

robinschmid:
Good try of git shame though. I was laughing hard.

PR #972, opened 2022-11-03
Lib export refinement

SteffenHeu:
Only export top compound match, add quality parameters for the library export

### DIFF
--- a/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationTask.java
+++ b/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationTask.java
@@ -88,7 +88,7 @@ public class LibraryBatchGenerationTask extends AbstractTask {
   private final SpectralLibraryExportFormats format;
   private final Map<DBEntryField, Object> metadataMap;
   private final boolean handleChimerics;
-  private final LibraryExportQualityParameters qualityParameters;
+  private final MsMsQualityChecker msMsQualityChecker;
   private double allowedOtherSignalSum = 0d;
   private MZTolerance mzTolChimericsMainIon;
   private MZTolerance mzTolChimericsIsolation;
@@ -113,8 +113,8 @@ public class LibraryBatchGenerationTask extends AbstractTask {
         LibraryBatchGenerationParameters.metadata).getEmbeddedParameters();
     metadataMap = meta.asMap();
 
-    qualityParameters = parameters.getParameter(LibraryBatchGenerationParameters.quality)
-        .getEmbeddedParameters();
+    msMsQualityChecker = parameters.getParameter(LibraryBatchGenerationParameters.quality)
+        .getEmbeddedParameters().toQualityChecker();
 
     //
     handleChimerics = parameters.getValue(LibraryBatchGenerationParameters.handleChimerics);
@@ -193,15 +193,14 @@ public class LibraryBatchGenerationTask extends AbstractTask {
 //      final DataPoint[] dataPoints = spectra.get(i);
 
       final Scan msmsScan = scans.get(i);
-      final List<DataPoint> explainedSignals = qualityParameters.matchAndGetExplainedSignals(
+      final List<DataPoint> explainedSignals = msMsQualityChecker.matchAndGetExplainedSignals(
           msmsScan, match, row);
       if (explainedSignals == null) {
         continue;
       }
 
-      DataPoint[] dps =
-          qualityParameters.getValue(LibraryExportQualityParameters.exportExplainedSignalsOnly)
-              ? explainedSignals.toArray(DataPoint[]::new) : ScanUtils.extractDataPoints(msmsScan);
+      DataPoint[] dps = msMsQualityChecker.exportExplainedSignalsOnly() ? explainedSignals.toArray(
+          DataPoint[]::new) : ScanUtils.extractDataPoints(msmsScan);
 
       // add instrument type etc by parameter
       SpectralDBEntry entry = new SpectralDBEntry(msmsScan, match, dps);
--- a/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationTask.java
+++ b/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationTask.java
@@ -83,12 +83,12 @@ public class LibraryBatchGenerationTask extends AbstractTask {
 
   private static final Logger logger = Logger.getLogger(LibraryBatchGenerationTask.class.getName());
   private final ModularFeatureList[] flists;
-  private final int minSignals;
   private final ScanSelector scanExport;
   private final File outFile;
   private final SpectralLibraryExportFormats format;
   private final Map<DBEntryField, Object> metadataMap;
   private final boolean handleChimerics;
+  private final LibraryExportQualityParameters qualityParameters;
   private double allowedOtherSignalSum = 0d;
   private MZTolerance mzTolChimericsMainIon;
   private MZTolerance mzTolChimericsIsolation;
@@ -102,7 +102,6 @@ public class LibraryBatchGenerationTask extends AbstractTask {
   public LibraryBatchGenerationTask(final ParameterSet parameters, final Instant moduleCallDate) {
     super(null, moduleCallDate);
     flists = parameters.getValue(LibraryBatchGenerationParameters.flists).getMatchingFeatureLists();
-    minSignals = parameters.getValue(LibraryBatchGenerationParameters.minSignals);
     scanExport = parameters.getValue(LibraryBatchGenerationParameters.scanExport);
     format = parameters.getValue(LibraryBatchGenerationParameters.exportFormat);
     String exportFormat = format.getExtension();
@@ -113,6 +112,9 @@ public class LibraryBatchGenerationTask extends AbstractTask {
     LibraryBatchMetadataParameters meta = parameters.getParameter(
         LibraryBatchGenerationParameters.metadata).getEmbeddedParameters();
     metadataMap = meta.asMap();
+
+    qualityParameters = parameters.getParameter(LibraryBatchGenerationParameters.quality)
+        .getEmbeddedParameters();
 
     //
     handleChimerics = parameters.getValue(LibraryBatchGenerationParameters.handleChimerics);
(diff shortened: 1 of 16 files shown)


src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationTask.java
```diff
@@ -183,59 +185,55 @@ private void processRow(final BufferedWriter writer, final FeatureListRow row)
...
+    var match = matches.get(0);
+
+    // filter matches
+    for (int i = 0; i < scans.size(); i++) {
+//      final DataPoint[] dataPoints = spectra.get(i);
 
-    for (var match : matches) {
-      // first entry for the same molecule reflect the most common ion type, usually M+H
-      if (Objects.equals(match.getCompoundName(), lastName)) {
+      final List<DataPoint> dataPoints = qualityParameters.matches(scans.get(i), match, row);
```
robinschmid, line 195: rename variable to explainedSignals.
method maybe matchAndGetExplainedSignals

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
```diff
@@ -0,0 +1,164 @@
...
+import org.openscience.cdk.tools.manipulator.MolecularFormulaManipulator;
+
+public class LibraryExportQualityParameters extends SimpleParameterSet {
+
+  public static final OptionalParameter<IntegerParameter> minNumSignals = new OptionalParameter<>(
+      new IntegerParameter("Minimum number of signals",
+          "Number of signals an MS2 must contain to be exported.", 3), true);
+
+  public static final OptionalParameter<PercentParameter> minExplainedPeaks = new OptionalParameter<>(
+      new PercentParameter("Minimum explained peaks (%)",
```
robinschmid, line 60: lets stick to signals as before

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
```diff
@@ -0,0 +1,164 @@
...
+          "Minimum number of explained peaks in an MSMS.", 0.4d), false);
+
+  public static final OptionalParameter<PercentParameter> minExplainedIntensity = new OptionalParameter<>(
+      new PercentParameter("Minimum explained intensity (%)",
+          "Minimum explained intensity in an MSMS.", 0.4d), false);
+
+  public static final MZToleranceParameter formulaTolerance = new MZToleranceParameter(
+      "MS/MS Formula tolerance", "m/z tolerance to assign MSMS signals to a formula.", 0.003,
```
robinschmid, line 68: Formula m/z tolerance

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
```diff
@@ -0,0 +1,164 @@
...
+      10.0d);
+
+  public static final BooleanParameter exportExplainedPeaksOnly = new BooleanParameter(
+      "Export explained peaks only",
```
robinschmid, line 72: signals instead of peaks

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
```diff
@@ -0,0 +1,164 @@
...
+      "Only explained peaks will be exported to the library spectrum.", false);
+
+  public static final BooleanParameter exportFlistNameMatchOnly = new BooleanParameter(
+      "Match compound and feature list name",
+      "Only export MS/MS spectra if the feature list name contains the compound name.");
```
robinschmid, line 77: This one seems to be more a parameter outside this quality parameters.
Maybe in the export maybe in the matching itself?

robinschmid: similar to the raw data file matching in the CSV matcher

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
```diff
@@ -0,0 +1,164 @@
...
+    super(new Parameter[]{minNumSignals, minExplainedPeaks, minExplainedIntensity, formulaTolerance,
+        exportExplainedPeaksOnly, exportFlistNameMatchOnly});
+  }
+
+  /**
+   * @param msmsScan   The msms scan to evaluate
+   * @param annotation The annotation to base the evaluation on
+   * @return The list of explained peaks. Null if this spectrum did not match the quality
+   * parameters. Empty list if formula parameters are disabled but the number of peaks matched the
+   * requirements.
```
robinschmid, line 89: Why empty and not just all datapoints?
The name of the method does not really tell us what it does.

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
```diff
@@ -0,0 +1,164 @@
...
+   */
+  public List<DataPoint> matches(final Scan msmsScan, final FeatureAnnotation annotation,
+      FeatureListRow f) {
+
+    if (getValue(LibraryExportQualityParameters.minNumSignals)
```
robinschmid, line 94: I know this is not super performance depending step... but maybe it would be cleaner and easier to maintain later.
Can you have a method in the parameters class to convert to a record LibraryQualityFilter where all the parameter values are extracted.
This way we do not have to loop through the paramerters for every lookup.  
Also it is clearer what types the values are etc.

robinschmid: the match method would then be in the record

robinschmid: Interesting option might even be to wrap the minNumSignals in an Optional, this would be very desccriptive and maybe easy to handle.
boolean and double are also good

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
```diff
@@ -0,0 +1,164 @@
...
+
+    final String formula = annotation != null ? annotation.getFormula() : null;
+    if (formula == null || annotation.getCompoundName() == null) {
+      return null;
+    }
+
+    if (getValue(LibraryExportQualityParameters.exportFlistNameMatchOnly) && !f.getFeatureList()
+        .getName().contains(annotation.getCompoundName())) {
+      return null;
+    }
```
robinschmid, line 108: A similar thing is done in the CSV compound matching

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
```diff
@@ -0,0 +1,164 @@
...
+        LibraryExportQualityParameters.formulaTolerance);
+    final IMolecularFormula molecularFormula = MolecularFormulaManipulator.getMajorIsotopeMolecularFormula(
+        formula, SilentChemObjectBuilder.getInstance());
+    final List<DataPoint> explainedSignals = new ArrayList<>();
+
+    try {
+      FormulaUtils.replaceAllIsotopesWithoutExactMass(molecularFormula);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
robinschmid, line 120: Maybe just use the FormulaUtils.createMajorIsotopeMolFormula

SteffenHeu: For some reason it always created the formula with nonsense isotopes here. i tried to debug it, nothing helped and this fixes it.

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
SteffenHeu: yeah but not everyone uses it.

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
```diff
@@ -0,0 +1,164 @@
...
+    if (adductType.getCDKFormula() != null) {
+      molecularFormula.add(adductType.getCDKFormula());
+    }
+
+    final DataPoint[] dataPoints = ScanUtils.extractDataPoints(msmsScan);
+
+    if (getValue(LibraryExportQualityParameters.minExplainedIntensity)) {
+      MSMSScore intensityFormulaScore = MSMSIntensityScoreCalculator.evaluateMSMS(
+          msmsFormulaTolerance, molecularFormula, dataPoints, msmsScan.getPrecursorMz(),
+          msmsScan.getPrecursorCharge(), dataPoints.length);
```
robinschmid, line 132: why do we have two different calculators - they should be doign the same right? The score should just have both the matched signals and the matched intensity

SteffenHeu: could not be bother to update it because it's also used in the feature table and i'd have to update the load/save and make it compatible with existing projects. the intensity is not used anywhere else, so this was the easiest solution

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
robinschmid: Its exactly the same code though...

src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryExportQualityParameters.java
robinschmid: ok

PR #975, opened 2022-11-04
Distance of MS2 to apex

robinschmid:


src/main/java/io/github/mzmine/datamodel/features/types/numbers/RtMs2ApexDistanceType.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @NotNull
+  @Override
+  public List<RowBinding> createDefaultRowBindings() {
+    return List.of();
+  }
+
+  @Override
+  public boolean getDefaultVisibility() {
+    return true;
```
SteffenHeu, line 75: No :D

SteffenHeu:
What's the purpose of this metric?

src/main/java/io/github/mzmine/datamodel/features/types/numbers/RtMs2ApexDistanceType.java
robinschmid: That was a copy paste

robinschmid:
Danny needs it to estimate how well the methods work - so how close it triggers to the apex of the peak.
I think this might be interesting overall. Could be some quality criteria for us

PR #983, opened 2022-11-08
dont merge msms during export unless user wants to

SteffenHeu:


robinschmid (approved):
Better to give this option to the user. Thanks steffen

PR #984, opened 2022-11-08
Fixed issue where selected scan is null after normalization

tnguy271:
For this issue, https://github.com/mzmine/mzmine3/issues/925 the user was unable to do adap aligner after they normalized. This throws the error the selected scans is null.


src/main/java/io/github/mzmine/modules/dataprocessing/norm_linear/LinearNormalizerTask.java
```diff
@@ -108,6 +108,8 @@ public void run() {
         new ModularFeatureList(originalFeatureList + " " + suffix, getMemoryMapStorage(),
             originalFeatureList.getRawDataFiles());
 
+
+    originalFeatureList.getRawDataFiles().forEach(file-> normalizedFeatureList.setSelectedScans(file, file.getScans()));
```
SteffenHeu, line 112: the selected scans should be transferred from the originalFeatureList. Similar to the previous fixes.

`originalFeatureList.getRawDataFiles().forEach(file-> normalizedFeatureList.setSelectedScans(file, originalFeatureList.getSecletedScans(file)));`

SteffenHeu:
thanks @tnguy271

PR #986, opened 2022-11-09
MGF file format option for Batch and Single library export

robinschmid:
- added an MGF generator 
- added batch and single spectra support
- data points are formatted


src/main/java/io/github/mzmine/util/spectraldb/entry/DBEntryField.java
```diff
@@ -66,7 +67,7 @@ COLLISION_ENERGY, FRAGMENTATION_METHOD, ISOLATION_WINDOW, NUM_PEAKS(Integer.clas
   PRINCIPAL_INVESTIGATOR, DATA_COLLECTOR, SOFTWARE,
 
   // Dataset ID is for MassIVE or other repositories
-  DATASET_ID, USI, DATAFILE_COLON_SCAN_NUMBER, SPLASH,
+  DATASET_ID, USI, SCAN_NUMBER(Integer.class), DATAFILE_COLON_SCAN_NUMBER, SPLASH,
```
SteffenHeu, line 70: is having the scan number necessary? seems totally useless in a library entry

robinschmid: Its optional - but the idea is that the scan number, dataset_id, USI, can locate the scan in the original file in a public dataset - so that people can go back and find the original spectrum if wanted

PR #990, opened 2022-11-10
fix issue in alignment task where selected scans is null

tnguy271:


tnguy271:
@SteffenHeu Could you check this merge request?

src/main/java/io/github/mzmine/modules/dataprocessing/align_adap3/ADAP3AlignerTask.java
```diff
@@ -240,6 +240,10 @@ private FeatureList alignPeaks() {
 
     alignedPeakList.getAppliedMethods().add(new SimpleFeatureListAppliedMethod(
         ADAP3AlignerModule.class, parameters, getModuleCallDate()));
+
+
+    allDataFiles.forEach(file-> alignedPeakList.setSelectedScans(file, file.getScans()));
```
SteffenHeu, line 245: this is the same issue we discussed before, don't use all scans. Use the selected scans from the previous feautre lists used as input for this feature list.

tnguy271: @SteffenHeu I changed the logic, please let me know if it's correct. Thank you.

SteffenHeu: looks good, thanks

PR #991, opened 2022-11-11
Fixes

SteffenHeu:
- fix compound name matching in ms2 quality and library export
- fix cast exception in msms mirror
- add green-yellow paint scale as default (better contrast with black heatmap background)
- [add correct module during ScanFilteringTask](https://github.com/mzmine/mzmine3/pull/991/commits/752740377605ae3604a58cede9a553b3a20b8949)

### DIFF
--- a/src/main/java/io/github/mzmine/parameters/parametertypes/paintscale/PaintScalePaletteParameter.java
+++ b/src/main/java/io/github/mzmine/parameters/parametertypes/paintscale/PaintScalePaletteParameter.java
@@ -41,8 +41,8 @@ import org.w3c.dom.NodeList;
  *
  * @author SteffenHeu steffen.heuckeroth@gmx.de / s_heuc03@uni-muenster.de
  */
-public class PaintScalePaletteParameter
-    implements UserParameter<SimpleColorPalette, PaintScalePaletteComponent> {
+public class PaintScalePaletteParameter implements
+    UserParameter<SimpleColorPalette, PaintScalePaletteComponent> {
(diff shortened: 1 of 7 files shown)


src/main/java/io/github/mzmine/parameters/parametertypes/paintscale/PaintScalePaletteParameter.java
```diff
@@ -106,11 +107,16 @@ public void loadValueFromXML(Element xmlElement) {
 
     selected = (selected != -1) ? selected : 0;
 
+    if (!palettes.contains(SimpleColorPalette.GREEN_YELLOW)) {
+      palettes.add(SimpleColorPalette.GREEN_YELLOW);
```
robinschmid, line 111: As a default we should use someting that is colorblind friendly.
I think the default here is usually blue-red, blue to yellow, blue, to brown. Something like that. 

green, yellow, red is not so easy

robinschmid: Original colors
image

Protanopia - colorblindness
image

robinschmid: In this case green yellow is like a monochrome scale. Only the dark/light are different

SteffenHeu:
grafik
new default palette

robinschmid:
looks great

Issue #994, opened 2022-11-14
Unhandled Exception in Saving Feature List task


hhabra:
### Basic information

* My operating system and version: Mac OS X 12.6.1
* My MZmine version: 3.2.8

### What happened

Attempting to save my project led to the following error: 
![image](https://user-images.githubusercontent.com/8129799/201783208-294054b9-856b-4dc7-926e-3e76c0118e61.png)

**Project:** I have 48 DDA files whose MS/MS spectra I would like to gather into a single .mgf file. 

**Actions taken:** 
Raw data import → Mass Detection (MS = 1) and Mass Detection (MS = 2) → MSn feature list builder → 13C Isotope Filter → Join Aligner → Export as GNPS - Feature Based Molecular Networking

Attempting to save my project at this stage leads to the error in the screenshot.

SteffenHeu:
Hi hhabra, thanks for reporting the issue.

can you test the auto build from this PR and let us know if it works? 
https://github.com/mzmine/mzmine3/pull/995

hhabra:
Hello Steffen,
Unless I did things incorrectly, I'm afraid I'm getting the same error (I pulled repository with git and built using ./gradlew). The version of the app appears to be 3.2.53.

robinschmid:
Is it chromatography data? Then you should consider using the local minimum resolver or similar to split the chromatograms into features. 
Another way to download is to scroll down here:
https://github.com/mzmine/mzmine3/actions/runs/3468498870

hhabra:
Yes, it is liquid chromatography data. Are the peak resolving steps necessary if I only want to gather the MS2 data into a MGF file? My current process is: Raw data import → Mass Detection (MS = 1) and Mass Detection (MS = 2) → MSn feature list builder (MS level = 2) → 13C Isotope Filter → Join Aligner → Export as GNPS - Feature Based Molecular Networking
The issue is that MZmine is not able to save the results of MSn feature list builder as a project. Saving a batch file works fine I think, but not the chromatogram files.

robinschmid:
We will release a new version soon and hope this is fixed there but might be a more specific problem. So please try the version once pushed. Thanks you.

PR #995, opened 2022-11-15
add selected scans to MsnPeakPickingTask

SteffenHeu:
reported in #994


robinschmid (approved):
I pushed a fix.
So there was a mixup in the chromatogram construction with the leading and trailing zeros. Sometimes it would add a scan 2 times if there was only one zero between the scans.

SteffenHeu:
thanks

PR #1011, opened 2022-11-28
Update to JDK19, keep javafx16

robinschmid:
This combination seems to work fine. I also improved the buffering of the graphical types. Now we buffer the wrapper pane.

There is different bugs with the table on javafx>16
- Scrolling creates all cells
- the table jumps 
- if tested for TableRow.isVisible() it seems wrong state is provided and some cells are not rendered.


SteffenHeu:
You also have to update the github actions if we use 19 or not?

robinschmid:
True. I thought about it and then forgot to do it.

PR #1013, opened 2022-11-29
Targeted update

SteffenHeu:
- update targeted feature detection module
  - uses the csv utils and ImportTypes like the compound db search
  - supports ion mobility files
- add utility methods to CSVParsingUtils to easily extract CompoundDBAnnotations from a csv file
- update compound db search so we use some of the utility methods to avoid duplication


src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
```diff
@@ -95,6 +95,24 @@ static List<CompoundDBAnnotation> buildCompoundsWithAdducts(
...
+   *                       formula/smiles/neutral mass.
+   * @return True if the baseAnnotation contains a precursor m/z and useIonLibrary is false. Also
+   * true if useIonLibrary is true and the annotation contains a smiles, a formula or a neutral
+   * mass.
+   */
+  static boolean isBaseAnnotationValid(CompoundDBAnnotation baseAnnotation, boolean useIonLibrary) {
+    if (baseAnnotation.getPrecursorMZ() != null && !useIonLibrary) {
+      return true;
+    } else if (useIonLibrary && (baseAnnotation.get(NeutralMassType.class) != null
+        || baseAnnotation.getFormula() != null || baseAnnotation.getSmiles() != null)) {
```
robinschmid, line 110: How about we always calculate the neutral mass if its missing and use a priority of the sources
neutral mass > inchi > smiles > formula

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionModuleTask.java
```diff
@@ -0,0 +1,371 @@
...
+import java.util.logging.Logger;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+class TargetedFeatureDetectionModuleTask extends AbstractTask {
+
+  protected static final Range<Double> doubleInfiniteRange = Range.closed(Double.NEGATIVE_INFINITY,
+      Double.POSITIVE_INFINITY);
+  protected static final Range<Float> floatInfiniteRange = Range.closed(Float.NEGATIVE_INFINITY,
+      Float.POSITIVE_INFINITY);
```
robinschmid, line 78: those are not really needed there is Range.all()

robinschmid, line 78: empty comment

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionModuleTask.java
```diff
@@ -0,0 +1,371 @@
...
+
+    // Get all scans of this data file
+    if (dataFile instanceof IMSRawDataFile imsFile) {
+      if (processImsFile((List<? extends Gap>) gaps, imsFile)) {
+        return;
+      }
+    } else if (dataFile instanceof RawDataFile) {
+      if (!processLcmsFile(gaps)) {
+        return;
+      }
```
robinschmid, line 216: Thats a good one.
processIMS is true if canceled and processLcms is false...
maybe better not check the return state of these methods and rather check for isCanceled() after calling this methods. Looks redundant but is clear in purpose

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionParameters.java
```diff
@@ -0,0 +1,112 @@
...
+public class TargetedFeatureDetectionParameters extends SimpleParameterSet {
+
+  public static final RawDataFilesParameter rawDataFile = new RawDataFilesParameter();
+  public static final ScanSelectionParameter scanSelection = new ScanSelectionParameter(
+      new ScanSelection(1));
+  public static final StringParameter suffix = new StringParameter("Name suffix",
+      "Suffix to be added to feature list name", "detectedPeak");
+  public static final FileNameParameter featureListFile = new FileNameParameter("Feature list file",
+      "Name of the file that contains a list of peaks for targeted feature detection.",
+      FileSelectionType.OPEN);
```
robinschmid, line 69: Maybe use the same parameter name as for the file in local database search?

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionParameters.java
```diff
@@ -0,0 +1,112 @@
...
+  public static final StringParameter fieldSeparator = new StringParameter("Field separator",
+      "Character(s) used to separate fields in the database file", ",");
```
robinschmid, line 71: Maybe add use \t for tab separated data

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionParameters.java
```diff
@@ -0,0 +1,112 @@
...
+      new RTToleranceParameter());
+  public static final OptionalParameter<MobilityToleranceParameter> mobilityTolerance = new OptionalParameter<>(
+      new MobilityToleranceParameter());
+
+  public static final OptionalModuleParameter<IonLibraryParameterSet> ionLibrary = new OptionalModuleParameter<>(
+      "Calculate adduct masses",
+      "Ion types to search for. Either neutral mass, formula or smiles must be imported for every compound.",
+      new IonLibraryParameterSet());
+
+  private static final List<ImportType> importTypes = List.of(
```
robinschmid, line 85: Would be great if all the parameters of the local csv library were just a sub parameter set that is used in both the database search and here

src/main/java/io/github/mzmine/util/CSVParsingUtils.java
```diff
@@ -70,10 +93,118 @@ public static List<ImportType> findLineIds(List<ImportType> importTypes, String[
...
+          annotation.put(dt, Double.parseDouble(line[columnIndex]));
+        } else if (type.getDataType() instanceof StringType st) {
+          annotation.put(st, line[columnIndex]);
+        } else if (type.getDataType() instanceof IonAdductType iat) {
+          final IonType ionType = IonType.parseFromString(line[columnIndex]);
+          annotation.put(IonTypeType.class, ionType);
+        } else {
+          throw new RuntimeException(
+              "Don't know how to parse data type " + type.getDataType().getClass().getName());
+        }
```
robinschmid, line 131: Best use case for the instanceof pattern matching

robinschmid (changes requested):
Looks good to me. only a few comments

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionModuleTask.java
SteffenHeu: standard extract method :D

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionModuleTask.java
SteffenHeu: no, it is needed because Range.all() is unbound, so it leads to errors later on:
```
      final double upperBreakPoint = doubleToleranceRange.upperEndpoint();
      final double lowerBreakPoint = doubleToleranceRange.upperEndpoint();
```

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionParameters.java
SteffenHeu: yeah, would be great....

robinschmid: :D
Yeah can be a later thing

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionModuleTask.java
robinschmid: What does that mean?
So you cannot get the endpoints?

src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
```diff
@@ -103,52 +124,68 @@ static List<CompoundDBAnnotation> buildCompoundsWithAdducts(
...
-    final Double neutralMass = annotation.get(NeutralMassType.class);
+  static double calcMzForAdduct(@NotNull CompoundDBAnnotation annotation,
+      @NotNull IonType adduct) throws CannotDetermineMassException {
+
+    Double neutralMass = annotation.get(NeutralMassType.class);
+    if (neutralMass == null) {
+      // try to calc the neutral mass and keep it for subsequent calls.
+      neutralMass = CompoundDBAnnotation.calcNeutralMass(annotation);
+      annotation.put(NeutralMassType.class, neutralMass);
+    }
```
robinschmid, line 135: Is this only called in a single threaded environment?
I think it should. We load the library and create different ions to generate the final list of possible annotations.
Then this is fine. Otherwise we would need to synchronize?

src/main/java/io/github/mzmine/datamodel/features/compoundannotations/SimpleCompoundDBAnnotation.java
```diff
@@ -270,7 +270,7 @@ public boolean matches(FeatureListRow row, @Nullable MZTolerance mzTolerance,
         Math.abs(1 - (row.getAverageCCS() / ccs)) > percentCCSTolerance));
   }
 
-  public Float getScore(FeatureListRow row, @Nullable MZTolerance mzTolerance,
+  public Float calculateScore(FeatureListRow row, @Nullable MZTolerance mzTolerance,
```
robinschmid, line 273: See #1023 I have changed this method to use the new FeatureListUtils.getAlignmentScore method. I think there are also some weird things here. 
There were checks if the value falls within the range but the matches method should already do this.
If it were still somehow falling outside the tolerance - the value should still be added 0 score and scorers should be incremented otherwise this field (mz,rt,mobility) will not be considered. See my PR

src/main/java/io/github/mzmine/modules/dataprocessing/id_ccscalc/CCSUtils.java
```diff
@@ -132,7 +133,8 @@ public static List<CCSCalibrant> getCalibrantsFromCSV(final File file) throws IO
             new ImportType(true, "mobility", DataTypes.get(
                 io.github.mzmine.datamodel.features.types.numbers.MobilityType.class)), //
             new ImportType(true, "ccs", DataTypes.get(CCSType.class)), //
-            new ImportType(true, "charge", DataTypes.get(ChargeType.class))), content[0]);
+            new ImportType(true, "charge", DataTypes.get(ChargeType.class))), content[0],
+        new SimpleStringProperty());
```
robinschmid, line 137: What is this simple string property actually doing there?

robinschmid: its the error message but why is it set but not used?

src/main/java/io/github/mzmine/parameters/ParameterSet.java
```diff
@@ -49,11 +51,29 @@ default <V, T extends Parameter<V>> V getValue(T parameter) {
...
+      final UserParameter<V, ?> actualParam = getParameter(parameter).getEmbeddedParameter();
+      return actualParam == null ? null : actualParam.getValue();
+    }
+    return defaultValueSupplier != null ? defaultValueSupplier.get() : null;
+  }
+
+  default <T extends ParameterSet> ParameterSet getEmbeddedParameterValue(
+      OptionalModuleParameter<T> parameter) {
+    return getParameter(parameter).getEmbeddedParameters();
+  }
```
robinschmid, line 75: This one I dont understand. Its not the value but the embedded parameter so its not really useful

src/main/java/io/github/mzmine/util/RangeUtils.java
```diff
@@ -265,7 +265,15 @@ public static String formatRange(Range<? extends Number> range, NumberFormat for
...
+   * @param center the center of the range
+   * @param length the total length of the range, divided by 2 on each side to creat the range.
+   */
   public static Range<Double> rangeAround(double center, double length) {
     return Range.closed(center - length / 2, center + length / 2);
   }
+
+  public static Range<Float> rangeAround(float center, float length) {
+    return Range.closed(center - length / 2, center + length / 2);
+  }
```
robinschmid, line 278: that was close. was just about to create one of those

src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
SteffenHeu: Why does it have to be single threaded? There was some problem with the cdk right?

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_targeted/TargetedFeatureDetectionModuleTask.java
SteffenHeu: exactly, range.all is unbound.

src/main/java/io/github/mzmine/parameters/ParameterSet.java
SteffenHeu: It returns the parameter set of an OptionalModuleParameter. i wanted to keep the name the same as for `getEmbeddedParameterValue(OptionalParameter<T> parameter)` so it's easier to type

robinschmid: why not getEmbeddedParameters instead of value

src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
robinschmid: at least the data structure behind it is HashMap - so if this is called on multiple threads this will produce a concurrent mod.

SteffenHeu: ohh ok you are talking about the annotation itselfm not about the while method. 

Yeah i think this is not going to be used in a multithreaded environment

src/main/java/io/github/mzmine/modules/dataprocessing/id_ccscalc/CCSUtils.java
SteffenHeu: Cause it is newly added :D can do later

src/main/java/io/github/mzmine/parameters/ParameterSet.java
SteffenHeu: i wanted to keep the name the same as for other parameter types

src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
robinschmid: okay

src/main/java/io/github/mzmine/modules/dataprocessing/id_ccscalc/CCSUtils.java
robinschmid: :D

src/main/java/io/github/mzmine/parameters/ParameterSet.java
robinschmid: ok

PR #1014, opened 2022-11-29
Update wizard

SteffenHeu:
- add hilic presets
- update ims workflow in wizard


src/main/java/io/github/mzmine/modules/tools/batchwizard/defaults/DefaultLcParameters.java
```diff
@@ -45,6 +45,10 @@ public class DefaultLcParameters {
       3, 1, 50, new RTTolerance(0.03f, Unit.MINUTES), new RTTolerance(0.03f, Unit.MINUTES),
       new RTTolerance(0.08f, Unit.MINUTES));
 
+  public static final DefaultLcParameters hilic = new DefaultLcParameters(true,
+      Range.closed(1d, 10d), 5, 1, 10, new RTTolerance(0.1f, Unit.MINUTES),
```
robinschmid, line 49: this should be wider similar, to the other parameter sets
we said we do not want to concat by default. people should have to set those values if they really want

PR #1015, opened 2022-11-30
Tag modules from the wizard as IMS ready

SteffenHeu:
Only the adap chrom builder still has a warning message, i thought it was more modules. The message is suppressed in batch mode now.
In 3.2.8 the Rows filter and the isotopic peaks finder also did not have this tag.


robinschmid:
looks good. Maybe chromatogram builder should be unrestircted and rather put the messages in a box

PR #1021, opened 2022-12-06
Alignment scores after alignment

robinschmid:
Added new columns only to the rows.

image

### DIFF
--- a/src/main/java/io/github/mzmine/datamodel/features/types/AlignExtraFeaturesType.java
+++ b/src/main/java/io/github/mzmine/datamodel/features/types/AlignExtraFeaturesType.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2004-2022 The MZmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+ * OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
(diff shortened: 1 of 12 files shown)


robinschmid:
Needs to be added to the image aligner

src/main/java/io/github/mzmine/datamodel/features/types/AlignExtraFeaturesType.java
```diff
@@ -0,0 +1,54 @@
...
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+ * OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.datamodel.features.types;
```
SteffenHeu, line 26: i feel like we should put the alignment score types in a new package sicne they are currently only used there

src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
```diff
@@ -164,13 +171,50 @@ public String getTaskDescription() {
     return "Join aligner, " + featureListName + " (" + featureLists.size() + " feature lists)";
   }
 
+  private static int getCandidates(final Range<Double> mzRange, final Range<Float> rtRange,
```
SteffenHeu, line 174: add javadoc that the originals list needs to be sorted

src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
```diff
@@ -276,6 +325,82 @@ public void run() {
 
   }
 
+  /**
+   *
+   */
+  private void addAlignmentScores() {
```
SteffenHeu, line 331: refactor to utility or static method so we can use it in the lc image aligner too

src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
robinschmid: I push this to FeatureListUtils in a different PR. Might be cleaner to merge it this way and then open a separate PR

src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
robinschmid: Done in #1021

robinschmid:
All done in #1021
Because the clean up and utility methods was more complex than thought.
We should merge this and then see that the other one works

PR #1022, opened 2022-12-07
Clean up alignment scores and FeatureListUtils

robinschmid:
- harmonized ways to get feature candidates within ranges and the calclation of the alginment score and the AlignmentScores analysis

Scores from LC-MS and then imaging to LC alignment are merged

Need to check that the scores and sorting are correct.
Earlier the sorting in the Image aligner was reversed but the default sorting of RowVsRowScore is already large - low 

The alignment scores are just moved. 

Merge after #1021


SteffenHeu:
Let me know when you checked

> Need to check that the scores and sorting are correct.

robinschmid:
All done.
Seems to perform the same way.
Also added a test to MZmine master to check the number of aligned features in the small feature finding test

src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
```diff
@@ -414,22 +295,23 @@ private void alignRowsOnBaseRows(List<List<FeatureListRow>> unalignedRows,
...
-          mzWeight > 0 ? mzTolerance.getToleranceRange(row.getAverageMZ()) : Range.all();
+          mzWeight > 0 ? mzTolerance.getToleranceRange(rowToAdd.getAverageMZ()) : Range.all();
       final Range<Float> rtRange =
-          rtWeight > 0 ? rtTolerance.getToleranceRange(row.getAverageRT()) : Range.all();
+          rtWeight > 0 ? rtTolerance.getToleranceRange(rowToAdd.getAverageRT()) : Range.all();
       final Range<Float> mobilityRange =
-          compareMobility && row.getAverageMobility() != null ? mobilityTolerance.getToleranceRange(
-              row.getAverageMobility()) : Range.singleton(0f);
+          compareMobility && mobilityWeight > 0 && rowToAdd.getAverageMobility() != null
+              ? mobilityTolerance.getToleranceRange(rowToAdd.getAverageMobility()) : Range.all();
```
SteffenHeu, line 310: does changing from Range.singleton(0f) to Range.all() not change the outcome?

src/main/java/io/github/mzmine/util/FeatureListUtils.java
```diff
@@ -54,100 +66,267 @@ public static void copyPeakListAppliedMethods(FeatureList source, FeatureList ta
...
+
+    SortedList<FeatureListRow> rows = alignedFeatureList.getRows().sorted(MZ_ASCENDING);
+
+    // find the number of rows that match RT,MZ,Mobility in each original feature list
+    rows.stream().parallel().forEach(alignedRow -> {
+      AlignmentScores score = calculator.calcScore(alignedRow);
+      if (mergeScores) {
+        AlignmentScores oldScore = alignedRow.get(AlignmentMainType.class);
+        score = score.merge(oldScore);
+      }
```
SteffenHeu, line 263: why exactly is there an option to merge?

robinschmid: So we run the join aligner first and get the results
If you were to run the join aligner or the lc image aligner again - we need to merge the results.
So how many samples were detected, how many extra features etc...

src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinAlignerTask.java
robinschmid: All the other ones are Range.all() so for RT and mz. Only makes sense to handle it the same here.

src/main/java/io/github/mzmine/util/FeatureListUtils.java
SteffenHeu: ahh ok, makes sense

PR #1023, opened 2022-12-08
Compound annotation score uses alignment score

robinschmid:
With the new alignment score method we can also use it in the compound annotation score.

```java
 /**
   * Compare row average values to ranges (during alignment or annotation to other mz, rt, and
   * mobility values based on tolerances -> ranges). General score is SUM((difference
   * row-center(range)) / rangeLength * factor) / sum(factors)
   *
   * @param testMz         tested value
   * @param testRt         tested value
   * @param testMobility   tested value
   * @param mzRange        allowed range
   * @param rtRange        allowed range
   * @param mobilityRange  allowed range
   * @param mzWeight       weight factor
   * @param rtWeight       weight factor
   * @param mobilityWeight weight factor
   * @return the alignment score between 0-1 with 1 being a perfect match
   */
  public static double getAlignmentScore(Double testMz, Float testRt, Float testMobility,
      @Nullable Range<Double> mzRange, @Nullable Range<Float> rtRange,
      @Nullable Range<Float> mobilityRange, double mzWeight, double rtWeight,
      double mobilityWeight) {

    // don't score range.all, will distort the scoring.
    mzRange = mzRange == null || mzRange.equals(Range.all()) ? null : mzRange;
    rtRange = rtRange == null || rtRange.equals(Range.all()) ? null : rtRange;
    mobilityRange =
        mobilityRange == null || mobilityRange.equals(Range.all()) ? null : mobilityRange;

    int scorers = 0;

    double score = 0f;
    // values are "matched" if the given value exists in this class and falls within the tolerance.
    if (mzWeight > 0 && mzRange != null && testMz != null) {
      final double exactMass = RangeUtils.rangeCenter(mzRange);
      double diff = Math.abs(testMz - exactMass);
      double maxAllowedDiff = rangeLength(mzRange) / 2;
      // no negative numbers
      score += Math.max(0, (1 - diff / maxAllowedDiff) * mzWeight);
      scorers += (int) Math.round(mzWeight);
    }

    if (rtWeight > 0 && rtRange != null && testRt != null) {
      final Float rt = RangeUtils.rangeCenter(rtRange);
      float diff = Math.abs(testRt - rt);
      score += Math.max(0, 1 - (diff / (rangeLength(rtRange) / 2)) * rtWeight);
      scorers += (int) Math.round(rtWeight);
    }

    if (mobilityWeight > 0 && mobilityRange != null && testMobility != null) {
      final Float mobility = RangeUtils.rangeCenter(mobilityRange);
      float diff = Math.abs(testMobility - mobility);
      score += Math.max(0, 1 - (diff / (rangeLength(mobilityRange) / 2)) * mobilityWeight);
      scorers += (int) Math.round(mobilityWeight);
    }

    if (scorers == 0) {
      return 0f;
    }

    return score / scorers;
  }

```


robinschmid:
This one is an addition to #1022

SteffenHeu (approved):

PR #1025, opened 2022-12-08
Themeparameter

SteffenHeu:
Added some new themes at https://github.com/JabRef/jabref/, will ask there for permission to reuse first, though. It's MIT license, too.

grafik


SteffenHeu:
Ok with the other delepvers https://github.com/JabRef/jabref/issues/9438

robinschmid:
Looks good to me

PR #1026, opened 2022-12-09
More consistent theme, fix lipid class parameter

SteffenHeu:
- Made theming more consistent
- fixed some black font on black background stuff
- refactored our theme additions to a separate css file
- hopefully fixed the lipid parameter duping itself all the time


robinschmid (approved):
Thanks for looking into the themes

PR #1033, opened 2022-12-12
update MobilityScanMergerTask

SteffenHeu:
The noise level parameter did not clearly state if it is applied to mobiltiy scans or to the frame. It should be applied to the frame (but wasnt in the merger task, probably due to all the refactoring we did to the merge method). applying the noise level to mobscans again would duplicate the mass detector which wouldnt make sense.

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_mobilityscanmerger/MobilityScanMergerParameters.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_mobilityscanmerger/MobilityScanMergerParameters.java
@@ -43,7 +43,7 @@ public class MobilityScanMergerParameters extends SimpleParameterSet {
 
   public static final RawDataFilesParameter rawDataFiles = new RawDataFilesParameter();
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/mzmine/modules/dataprocessing/featdet_mobilityscanmerger/MobilityScanMergerParameters.java
```diff
@@ -43,12 +43,13 @@ public class MobilityScanMergerParameters extends SimpleParameterSet {
 
   public static final RawDataFilesParameter rawDataFiles = new RawDataFilesParameter();
 
-  public static final DoubleParameter noiseLevel = new DoubleParameter("Noise level",
-      "Data points below this threshold will be ignored.",
+  public static final DoubleParameter noiseLevel = new DoubleParameter("Frane noise level",
```
robinschmid, line 46: Frane - Frame

robinschmid (changes requested):
There is a type, rest looks good

PR #1035, opened 2022-12-12
Update buttons in the status bar and check for updates

robinschmid:


src/main/java/io/github/mzmine/gui/MZmineGUI.java
```diff
@@ -614,13 +642,14 @@ public void displayMessage(String title, String msg) {
...
 
-      final Text text = new Text();
-//      text.getStyleClass().add("label");
-//      text.getStyleClass().add("text-id");
+      TextFlow flow = new TextFlow(new Text(msg));
+      if (url != null) {
+        Hyperlink href = new Hyperlink(url);
+        flow.getChildren().add(href);
+        href.setOnAction(event -> WebUtils.openURL(url));
+      }
```
SteffenHeu, line 650: does this still have the correct color in dark mode?

src/main/java/io/github/mzmine/gui/MZmineGUI.java
```diff
@@ -592,18 +599,39 @@ public void setStatusBarText(String message, Color textColor) {
...
       }
-      statusBar.setText(message);
-      statusBar.setStyle("-fx-text-fill: " + FxColorUtil.colorToHex(textColor));
+      statusBar.setText(null);
+      if (statusLabel == null) {
+        statusLabel = new Label(message);
+        statusBar.getLeftItems().add(statusLabel);
+      } else {
+        statusLabel.setText(message);
+      }
```
SteffenHeu, line 608: the hack to remove the link could be creating a new label every time, but i don't think that's feasible. Why not have a method `public void setStatusBarText(String message, Color textColor, URL url)` that will put a url, and the normal `public void setStatusBarText(String message, Color textColor)` removes said url by default

SteffenHeu (changes requested):

src/main/java/io/github/mzmine/gui/MZmineGUI.java
robinschmid: I removed this hardcoded color. We now just use the default color. and the negative color

robinschmid:
I have done exactly what you said. Good idea with one method. 
The colors are now the default style and the negative color if update needed. 
One method and the link is removed

SteffenHeu (approved):

SteffenHeu:
thanks

PR #1041, opened 2022-12-15
extract subtype value from ListWithSubsType

SteffenHeu:
Added a method to extract all values of a given data type (e.g. FormulaType) from all annotations (e.g. Compound db, spectral db. formula, lipid...)


src/main/java/io/github/mzmine/util/FeatureUtils.java
```diff
@@ -519,4 +527,75 @@ public static FeatureAnnotation getBestFeatureAnnotation(ModularDataModel m) {
...
+      // get the actual value of the ListWithSubsType stored in the feature list, we know its a list
+      final List<?> annotationList = featureListRow.get(listType);
+
+      if (annotationList == null || annotationList.isEmpty()) {
+        continue;
+      }
+
+      // get the list of subTypes
+      final List<DataType> subDataTypeList = listType.getSubDataTypes();
+      final int subColIndex = subDataTypeList.indexOf(theType);
```
robinschmid, line 586: Maybe add some comment here because that is important.
// find the targetType in this annotation, e.g., FormulaType or SmilesType.

Rename theType to something like targetType or searchType

robinschmid (dismissed):
Great addition. Just one comment that clarifies

PR #1044, opened 2022-12-16
Add style file and action

robinschmid:


robinschmid:
Need to check if there are newer better alternatives to check style and code quality

PR #1054, opened 2022-12-21
Fix netCDF

robinschmid:


tomas-pluskal:
Thanks Robin! Could you please also fix the `1 < 2` condition? I'm pretty sure it was meant to be `j < 2`.

robinschmid:
There was a condition in the loop. I just moved the check into the for loop condition.

PR #1058, opened 2022-12-22
dark mode intro tab

SteffenHeu:
Added dark mode intro trab to not destroy your eyes at night

grafik


robinschmid (approved):
Great dark mode - this makes the dark mode even more compelling

PR #1061, opened 2022-12-28
seasons

robinschmid:
There is a problem with the width of the AnchorPane that is used to overlay snow. How to ahve it take the size of the underlaying panes


src/main/java/io/github/mzmine/gui/mainwindow/MainWindow.fxml
```diff
@@ -55,6 +55,7 @@
   <stylesheets>
     <URL value="@/themes/MZmine_default.css"/>
   </stylesheets>
+  <StackPane fx:id="stackPane">
```
SteffenHeu, line 58: what's this for?

src/main/java/io/github/mzmine/gui/SnowFallPane.java
```diff
@@ -0,0 +1,98 @@
...
+import javafx.scene.Node;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.Region;
+import javafx.scene.paint.Color;
+import javafx.util.Duration;
+
+/**
+ * Winter season theme
+ */
+public class SnowFallPane extends AnchorPane {
```
SteffenHeu, line 40: i think if you make this a stack pane it will grow

PR #1066, opened 2023-01-06
add selected scans in rt cal, update csv import

SteffenHeu:
- fix rt calibration 
- disable feature list import modules since they are broken currently

### DIFF
--- a/src/main/java/io/github/mzmine/modules/io/import_features_csv/CsvImportTask.java
+++ b/src/main/java/io/github/mzmine/modules/io/import_features_csv/CsvImportTask.java
@@ -36,6 +36,7 @@ import io.github.mzmine.datamodel.features.Feature;
 import io.github.mzmine.datamodel.features.ModularFeature;
 import io.github.mzmine.datamodel.features.ModularFeatureList;
 import io.github.mzmine.datamodel.features.ModularFeatureListRow;
+import io.github.mzmine.datamodel.features.SimpleFeatureListAppliedMethod;
 import io.github.mzmine.datamodel.impl.SimpleDataPoint;
 import io.github.mzmine.parameters.ParameterSet;
 import io.github.mzmine.taskcontrol.AbstractTask;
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/mzmine/modules/io/import_features_csv/CsvImportTask.java
```diff
@@ -135,16 +139,15 @@ public void run() {
...
-              if (dp.getMZ() == feature_mz) {
-                finalDataPoint[0] = new SimpleDataPoint(feature_mz, dp.getIntensity());
-                break;
-              }
-            }
+
+        final Scan s_no = rawDataFile.getScanNumberAtRT(feature_rt, 1);
+        if (s_no != null) {
+          representativeScan = s_no;
+          final int peakIndex = s_no.binarySearch(feature_mz, true);
```
robinschmid, line 146: very nice

src/main/java/io/github/mzmine/modules/io/import_features_csv/CsvImportTask.java
```diff
@@ -135,16 +139,15 @@ public void run() {
...
-          if (s_no.getRetentionTime() == feature_rt) {
-            representativeScan = s_no;
-            for (DataPoint dp : s_no) {
-              if (dp.getMZ() == feature_mz) {
-                finalDataPoint[0] = new SimpleDataPoint(feature_mz, dp.getIntensity());
-                break;
-              }
-            }
+
+        final Scan s_no = rawDataFile.getScanNumberAtRT(feature_rt, 1);
```
robinschmid, line 143: just looked at this method and it seems to be completely broken somehow. I will try to fix it in a separate PR. I will change it to get closest scan. So better check here if the feature is within the RT range.

robinschmid (commented):
Everything else looks good only the RT check

src/main/java/io/github/mzmine/modules/io/import_features_csv/CsvImportTask.java
robinschmid: Okay it was functional... I did not see one part of it.. Still gonna optimize

robinschmid: #1074

PR #1074, opened 2023-01-11
binarySearchClosestScan in RawDataFile

robinschmid:


src/main/java/io/github/mzmine/util/collections/BinarySearchHelper.java
```diff
@@ -0,0 +1,146 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.util.collections;
+
+import io.github.mzmine.util.ArrayUtils;
+import java.util.Arrays;
+
+public interface BinarySearchHelper {
```
SteffenHeu, line 31: Not a fan, it could be that an object has multiple things you want to binary search. Then the getValue cannot be implemented twice

robinschmid: Its just that we have to copy the whole binary search each time

src/main/java/io/github/mzmine/datamodel/RawDataFile.java
```diff
@@ -75,14 +75,15 @@ public interface RawDataFile {
    * The maximum number of centroid data points in all scans (after mass detection and optional
    * processing)
    *
-   * @return max number of data points in masslist
+   * @return data point with maximum intensity (centroided)
```
SteffenHeu, line 78: this new doc is wrong, why did you change it?

src/main/java/io/github/mzmine/datamodel/RawDataFile.java
```diff
@@ -75,14 +75,15 @@ public interface RawDataFile {
...
    */
   int getMaxCentroidDataPoints();
 
+
   /**
    * The maximum number of raw data points in all scans
    *
-   * @return max raw data points in scans
+   * @return data point with maximum intensity in unprocessed data points
```
SteffenHeu, line 86: same

SteffenHeu (changes requested):
adjust the comments then its good

src/main/java/io/github/mzmine/datamodel/RawDataFile.java
robinschmid: NIce catch. 
RawDataFIleImpl had also had javadoc.... but the interfaces also - and this was just copied from the RawDataFileImpl, so it was wrong all along :D
changed it back

src/main/java/io/github/mzmine/datamodel/RawDataFile.java
robinschmid: NIce catch. 
RawDataFIleImpl had also had javadoc.... but the interfaces also - and this was just copied from the RawDataFileImpl, so it was wrong all along :D
changed it back

SteffenHeu (approved):

PR #1078, opened 2023-01-17
Add gc workflow

tnguy271:


robinschmid:
Great thanks for working on this I will go through it now

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
+import io.github.mzmine.main.MZmineCore;
 import io.github.mzmine.modules.tools.batchwizard.WizardPart;
 import io.github.mzmine.modules.tools.batchwizard.WizardPreset.ChromatographyDefaults;
 import io.github.mzmine.parameters.Parameter;
 import io.github.mzmine.parameters.impl.SimpleParameterSet;
-import io.github.mzmine.parameters.parametertypes.BooleanParameter;
-import io.github.mzmine.parameters.parametertypes.ComboParameter;
-import io.github.mzmine.parameters.parametertypes.HiddenParameter;
-import io.github.mzmine.parameters.parametertypes.IntegerParameter;
+import io.github.mzmine.parameters.parametertypes.*;
```
robinschmid, line 34: Please use the MZmine code  style. can be automatically applied in an IDE

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
+      This does not describe the deviation of the accurate mass (measured) from the exact mass (calculated),
+      but the fluctuation of the accurate between two scans.""", 0.002, 10);
+
+  public static final IntegerParameter minimumScanSpan = new IntegerParameter(
+          "Min group size in # of scans", """
+      Minimum scan span over which some feature in the chromatogram must have (continuous) points above the noise level
+      to be recognized as a chromatogram.
+      The optimal value depends on the chromatography system setup. The best way to set this parameter
+      is by studying the raw data and determining what is the typical time span of chromatographic features.""",
+          5, true, 1, null);
```
robinschmid, line 63: This parameter already exists in this class as minNumberOfDataPoints

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
+  /**************************************************ADAP CHROMATOGRAMBUILDER***************************************************************/
+  public static final DoubleParameter minHighestPoint = new DoubleParameter("Min highest intensity",
+          "Points below this intensity will not be considered in starting a new chromatogram",
+          MZmineCore.getConfiguration().getIntensityFormat());
+
+  public static final MZToleranceParameter mzTolerance = new MZToleranceParameter(
+          "Scan to scan accuracy (m/z)", """
+      m/z tolerance of the same compound between two scans.
+      This does not describe the deviation of the accurate mass (measured) from the exact mass (calculated),
+      but the fluctuation of the accurate between two scans.""", 0.002, 10);
```
robinschmid, line 55: these two parameters are defined in WizardMassSpectrometerParameters.
The idea is that the intensity and mz tolerance / accuracy is dependent on the mass spec not on the GC system

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
 import io.github.mzmine.parameters.parametertypes.ranges.RTRangeParameter;
+import io.github.mzmine.parameters.parametertypes.tolerances.MZTolerance;
+import io.github.mzmine.parameters.parametertypes.tolerances.MZToleranceParameter;
 import io.github.mzmine.parameters.parametertypes.tolerances.RTTolerance;
 import io.github.mzmine.parameters.parametertypes.tolerances.RTTolerance.Unit;
 import io.github.mzmine.parameters.parametertypes.tolerances.RTToleranceParameter;
 
+import java.text.NumberFormat;
+
 public class WizardChromatographyParameters extends SimpleParameterSet {
```
robinschmid, line 45: One important aspect of the wizard is that we want to reduce the stress of setting up the general workflows by reducing the number of parameters. Sometimes this can be done by using parameters like the full width at half maximum, easily accessed from a feature table - so the user can actually know. Maybe we can find a way to reuse some of those parameters that we already defined - but it definitely seems like the GC should even get its own parameter set

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
+  public static final IntegerParameter MIN_CLUSTER_SIZE = new IntegerParameter(
+          "Minimum Number of Peaks", "Minimum number of peaks that can form a component", 1);
+
+  /**************************************************ADAP ALIGNER***************************************************************/
+  public static final DoubleParameter SAMPLE_COUNT_RATIO = new DoubleParameter(
+          "Min confidence (between 0 and 1)",
+          "A fraction of the total number of samples. An aligned feature must be detected at "
+                  + "least in several samples.\nThis parameter determines the minimum number of samples where a "
+                  + "feature must be detected.",
+          NumberFormat.getInstance(), 0.7, 0.0, 1.0);
```
robinschmid, line 94: We have something similar in the WizardFilterParameters in minNumberOfSamples
It can be both the absolute number oif samples and the ratio though... 
Need to see if the ADAP aligner works with both - otherwise maybe we could make it work

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
+          NumberFormat.getNumberInstance(), 0.2);
+
+  public static final DoubleParameter RET_TIME_TOLERANCE = new DoubleParameter(
+          "Retention time tolerance (min)",
+          "Retention time tolerance value (between 0 and 1) is used for determine the number of components"
+                  + " in a window. The larger tolerance, the smaller components are determined.",
+          NumberFormat.getNumberInstance(), 0.05, 0.0, Double.MAX_VALUE);
 
+  public static final IntegerParameter MIN_CLUSTER_SIZE = new IntegerParameter(
+          "Minimum Number of Peaks", "Minimum number of peaks that can form a component", 1);
```
robinschmid, line 86: maybe its good to revise this and name it Minimum number of signals? 
And then in the description call it minimum number of signals to form a pseudo spectrum

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
+  /**************************************************MULTICURVE RESOLUTION***************************************************************/
+  public static final DoubleParameter PREF_WINDOW_WIDTH = new DoubleParameter(
+          "Deconvolution window width (min)", "Preferred width of deconvolution windows (in minutes).",
+          NumberFormat.getNumberInstance(), 0.2);
+
+  public static final DoubleParameter RET_TIME_TOLERANCE = new DoubleParameter(
+          "Retention time tolerance (min)",
+          "Retention time tolerance value (between 0 and 1) is used for determine the number of components"
+                  + " in a window. The larger tolerance, the smaller components are determined.",
+          NumberFormat.getNumberInstance(), 0.05, 0.0, Double.MAX_VALUE);
```
robinschmid, line 83: There are 3 RT tolerances below - but given as RT tolerances. But the name here also states RT tolerance in minutes... so that should work
Why does this have to be between 0-1 and then the Double.MAX_VALUE is the actual max?
Maybe you can just use the intraSampleRtTolerance - its meant to be relatively narrow to account for ions generated by the same compound

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
+
+  public static final DoubleRangeParameter RT_FOR_CWT_SCALES_DURATION = new DoubleRangeParameter(
+          "RT wavelet range",
+          "Upper and lower bounds of retention times to be used for setting the wavelet scales. Choose a range that that simmilar to the range of peak widths expected to be found from the data.",
+          MZmineCore.getConfiguration().getRTFormat(), true, true, Range.closed(0.001, 0.1));
+
+  /**************************************************MULTICURVE RESOLUTION***************************************************************/
+  public static final DoubleParameter PREF_WINDOW_WIDTH = new DoubleParameter(
+          "Deconvolution window width (min)", "Preferred width of deconvolution windows (in minutes).",
+          NumberFormat.getNumberInstance(), 0.2);
```
robinschmid, line 77: if its minutes it should be an RT tolerance.
There is the approximateChromatographicFWHM which seems to fit here

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
+  public static final DoubleParameter SAMPLE_COUNT_RATIO = new DoubleParameter(
+          "Min confidence (between 0 and 1)",
+          "A fraction of the total number of samples. An aligned feature must be detected at "
+                  + "least in several samples.\nThis parameter determines the minimum number of samples where a "
+                  + "feature must be detected.",
+          NumberFormat.getInstance(), 0.7, 0.0, 1.0);
+
+  public static final RTToleranceParameter RET_TIME_RANGE = new RTToleranceParameter();
+
+  public static final MZToleranceParameter MZ_RANGE = new MZToleranceParameter();
```
robinschmid, line 98: There are three mz tolerances in WizardMassSpectrometerParamters

src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WizardChromatographyParameters.java
```diff
@@ -26,21 +26,83 @@
...
+
+  public static final DoubleParameter SCORE_TOLERANCE = new DoubleParameter(
+          "Score threshold (between 0 and 1)",
```
robinschmid, line 101: Name it Alignment score threshold 
you can omit the between 0 and 1 because the parameter enforces it with you set min max.
You can point it out in the description though

robinschmid:
Maybe we have to separate the workflows more.  What do you think @SteffenHeu 
GC-EI
LC-MS + GC-CI-MS + IMS 
Imaging-MS + Imaging-IMS-MS
Direct infusion

tnguy271:
@robinschmid I saw your new changes to the wizard, should I only resolve this pr now?

robinschmid:
I think the easiest might be if you pull my branch (https://github.com/robinschmid/mzmine3/tree/wizard_workflows)
and apply the changes directly there. 
There is also a comment from me in which classes you need to work
https://github.com/mzmine/mzmine3/pull/1087

I think its much easier now and I just pushed some last changes to make it flexible but restrict for example the GC-EI workflows to the EI-Spectral deconvolution

robinschmid:
I also sent some more information on Slack. Not sure which platform is best to communicate

robinschmid:
merged as #1143

PR #1080, opened 2023-01-18
add option to keep/remove all intermediate feature lists in batch mode

SteffenHeu:
grafik

### DIFF
--- a/src/main/java/io/github/mzmine/modules/batchmode/BatchComponent.fxml
+++ b/src/main/java/io/github/mzmine/modules/batchmode/BatchComponent.fxml
@@ -118,7 +118,7 @@
       GridPane.rowIndex="4">
       <children>
         <Button fx:id="btnSetHandleIntermediateFlists" mnemonicParsing="false"
-          onAction="#onHandleIntermediateFlistsPressed" text="Set"/>
+          onAction="#onHandleIntermediateFlistsPressed" text="Apply to batch"/>
       </children>
     </FlowPane>
     <FlowPane alignment="CENTER_RIGHT" GridPane.halignment="RIGHT" GridPane.rowIndex="4"
@@ -127,7 +127,7 @@
         <Label alignment="CENTER" text="Intermediate feature lists"/>
       </children>
     </FlowPane>
-    <ComboBox fx:id="cmbHandleFlists" prefWidth="150.0" GridPane.columnIndex="1"
+    <ComboBox fx:id="cmbHandleFlists" GridPane.columnIndex="1"
       GridPane.rowIndex="4">
       <padding>
         <Insets top="5.0"/>
--- a/src/main/java/io/github/mzmine/modules/batchmode/BatchComponent.fxml
+++ b/src/main/java/io/github/mzmine/modules/batchmode/BatchComponent.fxml
@@ -24,15 +24,14 @@
   ~ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
   ~ OTHER DEALINGS IN THE SOFTWARE.
   -->
-<?import io.github.mzmine.parameters.parametertypes.filenames.LastFilesButton?>
 
+<?import io.github.mzmine.parameters.parametertypes.filenames.LastFilesButton?>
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/mzmine/modules/batchmode/BatchComponent.fxml
```diff
@@ -41,80 +40,110 @@
...
+      <Button fx:id="btnRemoveModule" mnemonicParsing="false" onAction="#onRemoveModulePressed"
+        text="&lt;&lt;"/>
+      <GridPane.margin>
+        <Insets/>
+      </GridPane.margin>
+      <padding>
+        <Insets top="5.0"/>
+      </padding>
+    </VBox>
+    <FlowPane alignment="CENTER_LEFT" prefHeight="200.0" prefWidth="200.0" GridPane.columnIndex="2"
```
robinschmid, line 117: need to be defined? or can it just be the calculated size

src/main/java/io/github/mzmine/modules/batchmode/BatchComponent.fxml
```diff
@@ -41,80 +40,110 @@
...
+      GridPane.rowIndex="4">
+      <children>
+        <Button fx:id="btnSetHandleIntermediateFlists" mnemonicParsing="false"
+          onAction="#onHandleIntermediateFlistsPressed" text="Set"/>
```
robinschmid, line 121: Maybe Apply to batch

src/main/java/io/github/mzmine/modules/batchmode/BatchComponent.fxml
```diff
@@ -41,80 +40,110 @@
...
+      </children>
+    </FlowPane>
+    <FlowPane alignment="CENTER_RIGHT" GridPane.halignment="RIGHT" GridPane.rowIndex="4"
+      GridPane.valignment="CENTER">
+      <children>
+        <Label alignment="CENTER" text="Intermediate feature lists"/>
+      </children>
+    </FlowPane>
+    <ComboBox fx:id="cmbHandleFlists" prefWidth="150.0" GridPane.columnIndex="1"
```
robinschmid, line 130: maybe remove prefWidth - should be calculated

src/main/java/io/github/mzmine/modules/batchmode/BatchComponent.fxml
SteffenHeu: that's just how it comes out of the scene builder

src/main/java/io/github/mzmine/modules/batchmode/BatchComponentController.java
```diff
@@ -501,4 +511,31 @@ private TreeItem<Object> cloneTreeItem(TreeItem<Object> item) {
...
+  void onHandleIntermediateFlistsPressed() {
+    final OriginalFeatureListOption option = cmbHandleFlists.getValue();
+    setHandleOriginalFeatureLists(option);
+    var str = switch (option) {
+      case KEEP -> "keep intermediate feature lists.";
+      case REMOVE -> "remove intermediate feature lists.";
+      case PROCESS_IN_PLACE -> "process on feature lists if possible and remove otherwise.";
+    };
+    MZmineCore.getDesktop()
+        .displayMessage("Batch mode updated", "Updated all batch steps to " + str);
```
robinschmid, line 525: currently this does not log the message. Maybe we should in general also log those messages?
Or just log it here as well

robinschmid (approved):
Finally... was waiting for this feature for so long

SteffenHeu:
> Each module already has its own parameter "keep previous feature list?", doesn't it? Doesn't this make it redundant?

yes, but when you have a huge batch you dont want to update all options individually. E.g. when coming from an optimisation, you usually have all options on "KEEP". Now you can set all to REMOVE

SteffenHeu:
It just updates.

robinschmid:
Makes total sense.... This replaces all the values in all modules. We changed it to a specific parameter type so that we can do this.
The idea is that during optimization you want to keep all lists and then switch to the optimized version that processes in place or removes old lists

PR #1086, opened 2023-01-20
Let parameter components grow in column and rows

robinschmid:
Parameters can define if the component grows


src/main/java/io/github/mzmine/parameters/dialogs/ParameterSetupPane.java
```diff
@@ -246,6 +245,7 @@ protected void callCancelButton() {
   @NotNull
   public GridPane createParameterPane(@NotNull Parameter<?>[] parameters) {
     GridPane paramsPane = new GridPane();
+    paramsPane.setPadding(new Insets(5));
```
SteffenHeu, line 248: nice i also wanted to do this

PR #1093, opened 2023-01-23
Save mzmine version to batch file and log if versions mismatch

robinschmid:
Show dialog when importing a batch to the batch dialog

image

Added an attribute to the batch element
<?xml version="1.0" encoding="UTF-8"?><batch mzmine_version="3.3.19">

### DIFF
--- a/src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
+++ b/src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
@@ -85,12 +85,12 @@ public class BatchQueue extends ArrayObservableList<MZmineProcessingStep<MZmineP
 
       String vstring = switch (version.compareTo(MZmineCore.getMZmineVersion())) {
         case -1 ->
-            "an older version of MZmine ({0}). Make sure all parameters are the same and set potential new parameters.";
-        case 0 -> "the same version of MZmine ({0}).";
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
```diff
@@ -52,6 +54,23 @@ public class BatchQueue extends ArrayObservableList<MZmineProcessingStep<MZmineP
 
   // Method element name.
   private static final String METHOD_ELEMENT = "method";
+  // attr of the main xmlElement
+  public static final String XML_MZMINE_VERSION_ATTR = "mzmine_version";
+
+  // added in MZmine 3.4.0 - null before that
+  @Nullable
+  private final Semver mzmineVersionCreatedIn;
```
SteffenHeu, line 62: What happens if the user corrects the values (if there are changes)? There's no way of updating the value to the current version. the error will be there forever.

robinschmid: Oh if you import and just dismiss the warning and save the batch again the batch is in the newest version

robinschmid: in saving its always saving the latest version.

robinschmid:
Closed in favor of #1094

PR #1097, opened 2023-01-24
Rewrite sirius export

SteffenHeu:
started to refactor the sirius export to use the export we have for spectral libraries. added some dbentries and made the structure reuse code we have to avoid duplication.

not done and tested yet

### DIFF
--- a/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
+++ b/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
@@ -49,6 +49,7 @@ import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsParamete
 import io.github.mzmine.parameters.parametertypes.submodules.OptionalModuleParameter;
 import io.github.mzmine.parameters.parametertypes.tolerances.MZToleranceParameter;
 import io.github.mzmine.util.ExitCode;
+import java.util.Collection;
 import java.util.List;
 import javafx.stage.FileChooser.ExtensionFilter;
 import org.jetbrains.annotations.NotNull;
@@ -60,37 +61,37 @@ public class SiriusExportParameters extends SimpleParameterSet {
   );
 
 
-  public static final OptionalModuleParameter<MsMsSpectraMergeParameters> MERGE_PARAMETER =
-      new OptionalModuleParameter<>("Merge MS/MS",
-          "Merge high qualitative MS/MS into one spectrum instead of exporting all MS/MS separately.",
-          new MsMsSpectraMergeParameters(), true);
+  public static final OptionalModuleParameter<MsMsSpectraMergeParameters> MERGE_PARAMETER = new OptionalModuleParameter<>(
+      "Merge MS/MS",
+      "Merge high qualitative MS/MS into one spectrum instead of exporting all MS/MS separately.",
+      new MsMsSpectraMergeParameters(), true);
   /**
    * MZTolerance to exclude duplicates in correlated spectrum
    */
   public static final MZToleranceParameter MZ_TOL = new MZToleranceParameter("m/z tolerance",
       "m/z tolerance to exclude duplicates in correlated spectrum", 0.001, 5);
-  public static final BooleanParameter RENUMBER_ID =
-      new BooleanParameter("Renumber IDs", "Resets the IDs (uses the row ID otherwise)", false);
+  public static final BooleanParameter RENUMBER_ID = new BooleanParameter("Renumber IDs",
+      "Resets the IDs (uses the row ID otherwise)", false);
 
-  public static final BooleanParameter NEED_ANNOTATION =
-      new BooleanParameter("Only rows with annotation",
-          "Only export rows with an annotation (run MS annotate or metaMSEcorrelate)", false);
+  public static final BooleanParameter NEED_ANNOTATION = new BooleanParameter(
+      "Only rows with annotation",
+      "Only export rows with an annotation (run MS annotate or metaMSEcorrelate)", false);
   public static final BooleanParameter EXCLUDE_EMPTY_MSMS = new BooleanParameter(
       "Exclude empty MS/MS spectra",
       "Do not export empty MS/MS spectra (only features with MS/MS spectrum with at least 1 signal are exported)",
       false);
-  public static final BooleanParameter EXCLUDE_MULTICHARGE =
-      new BooleanParameter("Exclude multiple charge", "Do not export multiply charged rows", false);
+  public static final BooleanParameter EXCLUDE_MULTICHARGE = new BooleanParameter(
+      "Exclude multiple charge", "Do not export multiply charged rows", false);
   public static final BooleanParameter EXCLUDE_MULTIMERS = new BooleanParameter("Exclude multimers",
       "Do not export rows that were annotated as multimers (2M) (run MS annotate or metaMSEcorrelate)",
       false);
   public static final FeatureListsParameter FEATURE_LISTS = new FeatureListsParameter();
   public static final FileNameParameter FILENAME = new FileNameParameter("Filename",
       "Name of the output MGF file. "
-      + "Use pattern \"{}\" in the file name to substitute with feature list name. "
-      + "(i.e. \"blah{}blah.mgf\" would become \"blahSourceFeatureListNameblah.mgf\"). "
-      + "If the file already exists, it will be overwritten.",
-      extensions, FileSelectionType.SAVE);
+          + "Use pattern \"{}\" in the file name to substitute with feature list name. "
+          + "(i.e. \"blah{}blah.mgf\" would become \"blahSourceFeatureListNameblah.mgf\"). "
+          + "If the file already exists, it will be overwritten.", extensions,
+      FileSelectionType.SAVE);
 
   public SiriusExportParameters() {
     super(new Parameter[]{FEATURE_LISTS, FILENAME, MERGE_PARAMETER, MZ_TOL, RENUMBER_ID,
@@ -111,14 +112,28 @@ public class SiriusExportParameters extends SimpleParameterSet {
   @Override
   public ExitCode showSetupDialog(boolean valueCheckRequired) {
     String message = "<html>SIRIUS Module Disclaimer:" + "<ul>"
-                     + "<li>If you use the SIRIUS export module, cite <a href=\"https://bmcbioinformatics.biomedcentral.com/articles/10.1186/1471-2105-11-395\">MZmine2 paper</a> and the following article:<br>"
-                     + "<a href=\"http://dx.doi.org/10.1038/s41592-019-0344-8\">K. Duhrkop, et al., Sirius 4: a rapid tool for turning tandem mass spectra into metabolite structure information, Nature Methods, 2019.</a>"
-                     + "<li>Sirius can be downloaded at the following address: <a href=\"https://bio.informatik.uni-jena.de/software/sirius/\">https://bio.informatik.uni-jena.de/software/sirius/</a>"
-                     + "<li>Sirius results can be mapped into <a href=\"http://gnps.ucsd.edu/\">GNPS</a> molecular networks. <a href=\"https://bix-lab.ucsd.edu/display/Public/Mass+spectrometry+data+pre-processing+for+GNPS\">See the documentation</a>."
-                     + "</ul>";
+        + "<li>If you use the SIRIUS export module, cite <a href=\"https://bmcbioinformatics.biomedcentral.com/articles/10.1186/1471-2105-11-395\">MZmine2 paper</a> and the following article:<br>"
+        + "<a href=\"http://dx.doi.org/10.1038/s41592-019-0344-8\">K. Duhrkop, et al., Sirius 4: a rapid tool for turning tandem mass spectra into metabolite structure information, Nature Methods, 2019.</a>"
+        + "<li>Sirius can be downloaded at the following address: <a href=\"https://bio.informatik.uni-jena.de/software/sirius/\">https://bio.informatik.uni-jena.de/software/sirius/</a>"
+        + "<li>Sirius results can be mapped into <a href=\"http://gnps.ucsd.edu/\">GNPS</a> molecular networks. <a href=\"https://bix-lab.ucsd.edu/display/Public/Mass+spectrometry+data+pre-processing+for+GNPS\">See the documentation</a>."
+        + "</ul>";
     ParameterSetupDialog dialog = new ParameterSetupDialog(valueCheckRequired, this, message);
     dialog.showAndWait();
     return dialog.getExitCode();
+  }
+
+  @Override
+  public boolean checkParameterValues(Collection<String> errorMessages) {
+    boolean superCheck = super.checkParameterValues(errorMessages);
+
+    if (getValue(FEATURE_LISTS).getMatchingFeatureLists().length > 1 && !getValue(
+        FILENAME).getName().contains(SiriusExportTaskNew.plNamePattern)) {
+      errorMessages.add("More than one feature list selected but \"{}\" pattern not found in name."
+          + "Please add the name pattern to create individual files.");
+      superCheck = false;
+    }
+
+    return superCheck;
   }
 
   @Override
--- a/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
+++ b/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
@@ -1,0 +1,290 @@
+/*
+ * Copyright (c) 2004-2022 The MZmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
+ * conditions:
+ *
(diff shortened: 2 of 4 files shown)


src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
```diff
@@ -111,16 +112,30 @@ public SiriusExportParameters() {
...
     dialog.showAndWait();
     return dialog.getExitCode();
   }
 
+  @Override
+  public boolean checkParameterValues(Collection<String> errorMessages) {
+    boolean superCheck = super.checkParameterValues(errorMessages);
+
+    if (getValue(FEATURE_LISTS).getMatchingFeatureLists().length > 1 && !getValue(
+        FILENAME).getName().contains(SiriusExportTaskNew.plNamePattern)) {
```
robinschmid, line 130: when exposing this pattern variable, make it static final and all upper case to signal its final. Give it a meaningful name. MULTI_NAME_PATTERN or similar.

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
```diff
@@ -111,16 +112,30 @@ public SiriusExportParameters() {
...
+      errorMessages.add("More than one feature list selected but \"{}\" pattern not found in name."
```
robinschmid, line 131: you could use the NAME PATTERN var here and just add it into the string instead of {} just in case it would change at one point

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class SiriusExportTaskNew extends AbstractTask {
```
robinschmid, line 72: Maybe copy all code over to the old Sirius export classes to keep the git history? optional

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+  private final MsMsSpectraMergeParameters mergeParameters;
+  private final double minimumRelativeNumberOfScans;
+  private final MZTolerance mzTol;
+  private final Boolean excludeEmptyMSMS;
+  private final Boolean excludeMultiCharge;
+  private final Boolean excludeMultimers;
+  private final Boolean needAnnotation;
+  private final Boolean renumberID;
+  private final int totalRows;
+  public static final String plNamePattern = "{}";
```
robinschmid, line 89: code format should push this to the top right?

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+  final NumberFormats format = MZmineCore.getConfiguration().getExportFormats();
```
robinschmid, line 90: private

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+  private final MergeMode mergeMode;
+  private int exportedRows = 0;
+
+
+  protected SiriusExportTaskNew(ParameterSet parameters, @NotNull Instant moduleCallDate) {
+    super(null, moduleCallDate);
+
+    this.parameters = parameters;
+    this.featureLists = parameters.getParameter(SiriusExportParameters.FEATURE_LISTS).getValue()
```
robinschmid, line 99: just because its a new module and you are already changing it all...
Could you change to all the convenience methods like getValue

in the wizard PR I actually added a short hand for OptionalParameter. Maybe we could add this also as a default to the parameter set. its something like: 

```java
OptionalParameter<T> op = params.getParameter(parameter);
return new OptionalValue<>(op.getValue(), op.getEmbeddedParameter().getValue());
```
This eliminates the need to use the static parameters that we always use to access the values twice.

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+  @Override
+  public void run() {
+    setStatus(TaskStatus.PROCESSING);
+
+    // Shall export several files?
+    boolean substitute = fileName.getPath().contains(plNamePattern);
+
+    int totalExported = 0;
+    // Process feature lists
+    for (FeatureList featureList : featureLists) {
```
robinschmid, line 143: add isCancelled check in this loop

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+
+  private int exportFeatureList(FeatureList featureList, BufferedWriter writer) {
+
+    for (FeatureListRow row : featureList.getRows()) {
+      if (isCanceled()) {
+        return exportedRows;
+      }
+
+      if (!checkFeatureCriteria(row)) {
+        exportedRows++;
```
robinschmid, line 180: its not exported if the criteria is not met...
So maybe increment filtered rows or similar

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+        continue;
+      }
+
+      List<SpectralLibraryEntry> entries = new ArrayList<>();
```
robinschmid, line 184: Add a comment that we are reusing the entries as an easy way to form mgf entries

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+
+      // export best MS1
+      final MassList massList = Objects.requireNonNull(row.getBestFeature().getRepresentativeScan())
+          .getMassList();
+      if (massList == null) {
+        throw new MissingMassListException(row.getBestFeature().getRepresentativeScan());
+      }
+      spectrumToEntry(MsType.MS, massList, row.getBestFeature());
```
robinschmid: you are passing in massList here but are not handling it in the pattern matching switch.
I think its fair to pass in the scan because you are automatically extracting the masslist there

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+            .toList();
+        entries.addAll(ms2Entries);
+      }
+
+      final String fileContent = entries.stream().map(MGFEntryGenerator::createMGFEntry)
+          .collect(Collectors.joining("\n\n"));
+
+      try {
+        writer.write(fileContent);
+      } catch (IOException e) {
```
robinschmid, line 208: You are catching the IOException in the caller... so no need to catch it here. Just add to method declaration

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,404 @@
...
+      case MSMS -> entry.putIfNotNull(DBEntryField.MS_LEVEL, 2);
+    }
+
+    if (spectrum instanceof MergedSpectrum spec) {
+      putMergedSpectrumFieldsIntoEntry(spec, entry);
+    }
+
+    return entry;
+  }
+
```
robinschmid, line 247: maybe put this into a constructor of SpectralLibraryEntry that uses the MergedSpectrum?

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,404 @@
...
+    PolarityType polarity = f.getRepresentativeScan().getPolarity();
+    if (f.getRow().getRowCharge() != null) {
+      charge = f.getRow().getRowCharge();
+    } else if (f.getMostIntenseFragmentScan().getMsMsInfo() instanceof DDAMsMsInfo dda) {
+      charge = dda.getPrecursorCharge() != null ? dda.getPrecursorCharge() : charge;
+    }
+
+    entry.putIfNotNull(DBEntryField.FEATURE_ID, f.getRow().getID());
+    entry.putIfNotNull(DBEntryField.PRECURSOR_MZ, f.getMZ());
+    entry.putIfNotNull(DBEntryField.CHARGE,
```
robinschmid, line 269: Maybe it would be nice to have some convenience method for this as we need the charge in many different places.

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+      FeatureListRow row) {
+
+    List<SpectralLibraryEntry> entries = new ArrayList<>();
+    final MsMsSpectraMergeModule merger = MZmineCore.getModuleInstance(
+        MsMsSpectraMergeModule.class);
+
+    switch (mergeMode) {
+      case SAME_SAMPLE -> {
+        for (Feature f : row.getFeatures()) {
+          if (f.getFeatureStatus() == FeatureStatus.DETECTED) {
```
robinschmid, line 285: maybe clean up and do (same in CONSECUTIVE_SCANS case)
if != DETECTED) continue;

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    if (excludeMultimers && adduct != null && adduct.getIonType().getMolecules() > 1) {
+      return false;
+    }
+
+    return true;
+  }
+
+  private boolean checkMassList(Scan scan, MassList massList) {
```
robinschmid, line 361: can be removed and just throw MissingMassListExceptions everywhere

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,404 @@
...
+
+  private boolean checkFeatureCriteria(final FeatureListRow row) {
+
+    if (!row.hasMs2Fragmentation()) {
+      return false;
+    }
+
+    if (excludeMultiCharge && row.getRowCharge() > 1) {
+      return false;
+    }
```
robinschmid, line 347: Just do one return statement:
```java
return row.haseMs2Fragmentation() 
// SIRIUS does not support multi charge
&& !(excludeMultiCharge && row.getRowCharge() > 1)
// handle adducts and multimers
&& !(needAnnotation && adduct == null)
&& !(excludeMultimers && adduct != null && adduct.getIonType().getMolecules() > 1);
```

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,404 @@
...
+            if (bestMS2 == null) {
+              continue;
+            }
+            if (!checkMassList(bestMS2, bestMS2.getMassList())) {
+              return null;
+            }
+            if (excludeEmptyMSMS && bestMS2.getMassList().getNumberOfDataPoints() <= 0) {
+              continue;
+            }
+
```
robinschmid, line 296: as mentioned below, remove method and throw MissingMassListException here

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
```diff
@@ -0,0 +1,408 @@
...
+            }
+          }
+        }
+      }
+
+      case ACROSS_SAMPLES -> {
+        // merge everything into one
+        MergedSpectrum spectrum = merger.mergeAcrossSamples(mergeParameters, row)
+            .filterByRelativeNumberOfScans(minimumRelativeNumberOfScans);
+        entries.add(spectrumToEntry(MsType.MSMS, spectrum, row.getBestFeature()));
```
robinschmid, line 332: check for empty spectrum here

src/main/java/io/github/mzmine/util/spectraldb/entry/DBEntryField.java
```diff
@@ -340,6 +350,11 @@ public String getMgfID() {
       case DATAFILE_COLON_SCAN_NUMBER -> "DATAFILE_SCANNUMBER";
       case QUALITY_CHIMERIC -> "QUALITY_CHIMERIC";
       case DATASET_ID -> "DATASET_ID";
+      case FEATURE_ID -> "FEATURE_ID";
+      case SIRIUS_SPEC_TYPE -> "SPECTYPE";
```
robinschmid, line 354: maybe we could just use  the MERGED_SPEC_TYPE
and change the mgf string to SPECTYPE.
This is a custom field that i created

robinschmid (commented):
I like the new Export. Very clean

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
SteffenHeu: maybe when im done

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
SteffenHeu: dont know why it doesn't

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
SteffenHeu: the merged spectrum is only used here for now and not really intended to be used for anything else. I think this is fine to not encourage the usage anywhere else.

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
SteffenHeu: edited ModularFeature.getFeatures() to only return detected features.

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
SteffenHeu: no that's afwul to read.

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
robinschmid: ok

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
robinschmid: although its only doing it if you activate reorganize.
image

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
robinschmid: Also in GNPS export... but true

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
robinschmid: Why? 
its like 6 lines vs 15 or so... 
I thought its quite nice compared to multiple return false statements. 
But your call

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
robinschmid: just meant to rather do a gatekeeper if here with continue instead of wrapping it into another level of {}.
Just wanted to be the NO guy... but getFeatures now gives you all not UNKOWN. Thats good so detected and gap filled is considered.
I think its also good to consider gap filled here.

robinschmid:
Changes look good.

src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTaskNew.java
SteffenHeu: Less code does not mean it's easier to understand, especially if every statement consists of multiple conditions on its own.

SteffenHeu:
@robinschmid The getFeatures() Method in the row was used in the bindings. This lead to a change because the status was not set in the test. 

So before also values of features that were not set to != UNKNOWN before were used for bindings. I think this is the correct solution but we have to watch out for modules that did not do it correctly, because they wont be used anymore to evaluate bindings,.

PR #1100, opened 2023-01-25
featureML export module

christophuv:
Version 1 of an export module for visualizing the results of a feature list. The feature list is exported to the featureML file format, which can be loaded into TOPPView. Each detected feature is illustrated as a box in TOPPView and on highlight different sub-boxes are shown for the features of the different samples. Feature consensus areas can easily be inspected allowing to optimize the feature detection step with MzMine graphically towards a high number of features. Furthermore, users can visually identify regions in there chromatograms with peaks that have not been found with their current settings.


src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+    double maxMZ = 0.;
+    double meanMZ = 0.;
+    
+    double x = 0.;
+    
+    for(FeatureListRow row : rows) {
+    	
+    	// find earliest and latest RTs, no function available (or I did not find it)
+    	
+    	minRT = 1000000.;
```
SteffenHeu, line 222: The indentation is 4 lines, we are using two. you can use the formatter we supply with the project (mzmine-intellij-code-formater.xml). See here how to set it up: https://mzmine.github.io/mzmine_documentation/contribute_intellij.html#code-formatter

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+    for(FeatureListRow row : rows) {
+    	
+    	// find earliest and latest RTs, no function available (or I did not find it)
+    	
+    	minRT = 1000000.; 
+    	maxRT = 0.;
+    	
+    	for (RawDataFile rawFile : rawDataFiles) {
+        	
+        	if(row.getFeature(rawFile) != null) {
```
SteffenHeu, line 227: you basically want to use the widest range of all detected features here, right? I think this should be equivalent to the rt range of the row.

I noticed we somehow don't have a convenience method for that, but you can use
`row.get(RTRangeType.class)`

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+        writer.write(String.format("          <pt x='%f' y='%f' />", minRT, maxMZ));
+        writer.newLine();
+        writer.write(String.format("          <pt x='%f' y='%f' />", maxRT, maxMZ));
+        writer.newLine();
+        writer.write(String.format("          <pt x='%f' y='%f' />", maxRT, minMZ));
+        writer.newLine();
+        writer.write(String.format("        </convexhull>"));
+        writer.newLine();
+
+        int hulNum = 1;
```
SteffenHeu, line 272: what does hulNum mean? best write out the full name

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+    	
+    	minMZ = row.getMZRange().lowerEndpoint();
+    	maxMZ = row.getMZRange().upperEndpoint();
+    	meanMZ = row.getAverageMZ();
+    	
+    	minRT = minRT * 60;
+    	maxRT = maxRT * 60;
+    	meanRT = row.getAverageRT() * 60;
+
+        writer.write(String.format("      <feature id='%d'>", row.getID()));
```
SteffenHeu, line 243: since this is an xml based format, did you intentionally write it with a normal file writer? It's fine with me, but i feel like it would be way easier to maintain if it used an xml writer sich as the `javax.xml.stream.XMLStreamWriter;`.

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+      } catch (IOException e) {
+        setStatus(TaskStatus.ERROR);
+        setErrorMessage("Could not open file " + curFile + " for writing.");
+        logger.log(Level.WARNING, String.format(
+            "Error writing featureML format to file: %s for feature list: %s. Message: %s",
+            curFile.getAbsolutePath(), featureList.getName(), e.getMessage()), e);
+        return;
+      }
+
+      checkConcurrentModification(featureList, numRows, numFeatures, numMS2);
```
SteffenHeu, line 172: This is a relict from the GNPS export which exports multiple files for the same feature list. i don't think it's needed here.

SteffenHeu (changes requested):
Thanks, i think this is a great addition. I added some comments

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularParameters.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class FeatureMLExportModularParameters extends SimpleParameterSet {
+
+  public static final FeatureListsParameter featureLists = new FeatureListsParameter(1);
+  public static final ComboParameter<FeatureListRowsFilter> filter = new ComboParameter<>(
+      "Filter rows", "Limit the exported rows to those with MS/MS data (or annotated rows)",
+      FeatureListRowsFilter.values(), FeatureListRowsFilter.ALL);
+  private static final List<ExtensionFilter> extensions = List.of( //
+      new ExtensionFilter("featureML", "*.featureML")
+  );
```
robinschmid, line 48: Looks like python code. Please setup the code formatter to automatically apply after each save / commit. One of the reasons is that otherwise later those lines are automatically changed in the next edid and clutter the change list in PRs.
https://mzmine.github.io/mzmine_documentation/contribute_intellij.html#code-formatter

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularParameters.java
```diff
@@ -0,0 +1,61 @@
...
+import io.github.mzmine.parameters.parametertypes.StringParameter;
+import io.github.mzmine.parameters.parametertypes.filenames.FileNameParameter;
+import io.github.mzmine.parameters.parametertypes.filenames.FileSelectionType;
+import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsParameter;
+import java.util.List;
+import javafx.stage.FileChooser.ExtensionFilter;
+
+public class FeatureMLExportModularParameters extends SimpleParameterSet {
+
+  public static final FeatureListsParameter featureLists = new FeatureListsParameter(1);
```
robinschmid, line 42: You can remove the 1 and allow any number of feature lists. At least in the filename parameter you are allowing the {} pattern to be replaced with the feature list name. In the export you can then write one file per feature list.

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class FeatureMLExportModularTask extends AbstractTask implements ProcessedItemsCounter {
+
+  public static final String DATAFILE_PREFIX = "datafile";
```
robinschmid, line 71: unused?

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+  private static final Logger logger = Logger.getLogger(FeatureMLExportModularTask.class.getName());
+  private final ModularFeatureList[] featureLists;
+  // parameter values
+  private final File fileName;
+  private final String headerSeparator = ":";
```
robinschmid, line 76: unused

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
robinschmid: can be removed

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+  /**
+   * @return true if type should be exported
+   */
+  public boolean filterType(DataType type) {
+    return !(type instanceof NoTextColumn || type instanceof NullColumnType
+        || type instanceof LinkedGraphicalType);
+  }
+  
+
+  private void checkConcurrentModification(FeatureList featureList, int numRows, long numFeatures,
```
robinschmid, line 321: this method can be deleted

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+      throws IOException {
+    final List<FeatureListRow> rows = flist.getRows().stream().filter(rowFilter::accept)
+        .sorted(FeatureListRowSorter.DEFAULT_ID).toList();
+    List<RawDataFile> rawDataFiles = flist.getRawDataFiles();
+
+    List<DataType> rowTypes = flist.getRowTypes().values().stream().filter(this::filterType)
+        .collect(Collectors.toList());
+
+    // Write feature row headers
+    writer.write(String.format("<?xml version='1.0' encoding='ISO-8859-1'?>"));
```
robinschmid, line 196: can you try if UTF-8 works in TOPPView etc.
UTF-8 is a better standard and more widely used

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+    writer.newLine();
+    writer.write(String.format("  <featureMap version='1.4' id='fm_16311276685788915066' xsi:noNamespaceSchemaLocation='http://open-ms.sourceforge.net/schemas/FeatureXML_1_4.xsd' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'>"));
+    writer.newLine();
+    writer.write(String.format("    <dataProcessing completion_time='%s'>", new SimpleDateFormat("yyyy.MM.dd.HH.mm.ss").format(new java.util.Date())));
+    writer.newLine();
+    writer.write(String.format("      <software name='MzMine3' version='%s' />", "TODO add version here"));
```
robinschmid, line 202: MZmine 3
MZmineCore.getMZmineVersion().toString()

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+    writer.newLine();
+    writer.write(String.format("    </dataProcessing>"));
+    writer.newLine();
+    writer.write(String.format("    <featureList count='%d'>", rowTypes.size()));
```
robinschmid, line 206: what does this parameter refer to?
rowTypes is the number of columns

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+    writer.newLine();
+    
+    double minRT = 100000.;
```
robinschmid, line 209: for those things its best to just use Double.MAX_VALUE or POSITIVE_INFINITY, similar for minimum use NEGATIVE_INFITINY
But steffen already mentioned there is a RTRangeType column that gives you this information

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
robinschmid: Second option would be something like this:

```java
// Create the doc.
final Document doc = DocumentBuilderFactory.newInstance().newDocumentBuilder().newDocument();
final Element fmap = createElement(doc, "featureMap",// 
    "version", "1.4",// 
    "id", "fm_16311276685788915066", // 
    "xsi:noNamespaceSchemaLocation", "http://open-ms.sourceforge.net/schemas/FeatureXML_1_4.xsd",//
    "xmlns:xsi", "http://www.w3.org/2001/XMLSchema-instance");
    doc.appendChild(fmap);

    Element processing = createElement(doc, "dataProcessing", "completion_time", date);
    fmap.appendChild(processing);

    processing.appendChild(createElement(doc, "software", "name", "MZmine", MZmineCore.getMZmineVersion()))

// Create transformer.
final Transformer transformer = TransformerFactory.newInstance().newTransformer();
    transformer.setOutputProperty(OutputKeys.METHOD, "xml");
    transformer.setOutputProperty(OutputKeys.INDENT, "yes");
    transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-8");
    transformer.setOutputProperty("{http://xml.apache.org/xslt}indent-amount", "4");

    // Write to file and transform.
    transformer.transform(new DOMSource(doc), new StreamResult(new FileOutputStream(file)));
```

And here the method to create the elements with attributes
```java 
private Element createElement(final Document doc, final String id,
      final Object... attributeValuePairs) {
    Element element = doc.createElement(id);
    for (int i = 0; i < attributeValuePairs.length; i += 2) {
      Object key = attributeValuePairs[i];
      Object value = attributeValuePairs[i + 1];
      element.setAttribute(key.toString(), Objects.requireNonNullElse(value.toString(), ""));
    }
    return element;
  }
```

src/main/java/io/github/mzmine/modules/io/export_features_featureML/TableFormat.java
```diff
@@ -0,0 +1,42 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.io.export_features_featureML;
+
+/**
+ * @author Robin Schmid (https://github.com/robinschmid)
+ */
+public enum TableFormat {
```
robinschmid, line 31: You are not using this class right?
And its just copied

robinschmid (commented):
Looks very good. 
Just a few comments to make it better but it seems that everything is in place so far.

Steffen and I started to look through it at the same time so there were some duplicate comemnts, sorry for that.

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
SteffenHeu: For files that can get really big the DOC based xml readers/writers are not good because they map the whole file into ram when reading writing. The Stream readers/writers are better suited for that.

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
SteffenHeu: probably featureList.getNumberOfRows()

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,343 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class FeatureMLExportModularTask extends AbstractTask implements ProcessedItemsCounter {
```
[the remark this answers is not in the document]
christophuv: Thank you, did that

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, did that and reformatted the code in the last commit

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, that was what I was looking for

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, renamed the variable

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, changed to that writer

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, did that

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularParameters.java
christophuv: Thank you, did that and formatted the code in the last commit

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularParameters.java
christophuv: Thank you, did that

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, removed it

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, removed the separator

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, did that

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, good idea

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, that was what I was looking for

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, that was a bug, replaced by the number of rows

christophuv: sorry, the number of files the feature was detected in

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, it is no longer necessary as the range is now derived from a function

src/main/java/io/github/mzmine/modules/io/export_features_featureML/TableFormat.java
christophuv: Thank you, did that

christophuv:
Thanks for your comments and as well for your thorough look into my code. 
I addressed them and updated the code. Please find the new commits here

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
robinschmid: True but they are also harder to do right?
I am fine with both.

robinschmid:
FYI I  activated the GitHub actions that run the builds and tests on all three major platforms Windows, macOS, Ubuntu
An easy way to see if everything runs smoothly

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularParameters.java
```diff
@@ -0,0 +1,53 @@
...
+import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsParameter;
+
+import java.util.List;
+
+import javafx.stage.FileChooser.ExtensionFilter;
+
+public class FeatureMLExportModularParameters extends SimpleParameterSet {
+
+    public static final FeatureListsParameter featureLists = new FeatureListsParameter();
+    public static final ComboParameter<FeatureListRowsFilter> filter = new ComboParameter<>("Filter rows", "Limit the exported rows to those with MS/MS data (or annotated rows)", FeatureListRowsFilter.values(), FeatureListRowsFilter.ALL);
```
robinschmid, line 43: Did you apply the code formatter to all files?
its in the main directory
mzmine-intellij-code-formater.xml

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,337 @@
...
+import java.util.regex.Pattern;
+import javax.xml.stream.XMLOutputFactory;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamWriter;
+
+import org.jetbrains.annotations.NotNull;
+
+// Export results to featureML format for visualization in TOPPView
+// schema available at
+// https://github.com/OpenMS/OpenMS/blob/7a2e4a41d4c9f511306afcb8bb4f1b773ace9b9a/share/OpenMS/SCHEMAS/FeatureXML_1_9.xsd
```
robinschmid, line 67: Thats okay like this. You could also use JavaDoc here 
/**
*
*/

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,337 @@
...
+                String newFilename = fileName.getPath().replaceAll(Pattern.quote(plNamePattern), cleanPlName);
+                curFile = new File(newFilename);
+            }
+            curFile = FileAndPathUtil.getRealFilePath(curFile, "featureML");
+
+            // Open file
+
+            try {
+
+                BufferedWriter writer = Files.newBufferedWriter(curFile.toPath(), StandardCharsets.UTF_8);
```
robinschmid, line 152: Best way is to do a try with resources. BufferedWriter is auto closable and this way you are safe that the locks on the files are released. Previously this was done with a try-catch-finally block where you do writer.close in the finally - thats unsafe and might in some cases not be closed
```java
try(BufferedWriter writer = Files.newBufferedWriter(new File("").toPath(), StandardCharsets.UTF_8)) {
```

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,337 @@
...
+                // Create XMLStreamWriter object from xmlOutputFactory.
+                XMLStreamWriter xmlStreamWriter = new IndentingXMLStreamWriter(xmlOutputFactory.createXMLStreamWriter(writer));
+
+                exportFeatureList(featureList, xmlStreamWriter);
+
+                xmlStreamWriter.flush();
+                xmlStreamWriter.close();
+
+                writer.flush();
+                writer.close();
```
robinschmid, line 165: remove the writer.close() when using try with resources

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,337 @@
...
+    private void exportFeatureList(ModularFeatureList flist, XMLStreamWriter xmlWriter) throws IOException, XMLStreamException {
+        final List<FeatureListRow> rows = flist.getRows().stream().filter(rowFilter::accept).sorted(FeatureListRowSorter.DEFAULT_ID).toList();
+        List<RawDataFile> rawDataFiles = flist.getRawDataFiles();
+
+        double minRT = 100000.;
+        double maxRT = 0.;
+        double meanRT = 0.;
+        double minMZ = 0.;
+        double maxMZ = 0.;
+        double meanMZ = 0.;
```
robinschmid, line 200: you could inline those values now. 
You are setting them in the for loop. 
Optional

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,337 @@
...
+
+        // write featureML header
+        xmlWriter.writeStartDocument();
+
+        // featureMap
+        this.generateElement(xmlWriter, "featureMap", new String[]{"version", "1.4", "id", "fm_16311276685788915066", "xsi:noNamespaceSchemaLocation", "http://open-ms.sourceforge.net/schemas/FeatureXML_1_4.xsd", "xmlns:xsi", "http://www.w3.org/2001/XMLSchema-instance"});
```
robinschmid, line 206: apply code formatter

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,337 @@
...
+
+        // dataProcessing
+        this.generateElement(xmlWriter, "dataProcessing", new String[]{"completion_time", new SimpleDateFormat("yyyy.MM.dd.HH.mm.ss").format(new java.util.Date())});
+
+        // software
+        this.generateEmptyElement(xmlWriter, "software", new String[]{"name", "MzMine3", "version", MZmineCore.getMZmineVersion().toString()});
```
robinschmid, line 212: Maybe use as tool name MZmine     without 3 and MZ upper   mine lower case

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
```diff
@@ -0,0 +1,337 @@
...
+        xmlWriter.writeEndElement();
+    }
+
+
+    /**
+     * @return true if type should be exported
+     */
+    public boolean filterType(DataType type) {
+        return !(type instanceof NoTextColumn || type instanceof NullColumnType || type instanceof LinkedGraphicalType);
+    }
```
robinschmid, line 336: This one can be removed. it seems unused

robinschmid:
Only a few last comments. Thanks for using the XML stream builder. definitely the best option here. And nice solution. 

Some of the comments are optional. 
But these two are important:

- [ ] try with resources for the BufferedWriter 
- [ ] code formatter applied to all your files - somehow it seems like it did not work

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularParameters.java
christophuv: Oh, apologies, something went wrong. I resetted the settings and applied it, now it worked.

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, changed it

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, my Java skills are a bit rusty, changed it accordingly

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, moved it to the loop

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, did that

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, changed it accordingly

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
christophuv: Thank you, indeed, removed it

christophuv:
Thank you for the feedback, great suggestions, I incorporated them in the code

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularTask.java
robinschmid: No worries you will find this also throughout the codebase... I guess mzmine started with Java 5 or 6 .... so there is a lot of legacy things in here. 
And Java got so much nicer in java 8 and also since then.

christophuv:
Thanks for code and the multiples files error message

robinschmid (approved):
All done, can be merged

christophuv:
Great, thanks for your help!

PR #1102, opened 2023-01-26
fix concurrent mod exception in ImsExpanderTask

SteffenHeu:
noticed a rarely occurring issue in the stream, so i replaced it with a standard for loop


src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderTask.java
```diff
@@ -110,10 +109,14 @@ public String getTaskDescription() {
 
   @Override
   public double getFinishedPercentage() {
-    return
-        0.4 * tasks.stream().mapToDouble(AbstractTask::getFinishedPercentage).sum() / tasks.size()
-            + 0.4 * (processedRows.get() / (double) totalRows)
-            + 0.2 * createdRows / (double) totalRows;
+    double sum = 0.0;
+    for (int i = 0; i < tasks.size(); i++) {
```
robinschmid, line 113: maybe add a little comment as a reminder. Just in general interesting.
Its coming from adding new elements to the stream thats why changing to for loop actually works. If removal was a thing this would not work.
Good catch

robinschmid:
Can be merged. You can decide if you add the comment

PR #1113, opened 2023-02-01
check for the presence of the substitute-pattern {} in CSVExporModule

christophuv:
added check for the presence of the substitute-pattern {} in CSVExportModule when multiple features lists are selected 
(based on the code suggestion of robinschmid from PR#1100, commit ed5aa04e138c9c9f57632f0f1ffa1ba7bd8b40e2)
// Code formatted and tested

### DIFF
--- a/src/main/java/io/github/mzmine/modules/io/export_features_csv/CSVExportModularParameters.java
+++ b/src/main/java/io/github/mzmine/modules/io/export_features_csv/CSVExportModularParameters.java
@@ -76,7 +76,7 @@ public class CSVExportModularParameters extends SimpleParameterSet {
     String plNamePattern = "{}";
     boolean substitute = filename.getValue().getPath().contains(plNamePattern);
 
-    if (!substitute && featureLists.getValue().getSpecificFeatureLists().length > 1) {
+    if (!substitute && this.getValue(featureLists).getMatchingFeatureLists().length > 1) {
       errorMessages.add("""
           Cannot export multiple feature lists to the same CSV file. Please use "{}" pattern in filename. \
           This will be replaced with the feature list name to generate one file per feature list.
--- a/src/main/java/io/github/mzmine/modules/io/export_features_csv/CSVExportModularParameters.java
+++ b/src/main/java/io/github/mzmine/modules/io/export_features_csv/CSVExportModularParameters.java
@@ -74,7 +74,7 @@ public class CSVExportModularParameters extends SimpleParameterSet {
 
     // Check if substitute pattern is present in filename if several feature lists are selected by the user
     String plNamePattern = "{}";
-    boolean substitute = filename.getValue().getPath().contains(plNamePattern);
+    boolean substitute = this.getValue(filename).getPath().contains(plNamePattern);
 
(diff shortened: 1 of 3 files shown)


SteffenHeu:
Thanks for the contribution!

I'd move this check to the checkParameterValues(List<String> errorMessages) method in the CSVExportModularParameters (override). That way we dont have to create a task and the setup dialog stays open, which makes it less annoying for the user.

christophuv:
Thanks for the hint. That function was implemented. However, I don't think it is called, maybe I am missing some setting in the parameters or the module description. Could you please double-check that? 
Even if I set the check to run regardless of isValueCheckRequired() in EmptyParameterSetupDialogBase.closeDialog, it is not executed.

SteffenHeu:
try

```
@Override
  public boolean checkParameterValues(Collection<String> errorMessages) {
    final boolean superCheck = super.checkParameterValues(errorMessages);

    // Check if substitute pattern is present in filename if several feature lists are selected by the user
    String plNamePattern = "{}";
    final File value = this.getValue(filename); // get the value of the parameter in this instance of the parameter set
    boolean substitute = value.getPath().contains(plNamePattern);

    if (!substitute && this.getValue(featureLists).getMatchingFeatureLists().length > 1) {
      errorMessages.add("""
          Cannot export multiple feature lists to the same CSV file. Please use "{}" pattern in filename.
          This will be replaced with the feature list name to generate one file per feature list.
          """);
    }

    return superCheck && errorMessages.isEmpty();
  }
```

Basically, the parameter sets are a bit tricky to understand. Since all the parameters are declared as static fields, the exist as a static parameter in the ParameterSet class. The parameter sets passed to the modules are always instances of the ParameterSet of the respective module. These instances are created by cloning and create a non-static version of the parameter in the cloning process. Therefore, you have to explicitly query the value of the parameter in the instance of the parameter set to make it work. 

See the "this.getValue()...." calls 

I hope that was clear to understand, otherwise feel free to ask.

christophuv:
Thank you. Worked. 
Yes, I understand your explanation. Thank you

src/main/java/io/github/mzmine/modules/io/export_features_csv/CSVExportModularParameters.java
```diff
@@ -76,7 +76,7 @@ public boolean checkParameterValues(Collection<String> errorMessages) {
     String plNamePattern = "{}";
     boolean substitute = filename.getValue().getPath().contains(plNamePattern);
```
SteffenHeu, line 77: also make this use getValue(filename)

src/main/java/io/github/mzmine/modules/io/export_features_featureML/FeatureMLExportModularParameters.java
```diff
@@ -67,7 +67,7 @@ public boolean checkParameterValues(Collection<String> errorMessages) {
     String plNamePattern = "{}";
     boolean substitute = filename.getValue().getPath().contains(plNamePattern);
```
SteffenHeu, line 68: same

christophuv:
@SteffenHeu Thank you, implemented the changes
code formatted and tested

PR #1114, opened 2023-02-01
Change to MinimumSamplesParameter with absolute and relative

robinschmid:
image
image


src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterParameters.java
```diff
@@ -47,7 +47,6 @@
 import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsParameter;
 import io.github.mzmine.parameters.parametertypes.submodules.OptionalModuleParameter;
 import io.github.mzmine.util.ExitCode;
-import java.text.DecimalFormat;
 import org.jetbrains.annotations.NotNull;
 
 public class RowsFilterParameters extends SimpleParameterSet {
```
SteffenHeu, line 52: increment parameter set version!

src/main/java/io/github/mzmine/parameters/parametertypes/MinimumFeaturesFilterParameters.java
```diff
@@ -32,22 +32,19 @@
...
+import io.github.mzmine.parameters.parametertypes.absoluterelative.AbsoluteAndRelativeInt;
 import io.github.mzmine.parameters.parametertypes.tolerances.RTToleranceParameter;
 import io.github.mzmine.util.ExitCode;
-import javafx.application.Platform;
-
-import java.awt.Window;
 import java.util.List;
+import javafx.application.Platform;
 
 public class MinimumFeaturesFilterParameters extends SimpleParameterSet {
```
SteffenHeu, line 41: increment version

SteffenHeu (changes requested):
Use your own functionality :D

src/main/java/io/github/mzmine/parameters/parametertypes/MinimumFeaturesFilterParameters.java
robinschmid: But then we also need to change the message. Currently old MZMine users will not see that the version  has changed because their batch did not track

robinschmid: will do

robinschmid:
all done

SteffenHeu (approved):

PR #1120, opened 2023-02-03
Module refine MS2 groupings by minimum relative feature height

robinschmid:
This is the output that I get for some DOM data with mz 0.2 isolation width.
Total is the number of MS2 grouped with feature
Unique is only the unique ones.
and removed is the number of MS2 that were removed from a feature because said feature had <25% of the highest feature
```
INFO: Refinement of fragmentation scan-feature assignment (total=26847; unique scans=21616) has removed 2047 scans (7.6 %) from lower abundant feature with a relative intensity threshold of 25.0 %.
The absolute minimum intensity threshold of 0.0E0 does not count to this value. 
```

### DIFF
--- a/src/main/java/io/github/mzmine/datamodel/features/Feature.java
+++ b/src/main/java/io/github/mzmine/datamodel/features/Feature.java
@@ -113,8 +113,8 @@ public interface Feature {
   }
 
   /**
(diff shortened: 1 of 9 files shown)


robinschmid:
this PR is the basis of #1123, which adds this step into the MS2 grouping done in the resolver.

PR #1122, opened 2023-02-03
Show message for updated ParamterSets in old batches

robinschmid:


src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/GeneralResolverParameters.java
```diff
@@ -85,4 +85,9 @@ public GeneralResolverParameters(Parameter[] parameters, String url) {
   public Resolver getResolver(ParameterSet parameterSet, ModularFeatureList flist) {
     return null;
   }
+
+  @Override
+  public int getVersion() {
+    return 2;
```
SteffenHeu, line 91: what did we change? You mean the MS2 grouping?

robinschmid: Yes. we also have to reflect this to the resolver otherwise this will not be picked up by the batch import

PR #1128, opened 2023-02-06
add duplicate filter for ims

SteffenHeu:


src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilder.java
```diff
@@ -223,7 +225,14 @@ protected static void makeAndAddDuplicateRowFilterStep(final BatchQueue q,
     // going back into scans so rather use scan mz tol
     param.setParameter(DuplicateFilterParameters.mzDifferenceMax, mzTol);
     param.setParameter(DuplicateFilterParameters.rtDifferenceMax, rtTol);
+//    param.setParameter(DuplicateFilterParameters.mobilityDifferenceMax,
+//        mobilityType != MobilityType.NONE);
+//    param.getParameter(DuplicateFilterParameters.mobilityDifferenceMax).getEmbeddedParameter()
+//        .setValue(new MobilityTolerance(mobilityType == MobilityType.TIMS ? 0.008f : 1f));
```
robinschmid, line 231: remove

src/main/java/io/github/mzmine/modules/tools/batchwizard/BatchWizardTab.java
```diff
@@ -408,7 +408,7 @@ public void createBatch() {
     }
 
     BatchModeParameters batchModeParameters = (BatchModeParameters) MZmineCore.getConfiguration()
-        .getModuleParameters(BatchModeModule.class);
+        .getModuleParameters(BatchModeModule.class).cloneParameterSet();
```
robinschmid, line 411: if we clone here we do not have to clone when calling runMZmineModule... with the parmaeters.
Can you remove the second call

robinschmid (dismissed):
only some small comments. looks good otherwise

src/main/java/io/github/mzmine/modules/tools/batchwizard/BatchWizardTab.java
SteffenHeu: that was just an attempt to fix the optional parameter not set correctly. will revert.

PR #1139, opened 2023-02-12
New parameters for the blank subtraction module

christophuv:
The quantifier feature can be chosen (height or area) as well as the ratio type (maximum or average of blank groups). Furthermore, the removed features can now also be exported to a separate feature list. This might help with optimizing the parameters of the blank subtraction module Code tested and formatted.


src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -28,37 +28,64 @@
...
 import io.github.mzmine.parameters.impl.SimpleParameterSet;
+import io.github.mzmine.parameters.parametertypes.BooleanParameter;
+import io.github.mzmine.parameters.parametertypes.ComboParameter;
 import io.github.mzmine.parameters.parametertypes.IntegerParameter;
 import io.github.mzmine.parameters.parametertypes.OptionalParameter;
 import io.github.mzmine.parameters.parametertypes.PercentParameter;
 import io.github.mzmine.parameters.parametertypes.StringParameter;
 import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsParameter;
 import io.github.mzmine.parameters.parametertypes.selectors.RawDataFilesParameter;
+import javolution.io.Struct.Bool;
```
robinschmid, line 39: I guess this is by accident?

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -28,37 +28,64 @@
...
-          "Specifies in how many of the blank files a peak has to be detected.");
+  public static final IntegerParameter minBlanks = new IntegerParameter(
+      "Minimum # of detection in blanks",
+      "Specifies in how many of the blank files a peak has to be detected.");
 
-  public static final OptionalParameter<PercentParameter> foldChange =
-      new OptionalParameter<>(new PercentParameter("Fold change increase",
+  public static final ComboParameter quantType = new ComboParameter("Quantification qualifier",
+      "Use either the features' height or area for the subtraction", new String[]{"Height", "Area"},
+      "Height");
```
robinschmid, line 56: you can reuse and modify this, we have an enum for area/ height
```java
  public static final ComboParameter<FeatureMeasurementType> FEATURE_INTENSITY = new ComboParameter<>(
      "Feature intensity", "Intensity in the quantification table (csv).",
      FeatureMeasurementType.values(), FeatureMeasurementType.AREA);
```

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -28,37 +28,64 @@
...
               + " filtered increases more than the given percentage to the blank, it will not be deleted from "
-              + "the feature list.",
-          3.0, 1.0, 1E5));
+              + "the feature list.", 3.0, 1.0, 1E5));
+
+  public static final BooleanParameter keepBackgroundFeatures = new BooleanParameter(
+      "Keep features",
+      "When this option is activated, any feature in a sample that is likely a background will still be kept"
+          + "if the same feature is not a background in another samples. This option does not change the generated "
+          + "feature list, but keeps the features in the blank samples and in those where these are similar to the blank ones.",
```
robinschmid, line 72: for long text you can use textblocks
```java
"""
rkawrhkdaw
dakjwrajwr
"""
```

robinschmid: line breaks are recognized though. 
You can add \ at the end of a line to remove the line break

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -28,37 +28,64 @@
...
+  public static final OptionalParameter<PercentParameter> foldChange = new OptionalParameter<>(
+      new PercentParameter("Fold change increase",
           "Specifies a percentage of increase of the intensity of a feature. If the intensity in the list to be"
               + " filtered increases more than the given percentage to the blank, it will not be deleted from "
-              + "the feature list.",
-          3.0, 1.0, 1E5));
+              + "the feature list.", 3.0, 1.0, 1E5));
+
+  public static final BooleanParameter keepBackgroundFeatures = new BooleanParameter(
+      "Keep features",
```
robinschmid, line 69: maybe we can find a more descriptive name?

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -28,37 +28,64 @@
...
+          + "if the same feature is not a background in another samples. This option does not change the generated "
+          + "feature list, but keeps the features in the blank samples and in those where these are similar to the blank ones.",
+      false);
+
   public static final StringParameter suffix = new StringParameter("Suffix",
       "The suffix for the new feature list.", "subtracted");
 
+  public static final BooleanParameter createDeleted = new BooleanParameter(
+      "Create deleted feature list",
+      "Indicates whether an additional feature list containing all non-used (deleted) features should be saved.");
```
robinschmid, line 80: set default to false

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -28,37 +28,64 @@
...
+
+  public static final StringParameter suffixDeleted = new StringParameter(
+      "Suffix deleted feature list",
+      "The suffix for the feature list containing the not-used features.", "backgroundFeatures");
```
robinschmid, line 84: maybe we could remove this parameter, just use the suffix on top and add something to it like "background"

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
robinschmid: just renamed the enum to AbundanceMeasure. You need to merge the master with your branch

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -28,37 +28,64 @@
...
 
-  public static final OptionalParameter<PercentParameter> foldChange =
-      new OptionalParameter<>(new PercentParameter("Fold change increase",
+  public static final ComboParameter quantType = new ComboParameter("Quantification qualifier",
+      "Use either the features' height or area for the subtraction", new String[]{"Height", "Area"},
+      "Height");
+
+  public static final ComboParameter ratioType = new ComboParameter("Ratio type",
+      "Use either the maximum or the average blank value for calculating the blank-ratio",
+      new String[]{"Maximum", "Average"}, "Maximum");
```
robinschmid, line 60: you have an enum for this. You can use ComboParameter<myenum> .... 
see the previous comment

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -85,13 +89,27 @@ public FeatureListBlankSubtractionTask(MZmineProject project,
...
+    this.suffixDeleted = parameters.getParameter(
+        FeatureListBlankSubtractionParameters.suffixDeleted).getValue();
     checkFoldChange = parameters.getParameter(FeatureListBlankSubtractionParameters.foldChange)
         .getValue();
     foldChange = parameters.getParameter(FeatureListBlankSubtractionParameters.foldChange)
         .getEmbeddedParameter().getValue();
-    intensityType = BlankIntensityType.Maximum;
+    this.ratioType =
+        parameters.getParameter(FeatureListBlankSubtractionParameters.ratioType).getValue()
+            .equals("Maximum") ? RatioType.Maximum : RatioType.Average;
```
robinschmid, line 105: no need to map string to enum  when using the appropriate ComboParmaeter<RatioType> 
getValue will give you a RatioType then.

In general you can also use RatioType.valueOf(String) in other places

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -85,13 +89,27 @@ public FeatureListBlankSubtractionTask(MZmineProject project,
         .getMatchingFeatureLists()[0];
     this.minBlankDetections = parameters.getParameter(
         FeatureListBlankSubtractionParameters.minBlanks).getValue();
+    this.keepBackgroundFeatures = parameters.getParameter(
+        FeatureListBlankSubtractionParameters.keepBackgroundFeatures).getValue();
```
robinschmid, line 93: not important here just a hint for the future.
We now have convenience methods like
parameter.getValue()
so no need to chain getParameter.getValue

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
christophuv: thanks, it was

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
christophuv: Thanks, wanted to keep it default to the current settings, forgot this.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -230,7 +318,11 @@ private boolean checkBlankSelection(FeatureList aligned, List<RawDataFile> blank
     return true;
   }
 
-  enum BlankIntensityType {
+  enum RatioType {
     Average, Maximum
   }
+
+  enum QuantType {
```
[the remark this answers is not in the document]
christophuv: Great, used, thanks

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
christophuv: thanks, will do

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -120,57 +138,116 @@ public void run() {
 
     // get the files that are not considered as blank
     final List<RawDataFile> nonBlankFiles = new ArrayList<>();
+    final List<RawDataFile> blankFiles = new ArrayList<>();
     for (RawDataFile file : originalFeatureList.getRawDataFiles()) {
       if (!blankRaws.contains(file)) {
         nonBlankFiles.add(file);
+      } else {
+        blankFiles.add(file);
```
robinschmid, line 146: why not just use the blankRaws

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
christophuv: I agree, this is hard. What about "Keep background samples"?

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -120,57 +138,116 @@ public void run() {
...
 
     final ModularFeatureList result = new ModularFeatureList(
-        originalFeatureList.getName() + " " + suffix, getMemoryMapStorage(), nonBlankFiles);
+        originalFeatureList.getName() + " " + suffix, getMemoryMapStorage(),
+        keepBackgroundFeatures ? originalFeatureList.getRawDataFiles() : nonBlankFiles);
+    final ModularFeatureList blankResult = new ModularFeatureList(
+        originalFeatureList.getName() + " " + suffixDeleted, getMemoryMapStorage(),
+        originalFeatureList.getRawDataFiles());
+
     originalFeatureList.getRowTypes().values().forEach(result::addRowType);
```
robinschmid, line 159: have to do the same thing with the blankResults list I think

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
christophuv: Great, thanks

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
christophuv: Perfect, of course, thanks for the hint!

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -120,57 +138,116 @@ public void run() {
...
-    nonBlankFiles.forEach(f -> result.setSelectedScans(f, originalFeatureList.getSeletedScans(f)));
+    if (!keepBackgroundFeatures) {
+      originalFeatureList.getRawDataFiles()
+          .forEach(f -> result.setSelectedScans(f, originalFeatureList.getSeletedScans(f)));
+    } else {
+      nonBlankFiles.forEach(
+          f -> result.setSelectedScans(f, originalFeatureList.getSeletedScans(f)));
+    }
```
robinschmid, line 166: you can just loop through 
```java
result.getRawDataFiles().forEach(f -> result.setSelectedScans....)
```

robinschmid: this removes the if else

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
christophuv: Thanks, did it and added the suffix to the description

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
christophuv: Thanks, changed it to a template with the enum and used the values as suggested above directly.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
christophuv: Thanks

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
christophuv: Thanks, the variable is not used further at all. I guess it was a relic from a previous iteration. Removed

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
christophuv: Sorry, I don't get this one. Here either the blank samples should be added or not depending on the users choice. Or is the code in your comment just a different way to write that?

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
christophuv: same as with the next

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -28,37 +28,64 @@
...
+
+  public static final OptionalParameter<PercentParameter> foldChange = new OptionalParameter<>(
+      new PercentParameter("Fold change increase",
           "Specifies a percentage of increase of the intensity of a feature. If the intensity in the list to be"
               + " filtered increases more than the given percentage to the blank, it will not be deleted from "
-              + "the feature list.",
-          3.0, 1.0, 1E5));
+              + "the feature list.", 3.0, 1.0, 1E5));
+
+  public static final BooleanParameter keepBackgroundFeatures = new BooleanParameter(
```
SteffenHeu, line 68: I dont really understand what this parameter will do. Whats the difference of not running this module at all and enabling this option? From what i understand the features that would be removed are not removed, so basically everything stays the same?

christophuv: Thanks

I will try to explain my idea. For this, a compound is a collection of features across samples and a feature a single chromatographic peak in a particular sample
The idea is that a compound can be a background signal and then all features are not higher than in the blank samples. In this case the compound is removed. However, if the compound is a true metabolite, some of its features will be higher than in the blank samples and thus it is retained in the next feature list.
 
This parameter does not change this task. If all features of a compounds are indistinguishable from the background, the compound is still removed. But if the compound has any feature that makes it a non-background compound, all features (including those that are indistinguishable from the background, are kept in the new feature list. This might allow the user to better optimize the background subtraction parameters and give them an idea of how much larger a compound is compared to its background feature.s 

Does this help? 
I agree that the explanation needs some more work and will try to improve it.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
SteffenHeu: i think what robin meant was to add "background" to normal suffix and removethis parameter

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
SteffenHeu: *paramterS.getValue

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -120,57 +138,116 @@ public void run() {
...
+      final ModularFeatureListRow filteredRow = new ModularFeatureListRow(result,
+          originalRow.getID(), (ModularFeatureListRow) originalRow, false);
+      final ModularFeatureListRow removedRow = new ModularFeatureListRow(blankResult,
+          originalRow.getID(), (ModularFeatureListRow) originalRow, false);
+
       for (RawDataFile blankRaw : blankRaws) {
         if (originalRow.hasFeature(blankRaw)) {
+          // save blank detections to a blank-list
+          final Feature f = originalRow.getFeature(blankRaw);
+          removedRow.addFeature(blankRaw, new ModularFeature(blankResult, f));
```
SteffenHeu, line 187: this adds the features to the blank feature list even if the numBlankDetections is not reached

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -120,57 +138,116 @@ public void run() {
...
         for (RawDataFile file : nonBlankFiles) {
           final Feature f = originalRow.getFeature(file);
           // check if there's actually a feature
           if (f != null && f.getFeatureStatus() != FeatureStatus.UNKNOWN) {
             // check validity
-            if (!checkFoldChange || f.getHeight() / blankIntensity >= foldChange) {
-              filteredRow.addFeature(file, new ModularFeature(result, f));
+            double quant = getFeatureQuantifier(f, quantType);
+            if (!checkFoldChange || quant / blankIntensity >= foldChange) {
+              if (!keepBackgroundFeatures) {
```
SteffenHeu, line 203: i dont really understand what the purpose of this parameter is.

here, it would stay in the feature list because it was not below the fold change or not detected in blanks at all.

so it's not considered a background feature. if the keepBackgroundFeatures is true, nothing is added to the feature list anymore?

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -120,57 +138,116 @@ public void run() {
...
+
+      // copy row types
+      if (numFeatures > 0) {
+        // use feature in the new results feature list
+        if (keepBackgroundFeatures) {
+          for (RawDataFile file : originalFeatureList.getRawDataFiles()) {
+            final Feature f = originalRow.getFeature(file);
+            if (f != null && f.getFeatureStatus() != FeatureStatus.UNKNOWN) {
+              filteredRow.addFeature(file, new ModularFeature(result, f));
+            }
```
SteffenHeu, line 223: why is everything added to the feature list again now? I really don't know what you're trying to do :sweat_smile:

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -120,57 +138,116 @@ public void run() {
...
+      if (numBlankDetections > 0 && numFeatures == 0) {
+        // put not-used features on the removed list
+        removedRows.add(removedRow);
       }
 
       processedRows.getAndIncrement();
     }
 
+    // create the list with not-used features first so the used features are the last list to be
+    // created and can be used in the next step when the "last-list" option is used
```
SteffenHeu, line 238: this is a flaw, the last-list option will use both lists

SteffenHeu (commented):
some comments

christophuv:
Thank you for the comments so far. 
I agree that the code is not very clear, not seems the motivation behind it. I will try to optimize it and provide a detailed help about what these parameters do any why. I apologize for this inconvenience.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
christophuv: It is indeed. Is there a possibility to avoid this?

christophuv: But then, these options are merely intended to help optimize these background-subtraction parameters. Once this has been done, only the subtracted features list can be kept.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
christophuv: sorry, I will try to clarify in the code and with a help

christophuv:
I apologize for the confusion. My motivation behind the changes was to help users with optimizing / understanding the parameters of the background subtraction. While it might be easy to understand the parameters conceptually (fold-increase, number of detections), it is a bit harder to judge whether a certain set of parameters is well suited for a dataset. 

I did not mean to change the algorithm for the blank subtraction itself. When the default parameters for the new settings are used, the algorithm should perform as before and thus not break compatibility with any batch-mode. Even if the new parameters are activated, the calculation will stay the same. However, the results will be presented a bit different. 

For the explanation I will refer to a FeatureRow as an aligned consensus-feature across the different samples. Such a consensus feature can thus have many features, at max one in each of the samples. The background subtraction module will remove all blank samples and those where the features are indistinguishable from the background. In the following example this keeps only the green (or orange) features (plants and QCs), but not any black ones (blank samples). Thus, the number of FeatureRows is reduced from previously n to m (with m <= n). 

1_current

Using the first new parameter (keep background samples): Here any FeatureRow that has at least one feature that is not a background is also kept and still a new table with m FeatureRows is generated. However, all Features of a FeatureRow are kept even if these are background features.
In the following figure, two examples are shown where a FeatureRow was also detected in the blank samples, but at much lower abundances. In the first row, no signal for the blank samples can be seen, but in rows 2 and 3 there a slight abundances for the blank samples (blank). This confirms (visually for the user) that the background subtraction is working and it also provides the user with a feeling for the difference in height or area. By scrolling through the list or sorting by the abundance of the blank samples, they should be able to find examples with high/low folds. 

2_keepBlanks

Additionally, keeping the features that are indistinguishable from the background of a FeatureRow with at least one feature not being the background also has the advantage that the background features can be used for statistical analysis, visualization of the blank, parameter optimization, etc.*1.

Using the new parameter (create deleted list) all FeatureRows that have not at least one feature that is distinguishable from the background are put into a new featureList. This also allows the user to see how the background subtraction worked. In the example below such a background-feature list is shown. The first example clearly shows a featureRow that seems to be a correct feature with slight abundances in the blanks samples. 

3_subtracted

It was nevertheless removed as its abundances were below the necessary fold of 1000%. 1.3E5 in the maximum blank, 1.3E6 in the maximum plant sample < 1000%. 

4_abundances

Using this review of the blank subtraction module and optimizing the parameters, the fold was decreased to 500%, which would keep this feature in the non-background FeatureTable (confirmed, not shown). 

Furthermore, having the removed features also helps to see small differences, inspect these features, might be good for the generation of MSMS exlcusion lists, manual inspection and targeted evaluation, etc..

Finally, once the user has a good understanding about these parameters and optimized them, the module can be included into a batch-processing with the default parameters, which will remove all blank samples and also features indistinguishable from noise. Or they can chose to include the blank sample features for and featureRow that has at least one feature distinguishable from the background for statistical analysis, etc..

Furthermore, two other parameters allow to select either the maximum or average in the blank samples as well as the quantifier (area or height). 

In summary, these new parameters are primarily a means to optimize the blank subtraction itself rather than a true evolution of the algorithm. My motivation behind it was also clear, a colleague of mine is struggling a bit finding some quite low abundant compounds in the dataset and wanted to see the effect of different parameters in this blank subtraction module upon her metabolites. 

I hope this explanation helps understand the rationale behind my suggested cod and implementation. I am happy to provide further details on that if it is still unclear. 

If you think these changes are unnecessary or not relevant, of course please feel free to deny the PR. 


*1 
Additionally, I think there are also different concepts about blank subtraction. I understand your motivation behind it and that you want to remove the features indistinguishable from the background entirely. 
I always understood the blank subtraction in a bit different way that it should assign for each FeatureRow (i.e., chemical compound in the dataset) whether or not it is a background compound (i.e., in almost all samples approximately similarly abundant) and thus the compound can be ignored/removed or if it is slightly abundant in the blank samples but much higher in the biological samples (i.e., a truly biological metabolite). In the latter case I also preferred to keep the abundances in the blank samples and any biological condition the compound was not detected to be able to perform a comparison of the biological conditions (similar to a re-integration or gap filling)

christophuv:
Oh no, the compilation fails with the FeatureMeasurementType enum not being found. On my machine this works. I apologize, I guess I have too little experience to resolve this myself

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -25,40 +25,69 @@
...
+      features distinguishable from the background and
+      features indistinguishable from the background in all non-blank samples.
+      Furthermore, if the option is activated, abundances for the blank samples will be retained.""",
+      false);
+
   public static final StringParameter suffix = new StringParameter("Suffix",
       "The suffix for the new feature list.", "subtracted");
 
+  public static final BooleanParameter createDeleted = new BooleanParameter(
+      "Create deleted feature list", """
```
SteffenHeu, line 83: I think "Create list of deleted features" is a better name.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -25,40 +25,69 @@
...
+              + "the feature list.", 3.0, 1.0, 1E5));
+
+  public static final BooleanParameter keepBackgroundFeatures = new BooleanParameter(
+      "Keep background samples/features", """
+      When this option is activated, the rational of the blank subtraction is changed slightly. Background
+      features are not removed, but aligned features are sorted into those that have
+      features distinguishable from the background and
+      features indistinguishable from the background in all non-blank samples.
+      Furthermore, if the option is activated, abundances for the blank samples will be retained.""",
+      false);
```
SteffenHeu, line 77: Trying to think of a better name here... I think what would be good is to change this to a ComboParameter and use a new enum which directly explains a bit what the influence of this parameter is.

The parameter should be named that concisely paraphrases "How to handle rows with features above and below fold change". The options would be "Keep all features in that row" and "Only keep features above fold change"

SteffenHeu (commented):
Sorry, somehow i did not see the notification that you changed something.

Looks way clearer now, thank you. I added some more comments, let me know what you think.

One more question:

> In the following figure, two examples are shown where a FeatureRow was also detected in the blank samples, but at much lower abundances. In the first row, no signal for the blank samples can be seen, but in rows 2 and 3 there a slight abundances for the blank samples (blank). **This confirms (visually for the user) that the background subtraction is working** and it also provides the user with a feeling for the difference in height or area. By scrolling through the list or sorting by the abundance of the blank samples, they should be able to find examples with high/low folds.

I don't really understand what you mean with the highlighted/bold text. If the blank raw data files are still in the subtracted feature list, it is not really clear to me how the user would see that the module is working. maybe it would be good to annotate these rows where one feature was above the threshold in that case?

christophuv:
Thank you for the comments. I will implement them. 

> In the following figure, two examples are shown where a FeatureRow was also detected in the blank samples, but at much lower abundances. In the first row, no signal for the blank samples can be seen, but in rows 2 and 3 there a slight abundances for the blank samples (blank). **This confirms (visually for the user) that the background subtraction is working and it also provides the user with a feeling for the difference in height or area**. By scrolling through the list or sorting by the abundance of the blank samples, they should be able to find examples with high/low folds.

With the bold text I refer to both the abundances in the blank samples as well as the visual representation in the feature lists. The user can color-code the blank samples accordingly and get a feeling how different fold-change parameters change the background subtraction. Together with the peak area/height values in the table this gives them an impression on how strict or relaxed the fold-change was set. Together with the secondary list (i.e., the one containing all rows with no feature being indistinguishable from the background) I think this will help. 
 

The annotation is certainly a good idea. Maybe this can be the number of blank detections, their average and std and the mean values in any non-background samples for features having been removed or kept. I will though have to dive into the code on how to add such a thing.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
christophuv: Thank you, changed and adapted it a bit

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
christophuv: Thank you, has been changed similar to other such parameters in MZmine modules (KEEP - desc, REMOVE - desc)
dialog

christophuv:
Oh no, the compilation fails with the FeatureMeasurementType enum not being found. On my machine this works. I apologize, I guess I have too little experience to resolve this myself

christophuv:
The FeatureMeasurementType as moved or renamed since the last sync. May I ask for the name of the new enum please?

SteffenHeu:
Sorry, i was too busy again. The changes look good to me. Do you want to dive into the annotation model before we merge this PR?

christophuv:
Thanks for looking at the changes. 
I am happy either way, but I will certainly look into the annotations and how to implement that. But this will take some more time.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -224,6 +235,24 @@ public void run() {
...
+                notBackgroundFeaturesOfCurrentRow.size(), nonBlankRaws.size(),
+                notBackgroundFeaturesOfCurrentRow.size() / nonBlankRaws.size() * 100.));
+            sb.append(String.format(" but not in any of the %3d blank samples", blankRaws.size()));
+          } else {
+            sb.append(String.format(" found in %3d / %3d (%4.1f%%) samples",
+                notBackgroundFeaturesOfCurrentRow.size(), nonBlankRaws.size(),
+                notBackgroundFeaturesOfCurrentRow.size() / nonBlankRaws.size() * 100.));
+            sb.append(String.format(" and in %3d / %3d (%4.1f%%) background samples (abundance %s)",
+                foundInNBlanks, blankRaws.size(), foundInNBlanks / blankRaws.size() * 100.,
+                MZmineCore.getConfiguration().getExportFormats().intensity(blankAbundance)));
```
SteffenHeu, line 252: same, should use the guiFormat.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionParameters.java
```diff
@@ -90,9 +90,15 @@ or if the background is of interest in subsequent data processing steps (i.e., s
       All features removed by this step will then be saved to a new feature list with the suffix 'subtractedBackground'.""",
       false);
 
+  public static final BooleanParameter verboseColumn = new BooleanParameter(
```
SteffenHeu, line 93: Would it make sense to always enable the annotation when the removed feature list is created and disable it otherwise? what do you think?

src/main/java/io/github/mzmine/modules/dataprocessing/filter_blanksubtraction/FeatureListBlankSubtractionTask.java
```diff
@@ -237,6 +266,15 @@ public void run() {
         backgroundFeaturesOfCurrentRow.forEach(f -> featureListRow.addFeature(f.getRawDataFile(),
             new ModularFeature(backgroundAlignedFeaturesList, f)));
 
+        if (this.createVerboseAnnotation) {
+          final StringBuilder sb = new StringBuilder();
+          sb.append(String.format(
+              "Background: Found in %3d / %3d (%4.1f%%) background samples (abundance %s) but not in any samples with higher abundances",
+              foundInNBlanks, blankRaws.size(), foundInNBlanks / blankRaws.size() * 100.,
+              MZmineCore.getConfiguration().getExportFormats().intensity(blankAbundance)));
```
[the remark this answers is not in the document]
christophuv: Thanks, will do

PR #1141, opened 2023-02-13
Remove netCDF from mass detection step 

robinschmid:
@tomas-pluskal @asmirn1 
Can you confirm that we can remove this option?

This PR removes the option to export netCDF right after mass detection. People should use the module instead. 

However we would need to enable that people can export mass lists instead of the original raw data in the module.

image


robinschmid:
@tomas-pluskal @asmirn1 
Can you confirm that we can remove this option?

This PR removes the option to export netCDF right after mass detection. People should use the module instead

image

robinschmid:
Oh there are more problems with this. 
We would then also need to add a ScanFilter to the netCDF export, but the method itself is MSDK

robinschmid:
@asmirn1, can you comment on this? We would like to remove the option from mass detection as an export does not belong with the fundamental data processing step

asmirn1:
@robinschmid I agree that the netCDF export isn't needed in the preprocessing step. I can remove it and make a merge request if you want me to.

ansgarkorf:
Thanks for confirming @asmirn1! No need to make an other PR. Would be great if you could review this one.

asmirn1:
The code looks good to me. `MassDetectionTask` is much cleaner now.

src/main/java/io/github/mzmine/modules/io/export_rawdata_netcdf/NetCDFExportTask.java
```diff
@@ -92,14 +85,16 @@ public void run() {
       if (outFilename.getName().toLowerCase().endsWith("mzml")) {
         msdkMethod = new MzMLFileExportMethod(msdkDataFile, outFilename, MzMLCompressionType.ZLIB,
             MzMLCompressionType.ZLIB);
+        msdkMethod.execute();
       }
 
       if (outFilename.getName().toLowerCase().endsWith("cdf")) {
+        if (useMassList) {
+
```
ansgarkorf, line 93: Something missing?

PR #1154, opened 2023-02-19
macOS signing

robinschmid:


robinschmid:
contains work from @roman-bushuiev in #716

PR #1159, opened 2023-02-20
Options to hide shapes columns in large datasets in FeatureTableFX

robinschmid:
Automatically applies when a new feautre list is loaded.
image

### DIFF
--- a/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableFX.java
+++ b/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableFX.java
@@ -60,6 +60,7 @@ import io.github.mzmine.datamodel.features.types.fx.ColumnID;
 import io.github.mzmine.datamodel.features.types.fx.ColumnType;
 import io.github.mzmine.datamodel.features.types.modifiers.ExpandableType;
 import io.github.mzmine.datamodel.features.types.modifiers.SubColumnsFactory;
+import io.github.mzmine.datamodel.features.types.numbers.AreaType;
 import io.github.mzmine.datamodel.features.types.numbers.CCSRelativeErrorType;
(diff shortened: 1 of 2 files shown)


robinschmid:
image

src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableFX.java
```diff
@@ -182,6 +184,49 @@ public FeatureTableFX() {
...
+  /**
+   * @return the maximum samples before deactivating the shapes columns
+   */
+  public int getMaximumSamplesForVisibleShapes() {
+    return parameters.getValue(FeatureTableFXParameters.deactivateShapesGreaterNSamples);
+  }
+
+  private void setShapeColumnsVisible(boolean state) {
+    setVisible(ColumnType.ROW_TYPE, FeatureShapeType.class, state);
+    setVisible(ColumnType.ROW_TYPE, FeatureShapeMobilogramType.class, state);
```
SteffenHeu, line 215: add FeatureShapeMobilogramHeatmapType, ImageType (FeatureTypes)

SteffenHeu (dismissed):

src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableFX.java
robinschmid: thought about this again. 
Currently this is called when small sample sets are loaded. 
Not sure if we want to activate FeatureType columns by default for small datasets... 
Even something like 10 samples might make the table unusable.
Maybe we could only activate those by default with 1 or 2 samples?

SteffenHeu: i would just deactivate them for large sample sizes and not change the setting for smaller sample sizes

PR #1164, opened 2023-02-20
Use OriginalFeatureListHandlingParameter's value instead of default v…

tnguy271:
…alue

### DIFF
--- a/src/main/java/io/github/mzmine/modules/dataprocessing/adap_mcr/ADAP3DecompositionV2Parameters.java
+++ b/src/main/java/io/github/mzmine/modules/dataprocessing/adap_mcr/ADAP3DecompositionV2Parameters.java
@@ -24,6 +24,7 @@
  */
 package io.github.mzmine.modules.dataprocessing.adap_mcr;
 
+import io.github.mzmine.parameters.parametertypes.OriginalFeatureListHandlingParameter;
 import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsParameter;
 import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsSelectionType;
 import java.awt.Window;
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderGcEiDeconvolution.java
```diff
@@ -175,9 +175,7 @@ private void makeMultiCurveResolutionStep(final BatchQueue q) {
     param.setParameter(ADAP3DecompositionV2Parameters.PEAK_LISTS,
         new FeatureListsSelection(FeatureListsSelectionType.BATCH_LAST_FEATURELISTS));
     param.setParameter(ADAP3DecompositionV2Parameters.ADJUST_APEX_RET_TIME, false);
-
-    // TODO replace with last OriginalFeatureListHandlingParameter
-    param.setParameter(ADAP3DecompositionV2Parameters.AUTO_REMOVE, true);
+    param.setParameter(ADAP3DecompositionV2Parameters.AUTO_REMOVE, handleOriginalFeatureLists.name().equals("REMOVE") ? true : false);
```
robinschmid, line 178: It is better to change the parameter type of ADAP3DecompositionV2Parameters.AUTO_REMOVE to be a
```java
OriginalFeatureListHandlingParameter handleOriginal = new OriginalFeatureListHandlingParameter(false);
```
then handle the two state in the corresponding task. Its an enum with REMOVE or KEEP so easy to convert to boolean

robinschmid: also you should not compare Strings when working with enums.
just do
handeOriginalFeatureLists != OriginalFeatureListOption.KEEP

tnguy271: I'll update the parameter

tnguy271: Currently I don't think the setParameter method works with OriginalFeatureListHandlingParameter. Correct me if I'm wrong.

src/main/java/io/github/mzmine/modules/dataprocessing/adap_mcr/ADAP3DecompositionV2Task.java
```diff
@@ -127,9 +128,8 @@ public void run() {
 
             // Add new peaklist to the project.
             project.addFeatureList(newPeakList);
-
             // Remove the original peaklist if requested.
-            if (parameters.getParameter(ADAP3DecompositionV2Parameters.AUTO_REMOVE).getValue()) {
+            if (parameters.getParameter(ADAP3DecompositionV2Parameters.HANDLE_ORIGINAL).getValue() == OriginalFeatureListOption.REMOVE) {
```
robinschmid, line 132: you can automatically run code formatting before commiting your changes

src/main/java/io/github/mzmine/modules/dataprocessing/adap_mcr/ADAP3DecompositionV2Parameters.java
```diff
@@ -86,13 +87,12 @@ public class ADAP3DecompositionV2Parameters extends SimpleParameterSet {
   public static final StringParameter SUFFIX = new StringParameter("Suffix",
       "This string is added to feature list name as suffix", "Spectral Deconvolution");
 
-  public static final BooleanParameter AUTO_REMOVE =
-      new BooleanParameter("Remove original feature lists",
-          "If checked, original chromomatogram and feature lists will be removed");
+  public static final OriginalFeatureListHandlingParameter HANDLE_ORIGINAL =
```
robinschmid, line 90: add this to the parameterset.
Now that we changed the name and datatype of this parameter we need to increment the version from 1 to 2 so that people will be alert.
```java

  @Override
  public int getVersion() {
    return 2;
  }
```

robinschmid (commented):
All good only one comment

robinschmid (approved):
Great

PR #1168, opened 2023-02-21
Batch copy DoubleBuffer to array

robinschmid:


src/main/java/io/github/mzmine/util/DataPointUtils.java
```diff
@@ -105,11 +105,8 @@ public static double[][] getDataPointsAsDoubleArray(DoubleBuffer mzValues,
    * subclasses. Usually, the data should be accessed directly via the buffer.
    */
   public static double[] getDoubleBufferAsArray(DoubleBuffer values) {
-    double[] data = new double[values.capacity()];
-
-    for (int i = 0; i < values.capacity(); i++) {
-      data[i] = values.get(i);
-    }
+    double[] data = new double[values.limit()];
+    values.get(data);
```
SteffenHeu, line 109: values.get is relative method, use the absolute `values.get(0, data);` instead. you can also patch `getDataPointsAsDoubleArray` to use the same

SteffenHeu (changes requested):

robinschmid:
Resolved with Steffen

PR #1176, opened 2023-02-22
Map old to new parameter names, change ADAP chrom builder parameters

robinschmid:
dark is the new parameter and light is the old names
image


SteffenHeu:
Looks good to me. Is it worth changing the parameter set to a map completely?

robinschmid:
I think its not worth. 
Most modules extract parameters as a first step and its efficient enough for small lists.
Might be a bigger memory impact with all those parametersets that we have flying around.

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imagebuilder/ImageBuilderParameters.java
```diff
@@ -53,17 +54,13 @@ public class ImageBuilderParameters extends SimpleParameterSet {
...
-      "m/z tolerance between scans to be placed in the same image.", 0.005, 15,
-      false);
+      "Scan to scan accuracy (m/z)", "m/z tolerance between scans to be placed in the same image.",
+      0.005, 15, false);
 
   public static final IntegerParameter minTotalSignals = new IntegerParameter(
-      "Min group size in # of scans",
-      "Minimum number of consecutive signals (data points) to form an image",
-      50);
+      "Minimum detected scans", "Minimum number of signals (data points) to form an image", 50);
```
SteffenHeu, line 61: its also consecutive since we use the same algorithm as adap

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imagebuilder/ImageBuilderParameters.java
```diff
@@ -53,17 +54,18 @@ public class ImageBuilderParameters extends SimpleParameterSet {
...
-      "m/z tolerance between scans to be placed in the same image.", 0.005, 15,
-      false);
+      "Scan to scan accuracy (m/z)", "m/z tolerance between scans to be placed in the same image.",
+      0.005, 15, false);
 
   public static final IntegerParameter minTotalSignals = new IntegerParameter(
-      "Min group size in # of scans",
-      "Minimum number of consecutive signals (data points) to form an image",
-      50);
+      "Minimum detected scans", "Minimum number of signals (data points) to form an image", 50,
```
SteffenHeu, line 61: Minimum total signals is a better name i think

SteffenHeu (commented):
some comments

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imagebuilder/ImageBuilderParameters.java
robinschmid: done

robinschmid:
I addressed all changes. Only the open one. I set to 0 to resort to the noise level in mass detection

PR #1188, opened 2023-02-27
Remodel ScanSelection and MS level filter

robinschmid:
This is a quite big pull request to harmonize SubModuleParameter, ParameterSetParameter, OptionalModuleParameter

There are options now to either show a sub ParameterSet in a sub panel or in a dialog.

There is no way to auto size a Stage (dialog) so we estimate some size changes here based on the number of new parameters that are shown.

image

image

**Correlation grouping**
image

image


robinschmid:
- Added support to load the scan selections from old parameter files
- Added string to describe the selected filters. Otherwise, people will miss that sometimes. Its only shown when selected.

image
image
image
image

src/main/java/io/github/mzmine/parameters/parametertypes/selectors/ScanSelectionComponent.java
```diff
@@ -25,246 +25,72 @@
...
-      scanMobilityRange = null;
-      polarity = null;
-      spectrumType = null;
-      msLevel = null;
-      scanDefinition = null;
-      updateRestrictionList();
+import javafx.scene.text.Text;
 
-    });
+public class ScanSelectionComponent extends OptionalModuleComponent {
```
SteffenHeu, line 33: OptionalModuleComponent -> Does this mean the scan selection always has a checkbox now?

src/main/java/io/github/mzmine/parameters/AbstractParameter.java
```diff
@@ -0,0 +1,53 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.parameters;
+
+import javafx.scene.Node;
+
+public abstract class AbstractParameter<ValueType, EditorComponent extends Node> implements
```
SteffenHeu, line 30: Why do we need this now?

SteffenHeu (approved):
only a few comments

src/main/java/io/github/mzmine/parameters/parametertypes/selectors/ScanSelectionComponent.java
robinschmid: yeah its always with checkbox... could be changed but in general you always get the ScanSelection.ALL_SCANS if its deselected

PR #1200, opened 2023-03-03
refactor pasef ms2 merging

SteffenHeu:
- delete spectramerging.mergemsmsspectra to use the mergespectra method for mergin pasef ms2 scans
- use fragment scan selection


robinschmid (approved):
looks good need to do the same for LC-MS

PR #1201, opened 2023-03-03
Rename more parameters

robinschmid:


src/main/java/io/github/mzmine/parameters/parametertypes/combowithinput/ComboWithInputComponent.java
```diff
@@ -85,4 +94,21 @@ public void setValue(final ComboWithInputValue<EnumValue, ?> newValue) {
           newValue.getEmbeddedValue());
     }
   }
+
+  @Override
+  public void addValueChangedListener(final Runnable onChange) {
```
SteffenHeu, line 99: a listener should be added in the parameter setup dialog or?

SteffenHeu (approved):

PR #1212, opened 2023-03-05
Fix parameter load: Embedded parametersets

robinschmid:
getElementsByTagName is evil.
It goes through the whole tree not only the first level - this lead to loading of  SubParameterSet parameters on the parent parameter level

### DIFF
--- a/src/main/java/io/github/mzmine/main/impl/MZmineConfigurationImpl.java
+++ b/src/main/java/io/github/mzmine/main/impl/MZmineConfigurationImpl.java
@@ -34,7 +34,6 @@ import io.github.mzmine.gui.preferences.UnitFormat;
 import io.github.mzmine.main.MZmineConfiguration;
 import io.github.mzmine.main.MZmineCore;
 import io.github.mzmine.modules.MZmineModule;
-import io.github.mzmine.parameters.Parameter;
 import io.github.mzmine.parameters.ParameterSet;
 import io.github.mzmine.parameters.impl.SimpleParameterSet;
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/mzmine/main/impl/MZmineConfigurationImpl.java
```diff
@@ -296,7 +295,13 @@ public void loadConfiguration(File file, boolean loadPreferences) throws IOExcep
...
 
           ParameterSet moduleParameters = getModuleParameters(moduleClass);
-          moduleParameters.loadValuesFromXML(moduleElement);
+          if(moduleParameters==null) {
+            logger.info(
+                "Module %s was in the config file but was not found in the current version of MZmine".formatted(
+                    moduleClass.getName()));
+          }
+          var parameterElement = (Element) moduleElement.getElementsByTagName("parameters").item(0);
+          moduleParameters.loadValuesFromXML(parameterElement);
```
SteffenHeu, line 304: it can still be null here or?

SteffenHeu (commented):
good catch

src/main/java/io/github/mzmine/main/impl/MZmineConfigurationImpl.java
robinschmid: It actually can. But then it describes that in the exception

robinschmid: in the try catch

PR #1213, opened 2023-03-05
New module for CompoundAnnotationsCSVExportTask

shamlymhd:
I hope to commit changes to this branch. The changes will create a new feature for exporting CompoundAnnotations into a CSV (comma-separated values) file.


src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
```diff
@@ -0,0 +1,119 @@
...
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+
+public class CompoundAnnotationsCSVExportTask extends AbstractTask implements ProcessedItemsCounter {
```
robinschmid, line 22: we can remove the ProcessedItemsCounter it was quite specific where we used it.

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
```diff
@@ -0,0 +1,119 @@
...
+                + " to CSV file(s) ";
+    }
+
+    @Override
+    public double getFinishedPercentage() {
+        return 0;
+    }
+
+    @Override
+    public void run() {
```
robinschmid, line 58: need to add logic here see issue for some details on how to get the annotations to export

robinschmid:
added some comments. The start is correct

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
shamlymhd: Got it 👍

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
shamlymhd: yeah, I thought the same. Thanks 😊

shamlymhd:
After annotating the compounds, new columns will be added to the feature list table. is this task to export these columns to a CSV file?

robinschmid:
Exactly. 
The data is already added to the ModularFeatureListRow which is collected in a FeatureList.

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
```diff
@@ -0,0 +1,187 @@
...
+    private void exportFeatureList(FeatureList featureList, BufferedWriter writer)
+            throws IOException {
+         // ToDO : Add headers and raws to csv
+        try {
+
+            // loop through all rows in the feature list
+            for (FeatureListRow row : featureList.getRows()) {
+                List<Object> featureAnnotations = row.getAllFeatureAnnotations();
+                for (Object object : featureAnnotations) {
+                    if (object instanceof FeatureAnnotation annotation) {
```
robinschmid, line 132: perfect

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
```diff
@@ -0,0 +1,187 @@
...
+                        // Export fields from the FeatureAnnotation object
+                        String CompoundName = annotation.getCompoundName();
+                        double precursorMZ = annotation.getPrecursorMZ();
+                        IonType adductType = annotation.getAdductType();
+                        Float mobility = annotation.getMobility();
+                        Float getCCS = annotation.getCCS();
+                        Float getRT = annotation.getRT();
+                        Float getScore = annotation.getScore();
```
robinschmid, line 140: this one has a convenience method annotation.getScoreString

robinschmid, line 140: try something like this:
`Stream.of(a,b,c,d).map(o -> ... null should be empty string and otherwise o.toString()  then map with CSVUtils.escape to handle values that contain the separator ,  then use collect join ","` 

this way you do one stream and end up with a joined string where each field is null checked and escaped.
So a value 1,000 would be wraped in "1,000" to escape the ,

robinschmid (changes requested):

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
shamlymhd: resolved

shamlymhd: yeah got it thank you

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
```diff
@@ -0,0 +1,187 @@
...
+
+
+    }
+    private void exportFeatureList(FeatureList featureList, BufferedWriter writer)
+            throws IOException {
+         // ToDO : Add headers and raws to csv
+        try {
+
+            // loop through all rows in the feature list
+            for (FeatureListRow row : featureList.getRows()) {
```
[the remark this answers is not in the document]
shamlymhd: done thank you

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
```diff
@@ -0,0 +1,187 @@
...
+//                line.append("partners").append(fieldSeparator);
+//            } else if (commonElements[i].equals(LegacyExportRowCommonElement.ROW_BEST_ANNOTATION)) {
+//                line.append("best ion").append(fieldSeparator);
+//            } else {
+//                name = commonElements[i].toString();
+//                name = name.replace("Export ", "");
+//                name = escapeStringForCSV(name);
+//                line.append(name).append(fieldSeparator);
+//            }
+//        }
```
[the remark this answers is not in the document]
shamlymhd: I cleaned the comments, and I did what you suggested before. please review

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
```diff
@@ -0,0 +1,174 @@
...
+                                .collect(Collectors.joining(","));
+
+                        // Export the fields as needed
+                        writer.write(result);
+
+                    }
+                }
+            }
+
+            writer.close();
```
robinschmid, line 168: no need to close the writer as its opened in a try with ressoruces block.

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Override
+    public String getTaskDescription() {
+        return "Exporting feature list(s) " + Arrays.toString(featureLists)
+                + " to CSV file(s) ";
+    }
+
+    @Override
+    public double getFinishedPercentage() {
+        return 0;
```
robinschmid, line 65: lets track the progress. 
Add a variable in the totalRows and calculate it by going through the featureLists.stream :: getNumberOfRows . sum
Then just increment another int finishedRows

src/main/java/io/github/mzmine/util/io/CSVUtils.java
```diff
@@ -100,4 +100,25 @@ public static String getListString(double[] list, String sep, NumberFormat forma
     return String.format("[%s]",
         Arrays.stream(list).mapToObj(format::format).collect(Collectors.joining(sep)));
   }
+
+  public static String escape(String value) {
```
robinschmid, line 104: why do you need this? and not just the method above where also the separator is escaped

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/TableFormat.java
```diff
@@ -0,0 +1,14 @@
+package io.github.mzmine.modules.io.export_compoundAnnotations_csv;
+
+public enum TableFormat {
```
robinschmid, line 3: this class can be deleted its a copy from somewhere

robinschmid (approved):
Thanks for the commitment. 
We are close to finishing this one. Only a few comments

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
shamlymhd: yeah got it thanks.

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
shamlymhd: Resolved

src/main/java/io/github/mzmine/util/io/CSVUtils.java
shamlymhd: compoundName, adductType, scoreType, precursorMZ, mobility, getCCS, and getRT, with any commas within these values properly escaped using CSVUtils.escape() method. it is important, right?

src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/TableFormat.java
shamlymhd: As a second point regarding the issue, you mentioned 'long table format, one row in CSV per annotation.' Therefore, I thought about creating a class and using it (copied from another module). However, it did not turn out to be useful. Thank you so much for your time and guidance.

src/main/java/io/github/mzmine/util/io/CSVUtils.java
robinschmid: There is a method above that has a similar signature:
CSVUtils.escape(value, separator) it takes care of different separators

shamlymhd: Done. thank you.

PR #1229, opened 2023-03-07
fix removed files reappear issue, remove raw data renaming

SteffenHeu:


robinschmid (approved):
Looks good to me. Would be safer to not rename files

PR #1237, opened 2023-03-08
Fix join aligner by adding @Nullable to all setToComponent

robinschmid:


src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/RTToleranceComponent.java
```diff
@@ -89,7 +90,13 @@ public RTTolerance getValue() {
 
   }
 
-  public void setValue(RTTolerance value) {
+  public void setValue(@Nullable RTTolerance value) {
+    if (value == null) {
+      toleranceField.setText("");
+      toleranceType.getSelectionModel().clearSelection();
```
SteffenHeu, line 96: i think we use enums for this or? Since it is probably set during loading/saving and uses Enum.valueOf i think defaulting to a certain value might be better.

src/main/java/io/github/mzmine/parameters/parametertypes/submodules/ModuleComboComponent.java
```diff
@@ -86,6 +86,10 @@ public int getSelectedIndex() {
   }
 
   public void setSelectedItem(MZmineProcessingStep<?> selected) {
+    if (selected == null) {
+      comboBox.getSelectionModel().clearSelection();
```
SteffenHeu, line 90: same here, i'd default to any of the values

robinschmid: This one is not an enum. Its an object and might be used in different modules with different steps.
clear selection just forces the user to select it.
this should also not happen to often.

PR #1239, opened 2023-03-09
add weak polarity check for spectral library matching

SteffenHeu:


src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/RowsSpectralMatchTask.java
```diff
@@ -411,10 +415,19 @@ public void matchRowToLibraries(List<SpectralLibraryEntry> entries, FeatureListR
       List<SpectralDBAnnotation> ids = null;
       // match against all library entries
       for (SpectralLibraryEntry ident : entries) {
+
+        final String entryPolarity = ident.getOrElse(DBEntryField.POLARITY, null);
```
robinschmid, line 419: could also check charge as this could also be 1 or -1 I think

robinschmid (dismissed):
looks good just two comments

src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/RowsSpectralMatchTask.java
```diff
@@ -598,4 +611,25 @@ public int getCount() {
...
+    if (polarity == null || polarity == PolarityType.UNKNOWN) {
+      return true;
+    }
+    return switch (polarityString) {
+      case null, default -> true;
+      case "+", "Positive", "POSITIVE", "positive", "pos", "+1" ->
+          polarity == PolarityType.POSITIVE;
+      case "-", "Negative", "NEGATIVE", "negative", "neg", "-1" ->
+          polarity == PolarityType.NEGATIVE;
+    };
```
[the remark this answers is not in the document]
robinschmid: also handle null outside the switch and then do String.toLowerCase before the switch to just check positive, pos, ...

src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/RowsSpectralMatchTask.java
SteffenHeu: decided against that, since we are not really sure how well the charge is harmonized in libraries. Some may put 1, but negative polarity, some may use -1 for that.

PR #1245, opened 2023-03-09
- corrected detectSpectrumType function

mhhur:
- corrected detectSpectrumType function for Waters's DDA and its mzML file converted.
- removed unnecessary code that checked the scan's retention time (in SimpleIonTimeSeries)

These corrections enable the correct reading of raw data files and displaying them. It was tested with Waters DDA data and mzML converted.

Please review the codes.

Thanks,
Manhoi Hur
@UC Riverside Metabolomics Core Facility


build.gradle
```diff
@@ -290,7 +290,8 @@ javafx {
 // currently there is an issue with the jfx>16 and the feature table on windows
 // scrolling lets the table jump
 // jfx19 fixes remote connection issues though
-    version = OperatingSystem.current().isMacOsX() ? "19" : "16"
+    //version = OperatingSystem.current().isMacOsX() ? "19" : "16"
+    version = "19"
```
SteffenHeu, line 294: please revert to original code. the issue with java 19 is described above.

src/main/java/io/github/mzmine/datamodel/featuredata/impl/SimpleIonTimeSeries.java
```diff
@@ -73,13 +73,14 @@ public SimpleIonTimeSeries(@Nullable MemoryMapStorage storage, @NotNull double[]
     if (mzValues.length != intensityValues.length || mzValues.length != scans.size()) {
       throw new IllegalArgumentException("Length of mz, intensity and/or scans does not match.");
     }
+    /*
```
SteffenHeu, line 76: What is you reason to remove this code? You are changing something deeply embedded in the MZmine data structure. This is a safety guard so we always have the soring guaranteed in chromatograms and features. This issue should absolutely never occur in regular processing. If this issue occurs something is wrong with the previous processing steps, which may not go unnoticed, because everything in MZmine assumes that chromatograms are sorted.

src/main/java/io/github/mzmine/util/scans/ScanUtils.java
```diff
@@ -678,16 +679,16 @@ public static MassSpectrumType detectSpectrumType(@NotNull double[] mzValues,
     // feature is more than 0.1% of the scan m/z range, it also indicates a
     // centroid spectrum. These criteria are empirical and probably not
     // bulletproof. However, it works for all the test cases we have.
-    if ((mainFeatureDataPointCount < 3) || (mainFeatureMzSpan > (scanMzSpan / 1000.0))) {
-      return MassSpectrumType.CENTROIDED;
+    if (!hasZeroDataPoint && ((mainFeatureDataPointCount < 3)) || (!hasZeroDataPoint && (mainFeatureMzSpan > (scanMzSpan / 1000.0)))) {
+        massSpectrumType =  MassSpectrumType.CENTROIDED;
```
SteffenHeu, line 683: Making the 'centroidness' of the spectrum solely rely on not having zero intensity datapoints is not feasible. Some old versions of MSConvert create mzml files with 0 intenisity datapoints after centroiding.

SteffenHeu (changes requested):
Please explain why you want to make these changes. They change some stuff deeply embedded to the data model. 

What exactly is the issue with waters files? Why does the detection fail currently?

build.gradle
mhhur: I reverted the changes to the Mzmine code on my computer, but I forgot to revert the changes when committing the code to GitHub. The reason for the initial edit was that there were many errors when setting up the Mzmine code on my computer. There are no errors now.

src/main/java/io/github/mzmine/datamodel/featuredata/impl/SimpleIonTimeSeries.java
mhhur: [MCF-report-Mzmine-Waters-raw-issue001.pptx](https://github.com/mzmine/mzmine3/files/10945207/MCF-report-Mzmine-Waters-raw-issue001.pptx)

Thanks for reviewing the code. Please find the attached PowerPoint file containing descriptions of the issues and my corresponding answers. Let me know your thoughts.

SteffenHeu: I'm sorry but I cannot merge this in this state.

If you want to adapt the waters import, you can do that. But this PR does too much internal changes to be merged. As i said, we assume that chromatograms are sorted. If that is not the case, we don't simly remove the check to not ensure that chromatograms are sorted. The correct way to approach this would be to adapt the waters import to sort the scans correctly and fix the issue on the import side.

There is this PR which we started as a GSOC project, which uses the new SDK, but i don't have the time for it. https://github.com/mzmine/mzmine3/pull/763

Otherwise, i suggest to convert the data to mzml with MSConvert or this tool https://github.com/AnP311/Waters2mzML, which has worked for most users so far.

mhhur: Currently, I am collaborating with Ian Morns at Waters to test their mzML conversion tool, TyneBridge. I am working on fixing an issue with the Waters import module and will be committing the updated code once it is resolved.

SteffenHeu: ok great.

thank you.

Issue #1271, opened 2023-03-22
Error while saving raw data files.


calderod:
### Basic information

* My operating system and version: **Windows 10**
* My MZmine version: **3.4.16**

### What happened

* Please describe what happened and how it differs from what you expected to happen.
* **After importing a 120 min HPLC method Thermo RAW file, I have tried saving the project both as a standalone file and as a a referencing file. Both tries give me the "Error while saving raw files" message. After downloading this MZMine program, I made a temporary file directory my C:\ drive like the software recommends, and this error still emerges even if  i try the two importing methods: Raw Data Import > Import MS Data and Raw Data Import > Thermo RAW.**
* Please include all parameter values for the MZmine modules you used. 
* **parameters are shown in the screenshots below.**
* Please upload screenshots of any errors **and** your log/mzmine.log file.
* **Below is a photo of the Set Preferences and the "Save Project As" Dialog boxes.**
![image](https://user-images.githubusercontent.com/91331571/226940291-f1e2c768-ad95-41ed-9ac7-79f553d75299.png)
**This next photo shows the error message.**
![image](https://user-images.githubusercontent.com/91331571/226942316-89244609-cf98-41d3-afee-54b75549f5a3.png)
**Here is the MZMine log file**
[3.22.2023 - Error while saving raw files mzmine_0_0.log](https://github.com/mzmine/mzmine3/files/11041347/3.22.2023.-.Error.while.saving.raw.files.mzmine_0_0.log)

* Please upload a standalone MZmine project file (or a batch + raw files) that demonstrates the problem you are reporting. You can use Google Drive, Dropbox, or a similar repository to upload the project.
* **Since I cannot save a MZMine project file, here is a link to a Google Drive folder that contains the .raw file that is causing this error message to occur. https://drive.google.com/drive/folders/1bZd51ltdEdJvi1g_c0UNEPpxbxWGdre6?usp=sharing**

**Please let me know if more context or files need to be provided to fix this issue.**

SteffenHeu:
Hi calderod, 

i can't really say from that error message and the log file is not really giving me any lead. I have updated the logging of errors in this version: 

Please try to recreate the issue and upload the new log file.

Some things that could be the issue although im not sure: (The log file says "access denied")
- You don't have write access to that directory 
- The path is too long

For these, you can try to save to a different, way shorter file path, e.g. just F:/ to test. (Given your user account has write permissions there). Otherwise try launching MZmine as an admin.

PR #1273, opened 2023-03-23
update logging while project save

SteffenHeu:
attempt to trace #1271


robinschmid (commented):
Thanks for improving the logs

PR #1286, opened 2023-03-29
spectral library tooltip

corinnabrungs:
tooltip for explained library intensity and matched signals


robinschmid (approved):
Thanks for looking into this

PR #1328, opened 2023-05-02
Try to allow for and test multiple java versions

bernt-matthias:
As a follow up on https://github.com/mzmine/mzmine3/issues/1310

No idea if this is the right way to do it.


robinschmid:
I think there is no need to test for multiple JDKs 
As we are not shipping a library but rather a tool

PR #1336, opened 2023-05-11
Add CLI argument to change files easily

robinschmid:
Added arguments i: input to change the input files in batch files easily. It just replaces all the files in the AllSpectralDataImportModule.

Either provide a txt file with paths or use glob patterns
```
-b D:\Data\batch\test.xml -i D:\Data\input_files.txt
-b D:\Data\batch\test.xml -i D:\Data\*.mzML
```

### DIFF
--- a/src/main/java/io/github/mzmine/main/MZmineArgumentParser.java
+++ b/src/main/java/io/github/mzmine/main/MZmineArgumentParser.java
@@ -25,7 +25,9 @@
 
 package io.github.mzmine.main;
 
+import io.github.mzmine.util.files.FileAndPathUtil;
 import java.io.File;
+import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.apache.commons.cli.BasicParser;
@@ -47,6 +49,7 @@ public class MZmineArgumentParser {
   private static final Logger logger = Logger.getLogger(MZmineArgumentParser.class.getName());
 
   private File batchFile;
+  private File[] overrideDataFiles;
   private File preferencesFile;
   private File tempDirectory;
   private boolean isKeepRunningAfterBatch = false;
@@ -61,6 +64,11 @@ public class MZmineArgumentParser {
     Option batch = new Option("b", "batch", true, "batch mode file");
     batch.setRequired(false);
     options.addOption(batch);
+
+    Option input = new Option("i", "input", true,
+        "input files. Either defined in a .txt text file with one file per line or ");
+    input.setRequired(false);
+    options.addOption(input);
 
     Option pref = new Option("p", "pref", true, "preferences file");
     pref.setRequired(false);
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/mzmine/main/MZmineArgumentParser.java
```diff
@@ -102,6 +110,20 @@ public void parse(String[] args) {
         logger.info(() -> "Batch file set by command line: " + sbatch);
         batchFile = new File(sbatch);
       }
+
+      String sinput = cmd.getOptionValue(input.getLongOpt());
+      if (sinput != null) {
+        logger.info(() -> "Input files were set to: " + sinput);
+        // search for files
+        try {
+          overrideDataFiles = FileAndPathUtil.parseFileInputArgument(sinput);
```
SteffenHeu, line 119: maybe only do this if it's headless mode?

src/main/java/io/github/mzmine/main/MZmineArgumentParser.java
```diff
@@ -190,5 +212,9 @@ public boolean isLoadTdfPseudoProfile() {
   public boolean isLoadTsfProfile() {
     return loadTsfProfile;
   }
+
+  public File[] getOverrideDataFiles() {
```
SteffenHeu, line 216: Nullable

src/main/java/io/github/mzmine/modules/batchmode/BatchModeModule.java
```diff
@@ -88,6 +102,16 @@ public static ExitCode runBatch(@NotNull MZmineProject project, File batchFile,
         }
       }
 
+      if (overrideDataFiles != null) {
+        if (!newQueue.setImportFiles(overrideDataFiles)) {
```
SteffenHeu, line 106: the method never returns false but throws an exception instead, so the code below is never reached.

src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
```diff
@@ -225,4 +229,27 @@ public void saveToXml(final Element xmlElement) {
...
     }
   }
+
+  /**
+   * Replace all import files in the {@link AllSpectralDataImportModule} - which needs to be the
+   * first step in batch
+   *
+   * @param allFiles replaces import files
+   */
+  public boolean setImportFiles(final File[] allFiles) {
```
SteffenHeu, line 239: does not return false but throws an exception. Refactor to do one of the two.

SteffenHeu (changes requested):
looks good, but some comments

src/main/java/io/github/mzmine/main/MZmineArgumentParser.java
```diff
@@ -62,6 +65,11 @@ public void parse(String[] args) {
     batch.setRequired(false);
     options.addOption(batch);
 
+    Option input = new Option("i", "input", true,
+        "input files. Either defined in a .txt text file with one file per line or ");
```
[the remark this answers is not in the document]
robinschmid: Changed to
```
 Option input = new Option("i", "input", true,
        """
    input data files. Either defined in a .txt text file with one file per line
    or by glob pattern matching. To match all .mzML files in a path: -i "D:\\Data\\*.mzML"
    """);
```

src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
robinschmid: returns false now. Logging was improved and jdocs as well

src/main/java/io/github/mzmine/modules/batchmode/BatchModeModule.java
robinschmid: returns false now. Logging was improved and jdocs as well

SteffenHeu (approved):

SteffenHeu:
looks good, thanks

PR #1351, opened 2023-05-25
Fixed bug reported by Sanchezillana

christophuv:
Fixed bug reported by Sanchezillana (https://github.com/mzmine/mzmine3/issues/1322#issue-1685003550)
Bug was integer division by <n / n * 100., which is always zero.
Fixed by casting <n to float.


SteffenHeu:
Great, thanks

PR #1363, opened 2023-06-02
Fix elements parameter

Sonja-analytik:
grouping elements in elements parameters


robinschmid (approved):
Looks good. Thanks for spotting this

PR #1391, opened 2023-06-20
Add threads and libraries cli arguments

robinschmid:
We added a libraries arguments to replace spectral libraries
and the threads argument to replace the number of threads easily

### DIFF
--- a/src/main/java/io/github/mzmine/main/MZmineArgumentParser.java
+++ b/src/main/java/io/github/mzmine/main/MZmineArgumentParser.java
@@ -51,6 +51,7 @@ public class MZmineArgumentParser {
 
   private File batchFile;
   private @Nullable File[] overrideDataFiles;
+  private @Nullable File[] overrideSpectralLibrariesFiles;
   private File preferencesFile;
   private File tempDirectory;
   private boolean isKeepRunningAfterBatch = false;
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -188,12 +188,31 @@ public static void main(final String[] args) {
...
+        } else {
+          try {
+            parameter.setValue(Integer.parseInt(numCores));
+          } catch (Exception ex) {
+            logger.log(Level.SEVERE,
+                "Cannot parse command line argument threads (int) set to " + numCores);
+            throw ex;
+          }
+        }
+      }
```
SteffenHeu, line 207: extract method

src/main/java/io/github/mzmine/main/MZmineArgumentParser.java
```diff
@@ -100,6 +109,12 @@ public void parse(String[] args) {
     keepInMemory.setRequired(false);
     options.addOption(keepInMemory);
 
+    Option numCores = new Option("c", "threads", true,
```
SteffenHeu, line 112: having a single letter argument that does not match the first letter of the full parameter is odd.

SteffenHeu: maybe only a long option since t is already taken?

src/main/java/io/github/mzmine/main/MZmineCore.java
robinschmid: done

Issue #1387, opened 2023-06-19
Sorting in scheduler


robinschmid:
i am pretty certain this cannot be correct.
sorting will change the order of the indices and sorting is often done in multiple iterations. 
So the indices will then point to the wrong intensities....
I think you will need to do something like
```java
record IndexedValue(int i, double value){}

IntStream.range(0, intensities.length).map().map(i -> new IndexedValue(i, intensities[i]).sorted(Comparator.comparingDouble(IndexedValue::value).reversed()))
```
I think this is the only good way of sorting two lists together. Similar in python where the lists are zipped together into a list of tuples

_Originally posted by @robinschmid in https://github.com/mzmine/mzmine3/pull/1144#discussion_r1232570529_

vivekd01:
i want to work on this issues please assign me this issue

PR #1402, opened 2023-06-22
Refactor simsef sorting

SteffenHeu:
adresses #1387. Does not change the outcome, worked implemented earlier. Refactored to utility method and updated Simsef scheduling task


src/main/java/io/github/mzmine/modules/tools/timstofmaldiacq/imaging/SimsefImagingSchedulerTask.java
```diff
@@ -336,7 +337,7 @@ private void createNewMsMsSpots(MobilityScanDataAccess access,
         continue;
       }
 
-      if (!checkChimerityScore(access, precursor, minChimerityScore, frame)) {
+      if (!checkPurityScore(access, precursor, minChimerityScore, frame)) {
```
robinschmid, line 340: just checking here. and also in all other cases. We now use purity score - so higher better. 
minChimerityScore is a wrong name then right? its minPurity

robinschmid (approved):
Looks good to me. Just that one variable name. Can be merged after that checked

PR #1427, opened 2023-07-13
check default and set temp dir on cleanup

SteffenHeu:


src/main/java/io/github/mzmine/main/TmpFileCleanup.java
```diff
@@ -59,18 +60,20 @@ public void run() {
...
-              "(.)*" + FeatureListLoadTask.TEMP_FLIST_DATA_FOLDER + "(.)*")) {
-            return true;
-          }
-          return false;
-        }
-      });
+      File[] tempDir = {FileAndPathUtil.getTempDir(),
+          new File(System.getProperty("java.io.tmpdir"))};
+      File remainingTmpFiles[] = Arrays.stream(tempDir)
+          .flatMap(f -> Arrays.stream(f.listFiles(new FilenameFilter() {
```
robinschmid, line 66: i think for this it is better / more readable to do 2 mappings
.map(f -> f.listFiles..)
.flatMap(Arrays::stream)

But we can merge it like this.

PR #1431, opened 2023-07-15
Add RepoRT export 

robinschmid:
- added option to have multiple Databases added to the CompoundDBAnnotation 
- this way we can specify the different IDs in a list


robinschmid:
More work required

PR #1447, opened 2023-07-31
Add speclib name

SteffenHeu:


src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralDBEntry.java
```diff
@@ -93,22 +113,15 @@ public static SpectralLibraryEntry loadFromXML(XMLStreamReader reader) throws XM
...
 
-    // TODO add library here somehow
-    return new SpectralDBEntry(null, mzs, intensities, fields);
-  }
-
-  @Override
-  public void putAll(Map<DBEntryField, Object> fields) {
-    this.fields.putAll(fields);
-  }
+    MZmineCore.getProjectManager().getCurrentProject().getCurrentSpectralLibraries();
```
robinschmid, line 116: why get without using it as a var

src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralDBEntry.java
```diff
@@ -93,22 +113,15 @@ public static SpectralLibraryEntry loadFromXML(XMLStreamReader reader) throws XM
...
-  @Override
-  public boolean putIfNotNull(DBEntryField field, Object value) {
-    if (field != null && value != null) {
-      fields.put(field, value);
-      return true;
+    var entry = new SpectralDBEntry(null, mzs, intensities, fields);
+    if (libraryFileName != null) {
+      final SpectralLibrary library = project.getCurrentSpectralLibraries().stream()
+          .filter(l -> l.getName().equals(libraryFileName)).findFirst().orElse(null);
+      entry.setLibrary(library);
```
robinschmid, line 122: maybe use the constructor with library. Can library field be final then?

robinschmid (changes requested):
Just few comments.

src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralDBEntry.java
SteffenHeu: the library is set when library.addEntry is called. Otherwise all library parsers need to be adapted.

robinschmid: ok

PR #1457, opened 2023-08-10
Scan signal filter - remove mz ranges

corinnabrungs:
This module removes mz ranges from all mass lists. Useful for internal calibrants, contaminants that we typically see in Thermo Orbitrap data in MSn, etc


src/main/java/io/github/mzmine/modules/dataprocessing/filter_scan_signals/ScanSignalRemovalModule.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  public @NotNull MZmineModuleCategory getModuleCategory() {
+    return MZmineModuleCategory.RAWDATAFILTERING;
+  }
+
+  @Override
+  public @NotNull Class<? extends ParameterSet> getParameterSetClass() {
+    return ScanSignalRemovalParameters.class;
+  }
```
SteffenHeu, line 81: I see you added methods for IMS files. if they are tested, you can override the getIonMobilitySupport() method

src/main/java/io/github/mzmine/modules/dataprocessing/filter_scan_signals/ScanSignalRemovalTask.java
```diff
@@ -0,0 +1,170 @@
...
+    for (int i = 0; i < scan.getNumberOfDataPoints(); i++) {
+      double mz = scan.getMzValue(i);
+      if (removeMzRanges.contains(mz)) {
+        continue;
+      }
+
+      double intensity = scan.getIntensityValue(i);
+
+      mzs.add(mz);
+      intensities.add(intensity);
```
SteffenHeu, line 160: I think this is fine for regular LC-MS files, but i think it will take super long for IMS files. Optimising this might be an option using a binary search. On the other hand, i wouldnt know who uses it for IMS files

robinschmid: Could be optimized but maybe not necessary. First, we have to see that IMS people actually use it and that it's too slow.

src/main/java/io/github/mzmine/modules/dataprocessing/filter_scan_signals/ScanSignalRemovalModule.java
corinnabrungs: I tried to make it possible for IMS. But doesn't work. Everything else works :)

robinschmid:
Now I know what was missing. We also need to apply to the frames. Or recompute the frames from the new mobility scans. 
@SteffenHeu what do you think is the best way to solve this? So if the mass lists of the mobility scans change, the frames would need to be recomputed at best right?

SteffenHeu:
i think it's fine to just apply it to the frame spectra.

robinschmid:
Done

PR #1464, opened 2023-08-12
Add WeakListeners to decouple views automatically

robinschmid:
WeakAdapter holds references to all listeners between javafx interfaces and generated WeakListeners as wrappers. Once a tab or stage is closed we need to call dispose and that enables garbage collection on the listeners. 

Easy safe way to get rid of all listeners.


SteffenHeu (dismissed):
looks good

PR #1467, opened 2023-08-16
Remove classes as keys - memory leaks

robinschmid:
Using classes as keys can lead to memory leaks if classes are reloaded.  Never use as keys in a map.  We had duplicates being added. 

Nothing serious happened though. 

https://stackoverflow.com/questions/2625546/is-using-the-class-instance-as-a-map-key-a-best-practice

I tested a couple of workflows. But we should setup more tests to test that everything works.


robinschmid:
updated to latest master

robinschmid:
Did some more testing and it seems to work

robinschmid:
@ansgarkorf @SteffenHeu

src/main/java/io/github/mzmine/datamodel/features/FeatureList.java
```diff
@@ -99,11 +100,48 @@ public interface FeatureList {
...
+    return getFeatureTypes().contains(type);
+  }
+
+  /**
+   * Checks if typeClass was added as a row type
+   *
+   * @param typeClass class of a DataType
+   * @return true if row type is available
+   */
+  default boolean hasRowType(Class typeClass) {
```
SteffenHeu, line 132: can we use templates here to avoid raw types?

src/main/java/io/github/mzmine/datamodel/features/ModularDataModel.java
```diff
@@ -68,8 +74,7 @@ default boolean isEmpty() {
    * @return
    */
   default <T> DataType<T> getTypeColumn(Class<? extends DataType<T>> tclass) {
-    DataType<T> type = getTypes().get(tclass);
-    return type;
+    return DataTypes.get(tclass);
```
SteffenHeu, line 77: Now it has nothing to do with the content of the data model but only with the DataTypes class.
Where do we use this method?

src/main/java/io/github/mzmine/datamodel/features/ModularDataModel.java
```diff
@@ -228,7 +232,18 @@ default <T extends Property<?>> String getFormattedString(Class<? extends DataTy
    * @return true if the new value is different than the old
    */
   default <T> boolean set(DataType<T> type, T value) {
-    return set((Class) type.getClass(), value);
+    Object old = getMap().put(type, value);
```
SteffenHeu, line 235: we don't check where the "type" object actually comes from. if a new instance is created in a loop, we save the new instance all the time, not the "true" instance from DataTypes. Over time this might use unnecessary memory.

src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
```diff
@@ -96,17 +96,23 @@ public class ModularFeature implements Feature, ModularDataModel {
...
+      if (change.wasAdded()) {
+        // do nothing for now
+      } else if (change.wasRemoved()) {
+        // remove type columns to maps
+        DataType type = change.getElementRemoved();
+        this.remove(type);
+      }
+    });
+    //
+    map.addListener((MapChangeListener<? super DataType, ? super Object>) change -> {
```
SteffenHeu, line 111: ah ok this explains why we had to do it manually before :D

src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
```diff
@@ -96,17 +96,23 @@ public class ModularFeature implements Feature, ModularDataModel {
...
+      if (change.wasAdded()) {
+        flist.addFeatureType(change.getKey());
```
SteffenHeu, line 113: i hope the add method is synchronised in the feature list, otherwise parallelised tasks such as speclib search may cause a concurrent modification

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
```diff
@@ -347,16 +347,16 @@ public void applyRowBindings(FeatureListRow row) {
    * @return feature types (columns)
    */
   @Override
-  public ObservableMap<Class<? extends DataType>, DataType> getFeatureTypes() {
+  public ObservableSet<DataType> getFeatureTypes() {
     return featureTypes;
   }
 
   @Override
   public void addFeatureType(Collection<DataType> types) {
```
SteffenHeu, line 355: i think this must be parallelised now

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
```diff
@@ -369,10 +369,10 @@ public void addFeatureType(@NotNull DataType<?>... types) {
   @Override
   public void addRowType(Collection<DataType> types) {
     for (DataType<?> type : types) {
-      if (!rowTypes.containsKey(type.getClass())) {
+      if (!hasRowType(type)) {
```
SteffenHeu, line 372: same

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
```diff
@@ -120,16 +120,21 @@ public ModularFeatureListRow(@NotNull ModularFeatureList flist, int id) {
...
+      } else if (change.wasRemoved()) {
+        // remove type columns to maps
+        DataType type = change.getElementRemoved();
+        this.remove(type);
+      }
+    });
+    //
+    map.addListener((MapChangeListener<? super DataType, ? super Object>) change -> {
+      if (change.wasAdded()) {
+        flist.addRowType(change.getKey());
```
SteffenHeu, line 135: same
also replace the change with the true instance from DataTypes

SteffenHeu (commented):
Got some comments

src/main/java/io/github/mzmine/datamodel/features/types/annotations/formula/FormulaListType.java
```diff
@@ -52,31 +52,22 @@
...
-  public boolean getDefaultVisibility() {
-    return true;
-  }
-
   // Unmodifiable list of all subtypes
   private static final List<DataType> subTypes = List.of(new FormulaListType(),
       new FormulaMassType(), new RdbeType(), new MzPpmDifferenceType(),
       new MzAbsoluteDifferenceType(), new IsotopePatternScoreType(), new MsMsScoreType(),
       new CombinedScoreType());
+  private static final Map<DataType, Function<ResultFormula, Object>> mapper = Map.ofEntries();
```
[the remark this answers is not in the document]
robinschmid: removed

src/main/java/io/github/mzmine/datamodel/features/ModularDataModel.java
robinschmid: removed

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
robinschmid: what do you mean? synchronized?

src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
robinschmid: not sure if it does. it is a set so it can only contain each element once and it would always result in the same type being added.

src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
robinschmid: in the best case this is already our single object. Most cases we use MZType.class.

robinschmid:
Only the synchronization is a question . but it did work so far

SteffenHeu:
did you forget to commit a file? the last commit says '[remove getRowType and getFeatureType from FeatureList](https://github.com/mzmine/mzmine3/pull/1467/commits/fe7e923b9d5ca2785f6cff7888fcde1f88c2286b)' but only removes a call to the method and not the mehtod itself.

robinschmid:
The method was already removed

PR #1468, opened 2023-08-16
Batch - error on missing modules

robinschmid:
In CLI we keep the default behavior of cancelling all.

In the BatchMode dialog if queue is imported - just show error messages for all missing modules.


robinschmid:
Not sure feels like dangerous to let people run a batch file where parts are missing modules

src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
```diff
@@ -142,9 +141,13 @@ public static BatchQueue loadFromXml(final Element xmlElement,
           moduleFound = MZmineCore.getModuleInstance(
               (Class<MZmineModule>) Class.forName(methodName));
         } catch (ClassNotFoundException e) {
-          logger.warning(
-              "Module not found for class " + methodName + " (maybe recreate the batch file)");
-          throw new UnknownModuleNameException(methodName, e);
+          String warning =
+              "Module not found for class " + methodName + " (maybe recreate the batch file)";
```
SteffenHeu, line 145: maybe add the version of the batch and the mzmine version to this error message to give users a better hint what may be wrong.

robinschmid: Done

src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
```diff
@@ -141,8 +142,14 @@ public static BatchQueue loadFromXml(final Element xmlElement,
...
               (Class<MZmineModule>) Class.forName(methodName));
         } catch (ClassNotFoundException e) {
-          String warning =
-              "Module not found for class " + methodName + " (maybe recreate the batch file)";
+          String batchVersionStr =
+              batchMzmineVersion == null ? "of unspecified version" : batchMzmineVersion.toString();
+
+          String warning = """
+              Module not found for class %s (maybe recreate the batch file).
+              Current MZmine version: %s (batch was created with MZmine %s.""".formatted(methodName,
```
SteffenHeu, line 150: close bracket

robinschmid: done

PR #1484, opened 2023-08-31
Add applied methods

SteffenHeu:
CSV export and intra row similarity did not add applied methods


robinschmid (commented):
Thanks for adding this. Maybe we should bake it in more

PR #1498, opened 2023-09-25
remove mztab modules

SteffenHeu:
removes mzTab (not mzTab-m) modules since the repository is down all the time


robinschmid (dismissed):
Looks good to me. This was unfunctional either way right?

PR #1507, opened 2023-09-28
extension of the DatabaseIsotopeRefiner

Sonja-analytik:
with "refineAnnotationsByIsotopesDifferentResolutions" based on the IsotopePatternFinder:

- Calculation of isotope patterns at different resolutions
- comparison of these with the found isotope pattern
- highest isotope pattern score is output


src/main/java/io/github/mzmine/modules/tools/isotopeprediction/IsotopePatternCalculator.java
```diff
@@ -146,9 +148,70 @@ public static IsotopePattern calculateIsotopePattern(IMolecularFormula cdkFormul
     }
   }
 
+  public static HashMap <Double, IsotopePattern> calculateIsotopePatternForResolutions(IMolecularFormula cdkFormula,
```
SteffenHeu, line 151: This looks mostly like a copied method with another loop in it. Replace the copied code with a call to the original method for pattern calculation and just store the resulting pattern in the map. makes it easier to maintain in the end.

SteffenHeu (changes requested):

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/DatabaseIsotopeRefinerScanBased.java
```diff
@@ -88,7 +101,7 @@ public static void refineAnnotationsByIsotopes(List<FeatureListRow> rows, MZTole
    */
   public static void refineAnnotationsByIsotopes(FeatureListRow row, MZTolerance mzTolerance,
       double minIntensity, double minIsotopeScore) {
-    refineAnnotationsByIsotopes(row, mzTolerance, minIntensity, minIsotopeScore, new HashMap<>());
+    refineAnnotationsByIsotopesDifferentResolutions(row, mzTolerance, minIntensity, minIsotopeScore, new HashMap<>());
```
robinschmid, line 104: seems like here you want to call refineAnnotationByIsotopes not DifferentResolution

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/DatabaseIsotopeRefinerScanBased.java
```diff
@@ -152,6 +165,57 @@ public static void refineAnnotationsByIsotopes(FeatureListRow row, MZTolerance m
...
+          var measuredIsotopes = chargeIsotopeMap.getOrDefault(absCharge, null);
+          if (measuredIsotopes == null) {
+            // replace missing pattern by MS1 scan
+            measuredIsotopes = ms1DefaultPattern;
+          }
+          if (measuredIsotopes == null) {
+            // no isotope pattern and no default MS1 scan. e.g., if just a feature list loaded
+            return false;
+          }
+          var score = calculateIsotopeScoreDifferentResolutions(annotation, measuredIsotopes, minIntensity,
```
robinschmid, line 210: Maybe we should save this score somehow in the final annotation so that we can reuse it?

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/DatabaseIsotopeRefinerScanBased.java
```diff
@@ -195,19 +260,57 @@ private static double calculateIsotopeScore(final CompoundDBAnnotation annotatio
...
+      } catch (Exception ex) {
+        logger.log(Level.WARNING,
+            "Cannot match isotope pattern similarity. Maybe no adduct, formula information", ex);
+      }
+    }
+    annotation.put(IsotopePatternScoreType.class, finalScore);
+    return finalScore;
+  }
+  private static double[] getDoubleResolutions(){
+    return new double[]{0.0001,0.001,0.01,0.1};
```
robinschmid, line 306: we can finalize those later but maybe the smallest shoudl be something like 0.00025 or a bit higher?
would be interesting to know what you need to separate N O C S ...

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/DatabaseIsotopeRefinerScanBased.java
```diff
@@ -195,19 +260,57 @@ private static double calculateIsotopeScore(final CompoundDBAnnotation annotatio
...
   }
 
+  private static MZTolerance getMzToleranceFromDoubleResolution(double doubleResolution, double mass){
+    double tolerancePPM =
+        (doubleResolution / mass ) * 1000000;
+    return new MZTolerance(doubleResolution, tolerancePPM);
```
robinschmid, line 312: maybe we can strip this out and directly have a global final MZTolerance[] with all tolerances just as the absolute values (now found in the getDoubleResolutions method).  
The ppm can just be 0 to only rely on absolute tolerance.

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/DatabaseIsotopeRefinerScanBased.java
robinschmid: see comment below. Maybe just creating a global MZtolerance[] might be best

src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/DatabaseIsotopeRefinerScanBased.java
```diff
@@ -195,19 +260,57 @@ private static double calculateIsotopeScore(final CompoundDBAnnotation annotatio
...
+    }
+    // create ion formula
+    IMolecularFormula ionFormula = FormulaUtils.getIonizedFormula(annotation);
+    assert ionFormula != null;
+    // cache the ionformula to IsotopePattern to reuse isotope patterns for the same formula
+    float finalScore = 0;
+    for (double resolution : getDoubleResolutions()) {
+      try {
+        ionIsotopePatternMap.computeIfAbsent(ionFormula, key -> IsotopePatternCalculator.calculateIsotopePatternForResolutions(ionFormula, minIntensity,
+            getDoubleResolutions(), adductType.getCharge(), adductType.getPolarity(), false));
```
robinschmid, line 278: This should be done outside of the loop. Is only done once.
Also you can write `var predictedIsotopePatterns = ionIsotopePatterMap.computeIfAbsent(...);`

robinschmid:
looks good to me

PR #1516, opened 2023-10-17
Kendrick plot fix

kyrasolaris01:
Some fixes to the Kendrick plot module. Simplified the code and now works in dark mode.

image

image


SteffenHeu (changes requested):
The architecture is kind of weird in my opinion. You need the chart and the legend to be created outside the tab. I would like it more if the module created the tab with the parameters and the tab then created a KendrickPane which contains all the information. And ideally the KendrickPane launches tasks to compute the datasets. So you can create a Pane/Tab without manually creating the charts first.

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotChart.java
```diff
@@ -0,0 +1,110 @@
...
+    scaleAxis.setTickLabelPaint(axisPaint);
+    if (colorScaleLabel != null) {
+      scaleAxis.setLabel(colorScaleLabel);
+    }
+    PaintScaleLegend newLegend = new PaintScaleLegend(scale, scaleAxis);
+    newLegend.setPadding(5, 0, 5, 0);
+    newLegend.setStripOutlineVisible(false);
+    newLegend.setAxisLocation(AxisLocation.BOTTOM_OR_LEFT);
+    newLegend.setAxisOffset(5.0);
+    newLegend.setSubdivisionCount(500);
```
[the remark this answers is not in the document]
kyrasolaris01: It ensures a nice gradient without any borders. I copy pasted this from SimpleXYZScatterPlot

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotChart.java
```diff
@@ -0,0 +1,110 @@
...
+    scaleAxis.setLabelPaint(axisPaint);
+    scaleAxis.setTickLabelFont(axisTickLabelFont);
+    scaleAxis.setTickLabelPaint(axisPaint);
+    if (colorScaleLabel != null) {
+      scaleAxis.setLabel(colorScaleLabel);
+    }
+    PaintScaleLegend newLegend = new PaintScaleLegend(scale, scaleAxis);
+    newLegend.setPadding(5, 0, 5, 0);
+    newLegend.setStripOutlineVisible(false);
+    newLegend.setAxisLocation(AxisLocation.BOTTOM_OR_LEFT);
```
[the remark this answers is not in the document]
kyrasolaris01: No, we use BOTTOM_OR_LEFT, but putting the legend on the right side of the plot is not important. I will put it on the right using: newLegend.setPosition(RectangleEdge.RIGHT);

kyrasolaris01:
Changed architecture as requested and made small changes as suggested.

ansgarkorf:
@SteffenHeu can we merge?

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotModule.java
```diff
@@ -60,8 +61,11 @@ public class KendrickMassPlotModule implements MZmineRunnableModule {
   public ExitCode runModule(@NotNull MZmineProject project, @NotNull ParameterSet parameters,
       @NotNull Collection<Task> tasks, @NotNull Instant moduleCallDate) {
 
-    Task newTask = new KendrickMassPlotTask(parameters, moduleCallDate);
-    tasks.add(newTask);
+    // Create Kendrick mass plot Tab
+    MZmineCore.runLater(() -> {
+      KendrickMassPlotTab newTab = new KendrickMassPlotTab(parameters);
+      MZmineCore.getDesktop().addTab(newTab);
```
SteffenHeu, line 67: the add tab call does not need to be in runLater, it's done implicitly. Does the creation need to be on the fx thread?

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotAnchorPaneController.java
```diff
@@ -25,902 +25,347 @@
...
-      height++;
-    } else if (height == 5) {
-      height = 1;
-    } else if (height < 5 && height != 1) {
-      height++;
-    }
-    renderer.setBlockHeightPixel(height);
-    renderer.setBlockWidthPixel(height);
+    KendrickMassPlotXYZDataset kendrickMassPlotXYZDataset = new KendrickMassPlotXYZDataset(
+        parameters, 1, 1);
```
SteffenHeu, line 222: The dataset does not create a thread to calculate the data, does it? It calcs the data on change/when it is created. This will lead to calculation on the FX thread, or?

SteffenHeu (commented):
I think the dataset is not calculated on another thread, but on the fx thread. 
This might lead to freezers or so, i think calculating the dataset in another thread is a better solution. Then the dataset should be changed when a parameter is changed

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotModule.java
kyrasolaris01: done

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotAnchorPaneController.java
kyrasolaris01: now it does create its own thread

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotXYZDataset.java
```diff
@@ -205,13 +298,17 @@ public Number getY(int series, int item) {
 
   @Override
   public Number getZ(int series, int item) {
-    return zValues[item];
+    return colorScaleValues[item];
```
SteffenHeu, line 301: What happens if this is called before the run method is finished? maybe check for the status here and in the other getters?

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotAnchorPaneController.java
```diff
@@ -25,902 +25,350 @@
...
+        }
 
-        // if the divisor is round(R) switch to round(R)-1 for RKM plot
-        zAxisDivisor = checkDivisor(zAxisDivisor, useRKM_Z, customZAxisKMBase, false);
-        PaintScaleLegend legend = (PaintScaleLegend) getChart().getSubtitle(1);
-        legend.getAxis().setLabel("RKM(" + customZAxisKMBase + ")");
+        XYPlot plot = Objects.requireNonNull(getChart()).getXYPlot();
+        KendrickMassPlotXYZDataset newDataset = new KendrickMassPlotXYZDataset(parameters,
+            xAxisDivisor, xAxisCharge, yAxisDivisor, yAxisCharge);
+        newDataset.run();
```
SteffenHeu, line 339: calling the run method here does not run it on a separate thread. you have to add the task to the task controller. probably you need to add a status listener here and update the plot once the calculation is finished (status changed to finished).

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotAnchorPaneController.java
```diff
@@ -25,902 +25,350 @@
...
-
-    this.useCustomXAxisKMBase =
-        parameters.getParameter(KendrickMassPlotParameters.xAxisCustomKendrickMassBase).getValue();
-
-    if (useCustomXAxisKMBase == true) {
-      this.customXAxisKMBase =
-          parameters.getParameter(KendrickMassPlotParameters.xAxisCustomKendrickMassBase)
-              .getEmbeddedParameter().getValue();
+    this.useRKM_X = parameters.getParameter(
+            KendrickMassPlotParameters.xAxisValues).getValue()
```
robinschmid, line 120: use directly parameters.getValue()

SteffenHeu (approved):
looks good to me now 👍

PR #1521, opened 2023-10-25
fix formulae not sorted after flist prediction

SteffenHeu:


src/main/java/io/github/mzmine/modules/dataprocessing/id_formulapredictionfeaturelist/FormulaPredictionFeatureListTask.java
```diff
@@ -241,8 +241,11 @@ public void run() {
         return;
       }
 
+
       // Add the new formula entry top results
       if (!resultingFormulas.isEmpty()) {
+        FormulaUtils.sortFormulaList(resultingFormulas, sortPPMFactor, sortIsotopeFactor,
+            sortMSMSFactor);
         row.setFormulas(resultingFormulas.subList(0,
             Math.min(resultingFormulas.size() - 1, maxBestFormulasPerFeature)));
```
robinschmid, line 250: There is an error in the use of subList the -1 is incorrect.
toIndex is exclusive...

PR #1527, opened 2023-11-02
Parser for structure IDs (smiles, inchi)

mmattano:
A CDK-based parser to translate SMILES to InChI keys and vice-versa

### DIFF
--- a/src/main/java/io/github/mzmine/util/structure/StructureInputType.java
+++ b/src/main/java/io/github/mzmine/util/structure/StructureInputType.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2004-2022 The MZmine Development Team
+ *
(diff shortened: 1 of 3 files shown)


SteffenHeu:
Hi mmattano, thank you for contributing. 

I think the test is currently failing.

As far as i can see, the parser currently is not incorporated into any module.
Where do you see implementation of this parser?

In my opinion, this could be incorporated to the parsing process of csv files in the compound database search module, to add additional metadata.

mmattano:
Hi Steffen, thanks for checking! I worked on this with Robin for the BioHackathon and we were planning to use it in an MS2 fragment structure annotation tool

SteffenHeu:
That's great, thanks 👍

robinschmid:
I incorporated this into one of my branches with substructure support. Will create a new PR soon

PR #1531, opened 2023-11-06
add rt filter for biotransformer

SteffenHeu:


src/main/java/io/github/mzmine/modules/dataprocessing/id_biotransformer/BioTransformerSingleRowTask.java
```diff
@@ -110,12 +122,23 @@ public void run() {
...
+    }
+
     final ModularFeatureList flist = row.getFeatureList();
     for (CompoundDBAnnotation annotation : bioTransformerAnnotations) {
       flist.stream().forEach(r -> {
         final CompoundDBAnnotation clone = annotation.checkMatchAndCalculateDeviation(r,
-            mzTolerance, null, null, null);
+            mzTolerance, rtTolerance, null, null);
         if (clone != null) {
+          if (rowGroupFilter && !(row.getGroupID() == r.getGroupID())) {
```
robinschmid, line 137: Here you want to check if a row is correlated to another row - so grouped. 
The group is the subnetwork ID so being in the same group does not tell us if they are actually correlated.
```java 
    R2RMap<RowsRelationship> ms1Groups = flist.getMs1CorrelationMap();
    RowsRelationship correlation = ms1Groups.get(rowA, rowB);
    if (correlation != null) {
      // rows are grouped by MS1
    }
```

src/main/java/io/github/mzmine/modules/dataprocessing/id_biotransformer/BioTransformerTask.java
```diff
@@ -223,8 +256,14 @@ public void run() {
         for (CompoundDBAnnotation annotation : bioTransformerAnnotations) {
           flist.stream().filter(this::filterProductRow).forEach(r -> {
             final CompoundDBAnnotation clone = annotation.checkMatchAndCalculateDeviation(r,
-                mzTolerance, null, null, null);
+                mzTolerance, rtTolerance, null, null);
             if (clone != null) {
+
+              if (rowGroupFilter && !(row.getGroupID() == r.getGroupID())) {
```
robinschmid, line 262: Here you want to check if a row is correlated to another row - so grouped.
The group is the subnetwork ID so being in the same group does not tell us if they are actually correlated.
```java 
    R2RMap<RowsRelationship> ms1Groups = flist.getMs1CorrelationMap();
    RowsRelationship correlation = ms1Groups.get(rowA, rowB);
    if (correlation != null) {
      // rows are grouped by MS1
    }
```

robinschmid (changes requested):
Only a small comment

src/main/java/io/github/mzmine/modules/dataprocessing/id_biotransformer/BioTransformerSingleRowTask.java
SteffenHeu: The filter is meant to be rather loose, because it is combined with the predicted transformation products. In the gui the parameter is called "Row group filter", so it does what it's meant to do

robinschmid: I think this filter does not really makes sense then. 
The row group really does not say much. 
If you group only based on RT range you may get a single group that contains all rows of that feature list as all fall into the same network.
I think if used here it may also be copied later into other places and give a wrong idea to users and programmers of what this group id really is

robinschmid (approved):
Great

PR #1533, opened 2023-11-07
Add unknown adducts for neutral losses/clusters 

robinschmid:
When we have -18 we just represent this as [M+?]+ and [M-H2O+?]+

Still testing


robinschmid:
Ready

PR #1553, opened 2023-12-07
change chart colors on theme change

SteffenHeu:


kyrasolaris01 (commented):
When I switch from dark to dark custom I get the message box. Should only happen when switching dark to light or light to dark

ansgarkorf (approved):

PR #1562, opened 2023-12-15
speed up gap filling, fix ims gap filling

SteffenHeu:
Ims gap filling had different logic than normal gap filling. 
If the current peak was not finished, more scans outside the rt range were accepted leadingto wierd shapes in edge-cases


robinschmid (approved):
Great improvement

PR #1567, opened 2023-12-21
[FeatureTable] Harmonize all Panes / Nodes that depend on FeatureList or FeatureTable

robinschmid:
This introduces classes to harmonize panes that depend on feature lists.

All bound by weak listeners. If a new feature list is select and changes in featureTableFX -> trigger change of feature list. 

Uses rows for all rows in a feature list 
Uses selectedRows for those selected 

We have a Parent that propagates changes to children in a tree.

So far I only introduce the classes and show how they could be used to simplify the SpectralIdentificationResultsPane and Tab.
Those are not yet used across the software - we would need to do the same with the network etc and then pass around the ParenGroup to set up listeners etc


ansgarkorf:
Sounds very good, I will test it over the next days.

ansgarkorf:
Looks good now!

PR #1572, opened 2024-01-03
Fix MSn tree, reduce to merged or individual scans

robinschmid:
image


robinschmid:
Merged with @ansgarkorf @SteffenHeu

PR #1577, opened 2024-01-04
DataTypes.getAll(Class classes...)

robinschmid:
A way to get multiple datatype class instances at once


src/main/java/io/github/mzmine/datamodel/features/types/DataTypes.java
```diff
@@ -139,6 +139,14 @@ public static <T> DataType<T> get(Class<? extends DataType<T>> clazz) {
     return TYPES.get(clazz);
   }
 
+  /**
+   * Get a list of the singleton datatypes
```
SteffenHeu, line 143: if the order is preserved as the test implies, i would mention that in the javadoc.

PR #1584, opened 2024-01-05
add package to build

SteffenHeu:


robinschmid:
I think we may not need this. At least so far its all running

PR #1610, opened 2024-01-16
add iintest ions, fix ion type in csv search

SteffenHeu:


SteffenHeu:
@robinschmid i can't get the parsing to work, can you check?

robinschmid:
Should work now

PR #1613, opened 2024-01-18
Dont allow multivariate curve resolution module on macos

kyrasolaris01:


robinschmid:
@ansgarkorf this might be a better option?
JNA seems old
https://github.com/mzmine/mzmine3/pull/1614

ansgarkorf:
Thanks Robin, looks also shorter :-)
I will test it quickly.

PR #1614, opened 2024-01-18
Dont allow multivariate curve resolution module on macos

robinschmid:
Maybe lets rather use apache SystemUtils or PlatformUtils from javafx to check OS.


ansgarkorf:
Looks good to me. Can you change it also in TSFUtils, TDFUtils, and ThermoRawImportTask for consistency?

PR #1615, opened 2024-01-19
MassDetectors

robinschmid:
@ansgarkorf can you review. 
Still need to fix the FeatureFindingTest.
Somehow the number of datapoints in mzXML have changed


ansgarkorf:
Looks good to me, can you switch from draft to regular PR?
Can you also make an issue that we have to make an extra module for the isotope mass detection? I suggest to tag Roman in that issue.

PR #1626, opened 2024-01-25
replace bytebuffers by project panama

SteffenHeu:
- drop in replacement, using project panama's memory segments.
- memory segment for now converted to bytebuffer to serve as drop-in


robinschmid:
Still in preview right?
So it may change. 
But this will be amazing to change some of the datastructures.

Does this resolve the file deletion on windows

SteffenHeu:
It's becoming final in JDK 22, i think there are only minor changes between jdk 21 and 22.

> This JEP proposes to finalize the FFM API with further small refinements based upon continued experience and feedback.

Don't think it solves the file deletion issue on windows because they are still mapped files. But we can make temp files larger than 2 gb.
Also can be used to store actual data points on the disk if needed.

PR #1633, opened 2024-02-02
Online reactivity

robinschmid:
## OnlineReactivityModule

This started as an implementation of  online reactivity checks - so if a unknown/known feature reacts in an applied reaction defined as smarts and delta mz.

## ObjectMapper to json and csv

For this I added utilities to export POJOs (e.g. records) to CSV or JSON with jackson object mappers. I updated to use the WriteOptions introduced by @SteffenHeu.

I added the CSVWriter and CSVReader to easily map objects or lists of objects.

Today, during the workshop, we did some live coding, and I used this branch to show how easily we can save a CSV in one line of code. Thats why there are more important changes that I added to this branch.

## Starter example module

I updated the StarterFeatureListModule to be a super easy example for people to start developing in MZmine. This included (finally) some abstract classes for

- AbstractProcessingModule: Easy definition of modules without the need to overwrite the same methods all the time
- TaskPerFeatureListModule: Creates one task per feature list
- TaskPerRawDataFileModule: ...

Tasks
- AbstractFeatureListTask: Easy to use task to process feature lists
- AbstractRawFileTask: ...

Those classes greatly reduce error potential

## File Extension filters

Filters were pushed to ExtensionFilters to reduce the redundancy. Later we should just switch over to use them all over mzmine

@SteffenHeu I think we should get this in


src/main/java/io/github/mzmine/modules/impl/AbstractMZmineModule.java
```diff
@@ -0,0 +1,50 @@
...
+package io.github.mzmine.modules.impl;
+
+import io.github.mzmine.parameters.ParameterSet;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class AbstractMZmineModule {
+
+  protected final @NotNull String name;
+  protected final @Nullable Class<? extends ParameterSet> parameterSetClass;
```
SteffenHeu, line 35: Having this nullable leads to problems at several locations. for example in MZmineCore.setUpAndRunModule. Should be NotNull

src/main/java/io/github/mzmine/modules/impl/AbstractRunnableModule.java
```diff
@@ -0,0 +1,65 @@
...
+  protected final boolean useMemoryMapping;
+
+  public AbstractRunnableModule(@NotNull final String name,
+      @Nullable final Class<? extends ParameterSet> parameterSetClass,
+      @NotNull final String description, @NotNull final MZmineModuleCategory moduleCategory,
+      final boolean useMemoryMapping) {
+    super(name, parameterSetClass);
+    this.description = description;
+    this.moduleCategory = moduleCategory;
+    this.useMemoryMapping = useMemoryMapping;
```
SteffenHeu, line 53: For simplicities sake i would not introduce a boolean for this. A learner will probably not know what to do with it. it should just be hinting at the MemoryMapStorage.forXY methods in the javadoc. (which should be added to the constructor)

src/main/java/io/github/mzmine/modules/impl/TaskPerFeatureListModule.java
```diff
@@ -0,0 +1,103 @@
...
+    if (nFlistParams > 1) {
+      MZmineCore.getDesktop().displayErrorMessage(
+          STR."There are too many (\{nFlistParams}) FeatureListsParameter in class \{getClass().getName()}. Can only have 1. Coding error.");
+      return ExitCode.ERROR;
+    }
+    // exactly one parameter for feature lists found
+    FeatureList[] featureLists = parameters.getValue(featureListParameters.getFirst())
+        .getMatchingFeatureLists();
+
+    MemoryMapStorage storage = useMemoryMapping ? MemoryMapStorage.forFeatureList() : null;
```
SteffenHeu, line 90: the boolean can be true but this can still be null, i think that's counterintuitive.

src/main/java/io/github/mzmine/modules/impl/TaskPerRawDataFileModule.java
```diff
@@ -0,0 +1,103 @@
...
+    if (nRawFiles > 1) {
+      MZmineCore.getDesktop().displayErrorMessage(
+          STR."There are too many (\{nRawFiles}) RawDataFilesParameter in class \{getClass().getName()}. Can only have 1. Coding error.");
+      return ExitCode.ERROR;
+    }
+    // exactly one parameter for RawDataFiles found
+    var rawFiles = parameters.getValue(rawFilesParameter.getFirst()).getMatchingRawDataFiles();
+
+    // raw data processing modules memory map to mass list storage
+    MemoryMapStorage storage = useMemoryMapping ? MemoryMapStorage.forMassList() : null;
```
SteffenHeu, line 90: same

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotXYZDataset.java
```diff
@@ -49,7 +52,9 @@
  */
 public class KendrickMassPlotXYZDataset extends AbstractXYZDataset implements Task,
     XYZBubbleDataset {
+  // TODO replace with getTask method or AbstractTaskXYZDataset
```
SteffenHeu, line 55: i guess we do that later?

src/main/java/io/github/mzmine/parameters/ParameterSet.java
```diff
@@ -239,4 +253,9 @@ default boolean hasParameter(Parameter<?> p) {
     return Arrays.stream(getParameters()).map(Parameter::getName)
         .anyMatch(name -> Objects.equals(p.getName(), name));
   }
+
+  @SuppressWarnings("unchecked")
+  default <T extends Parameter<?>> Stream<T> streamForClass(Class<T> parameterClass) {
```
SteffenHeu, line 258: `V, T extends Parameter<V>`?

src/main/java/io/github/mzmine/taskcontrol/AbstractTask.java
```diff
@@ -29,42 +29,19 @@
...
-import javafx.beans.property.StringProperty;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 /**
  * An abstract implementation of task which defines common methods to make Task implementation
  * easier. Added task status listener
  */
-public abstract class AbstractTask implements Task {
+public abstract class AbstractTask extends AbstractModifiableTask {
```
SteffenHeu, line 40: does this add anything to AbstracModifiableTask? Maybe we can remove one layer of inheritance by deleting one class or the other?

src/main/java/io/github/mzmine/taskcontrol/operations/ParentTaskDependent.java
```diff
@@ -25,20 +25,20 @@
 
 package io.github.mzmine.taskcontrol.operations;
 
-import io.github.mzmine.taskcontrol.AbstractTask;
+import io.github.mzmine.taskcontrol.AbstractModifiableTask;
```
SteffenHeu, line 28: nothing changed here except the class name, so they seem like carbon-copies to me

src/main/java/io/github/mzmine/util/files/ExtensionFilters.java
```diff
@@ -0,0 +1,102 @@
...
+  // LISTS
+  public static final List<ExtensionFilter> ALL_LIBRARY = List.of(ALL_SPECTRAL_LIBRARY_FILTER,
+      JSON_LIBRARY, MGF, MSP, JDCAMX, ALL_FILES);
+  private static final ExtensionFilter MZML = new ExtensionFilter("mzML MS data", "*.mzML",
+      "*.mzml");
+  private static final ExtensionFilter MZXML = new ExtensionFilter("mzXML MS data", "*.mzXML",
+      "*.mzxml");
+  private static final ExtensionFilter IMZML = new ExtensionFilter("imzML MS imaging data",
+      "*.imzML", "*.imzml");
+  private static final ExtensionFilter BRUKER_D = new ExtensionFilter("Bruker tdf files", "*.d");
```
SteffenHeu, line 76: the bruker "files" are not actual files but folders, so you cannot make a filter for it. You could do the .tdf filter, but that's not added here. also there's the .tsf files.

src/main/java/io/github/mzmine/util/io/CsvWriter.java
```diff
@@ -0,0 +1,68 @@
...
+import com.fasterxml.jackson.dataformat.csv.CsvMapper;
+import com.fasterxml.jackson.dataformat.csv.CsvSchema;
+import io.github.mzmine.util.files.FileAndPathUtil;
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import org.jetbrains.annotations.NotNull;
+
+public class CsvWriter {
```
SteffenHeu, line 38: we now have CsvUtils, CsvParsingUtils, CsvWriter, and CsvReader. This should be harmonised to maximum two classes, one for reading and one for writing and moved to a single package to increase visibility of the methods.

It's too scattered like this.

src/main/java/io/github/mzmine/modules/impl/TaskPerFeatureListModule.java
SteffenHeu: just do the forFeatureList here. 

Or rename the boolean to requiresMemoryMapping rather than useMemoryMapping. Might be the best option to cover modules that dont create new data.

SteffenHeu (commented):
looks good, but i have some comments.

src/main/java/io/github/mzmine/datamodel/features/types/TypeStringType.java
```diff
@@ -23,25 +23,21 @@
...
+package io.github.mzmine.datamodel.features.types;
 
-import io.github.mzmine.parameters.Parameter;
-import io.github.mzmine.parameters.impl.SimpleParameterSet;
-import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsParameter;
+import io.github.mzmine.datamodel.features.types.abstr.StringType;
+import org.jetbrains.annotations.NotNull;
 
-public class EmptyFeatureListParameters extends SimpleParameterSet {
+public class TypeStringType extends StringType {
```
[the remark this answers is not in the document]
robinschmid: I added some javadoc. But its really meant to be just a type within a subcolumns factory. e.g., OnlineReaction.Type or Annotation.Type or what ever

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/datasets/AbstractTaskXYDataset.java
```diff
@@ -32,15 +32,20 @@
...
 import javafx.beans.property.Property;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.property.SimpleStringProperty;
 import javafx.beans.property.StringProperty;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.jfree.data.xy.AbstractXYDataset;
 
 public abstract class AbstractTaskXYDataset extends AbstractXYDataset implements Task {
+  // TODO replace with internal getTask method and use AbstractModifiableTask
```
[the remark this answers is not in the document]
robinschmid: just a note of tech debt :D
The implements Task would be nice to convert to an internal Task and using one of the abstract tasks that alrady define all the functions.

src/main/java/io/github/mzmine/modules/dataprocessing/id_online_reactivity/OnlineLcReactivityModule.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @Override
+  public @Nullable Class<? extends ParameterSet> getParameterSetClass() {
+    return OnlineLcReactivityParameters.class;
+  }
+
+  @Override
+  public @NotNull String getDescription() {
+    // TODO add later
```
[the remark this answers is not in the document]
robinschmid: done

src/main/java/io/github/mzmine/modules/dataprocessing/id_online_reactivity/OnlineReactionMatchExportDto.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Used to export matches in SIRIUS and GNPS mgf: {@link SiriusExportTask}
+ * {@link GnpsFbmnExportAndSubmitTask}
+ *
+ * @param reaction  the reaction
+ * @param linkedIds all linked {@link FeatureListRow} IDs
+ */
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+@JsonFormat(with = JsonFormat.Feature.ACCEPT_CASE_INSENSITIVE_PROPERTIES)
+public record OnlineReactionMatchExportDto(Type compoundType,
```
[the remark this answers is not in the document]
robinschmid: DTO Data transfer object.
Just to denote that this object is not used internally but only to transfer data or write to json.
Can also rename

src/main/java/io/github/mzmine/modules/dataprocessing/id_online_reactivity/ReactionMatchingRawFiles.java
```diff
@@ -0,0 +1,33 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.dataprocessing.id_online_reactivity;
+
+import io.github.mzmine.datamodel.RawDataFile;
+import java.util.List;
+
+public record ReactionMatchingRawFiles(OnlineReaction reaction, List<RawDataFile> raws) {
```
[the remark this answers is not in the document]
robinschmid: good catch. you made me think about it.
This is a record that is only used internally in the task for processing purposes. 
So it is not retained but intermediate. 

I changed the visibility to package private so that its used anywhere else. And I added javadoc explaining its use

src/main/java/io/github/mzmine/modules/dataprocessing/id_online_reactivity/OnlineReactionMatchExportDto.java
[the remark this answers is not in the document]
robinschmid: considering the next comment I made this package private

src/main/java/io/github/mzmine/modules/impl/AbstractMZmineModule.java
robinschmid: I know what you mean. MZmineModule.getParameterSetClass is Nullable. Thats why I just overloaded that. 
We could make this notnull but the interface defines it as Nullable. 

What do you think we should do?
Just make it @Notnull and try to find places where it may be null and add parameters?

src/main/java/io/github/mzmine/modules/impl/AbstractRunnableModule.java
```diff
@@ -0,0 +1,65 @@
...
+public abstract class AbstractRunnableModule extends AbstractMZmineModule implements
+    MZmineRunnableModule {
+
+  protected final @NotNull String description;
+  protected final @NotNull MZmineModuleCategory moduleCategory;
+  protected final boolean useMemoryMapping;
+
+  public AbstractRunnableModule(@NotNull final String name,
+      @Nullable final Class<? extends ParameterSet> parameterSetClass,
+      @NotNull final String description, @NotNull final MZmineModuleCategory moduleCategory,
```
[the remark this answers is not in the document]
robinschmid: actually it may be nice to put the description last so that a multi line text block can be used without trailing arguments. What do you think?
Both would be good not sure what I prefer

src/main/java/io/github/mzmine/modules/impl/AbstractRunnableModule.java
robinschmid: I did not fully understand. So what should be the input by the developer into the constructor?
A MemoryMapStorage? like MemoryMapStorage.forFeatureList()

Okay this would remove some abstraction but would also make sure that we actually create a storage. You might be right. I will change it and maybe you can get back to me and tell me if this was the intent.

src/main/java/io/github/mzmine/modules/impl/TaskPerFeatureListModule.java
robinschmid: hm. I just changed the constructor to take a memory map storage but I think thats also not good. If the settings change then we should not do any memory mapping so I guess we need a method to create memory map storage just in the modules that need it. I will change that in the TaskPerFeatureListModule etc.

src/main/java/io/github/mzmine/modules/impl/AbstractRunnableModule.java
robinschmid: SAME AS NEXT COMMENT:

hm. I just changed the constructor to take a memory map storage but I think thats also not good. If the settings change then we should not do any memory mapping so I guess we need a method to create memory map storage just in the modules that need it. I will change that in the TaskPerFeatureListModule etc.

src/main/java/io/github/mzmine/modules/impl/AbstractMZmineModule.java
SteffenHeu: would be the best option i think. but i also think i must have tried to do that already and then decided to not do it for a reason. 😅

src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotXYZDataset.java
robinschmid: Yes thats really just a marker that we need to remove the implements task and change to an internal task object

src/main/java/io/github/mzmine/parameters/ParameterSet.java
robinschmid: True, added.
The T extends Parameter also already gave me what I needed because that parameter knew the type. But yours is better

src/main/java/io/github/mzmine/modules/impl/AbstractRunnableModule.java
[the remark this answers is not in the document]
SteffenHeu: Also true. maybe put it last then

src/main/java/io/github/mzmine/modules/impl/TaskPerFeatureListModule.java
SteffenHeu: I know what you mean. I think renaming the boolean to requiresMemoryMapping is the best option. Then it's clear that not every task needs memory mapping. "useMemoryMapping" sounds more like every task could make use of it, but it's not always good to use it. "requires" is more "some tasks dont need this"

src/main/java/io/github/mzmine/parameters/ParameterSet.java
SteffenHeu: yeah ,but it should be easier to access the value with the extra template, or?

src/main/java/io/github/mzmine/taskcontrol/AbstractTask.java
robinschmid: not sure what my idea was there. I think I wanted a task that has no storage and module call date and really just defines all from the Task interface but this was only used in AbstractTask.
I removed the AbstractModifiableTask

src/main/java/io/github/mzmine/taskcontrol/operations/ParentTaskDependent.java
robinschmid: My initial idea was to have an abstract class without storage and call date - real simple that implements the Task interface strictly. but I reverted that changes and now its Task

src/main/java/io/github/mzmine/util/files/ExtensionFilters.java
robinschmid: I think we generate the All .d buttons with this filter in the Data import parameter dialog. 
This was copied from the AllSpectralDataImportParameters.java.

Maybe we could check in a separate PR how we can improve the capture of bruker file formats? We could do a single filter for tdf / tsf if that makes sense

robinschmid: This was the old code:
```java
  public static final ExtensionFilter ALL_MS_DATA_FILTER = new ExtensionFilter("MS data", "*.mzML",
      "*.mzml", "*.mzXML", "*.mzxml", "*.imzML", "*.imzml", "*.d", "*.raw", "*.RAW", "*.mzData",
      "*.netcdf", "*.mzdata", "*.aird");

  public static final List<ExtensionFilter> extensions = List.of( //
      new ExtensionFilter("mzML MS data", "*.mzML", "*.mzml"), //
      new ExtensionFilter("mzXML MS data", "*.mzXML", "*.mzxml"), //
      new ExtensionFilter("imzML MS imaging data", "*.imzML", "*.imzml"), //
      new ExtensionFilter("Bruker tdf files", "*.d"), //
      new ExtensionFilter("Thermo RAW files", "*.raw", "*.RAW"), //
      new ExtensionFilter("Waters RAW folders", "*.raw", "*.RAW"), //
      new ExtensionFilter("mzData MS data", "*.mzData", "*.mzdata"),
      new ExtensionFilter("aird MS data", "*.aird", "*.Aird", "*.AIRD"), //
      new ExtensionFilter("netCDF", "*.cdf", "*.CDF", "*.netcdf", "*.NETCDF", "*.nc", "*.NC"), //
      new ExtensionFilter("zip", "*.zip", "*.gz"), //
      new ExtensionFilter("All files", "*.*") //
  );

  public static final List<ExtensionFilter> extensionsFolders = List.of( //
      new ExtensionFilter("Bruker tdf files", "*.d"), //
      new ExtensionFilter("Waters RAW folders", "*.raw"), //
      new ExtensionFilter("All files", "*.*") //
  );

```

src/main/java/io/github/mzmine/util/io/CsvWriter.java
robinschmid: Thats what I also thought. We developed these structures in parallel because this was not merged yet. But as soon as its merged we can harmonize it all

src/main/java/io/github/mzmine/modules/impl/AbstractRunnableModule.java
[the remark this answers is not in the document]
robinschmid: done

src/main/java/io/github/mzmine/parameters/ParameterSet.java
robinschmid: Done

src/main/java/io/github/mzmine/util/io/CsvWriter.java
robinschmid: I will create an issue

src/main/java/io/github/mzmine/modules/impl/AbstractMZmineModule.java
robinschmid: At least the null is checked in all places where its used

src/main/java/io/github/mzmine/modules/impl/TaskPerFeatureListModule.java
robinschmid: The memory mapping is defined by the config so I removed this option completely and now we are just calling MemoryMapStorage.forFeatureList etc in the specific modules. 
All other modules can also do this still

robinschmid:
Everything pushed

PR #1653, opened 2024-02-16
Add more columns to spectral library matches

corinnabrungs:
- rt and mz errors
- explained intensity
- comment
- new column order
image

### DIFF
--- a/src/main/java/io/github/mzmine/datamodel/features/types/annotations/SpectralLibraryMatchesType.java
+++ b/src/main/java/io/github/mzmine/datamodel/features/types/annotations/SpectralLibraryMatchesType.java
@@ -40,6 +40,8 @@ import io.github.mzmine.datamodel.features.types.modifiers.AnnotationType;
 import io.github.mzmine.datamodel.features.types.numbers.CCSRelativeErrorType;
 import io.github.mzmine.datamodel.features.types.numbers.CCSType;
 import io.github.mzmine.datamodel.features.types.numbers.MatchingSignalsType;
+import io.github.mzmine.datamodel.features.types.numbers.MzAbsoluteDifferenceType;
+import io.github.mzmine.datamodel.features.types.numbers.MzPpmDifferenceType;
 import io.github.mzmine.datamodel.features.types.numbers.NeutralMassType;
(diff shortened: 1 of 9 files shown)


SteffenHeu (commented):
Looks good, only two minor comments

src/main/java/io/github/mzmine/datamodel/features/types/annotations/SpectralLibraryMatchesType.java
```diff
@@ -65,10 +69,24 @@ public class SpectralLibraryMatchesType extends ListWithSubsType<SpectralDBAnnot
...
+      new CompoundNameType(), //
+      new CosineScoreType(),//
+      new MatchingSignalsType(),//
+      new ExplainedIntensityPercentType(),//
+      new IonAdductType(), //
+      new FormulaType(),//
+      new SmilesStructureType(),//
+      new InChIStructureType(),//
+      new PrecursorMZType(),//
+      new NeutralMassType(),//
```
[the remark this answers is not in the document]
corinnabrungs: how about: neutral mass, precursor mz, diffs,.. it's nice to have neutral mass and precursor mz together

src/main/java/io/github/mzmine/datamodel/features/types/numbers/RtRelativeErrorType.java
```diff
@@ -42,6 +41,6 @@ public RtRelativeErrorType() {
 
   @Override
   public @NotNull String getHeaderString() {
-    return UnitFormat.DIVIDE.format("Rt error", "%");
+    return "Rt error";
```
corinnabrungs, line 44: so I guess this should also be changed without percent

corinnabrungs: @SteffenHeu

SteffenHeu: Looks good. I just double-checked the csv export method and it uses the unique ID, so that's good. Would have been a problem if that used the header, bc then you wouldnt know which unit it is

Issue #1655, opened 2024-02-20
PDF export of javafx nodes / charts


robinschmid:
PDF export is based on itext

change to jfree or openpdf which is an open fork of itext
https://github.com/LibrePDF/OpenPDF

PR #1656, opened 2024-02-20
Pdf export openpdf

SteffenHeu:
Completes #1655

### DIFF
--- a/src/main/java/io/github/mzmine/util/DimensionUnitUtil.java
+++ b/src/main/java/io/github/mzmine/util/DimensionUnitUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2022 The MZmine Development Team
+ * Copyright (c) 2004-2024 The MZmine Development Team
  *
  * Permission is hereby granted, free of charge, to any person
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/mzmine/util/DimensionUnitUtil.java
```diff
@@ -27,52 +27,40 @@
...
-      case PT:
-        return value;
-    }
-    return Float.NaN;
+  public static float getSizeInUnit(float pixel, DimUnit unit) {
+    return switch (unit) { // values derived from unitconverters.net
+      case CM -> getSizeInUnit(pixel, DimUnit.MM) / 10;
+      case MM -> (float) (pixel / 0.352778);
+      case INCH -> (float) (pixel / 0.0138889);
+      case PX, PT -> pixel;
```
robinschmid, line 44: thanks java

robinschmid (approved):
Looks great to me.
Just in the batch export:
- spectral library entries work
- feature shape is empty pdf

Also we should in the future show a dialog or something for a second that says success or failed

PR #1661, opened 2024-02-23
Add volcano plot

SteffenHeu:
Tried to follow the mvci framework as good as possible


src/main/java/io/github/mzmine/datamodel/features/types/numbers/stats/AnovaPValueType.java
```diff
@@ -0,0 +1,60 @@
...
+import io.github.mzmine.datamodel.features.types.numbers.abstr.DoubleType;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
+import org.jetbrains.annotations.NotNull;
+
+public class AnovaPValueType extends DoubleType {
+
+  private static final DecimalFormat format = new DecimalFormat("0.###");
+
+  protected AnovaPValueType() {
```
ansgarkorf, line 37: Why protected? I think it must be public, otherwise the test fails.

src/main/java/io/github/mzmine/datamodel/AbundanceMeasure.java
```diff
@@ -25,9 +25,32 @@
...
+  public Class<? extends DataType<Float>> type() {
+    return type;
+  }
+
+  public Float get(ModularDataModel featureOrRow) {
+    if(featureOrRow == null) {
+      return null;
+    }
+    return featureOrRow.get(type);
+  }
```
robinschmid, line 54: I think its better to have this right on the ModularDataModel?
This would grant the most visible access to the abundance

src/main/java/io/github/mzmine/datamodel/features/FeatureAnnotationPriority.java
```diff
@@ -34,7 +43,43 @@
...
+  /**
+   * Returns the priority ordinal of the given annotation type in this enum or the lowest priority
+   * specified by this enum + 1 (the length of this enum). The smaller the integer, the higher the
+   * priority.
+   */
+  public static <T extends DataType<?> & AnnotationType> int getPriority(T clazz) {
+    final FeatureAnnotationPriority prio = Arrays.stream(FeatureAnnotationPriority.values())
+        .filter(p -> p.type.getClass().getName().equalsIgnoreCase(clazz.getClass().getName()))
+        .findFirst().orElse(null);
+    return prio != null ? prio.ordinal() : FeatureAnnotationPriority.values().length;
```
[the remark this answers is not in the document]
robinschmid: The method is also unused right?

src/main/java/io/github/mzmine/datamodel/features/FeatureAnnotationPriority.java
```diff
@@ -34,7 +43,43 @@
  * {@link FeatureAnnotationIterator}
  */
 public enum FeatureAnnotationPriority {
-  MANUAL, SPECTRAL_LIBRARY, LIPID, EXACT_COMPOUND, FORMULA;
+  MANUAL(ManualAnnotationType.class), LIPID(LipidMatchListType.class), SPECTRAL_LIBRARY(
+      SpectralLibraryMatchesType.class), EXACT_COMPOUND(CompoundDatabaseMatchesType.class), FORMULA(
+      FormulaListType.class);
+  private static final DataType<?>[] dataTypes = Arrays.stream(FeatureAnnotationPriority.values())
+      .map(FeatureAnnotationPriority::getAnnotationType).map(DataTypes::get)
+      .toArray(DataType<?>[]::new);
```
robinschmid, line 51: maybe better to do this on demand in the method?
And the interactors use of this array could be non static so that its actually only in memory if the plot is open. 

but optional

src/main/java/io/github/mzmine/datamodel/features/types/numbers/stats/AnovaPValueType.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public @NotNull String getUniqueID() {
+    return "anova_p_value";
+  }
+
+  @Override
+  public @NotNull String getHeaderString() {
+    return "ANOVA p value";
```
robinschmid, line 48: I think ANOVA p should suffice. Better shorter names.
Not sure if we also have to provide the F value?
so we could also do an ANOVA type and then PValueType und FValueType als sub columns.

src/main/java/io/github/mzmine/datamodel/features/FeatureAnnotationPriority.java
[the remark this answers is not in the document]
SteffenHeu: exhaustive switch will lead to exceptions during sorting when the MissValueType is included

src/main/java/io/github/mzmine/modules/dataanalysis/significance/RowSignificanceTest.java
```diff
@@ -0,0 +1,39 @@
...
+import io.github.mzmine.datamodel.AbundanceMeasure;
+import io.github.mzmine.datamodel.features.FeatureListRow;
+import org.jetbrains.annotations.Nullable;
+
+public interface RowSignificanceTest {
+
+  /**
+   * @return The result of this test or null of the test failed.
+   */
+  @Nullable RowSignificanceTestResult test(FeatureListRow row, AbundanceMeasure abundanceMeasure);
```
robinschmid, line 37: maybe we could use optional here?
I really like the optional pattern. And in the future they may optimize optional even further into a value type with valhalla

src/main/java/io/github/mzmine/modules/dataanalysis/significance/RowSignificanceTestModule.java
```diff
@@ -0,0 +1,74 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * @param <T> The value type required to construct an instance of this test using
+ *            {@link this#getInstance(PropertyComponent)}
+ */
+public interface RowSignificanceTestModule<T> extends MZmineModule {
+
+  enum TESTS {
```
robinschmid, line 45: maybe extract this into its own enum class. 
SignificanceTests or similar.

src/main/java/io/github/mzmine/modules/dataanalysis/significance/RowSignificanceTestProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+import io.github.mzmine.taskcontrol.operations.TaskSubSupplier;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.logging.Logger;
+import org.jetbrains.annotations.NotNull;
+
+public class RowSignificanceTestProcessor<T> extends AbstractTaskSubProcessor implements
+    TaskSubSupplier<List<RowSignificanceTestResult>> {
```
robinschmid, line 40: why not use a SubFunction instead?
input List<FeatureListRow> or even just FeatureListRow and output significance or whatever.
This way you could reuse this object for multiple tests

src/main/java/io/github/mzmine/datamodel/AbundanceMeasure.java
SteffenHeu: Also the modular data model is also used for Database annotations, where it is of no use

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/renderers/ColoredAreaShapeRenderer.java
```diff
@@ -0,0 +1,141 @@
...
+      final int series, final int item, final CrosshairState crosshairState, final int pass) {
+
+    currentDataset = dataSet;
+
+    /*Paint seriesColor = (dataSet instanceof ColoredXYDataset) ?
+        ((ColoredXYDataset) dataSet).getAWTColor() : getItemPaint(series, item);
+    System.out.println(
+        "Drawing dataset " + dataSet.getSeriesKey(0) + "with color " + seriesColor.toString());
+
+    g2.setPaint(seriesColor);*/
```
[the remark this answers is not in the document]
SteffenHeu: i just renamed the class, it was there before. dont remember why

src/main/java/io/github/mzmine/modules/dataanalysis/significance/RowSignificanceTestProcessor.java
SteffenHeu: that would just be a wrapper around the RowSignificanceTest object. THe "run" logic of this class is already super simple

src/main/java/io/github/mzmine/datamodel/AbundanceMeasure.java
robinschmid: Ok can also stay here.

src/main/java/io/github/mzmine/datamodel/features/FeatureAnnotationPriority.java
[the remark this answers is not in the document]
robinschmid: How so? I guess the missing value type should then be sorted to the start or ending. 
But I am not so deep into the code

src/main/java/io/github/mzmine/datamodel/features/FeatureAnnotationPriority.java
robinschmid: ok

src/main/java/io/github/mzmine/gui/chartbasics/simplechart/renderers/ColoredAreaShapeRenderer.java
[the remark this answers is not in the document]
robinschmid: ok

src/main/java/io/github/mzmine/datamodel/features/FeatureAnnotationPriority.java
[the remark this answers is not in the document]
SteffenHeu: because it's not an annotation, so we should not put it there

src/main/java/io/github/mzmine/modules/dataanalysis/significance/RowSignificanceTestModule.java
SteffenHeu: What do you mean? It is an enum

robinschmid: just to extract this into a separate file so that its not RowSignificanceTestModule.TESTS
Maybe SignificanceTestModules.

src/main/java/io/github/mzmine/modules/dataanalysis/significance/anova/AnovaResult.java
```diff
@@ -0,0 +1,34 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.dataanalysis.significance.anova;
+
+import io.github.mzmine.datamodel.features.FeatureListRow;
+import io.github.mzmine.modules.dataanalysis.significance.RowSignificanceTestResult;
+
+public record AnovaResult(FeatureListRow row, String groupingColumn, Double pValue) implements
```
robinschmid, line 31: pvalue can be null? or is there a good default? Otherwise annotate or turn to primitive with default value. I hav no idea but is the F value also provided as a return? then maybe add.

src/main/java/io/github/mzmine/modules/dataanalysis/significance/anova/AnovaTask.java
```diff
@@ -0,0 +1,99 @@
...
+      }
+        processed++;
+      return calc.test(row, AbundanceMeasure.Height);
+    }).filter(Objects::nonNull).toList();
+
+      anovaResults.forEach(r -> r.row().set(AnovaPValueType.class, r.pValue()));
+      flist.getAppliedMethods()
+          .add(new SimpleFeatureListAppliedMethod(AnovaModule.class, parameters, moduleCallDate));
+    } catch (MetadataColumnDoesNotExistException e) {
+      logger.log(Level.WARNING, e.getMessage(), e);
```
robinschmid, line 94: This task is only used from the module - so maybe this is worth an error? 
setErrorMessage and setError. 
Maybe even MZmineCore.getDesktop.displayErrorMessage().

Maybe also great to limit the try catch to the `calc = new AnovaTest()` part so that its clear that the error only comes from here.

src/main/java/io/github/mzmine/modules/dataanalysis/significance/anova/AnovaResult.java
SteffenHeu: refactored. Dont know if we need the f value, also don't think it is returned.

src/main/java/io/github/mzmine/modules/dataanalysis/significance/RowSignificanceTest.java
SteffenHeu: default value does not make sense imo. Not sure about the optional. It's easy to filter with Objects::nonNull in a stream.

src/main/java/io/github/mzmine/modules/dataanalysis/significance/anova/AnovaTask.java
SteffenHeu: done

src/main/java/io/github/mzmine/modules/dataanalysis/significance/anova/AnovaTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    final Map<?, List<RawDataFile>> fileGrouping = metadata.groupFilesByColumn(groupingColumn);
+    groupedFiles = fileGrouping.values().stream().toList();
+  }
+
+  @Nullable
+  private Double oneWayAnova(@NotNull double[][] intensityGroups) {
+
+    int numGroups = intensityGroups.length;
+    long numIntensities = Arrays.stream(intensityGroups).flatMapToDouble(Arrays::stream).count();
```
robinschmid, line 74: not important here but I think mapToInt or mapToLong with Array.length and then sum may be better in other instances to avoid stream concats

robinschmid: looking at all the streams below it feels like summary should be good to do it in one stream. Or just a simple loop:
```java

    Arrays.stream(intensityGroups).map(data -> Arrays.stream(data).summaryStatistics()).toList()
```

robinschmid: Just saw that there is a simple method to use in apache commons math. Maybe better than mangling the values so much. Or was there a reason to not use it? 
There is also a similar method to get the F value. The calling method in AnovaTest could also directly give you a List<double[]> instead of double[][] 

```java

    TestUtils.oneWayAnovaPValue(Arrays.stream(intensityGroups).toList())
```

src/main/java/io/github/mzmine/modules/dataanalysis/significance/anova/AnovaTest.java
```diff
@@ -0,0 +1,137 @@
...
+   */
+  private boolean checkConditions(double[]... groupAbundances) {
+    return groupAbundances.length > 2; // anova usually used for more than two groups
+  }
+
+  @Override
+  public AnovaResult test(FeatureListRow row, AbundanceMeasure abundanceMeasure) {
+    final double[][] intensityGroups = groupedFiles.stream()
+        .map(group -> StatisticUtils.extractAbundance(row, group, abundanceMeasure))
+        .toArray(double[][]::new);
```
robinschmid, line 127: in case you are using the TestUtils from apache math above then change this to a list and you can directly pass it to the function.
TestUtils.oneWayAnovaPValue()

src/main/java/io/github/mzmine/modules/dataanalysis/significance/anova/AnovaTest.java
SteffenHeu: i just reused the existing method. i dont know if there is a reason to not use apache math.

src/main/java/io/github/mzmine/modules/dataanalysis/significance/ttest/Student_tTest.java
```diff
@@ -0,0 +1,146 @@
...
+import io.github.mzmine.modules.dataanalysis.significance.RowSignificanceTest;
+import io.github.mzmine.modules.dataanalysis.significance.RowSignificanceTestResult;
+import io.github.mzmine.modules.dataanalysis.significance.StatisticUtils;
+import io.github.mzmine.modules.visualization.projectmetadata.table.MetadataTable;
+import io.github.mzmine.modules.visualization.projectmetadata.table.columns.MetadataColumn;
+import java.util.List;
+import java.util.Objects;
+import org.apache.commons.math3.stat.inference.TestUtils;
+
+public final class Student_tTest<T> implements RowSignificanceTest {
```
robinschmid, line 41: rename without _
StudentTtest?or TTest

robinschmid, line 41: add javadoc what T is

src/main/java/io/github/mzmine/modules/dataanalysis/significance/ttest/Student_tTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public Student_tTest(TTestSamplingConfig samplingConfig, MetadataColumn<T> column, T groupA,
+      T groupB) {
+    this.samplingConfig = samplingConfig;
+    this.column = column;
+    this.groupA = groupA;
+    this.groupB = groupB;
+
+    groupedFilesA = MZmineCore.getProjectMetadata().getMatchingFiles(column, this.groupA);
+    groupedFilesB = MZmineCore.getProjectMetadata().getMatchingFiles(column, this.groupB);
```
robinschmid, line 58: missing metadata group exception?

src/main/java/io/github/mzmine/modules/dataanalysis/significance/ttest/TTestResult.java
```diff
@@ -0,0 +1,34 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.dataanalysis.significance.ttest;
+
+import io.github.mzmine.datamodel.features.FeatureListRow;
+import io.github.mzmine.modules.dataanalysis.significance.RowSignificanceTestResult;
+
+public record TTestResult(FeatureListRow row, String groupingColumn, double pValue) implements
```
robinschmid, line 31: maybe you already changed that but anova results pValue was Double. Just in case check and also make primitive or annotation

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotController.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+package io.github.mzmine.modules.dataanalysis.volcanoplot;
+
+import io.github.mzmine.datamodel.features.FeatureList;
+import java.util.List;
+import javafx.collections.FXCollections;
+import javafx.scene.layout.Region;
+
+public class VolcanoPlotController {
```
robinschmid, line 33: extend FxController

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotController.java
```diff
@@ -0,0 +1,84 @@
...
+
+  private final VolcanoPlotModel model;
```
robinschmid, line 35: remove once FxController is extended which already defines protected model

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotController.java
```diff
@@ -0,0 +1,84 @@
...
+  private final VolcanoPlotInteractor interactor;
+  private final VolcanoPlotViewBuilder viewBuilder;
+  private final Region view;
+
+  public VolcanoPlotController(List<FeatureList> flists) {
+    final List<FeatureList> alignedLists = flists.stream()
+        .filter(flist -> flist.getNumberOfRawDataFiles() > 1).toList();
+
+    model = new VolcanoPlotModel();
+    model.setFlists(FXCollections.observableArrayList(alignedLists));
```
robinschmid, line 45: what was the idea of having multiple feature lsits here?
To switch fast between them?
Is this a common thing that should be handled within this mvci?
I guess the selection of the feature list could be handled from the checkbox on a tab or just from the parent?
I think most people will have just one feature list they are working on.

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotController.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  private void initialiseListeners() {
+    model.testProperty().addListener((_, _, newValue) -> {
+      if (newValue != null) {
+        computeDataset(() -> {
+        });
+      }
+    });
+    model.flistsProperty().addListener(_ -> {
```
robinschmid, line 60: oh nice so the unamed vars work now?

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotController.java
```diff
@@ -0,0 +1,84 @@
...
+    model.abundanceMeasureProperty().addListener(_ -> {
+      computeDataset(() -> {
+      });
+    });
+    model.selectedFlistProperty().addListener(_ -> {
+      computeDataset(() -> {});
+    });
+  }
+
+  private void computeDataset(Runnable runnable) {
```
robinschmid, line 73: runnable is always empty. So maybe remove and simplify the callers

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotInteractor.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import org.apache.commons.math.util.MathUtils;
+
+public class VolcanoPlotInteractor {
+
+  private static final DataType<?>[] annotationTypeHierarchy = FeatureAnnotationPriority.getDataTypesInOrder();
```
robinschmid, line 57: maybe make this a local var? its already static in the FeatureAnnotationPriority

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotInteractor.java
```diff
@@ -0,0 +1,155 @@
...
+      lastTask.cancel();
+    }
+
+    RowSignificanceTestProcessor<?> processor = new RowSignificanceTestProcessor<>(flist.getRows(),
+        model.getAbundanceMeasure(), test);
+    SimpleCalculation<RowSignificanceTestProcessor<?>> task = new SimpleCalculation<>(processor);
+
+    MZmineCore.getTaskController().addTask(task);
+    lastTask = task;
+    task.setOnFinished(() -> {
```
robinschmid, line 88: task control should be done within the FxController for example the onTaskThread method. 
One missing thing is the check if a task is still the latest active task. Maybe this can be incorporated into the FxController or into the implementation in VolcanoPlotController. What do you think?

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotInteractor.java
```diff
@@ -0,0 +1,155 @@
...
+      List<ProviderAndRenderer> datasets = new ArrayList<>();
+      colors.resetColorCounter(); // set color index to 0
+      for (Entry<DataType<?>, List<RowSignificanceTestResult>> entry : dataTypeMap.entrySet()) {
+        final DataType<?> type = entry.getKey();
+        final List<RowSignificanceTestResult> testResults = entry.getValue();
+
+        final List<RowSignificanceTestResult> significantRows = testResults.stream()
+            .filter(result -> result.pValue() < 0.05).toList();
+        final List<RowSignificanceTestResult> insignificantRows = testResults.stream()
+            .filter(result -> result.pValue() >= 0.05).toList();
```
robinschmid, line 112: would be great to add another property for this 0.05 threshold in the interface as well

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotModel.java
```diff
@@ -0,0 +1,109 @@
...
+
+public class VolcanoPlotModel {
+
+  private final ObjectProperty<FeatureList> selectedFlist = new SimpleObjectProperty<>();
+  private final ObjectProperty<ObservableList<FeatureList>> flists = new SimpleObjectProperty<>();
+  private final ObjectProperty<AbundanceMeasure> abundanceMeasure = new SimpleObjectProperty<>(
+      AbundanceMeasure.Height);
+  private final ObjectProperty<Collection<ProviderAndRenderer>> datasets = new SimpleObjectProperty<>(
+      List.of());
+  private final ObjectProperty<@Nullable RowSignificanceTest> test = new SimpleObjectProperty<>();
```
robinschmid, line 47: maybe there should be always one preselected? then Notnull?

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
```diff
@@ -0,0 +1,205 @@
...
+import org.jfree.chart.annotations.XYLineAnnotation;
+
+public class VolcanoPlotViewBuilder implements Builder<Region> {
+
+  private static final Logger logger = Logger.getLogger(VolcanoPlotViewBuilder.class.getName());
+
+  private final ObjectProperty<@Nullable RowSignificanceTest> test = new SimpleObjectProperty<>();
+
+  private final VolcanoPlotModel model;
+  private final int space = 5;
```
robinschmid, line 69: was thinking about this for a while now. Where could we put this?
FxLayouts.DEFAULT_SPACING or
FxPanes.....
?

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotModel.java
SteffenHeu: no, because  the settings of the test may be wrong or incomplete, so we need to deal with that.

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotController.java
SteffenHeu: That sounds like running circles. why not simply run the update of the model in a Platform.runlater in the interactor?

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
SteffenHeu: No clue if we have so many defaults that such a class would be worth it.

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
```diff
@@ -0,0 +1,205 @@
...
+  @NotNull
+  private HBox createFeatureListBox() {
+    final HBox featureListBox = new HBox(space);
+    final ComboBox<FeatureList> flistBox = new ComboBox<>();
+    flistBox.itemsProperty().bindBidirectional(model.flistsProperty());
+    flistBox.valueProperty().bindBidirectional(model.selectedFlistProperty());
+    featureListBox.getChildren().addAll(new Label("Feature list:"), flistBox);
+    if (!flistBox.getItems().isEmpty()) {
+      flistBox.getSelectionModel().select(0);
+    }
```
robinschmid, line 126: this would be a great candidate for a static factory method to make it easier and more concise in the future.
Like in MenuItems.create

Maybe we should add Fx as a pre or suffix to all to make them better searchable 

Maybe like below with a few overloaded methods to cover all default cases like the one used in createAbundanceBox. Per default  bidrirectional bindings should be great. 
```java
FxComboBoxes.create(label (nullable), itemsProperty, valueProperty, boolean selectFirst) 
```

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
```diff
@@ -0,0 +1,205 @@
...
+
+  @NotNull
+  private HBox createAbundanceBox() {
+    final ComboBox<AbundanceMeasure> abundanceCombo = new ComboBox<>(
+        FXCollections.observableList(List.of(AbundanceMeasure.values())));
+    abundanceCombo.setValue(model.getAbundanceMeasure());
+    model.abundanceMeasureProperty().bindBidirectional(abundanceCombo.valueProperty());
+
+    final HBox abundanceBox = new HBox(space);
+    abundanceBox.getChildren().addAll(new Label("Abundance measure:"), abundanceCombo);
```
robinschmid, line 149: Also I dont really know if this is useful but there is the label.setLabelFor(combo) method...
https://www.tutorialspoint.com/javafx-label-setlabelfor-method-example

Not sure if we should be using this at some places. Seems like you could activate shortcuts - not sure what else comes with it

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
```diff
@@ -0,0 +1,205 @@
...
+    return abundanceBox;
+  }
+
+  private Region createTestParametersPane() {
+
+    final ComboBox<RowSignificanceTestModules> testComboBox = new ComboBox<>(
+        FXCollections.observableList(RowSignificanceTestModules.TWO_GROUP_TESTS));
+
+    FlowPane controls = new FlowPane(Orientation.HORIZONTAL, 5, 5);
```
robinschmid, line 158: use space instead of 5?

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
```diff
@@ -0,0 +1,205 @@
...
+    });
+
+    // listen to external changes of the test and update the combo box accordingly
+    test.addListener((observableValue, rowSignificanceTest, newTest) -> {
+      if (newTest == null) {
+        return;
+      }
+//      if (testComboBox.valueProperty().get().getTestClass().isInstance(newTest)) {
+//        return;
+//      }
```
robinschmid, line 192: remove?

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_smoothing/SmoothingSetupDialog.java
```diff
@@ -79,7 +79,7 @@ public SmoothingSetupDialog(boolean valueCheckRequired, ParameterSet parameters)
     previewChart.setDomainAxisLabel(uf.format("Retention time", "min"));
     previewChart.setRangeAxisNumberFormatOverride(intensityFormat);
     previewChart.setMinHeight(400);
-    smoothedRenderer = new ColoredXYShapeRenderer();
+    smoothedRenderer = new ColoredAreaShapeRenderer();
```
robinschmid, line 82: why did you actually rename this class?
what does area refer to here? or is it just a filled shape that it refers to?

SteffenHeu: Its mostly used for feature areas. so you are not able to give it a specific shape to render, more like that it renders using .fill instead of .drawline

src/main/java/io/github/mzmine/taskcontrol/SimpleCalculation.java
```diff
@@ -0,0 +1,77 @@
...
+import java.time.Instant;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Represents a basic task that can be manged by the MZmine task controller but does not require the
+ * functionality of the more sophisticated {@link AbstractTask} or {@link AbstractSimpleTask}
+ * classes.
+ */
+public class SimpleCalculation<T extends TaskSubProcessor> extends AbstractTask {
```
robinschmid, line 38: Good idea to get rid of the complex constructor
How about adding Task as prefix so that we know that internally its a task.

src/main/java/io/github/mzmine/taskcontrol/SimpleCalculation.java
```diff
@@ -0,0 +1,77 @@
...
+    return processor.getFinishedPercentage();
+  }
+
+  @Override
+  public void run() {
+    setStatus(TaskStatus.PROCESSING);
+    try {
+      processor.process();
+    } catch (RuntimeException e) {
+      logger.log(Level.SEVERE, e.getMessage(), e);
```
robinschmid, line 69: I would say by default a not catched exception should be status error, what do you think?

src/main/java/io/github/mzmine/util/DataTypeUtils.java
```diff
@@ -134,5 +143,100 @@ public static void copyTypes(FeatureList source, FeatureList target, boolean fea
...
+    final T notAnnotatedType = (T) DataTypes.get(MissingValueType.class);
 
+    for (ANY any : collection) {
+      final T bestAnnotationType = getBestTypeWithValue(modularModelExtractor.apply(any),
+          mapMissingValues ? notAnnotatedType : null, allowedTypes);
+      if (bestAnnotationType != null) {
+        final List<ANY> rowsWithAnnotation = results.computeIfAbsent(bestAnnotationType,
+            a -> new ArrayList<>());
+        rowsWithAnnotation.add(any);
+      }
```
robinschmid, line 209: this is duplicated from the above method. Maybe extract this as a private method, use in both methods and as input M provided by extractor.apply(any)

src/main/java/io/github/mzmine/util/DataTypeUtils.java
```diff
@@ -134,5 +143,100 @@ public static void copyTypes(FeatureList source, FeatureList target, boolean fea
...
+   * @param <M>          A modular data model, such as
+   *                     {@link io.github.mzmine.datamodel.features.ModularFeatureListRow} or
+   *                     {@link io.github.mzmine.datamodel.features.ModularFeature}
+   * @return The annotation type or the default.
+   */
+  @Nullable
+  public static <T extends DataType<?>, M extends ModularDataModel> T getBestTypeWithValue(M row,
+      @Nullable T defaultType, T... allowedTypes) {
+    for (T allowedType : allowedTypes) {
+      final var type = (DataType<?>) DataTypes.get(allowedType);
```
robinschmid, line 232: This is not needed here as you already have the data type class. 
only good to call DataTypes.get with the DataType type object if you want to persist this object for long.

src/main/java/io/github/mzmine/modules/dataanalysis/significance/RowSignificanceTestProcessor.java
robinschmid: not sure what is better honestly. Just felt the Supplier + Process adds two method calls that need to go in sequence.

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotController.java
robinschmid: because you dont know if this task is still the latest task that was triggered.

robinschmid: The Controller orchestrates the tasks on the right thread. The interactor just has plain logic.
If the logic is super simple and just one line - you can also omit the interactor and only have a controller but there will always be a controller. 
Thats why the scheduling happens there and not in the interactor

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotInteractor.java
robinschmid: 0.05 or 0.01 or 0.001 is accepted in most fields depending on how large your error can be. 
So we could either have a drop down or just a text field. But it is not always 0.05. 
Some people even argue that those strict values are not even applicable to their field

robinschmid: also its nice in the volcano plot to control highlighting of the features.
Similar option is available in metaboanalyst

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotModel.java
robinschmid: makes sense

src/main/java/io/github/mzmine/parameters/parametertypes/metadata/MetadataGroupSelection.java
```diff
@@ -0,0 +1,89 @@
...
+
+  public static MetadataGroupSelection NONE = new MetadataGroupSelection("", "");
+
+  /**
+   * @return Checks if the current metadata table contains the specified column and the specified
+   * value. Case sensitive.
+   */
+  public boolean isValid() {
+    if (columnName == null || columnName.trim().isEmpty() || groupStr == null
+        || groupStr.isEmpty()) {
```
robinschmid, line 48: you can change trim.isEmpty to isBlank for both strings. This was finally added in jdk11
Also both values canot be null because you would get an exception for passing null to @NotNull?

src/main/java/io/github/mzmine/util/DataTypeUtils.java
SteffenHeu: the point is to return the list of type Any and not the rows

src/main/java/io/github/mzmine/parameters/parametertypes/metadata/MetadataGroupSelectionComponent.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public MetadataGroupSelectionComponent() {
+    super();
+    columnField.setPrefColumnCount(inputSize);
+    final AutoCompletionBinding<MetadataColumn<?>> colBinding = TextFields.bindAutoCompletion(
+        columnField, MZmineCore.getProjectMetadata().getColumns());
+
+    groupField.setPrefColumnCount(inputSize);
+    TextFields.bindAutoCompletion(groupField,
+        (Callback<ISuggestionRequest, Collection<? extends Object>>) iSuggestionRequest -> {
```
robinschmid, line 57: interesting. Never have seen that. 
Do you know if its called for every type event? 
I guess it should be fast enough for this even with 10k samples but just checking. The docs were not so conclusive

src/main/java/io/github/mzmine/parameters/parametertypes/metadata/MetadataGroupSelectionParameter.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @Override
+  public void setValueToComponent(MetadataGroupSelectionComponent metadataGroupSelectionComponent,
+      @Nullable MetadataGroupSelection newValue) {
+
+  }
+
+  @Override
+  public UserParameter<MetadataGroupSelection, MetadataGroupSelectionComponent> cloneParameter() {
+    return null;
```
robinschmid, line 113: implement?
Also the setValue methods above?

src/main/java/io/github/mzmine/parameters/parametertypes/metadata/MetadataGroupingComponent.java
```diff
@@ -0,0 +1,62 @@
...
+import io.github.mzmine.main.MZmineCore;
+import io.github.mzmine.modules.visualization.projectmetadata.ProjectMetadataColumnParameters.AvailableTypes;
+import io.github.mzmine.modules.visualization.projectmetadata.table.columns.MetadataColumn;
+import io.github.mzmine.parameters.PropertyComponent;
+import java.util.List;
+import javafx.beans.property.Property;
+import javafx.scene.control.TextField;
+import org.controlsfx.control.textfield.TextFields;
+
+public class MetadataGroupingComponent extends TextField implements PropertyComponent<String> {
```
robinschmid, line 37: I like the propertycomponent flag.
Maybe we could name it ValuePropertyComponent and later we could add more if there are others. Like EnabledPropertyComponent or similar?

SteffenHeu: renamed

src/main/java/io/github/mzmine/modules/dataanalysis/significance/anova/AnovaResult.java
SteffenHeu: f value is added, i created a new type for the feature list (AnovaResultsType) based on the SimpleSubColumnsType

robinschmid: F value is another method call I think. At least I saw it for ttest

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
robinschmid: Also not sure but it would be great to have the default spacing somewhere in a static field accessible. 
I am always switich between 4 and 5 :D

src/main/java/io/github/mzmine/modules/dataprocessing/featdet_smoothing/SmoothingSetupDialog.java
robinschmid: Maybe we could rename it then to ColoredFilledShapeRenderer? Or just ColoredShapeRenderer?

src/main/java/io/github/mzmine/util/DataTypeUtils.java
robinschmid: ok just felt like the code was duplicated in both methods

robinschmid:
Looks great so far. 
Just tell me if I should look at something again

src/main/java/io/github/mzmine/datamodel/features/FeatureAnnotationPriority.java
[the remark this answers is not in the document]
robinschmid: ok

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
```diff
@@ -0,0 +1,276 @@
...
+    final DoubleComponent pValueComponent = new DoubleComponent(100, 0.0, 1d,
+        new DecimalFormat("0.###"), 0.05);
+    Bindings.bindBidirectional(pValueComponent.getTextField().textProperty(),
+        model.pValueProperty(), new DecimalFormat("0.###"));
+    pValueBox.getChildren().addAll(label, pValueComponent.getTextField());
+    return pValueBox;
+  }
+
+  @NotNull
+  private FlowPane createControlsPane(Region... panes) {
```
ansgarkorf, line 142: Can you use an accordion?

src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
```diff
@@ -0,0 +1,276 @@
...
+    final FlowPane controls = new FlowPane(Orientation.HORIZONTAL);
+    controls.setHgap(space);
+    controls.setVgap(space);
+    controls.getChildren().addAll(panes);
+    controls.setPadding(new Insets(space));
+    controls.setAlignment(Pos.CENTER);
```
ansgarkorf, line 148: align left

PR #1667, opened 2024-02-26
add interfaces for mvci controllers with feature, row, file selections

SteffenHeu:


src/main/java/io/github/mzmine/gui/framework/fx/mvci/SelectedFeaturesController.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package io.github.mzmine.gui.framework.fx.mvci;
+
+import io.github.mzmine.datamodel.features.Feature;
+import java.util.List;
+import javafx.beans.property.ObjectProperty;
+
+public interface SelectedFeaturesController {
+  ObjectProperty<List<Feature>> selectedFeaturesProperty();
```
robinschmid, line 33: I guess we will see once we use them but do you think it should be OjectProperty or ObservableList?
The latter would force a nonnull list that is then bound to another or just provided in the constructor of objects.

PR #1670, opened 2024-02-27
fix min data points not set in gap filling wizard

SteffenHeu:


src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderLcDIA.java
```diff
@@ -119,7 +119,7 @@ public BatchQueue createQueue() {
     makeAndAddIsotopeFinderStep(q);
     makeAndAddJoinAlignmentStep(q, interSampleRtTol);
     makeAndAddRowFilterStep(q);
-    makeAndAddGapFillStep(q, interSampleRtTol);
+    makeAndAddGapFillStep(q, interSampleRtTol, minRtDataPoints);
```
robinschmid, line 122: good catch!
I usually use less data points here. 
Maybe minRTDataPoints/2?
or minRTDataPoints/5*3 or something like that?
What do you think?

robinschmid (dismissed):
Looks good just one decision to be made and you can then merge it directly

src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderLcDIA.java
SteffenHeu: 3/5 it is

PR #1694, opened 2024-03-04
Move components to framework module

robinschmid:


robinschmid:
depends on #1692

mzmine-community/src/main/java/io/github/mzmine/project/ProjectService.java
```diff
@@ -23,13 +23,26 @@
...
-  public static void loadMZmineCore() {
-    MZmineCore.main(new String[] {});
+/**
+ * Contains the project manager and current project. There is always just one project loaded
+ */
+public final class ProjectService {
+
+  @NotNull
+  public static ProjectManager getProjectManager() {
+    return ProjectManagerImpl.getInstance();
```
SteffenHeu, line 39: Currently the ProjectManagerImpl is also a singleton, so this class is basically just another way to the PojectManagerImpl instance.
I think this is one unnecessary layer, or?

mzmine-community/build.gradle
```diff
@@ -257,6 +258,16 @@ tasks.withType(Test) {
             "--enable-preview"
     ]
     useJUnitPlatform()
+
+//    beforeTest { descriptor ->
```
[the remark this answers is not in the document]
robinschmid: This was just for debugging purposes of the test. I could delete it or keep it in case we have other problems in the future. It was printing additional details

mzmine-community/src/main/java/io/github/mzmine/project/ProjectService.java
robinschmid: The project manager is an interface + implementation class. 
This is the safest way for now to decouple all places from the MZmineCore. If the implementation changes in the future we just have to change the instance in the service. 

In the future, we should look into depencency injection frameworks like dagger2 or similar

robinschmid: All other parts function in a similar way now. TaskService has a TaskControllerImpl - if we change that in the future we dont need to change all places. Hopefully

mzmine-community/build.gradle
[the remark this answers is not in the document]
robinschmid: Shall I delete the comment or just add another sentence to it

mzmine-community/src/main/java/io/github/mzmine/project/ProjectService.java
SteffenHeu: ok, let's hope so :D

mzmine-community/build.gradle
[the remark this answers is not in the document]
robinschmid: Added:
Below can be added for extensive logs to debug failing tests

PR #1699, opened 2024-03-06
revert jfreechart version

SteffenHeu:
- new jfreechart version 1.5.4 caused spectra plot to glitch out
- new jfreechart version 1.5.4 caused color to change after the first label was rendered in SimpleCharts
- Range axis sticky zero behaviour was disabled by default in EChartViewer, but the sticky zero behaviour was not applied to the axis. the default jfree behaviour was to have sticky zero enabled. I set the default in EChartViewer to true and actually applied it to the plot. Disabled sticky zero for Kendrick and VanKrevelen


robinschmid (approved):
Looks great. Will be easy to upgrade again in the future by changning just the JFreeChartUtils.getDatasetCountNullable again

PR #1705, opened 2024-03-07
Add PCA

SteffenHeu:
- grafik
- MVCI layout
- added more exposed properties to mvci, added binding factory


javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxComponentFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+import javafx.beans.property.Property;
+import javafx.collections.ObservableList;
+import javafx.scene.control.ComboBox;
+import javafx.scene.control.Label;
+import javafx.scene.layout.HBox;
+
+public class FxComponentFactory {
+
+  public static <T> HBox createLabelledComboBox(String label, ObservableList<T> values,
```
robinschmid, line 36: labeled with one l in american.
I am always confused which to use but just also looked at the LabeledProgressBar

mzmine-community/src/main/java/io/github/mzmine/gui/framework/fx/FxControllerBinding.java
```diff
@@ -0,0 +1,60 @@
...
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.gui.framework.fx;
+
+public sealed interface FxControllerBinding permits SelectedAbundanceMeasureController,
+    SelectedMetadataColumnController, SelectedRowsController, SelectedFeaturesController,
+    SelectedFilesController, SelectedFeatureListsController {
+
+  public static void bindExposedProperties(FxControllerBinding master, FxControllerBinding child) {
```
robinschmid, line 32: is this used somewhere. I like the idea a lot. 
Maybe you could use the same ending for all the classes like the interface:
Binding --> SelectedRowBinding
Or SelectedRowBinder

This would distinguish it from a controller.
Or PropProvider or somthing like that?

SteffenHeu: ok, will change that in the stats overview pr

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/ZCategoryProvider.java
```diff
@@ -0,0 +1,32 @@
...
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.gui.chartbasics.simplechart.providers;
+
+public interface ZCategoryProvider extends XYZValueProvider {
```
robinschmid, line 28: Why Z provider? XYZ already has a Z value - is the category connected to Z?
Maybe add javadoc.

Also do we need extend XYZValueProvider here? This could be standalone embracing composition over inheritance. 

Both implementing classes also implement XYZValueProvider itself

mzmine-community/src/main/java/io/github/mzmine/gui/framework/fx/FxControllerBinding.java
robinschmid: makes sense

mzmine-community/src/main/java/io/github/mzmine/gui/framework/fx/SelectedFilesController.java
```diff
@@ -29,6 +29,6 @@
 import java.util.List;
 import javafx.beans.property.ObjectProperty;
 
-public interface SelectedFilesController {
+public non-sealed interface SelectedFilesController extends FxControllerBinding {
```
robinschmid, line 32: why non-sealed?
Do you think we will extend this? Otherwise just final?

robinschmid: same for the other

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/MainMenu.fxml
```diff
@@ -524,6 +524,9 @@
       <MenuItem text="Volcano plot"
         onAction="#runModule"
         userData="io.github.mzmine.modules.dataanalysis.volcanoplot.VolcanoPlotModule"/>
+      <MenuItem text="Principal component analysis (PCA)"
+        onAction="#runModule"
+        userData="io.github.mzmine.modules.dataanalysis.pca_new.PCAModule"/>
```
robinschmid, line 529: just looked at the dataanalysis package. Maybe we can use common prefixes for the modules to sort them better? Like in the dataprocessing package.
Maybe mv_ for multivariate and uv univariate or multi / uni_

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/LoadingsProvider.java
```diff
@@ -0,0 +1,158 @@
...
+import java.awt.Color;
+import java.util.List;
+import java.util.TreeMap;
+import javafx.beans.property.Property;
+import org.apache.commons.math3.linear.RealMatrix;
+import org.jetbrains.annotations.Nullable;
+import org.jfree.chart.renderer.LookupPaintScale;
+import org.jfree.chart.renderer.PaintScale;
+
+public class LoadingsProvider extends SimpleXYProvider implements PlotXYZDataProvider,
```
robinschmid, line 50: PCALoadingsProvider? also for ScoresProvider

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/LoadingsProvider.java
```diff
@@ -0,0 +1,158 @@
...
+    ZCategoryProvider, XYItemObjectProvider<FeatureListRow> {
+
+  private final PCARowsResult result;
+  private final int loadingsY;
+  private final int loadingsX;
```
robinschmid, line 55: maybe pcIndexX or componentIndexX and Y?
also in the constructor.

I was kind of wondering what loadingsX is

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/LoadingsProvider.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  @Override
+  public void computeValues(Property<TaskStatus> status) {
+    final PCAResult pcaResult = result.pcaResult();
+
+    final RealMatrix loadingsMatrix = pcaResult.getLoadingsMatrix();
+
+    final TreeMap<DataType<?>, List<FeatureListRow>> groupedRows = (TreeMap<DataType<?>, List<FeatureListRow>>) DataTypeUtils.groupByBestDataType(
+        result.rows(), true, FeatureAnnotationPriority.getDataTypesInOrder());
```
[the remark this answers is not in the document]
robinschmid: or you convert the output already to something like
List<RowsByAnnotation> of type
record RowsByAnnotation (List rows, DataType annotationType)

This would make it even easier to grasp than TreeMap but both seems good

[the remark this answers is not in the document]
robinschmid: Okay the code below actually also seems to belong to the same logic. So maybe it would be nice to incorporate the value into the result type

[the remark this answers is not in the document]
robinschmid: Tried something here. 
See incoming PR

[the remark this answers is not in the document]
robinschmid: This PR adds some util methods to avoid indexOf lookups
https://github.com/SteffenHeu/mzmine3/pull/14

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/LoadingsProvider.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  @Override
+  public @Nullable PaintScale getPaintScale() {
+    return paintScale;
+  }
+
+  @Override
+  public double getZValue(int index) {
+    return zCategories[index];
```
robinschmid, line 126: ok thats why its named ZCategoryProvider.
Still feels like it could just be CategoryProvider and then it might also be a 4 D

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAController.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Override
+  public ObjectProperty<List<FeatureListRow>> selectedRowsProperty() {
+    return model.selectedRowsProperty();
+  }
+
+  public ObjectProperty<List<FeatureList>> featureListsProperty() {
+    return model.flistsProperty();
+  }
```
robinschmid, line 62: this is the same as selectedFeatureListsProperty
remove?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAController.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private void initListeners() {
+    model.flistsProperty().addListener(_ -> update());
+    model.domainPcProperty().addListener(_ -> update());
+    model.rangePcProperty().addListener(_ -> update());
+    model.abundanceProperty().addListener(_ -> update());
+    model.flistsProperty().addListener(_ -> update());
```
robinschmid, line 69: flistProperty is twice

robinschmid: I found a solution to the reoccuring update calls to change it to a single binding. This might be nice. Will finalize the PR. 

Also one _ is the invalidation listener. but here you would want to use the change listener. 
Its bad that they use the same method call... I nearly missed that.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCANotComputedException.java
```diff
@@ -0,0 +1,33 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.dataanalysis.pca_new;
+
+public class PCANotComputedException extends RuntimeException {
+
+  public PCANotComputedException() {
+    super("Computation of this PCA has not finished or has not started yet.");
```
robinschmid, line 31: can you add some more description when this happens? as javadoc or here in the exception

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAParameters.java
```diff
@@ -0,0 +1,57 @@
...
+        skipRawDataAndFeatureListParameters);
+    if (!superCheck) {
+      return superCheck;
+    }
+
+    final ModularFeatureList[] matchingFeatureLists = getValue(flist).getMatchingFeatureLists();
+    if (matchingFeatureLists[0].getNumberOfRawDataFiles() < 2) {
+      errorMessages.add("Selected feature list is not an aligned feature list.");
+      return false;
+    }
```
robinschmid, line 54: maybe a general question how we want to handle those errors. 
We could also just open the window and check there and then display an error message. ControlsFX has different options for this. Like a banner on the top (we use it for users to change the tmp directory) or in their OS message on the right

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAResult.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.commons.math3.linear.RealVector;
+import org.apache.commons.math3.linear.SingularValueDecomposition;
+import org.jetbrains.annotations.NotNull;
+
+public record PCAResult(RealMatrix data, RealMatrix dataMeanCentered,
+                        SingularValueDecomposition svd) {
+
+  public RealMatrix firstNComponents(int numComponents) {
+    return svd.getU().getSubMatrix(0, svd.getU().getRowDimension() - 1, 0,
+        numComponents - 1);
```
robinschmid, line 39: maybe add just a few comments. Now you remember:
describing what U is and Rows

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAResult.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  public RealMatrix principalComponentsMatrix() {
+    return svd.getU();
+  }
+
+  public RealMatrix projectDataToScores(int numComponents) {
+    final RealMatrix firstNComponents = firstNComponents(numComponents);
+    final RealMatrix subMatrixS = svd.getS()
```
robinschmid, line 48: same here for S in case you know

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAResult.java
```diff
@@ -0,0 +1,76 @@
...
+  public RealMatrix projectDataToScores(int domainColIndex, int rangeColIndex) {
+    final RealMatrix pcMatrix = pcMatrix(domainColIndex, rangeColIndex);
+    final RealMatrix projected = pcMatrix.multiply(svd.getS().getSubMatrix(0, 1, 0, 1));
+    return projected;
+  }
+
+  @NotNull
+  private RealMatrix pcMatrix(int domainColIndex, int rangeColIndex) {
+    final RealMatrix pcs = svd.getU();
+    final RealVector domainVector = pcs.getColumnVector(domainColIndex);
```
robinschmid, line 63: doc for column vector?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/ZCategoryProvider.java
SteffenHeu: This one just provides the category, not the z value. everything else is handled by the XYZValueProvider

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAController.java
SteffenHeu: I was about to fix that in the next pr. how does it make a difference if its one or the other listener?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAParameters.java
SteffenHeu: Next pr changes the featurelistsparameter.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAUpdateTask.java
```diff
@@ -0,0 +1,145 @@
...
+
+    final LoadingsProvider loadings = new LoadingsProvider(pcaRowsResult, "Loadings", Color.RED,
+        domainPcIndex, rangePcIndex);
+    final ColoredXYZDataset loadingsDS = new ColoredXYZDataset(loadings, RunOption.THIS_THREAD);
+    progressProvider.getAndIncrement();
+
+    loadingsDatasets.add(new DatasetAndRenderer(loadingsDS, new ColoredXYShapeRenderer()));
+    scoresDatasets.add(new DatasetAndRenderer(scoresDS, new ColoredXYShapeRenderer()));
+
+    for (int i = 1; i <= pcaRowsResult.pcaResult().principalComponentsMatrix().getRowDimension();
```
robinschmid, line 110: maybe add a method to pcaResult.getComponentsCount

robinschmid: and maybe directly listComponents

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAUpdateTask.java
```diff
@@ -0,0 +1,145 @@
...
+  protected void updateGuiModel() {
+    model.setScoresDatasets(scoresDatasets);
+    model.setLoadingsDatasets(loadingsDatasets);
+    model.setPcaResult(pcaRowsResult);
+    model.getAvailablePCs().setAll(components);
+
+    if (rangePcIndex < components.size()) {
+      model.setRangePc(rangePcIndex + 1);
+    } else {
+      model.setRangePc(components.getLast() + 1);
```
robinschmid, line 126: why would components.getLast + 1 be the component then?
shouldt it just be domain = 1 and range = 2 as a default?

robinschmid: components also seem to be already +1 so not index but component name

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAViewBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jfree.chart.LegendItemCollection;
+
+public class PCAViewBuilder extends FxViewBuilder<PCAModel> {
+
+  private static final int space = 5;
+
+  private final SimpleXYChart<?> scoresPlot = new SimpleXYChart<>("Scores plot", "PC1", "PC2");
+  private final SimpleXYChart<?> loadingsPlot = new SimpleXYChart<>("Loadings plot", "PC1", "PC2");
+
```
robinschmid, line 60: just keep Scores and Loadings without plot?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAViewBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  private void initDatasetListeners() {
+    model.loadingsDatasetsProperty().addListener(((_, _, newValue) -> {
+      loadingsPlot.applyWithNotifyChanges(false, () -> {
+        loadingsPlot.removeAllDatasets();
+        if (newValue == null || newValue.isEmpty()) {
+          return;
+        }
+        newValue.forEach(d -> loadingsPlot.addDataset(d.dataset(), d.renderer()));
```
robinschmid, line 128: it feels like this is doubled? there is a another call to forEach.addDataset.

Will create a PR that extracts a method from this. because both plots behave the same way I think. If not then no need to merge.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAViewBuilder.java
SteffenHeu: why does that matter?

robinschmid: thought its nice to keep the title short. but optional

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/ScoresProvider.java
```diff
@@ -0,0 +1,176 @@
...
+  @Override
+  public void computeValues(Property<TaskStatus> status) {
+    final PCAResult pcaResult = result.pcaResult();
+
+    final RealMatrix scores = pcaResult.projectDataToScores(pcX, pcY);
+
+    final List<RawDataFile> files = result.files();
+    final Map<?, List<RawDataFile>> groupedFiles =
+        groupingColumn != null ? MZmineCore.getProjectMetadata().groupFilesByColumn(groupingColumn)
+            : Map.of();
```
robinschmid, line 91: maybe make the function accept null and just return Map.of in that case. Might make it easier in the future.

Do we need to capture the missing metadata exception here or ist it ok to pass through to the caller?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/ScoresProvider.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  @Override
+  public int getNumberOfCategories() {
+    return numberOfCategories;
+  }
+
+  @Override
+  public String getLegendLabel(int category) {
+    return groupNames[category] != null ? groupNames[category] : "unnamed group";
```
robinschmid, line 169: when is the groupNames[] == null? 
is this a safe guard or does the map that initializes groupName contain a null key? 
Otherwise you could set the "unnamed group" before.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/rowsboxplot/RowBarDataset.java
```diff
@@ -0,0 +1,63 @@
...
+import io.github.mzmine.main.MZmineCore;
+import io.github.mzmine.modules.visualization.projectmetadata.table.MetadataTable;
+import io.github.mzmine.modules.visualization.projectmetadata.table.columns.MetadataColumn;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import org.jetbrains.annotations.Nullable;
+import org.jfree.data.statistics.DefaultBoxAndWhiskerCategoryDataset;
+
+public class RowBarDataset extends DefaultBoxAndWhiskerCategoryDataset {
```
robinschmid, line 41: better name it RowBoxPlotDataSet or RowBoxWiskersDataset.
easier to find

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/rowsboxplot/RowsBoxplotController.java
```diff
@@ -0,0 +1,83 @@
...
+        .addListener((_, _, n) -> updateDataset(model.getSelectedRows()));
+    model.groupingColumnProperty().addListener((_, _, n) -> updateDataset(model.getSelectedRows()));
+  }
+
+  private void updateDataset(List<FeatureListRow> n) {
+    if (n == null || n.isEmpty()) {
+      model.setDataset(null);
+      return;
+    }
+    model.setDataset(
```
robinschmid, line 60: onGuiThread 
generally model might changeon different thread so safe guard here

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/rowsboxplot/RowsBoxplotController.java
```diff
@@ -0,0 +1,83 @@
...
+public class RowsBoxplotController extends FxController<RowsBoxplotModel> implements
+    SelectedRowsController, SelectedMetadataColumnController, SelectedAbundanceMeasureController {
+
+  private final RowsBoxplotViewBuilder builder;
+
+  public RowsBoxplotController() {
+    super(new RowsBoxplotModel());
+    builder = new RowsBoxplotViewBuilder(model);
+
+    model.selectedRowsProperty().addListener((_, _, n) -> updateDataset(n));
```
robinschmid, line 49: instead of giving the list as argument just take it from the model within the updateDataset method. This avoids that you have to do it in every listener.
You could also use the new LastChangeProperty here. If you dont like to put this binding in the model, you could also create this property here locally just for the binding.
This would reduce it to one line of code

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/rowsboxplot/RowsBoxplotModel.java
```diff
@@ -0,0 +1,90 @@
...
+
+import io.github.mzmine.datamodel.AbundanceMeasure;
+import io.github.mzmine.datamodel.features.FeatureListRow;
+import io.github.mzmine.modules.visualization.projectmetadata.table.columns.MetadataColumn;
+import java.util.List;
+import javafx.beans.property.ObjectProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import org.jetbrains.annotations.Nullable;
+
+public class RowsBoxplotModel {
```
robinschmid, line 36: its so nice that I can just read the properties and then jump over the rest as its auto generated :D
So other classes are not so polluted

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/rowsboxplot/RowsBoxplotViewBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+import io.github.mzmine.javafx.mvci.FxViewBuilder;
+import io.github.mzmine.main.MZmineCore;
+import java.util.List;
+import javafx.scene.layout.BorderPane;
+import javafx.scene.layout.Region;
+import org.jfree.chart.ChartFactory;
+import org.jfree.chart.JFreeChart;
+import org.jfree.chart.renderer.category.BoxAndWhiskerRenderer;
+
+public class RowsBoxplotViewBuilder extends FxViewBuilder<RowsBoxplotModel> {
```
robinschmid, line 40: In future work it would be great to add teh control panels at the bottom in an accordion to also adjust and control the boxplot in case it is standalone. There could be a property in the model if the accordion should be visible or not.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/StatisticUtils.java
```diff
@@ -50,13 +61,106 @@ public static double[] calculateLog2FoldChange(List<RowSignificanceTestResult> t
...
+      for (int rowIndex = 0; rowIndex < rows.size(); rowIndex++) {
+        final FeatureListRow row = rows.get(rowIndex);
+        final Feature feature = row.getFeature(file);
+
+        final double abundance;
+        if (feature != null) {
+          abundance = measure.get((ModularDataModel) feature);
+        } else {
+          abundance = Double.NaN;
+        }
```
robinschmid, line 159: this will produce NaN for missing features and null for missing data type.
Maybe enough to just do this and always get null:
```java
// cast of null is allowed and will remain null
double abundance = measure.get(ModularDataModel) feature);
```

robinschmid (changes requested):
This is amazing. The code reads very well and this can be a blueprint for further development.

I added three isolated PR to your branch. 
https://github.com/SteffenHeu/mzmine3/pulls

mzmine-community/src/main/java/io/github/mzmine/gui/framework/fx/FxControllerBinding.java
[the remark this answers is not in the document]
SteffenHeu: i did not need to cast anything yet, but i can add a second method

mzmine-community/src/main/java/io/github/mzmine/gui/framework/fx/SelectedFilesController.java
SteffenHeu: because we would have to put all controllers in this package then

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/MainMenu.fxml
SteffenHeu: can do at some point. the thing is that the others will be removed anyway

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/LoadingsProvider.java
SteffenHeu: then we would need another dataset for it, because we cannot use the z value for the category anymore. Then we also need a new renderer. 
This is the better solution

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCANotComputedException.java
SteffenHeu: not used anymore

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAUpdateTask.java
SteffenHeu: last +1 does not make sense, yes.

> components also seem to be already +1 so not index but component name

where? 1 is subtracted in the constructor

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/ScoresProvider.java
SteffenHeu: the column is null if it does not exist due to the metadatagroupingcomponent. In the next PR it automaticall changes the property to MEtadatadaColumn instead of string. so wont hcange here.

SteffenHeu: adapted to return Map.of on null

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/ScoresProvider.java
SteffenHeu: just a safety measure.

SteffenHeu: i guess only if object.tostring returns null.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/rowsboxplot/RowBarDataset.java
SteffenHeu: next pr

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/rowsboxplot/RowsBoxplotViewBuilder.java
SteffenHeu: yes could do. but just a boxplot is al ittle few info, so i dont know if we will do that

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/StatisticUtils.java
```diff
@@ -50,13 +61,106 @@ public static double[] calculateLog2FoldChange(List<RowSignificanceTestResult> t
...
+
+    RealMatrix result = inPlace ? data
+        : new Array2DRowRealMatrix(data.getRowDimension(), data.getColumnDimension());
+
+    for (int col = 0; col < data.getColumnDimension(); col++) {
+      final RealVector columnVector = data.getColumnVector(col);
+//      double sum = 0;
+//      for (int row = 0; row < columnVector.getDimension(); row++) {
+//        sum += columnVector.getEntry(row);
+//      }
```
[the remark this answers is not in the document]
SteffenHeu: Will write a test for the methods in the future, so i want to test against this at some point

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/StatisticUtils.java
SteffenHeu: can a double be null? i dont think so. this is indeed a problem. the matrix cannot contain null, so we need to be able to know if a feature is missing. Double.NaN can do that. I propose to use Double.Nan also in line 156 if the type is missing.

mzmine-community/src/main/java/io/github/mzmine/gui/framework/fx/FxControllerBinding.java
SteffenHeu: done in dashboard pr

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/LoadingsProvider.java
robinschmid: The use here is fine. It was more about the naming of the ZCategoryProvider.
Just felt like the class just provides a category for an item. Like the ItemProvider provides an item like a row. 

I mean I could have an XYZ dataset and then still add the ZCategoryProvider as the 4th dimension or just additional info. So renaming would make that clearer that it does not need to be coupled to Z

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/ZCategoryProvider.java
robinschmid: So maybe the Z could be dropped from the name to make clear that it just provides a Category?
Also dropping the extends XYZValueProvider

mzmine-community/src/main/java/io/github/mzmine/gui/framework/fx/FxControllerBinding.java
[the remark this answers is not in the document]
robinschmid: Was just thinking because you are not using any specific method in here and the instanceof checks are kind of needed anyway

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/ScoresProvider.java
robinschmid: Nice

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/rowsboxplot/RowsBoxplotViewBuilder.java
robinschmid: Lets see when the need arises

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/StatisticUtils.java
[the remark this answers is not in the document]
robinschmid: ok thought it was duplicated

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/StatisticUtils.java
robinschmid: oh sorry. Double abundance of course. but if it needs notnull then I would prefer adding a getOrElse method to the AbundanceMeasure.
Could even simplify and add another method: getOrElseNaN(feature)

I guess this will be used in many places and always catching the different cases would be annoying?

SteffenHeu: I already added a get or nan method

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAController.java
robinschmid: I think repeating those calls like this invite copy paste error and are redundant. Thats why it would be nice to use a binding and just one line of code. 
Could also be a utils method that takes a Runnable and Observable... to run on changes. All better than repeating the whole listener.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/StatisticUtils.java
robinschmid: ok nice. It did not show but updated again.
You could still simplify the whole call to `double abundance = measure.get(ModularDataModel) feature);`
instead of if null else.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCATab.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Override
+  public @NotNull Collection<? extends RawDataFile> getRawDataFiles() {
+    return List.of();
+  }
+
+  @Override
+  public @NotNull Collection<? extends FeatureList> getFeatureLists() {
+    return controller.featureListsProperty().get();
```
robinschmid, line 55: all the calls to featureListProperty need to be changed to selectedFeatureListsProperty

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAUpdateTask.java
robinschmid: Ok makes sense.
The other part of the question was just if both should be set to getLast then or rather to PC1 and PC2? both are good though

robinschmid:
Looks great. Can be merged once the compile conflict is resolved and you took final decision on the last comments. Feel free to decide.

PR #1713, opened 2024-03-08
change average height and area to maximum

corinnabrungs:
contains PR #1712

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/RelativeHeightType.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/RelativeHeightType.java
@@ -30,8 +30,9 @@ import org.jetbrains.annotations.NotNull;
 
 /**
  * Relative height may be added to the Feature or FeatureListRow to identify the significance of a
- * feature. This is important for spectral library export to filter out minor features with relative
- * intensity cutoff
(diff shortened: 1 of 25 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/RelativeHeightType.java
```diff
@@ -0,0 +1,54 @@
...
+
+import io.github.mzmine.datamodel.features.types.numbers.abstr.PercentType;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Relative height may be added to the Feature or FeatureListRow to identify the significance of a
+ * feature. This is important for spectral library export to filter out minor features with relative
+ * intensity cutoff
+ */
+public class RelativeHeightType extends PercentType {
```
SteffenHeu, line 36: What is the plan for this type? It is currently not set anywhere. The javadoc should specify what the height is set in relation to. 

At the full tic of the most intense spectrum of the feature? 
Relative to the highest peak in the most intense spectrum?

or in the case of library building maybe at the time the ms2 was taken?

corinnabrungs: The plan is for the library building. If multiple features have the same m/z, how do we deal with those? So it's relative to the highest feature with the same m/z, and a customized cutoff can be set to remove those features for library export.

PR #1715, opened 2024-03-09
Gc aligner

robinschmid:
Based on PR #1710 by @kyrasolaris01 

Creating another PR to keep the two versions separate and comparable.

- Much of the heavy scheduling of the alignment of rows onto base rows was extracted into a BaseFeatureListAligner. 
- The different scoring for GC and Join is split into FeatureRowAlignScorer implementations. 
- FeatureCloner defines how features are cloned - either just cloned or reextracted in the case of GC if feature has a different mz than the row.
- The module and class of Gc and JoinAligner are now precisely the same, ignoring minor naming differences.


mzmine-community/src/main/java/io/github/mzmine/util/FeatureListUtils.java
```diff
@@ -426,6 +441,48 @@ public static double getAlignmentScore(Double testMz, Float testRt, Float testMo
...
+    // don't score range.all, will distort the scoring.
+    rtRange = rtRange == null || rtRange.equals(Range.all()) ? null : rtRange;
+
+    int scorers = 0;
+
+    double score = 0f;
+    // values are "matched" if the given value exists in this class and falls within the tolerance.
+
+    if (rtWeight > 0 && rtRange != null && testRt != null) {
+      final Float rt = RangeUtils.rangeCenter(rtRange);
```
SteffenHeu, line 468: Range.center does not work with range.all. I think we should disable the 0 weight option in the dialog.

SteffenHeu (dismissed):
2 comments only, looks good

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/align_gc/GCAlignerTask.java
```diff
@@ -0,0 +1,119 @@
...
+  protected void process() {
+
+    logger.info(() -> "Running parallel GC aligner on " + featureLists.size() + " feature lists.");
+
+    var mzTolerance = parameters.getValue(GCAlignerParameters.MZ_TOLERANCE);
+    FeatureCloner featureCloner = new ExtractMzMismatchFeatureCloner(mzTolerance);
+    // create the row aligner that handles the scoring
+    var rowAligner = new GcRowAlignScorer(parameters);
+    listAligner = new BaseFeatureListAligner(this, featureLists, featureListName,
+        getMemoryMapStorage(), rowAligner, featureCloner, MZ_ASCENDING);
```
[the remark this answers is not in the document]
robinschmid: ups. 
I checked if I added all changes multiple times but this one slipped my attention

mzmine-community/src/main/java/io/github/mzmine/util/FeatureListUtils.java
robinschmid: Good that you marked this. I looked through the scores again and the GC score was not correct yet. So I simplified the whole calculation, reducing redundant code into functions. We now accumulate the score in the ScoreAccumulator. 

The new GC score used 1-similarity but this should be just similarity.

robinschmid:
@ansgarkorf I fixed the GC score. It seems the similarity was inverted. 
Can you check the results on your side

SteffenHeu:
whats different than before?

ansgarkorf:
Speed is good, was something else! Looks all good!

PR #1719, opened 2024-03-13
add water loss as neutral loss option

corinnabrungs:


mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonModification.java
```diff
@@ -140,9 +145,11 @@ public class IonModification extends NeutralMolecule implements Comparable<IonMo
 
   // default values
   public static final IonModification[] DEFAULT_VALUES_POSITIVE = {H_NEG, M_PLUS, H, NA, K, NH4,
+      H2O, M_PLUS_H2O, H_H2O_1, H_H2O_2, H_H2O_3,
```
robinschmid, line 148: Can you reorder?
put H_NEG, and H2O at the end.
then put the water los of M+ or M+H before or after their respective counter part.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonModification.java
```diff
@@ -140,9 +145,11 @@ public class IonModification extends NeutralMolecule implements Comparable<IonMo
...
       M2plus, H2plus, CA, FE, MG, NA_H, NH4_H, K_H, Hneg_NA2, Hneg_CA, Hneg_FE,
       M_PLUS_ALREADY_CHARGED, Hneg_MG};
-  public static final IonModification[] DEFAULT_VALUES_NEGATIVE = {M_MINUS, H_NEG, NA_2H, NA, CL,
+  public static final IonModification[] DEFAULT_VALUES_NEGATIVE = {M_MINUS, H_NEG, NA_2H, H2O, NA,
```
robinschmid, line 151: same here, lets put H2O and Na at the end. Its not meant to be selected but only used for remixing adducts

robinschmid (dismissed):
Looks good just a small comment above and another for the wizard:

You could look into the BaseWizardBatchBuilder.createAndSetIonLibrary here are the default adducts and modifications selected. Maybe add the M-H2O+H and remove the general water loss here. This would be more conservative.

corinnabrungs:
Ah good to know where the default settings are.

robinschmid (approved):
Looks good to me. Thanks @corinnabrungs

PR #1724, opened 2024-03-13
Patch library match dialog

robinschmid:
still need to add the missing dependency in the future but it works for now.


SteffenHeu:
I wouldnt consider this a solution :D

robinschmid:
Postponed the real fix for home. 
Just updating the cdk to 2.9 fixes the issue. 
Structures work well and I added a ParserTest in PR #1727

PR #1727, opened 2024-03-14
Structure in table 

robinschmid:
Added StructureParser and StructureRenderService. We can use one renderer to render to multiple canvas.

I only quickly updated the Structure2DComponent - those old classes are a bit messy but I cleaned it a bit and  made it more javafx like with property. 

image


robinschmid:
Font size in spec match pane is adjusted
image

robinschmid:
Update of CDK to 2.9 fixes the missing log4j dep. 
Structures are still visible

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
```diff
@@ -0,0 +1,113 @@
...
+    return column;
+  }
+
+  @Override
+  public @Nullable Runnable getDoubleClickAction(@NotNull final ModularFeatureListRow row,
+      @NotNull final List<RawDataFile> file, @Nullable final DataType<?> superType,
+      final @Nullable Object value) {
+    return () -> FxThread.runLater(() -> {
+      if (value instanceof MolecularStructure structure) {
+        new MolStructureViewer("", structure.structure()).show();
```
SteffenHeu, line 104: i think this should open the spectral match pane or the comdb pane. You have the structre there too and you also have the other information.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/StructureTableCell.java
```diff
@@ -0,0 +1,29 @@
...
+
+import io.github.mzmine.datamodel.features.types.modifiers.GraphicalColumType;
+import io.github.mzmine.datamodel.structures.MolecularStructure;
+import java.util.logging.Logger;
+import javafx.beans.binding.Bindings;
+import javafx.scene.Node;
+import javafx.scene.control.ContentDisplay;
+import javafx.scene.control.TreeTableCell;
+
+public class StructureTableCell<S> extends TreeTableCell<S, Object> {
```
SteffenHeu, line 11: Ah ok this is how you are supposed to do it. Ill check if we can do that for the charts as well

SteffenHeu (commented):
Looks good, but please add a way to load and save the new data type and a test for it

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Override
+  public @Nullable Node createCellContent(@NotNull final ModularFeatureListRow row,
+      final MolecularStructure cellData, @Nullable final RawDataFile raw,
+      final AtomicDouble progress) {
+    if (cellData == null) {
+      return null;
+    }
+// not really used as we are setting a different cell factory
+    return new Structure2DComponent(cellData.structure());
```
SteffenHeu, line 86: if this is not supposed to happen we should throw an exception, or?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/CompoundNameType.java
```diff
@@ -62,6 +62,11 @@ public final String getUniqueID() {
 
   @Override
   public boolean getDefaultVisibility() {
-    return true;
+    return false;
+  }
+
+  @Override
+  public int getPrefColumnWidth() {
+    return 350;
```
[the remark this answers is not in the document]
robinschmid: I think we should find a way to hide it in the table but its needed for the export I would say to really get the library entry name there

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.List;
+import java.util.logging.Logger;
+import javafx.beans.property.Property;
+import javafx.beans.property.SimpleObjectProperty;
+import javafx.scene.Node;
+import javafx.scene.control.TreeTableColumn;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class MolecularStructureType extends DataType<MolecularStructure> implements
```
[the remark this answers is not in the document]
robinschmid: Good that you say that. I did not test it yet. I wanted to avoid export for this type because its calculated on demand for existing smiles and inchi. So it is more like a cached CDK structure for visualization and maybe other operations. 

Do we already have an interface that regulates that a type is not exported? 
Graphical is not exported right?

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/MolStructureViewer.java
```diff
@@ -66,7 +70,7 @@ public MolStructureViewer(String name, final URL structure2DAddress,
       final URL structure3DAddress) {
 
     setTitle("Structure of " + name);
-    setupViewer(name);
+    setupViewer(name, structure3DAddress!=null);
```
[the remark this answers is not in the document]
robinschmid: Yeah I mostly reused this old class to just have a way to look closer at a structure. 
But you are right if a user double clicks it would be nice to open the spectral library match or compound match. That is what you propose right?
Should be even possible we get the row and the super type so it could be easy to check and open the corresponding match tab instead.
```java
  public Runnable getDoubleClickAction(@NotNull ModularFeatureListRow row,
      @NotNull List<RawDataFile> file, @Nullable DataType<?> superType,
      @Nullable final Object value) {
```

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
robinschmid: Great suggestion. Was to fixated on seeing the structure.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
robinschmid: True

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/StructureTableCell.java
robinschmid: This is how I have implemented it now for the structure and also for the LabeledProgressBarCell. So I guess you could implement that and use that chart caching node. Then in the end we may have to double check if the correct chromatogram is set to the parent pane?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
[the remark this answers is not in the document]
robinschmid: I made sure this is not exported to any text format. 
I did override the saveToXml method to not export the value. Also added the tag interface NoTextColumn so that its not exported to csv etc.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/MolStructureViewer.java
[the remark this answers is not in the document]
robinschmid: I had to pass in a @Nullable FeatureTableFx here to actually open annotation tabs. 
Made it Nullable because down the road we might open it from different places. There was already the compound annotation pane that also uses the double click actions.

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralLibraryEntry.java
```diff
@@ -287,4 +288,9 @@ default Integer toInteger(Object v) {
   void setLibrary(@Nullable SpectralLibrary library);
 
   @Nullable String getLibraryName();
+
+  /**
+   * @return the structure parsed from smiles or inchi
+   */
+  MolecularStructure getStructure();
```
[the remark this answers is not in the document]
robinschmid: great idea. Done

mzmine-community/src/test/java/io/github/mzmine/datamodel/structures/StructureParserTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+class StructureParserTest {
+
+  @Test
+  void parseStructureSmiles() {
+    StructureParser parser = new StructureParser(false);
+    parser.parseStructure("CCCO", StructureInputType.SMILES);
```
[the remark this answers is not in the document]
robinschmid: yeah changed it to assertNotNull

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
robinschmid: Done

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
robinschmid: Throwing exception now explaining what to do instead

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/CompoundNameType.java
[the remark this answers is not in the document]
SteffenHeu: my point was that the column with is really large

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
```diff
@@ -96,11 +108,20 @@ public double getColumnWidth() {
...
+      @Nullable final DataType<?> superType, final @Nullable Object value) {
+    if (!(value instanceof MolecularStructure structure)) {
+      return null;
+    }
     return () -> FxThread.runLater(() -> {
-      if (value instanceof MolecularStructure structure) {
+      if (superType instanceof CompoundDatabaseMatchesType) {
+        CompoundDatabaseMatchTab tab = new CompoundDatabaseMatchTab(table);
+        tab.setFeatureRow(row);
+        MZmineCore.getDesktop().addTab(tab);
```
SteffenHeu, line 121: DesktopService? :D

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/CompoundNameType.java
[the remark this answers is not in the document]
robinschmid: Ah ok. I could make it smaller for now and later we could find a way to hide sub columns in SubColumnFactories

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
robinschmid: Yes :D
Copied from somewhere.

robinschmid: Oh in this case its harder to use DesktopService as it returns the reusable Desktop interface. 
Not the MZmineDesktop. So I guess here we have to use MZmineCore.getDesktop which was designed to return the right type.
If we see int he future that this class should be separated from MZmineCore, then we should find some better way of adding tabs.

robinschmid: What do you think?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/CompoundNameType.java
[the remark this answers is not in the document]
robinschmid: Changed it to 100.
Also checked the toggle library matches already hides this column as it is defualt visibility false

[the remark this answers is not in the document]
robinschmid: this is how it looks on my side if I actively add the compound columns now with 100 width
image

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
SteffenHeu: ok, sounds reasonable

PR #1737, opened 2024-03-22
Add MS2Deepscore to MZMine

niekdejonge:
Adding a module to MZMine that is able to make ms2deepscore (https://github.com/matchms/ms2deepscore) predictions.

- [x] Load model
- [x] Add test for making correct predictions for tensor
- [x] Implement embedding creation. (without directly calculating cosine)
- [x] Add method for tensorizing spectrum
- [x] Create flexible method for tensorizing spectrum, based on model settings (the additional metadata is still fixed)
- [x] Add test for correct tensorization
- [x] Add integration test for correct embedding generation from spectrum
- [x] Add dot product calculations to create similarity scores
- [x] Add module to MZMine to use ms2deepscore for molecular networking and library searching
- [x] Add a parameter to add the path to the ms2deepscore model (instead of default test model)
- [x] Add automatic download of ms2deepscore model (from zenodo?)
- [x] Add minSignals (maybe using SignalFiltersParameters)


robinschmid:
I will look at it tomorrow. Sounds great so far.

mzmine-community/build.gradle
```diff
@@ -156,6 +156,10 @@ dependencies {
     implementation(libs.commons.io)
     implementation(libs.commons.lang3)
     implementation(libs.fastutil)
+    implementation(libs.bundles.djl.pytorch)
+    implementation(libs.djl.api)
+    implementation(libs.djl.pytorch.engine)
+    testImplementation(libs.djl.pytorch.engine)
```
robinschmid, line 162: You can remove these three lines and only keep the bundle above. 
That was me when you had the error with the library but that should be resolved.
After doing that just reload gradle project. There will be an icon after the change in the top right corner or you open the gradle tab and there is a reload gradle project button

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
```diff
@@ -0,0 +1,62 @@
...
+                          NDArray spectrumNDArray2,
+                          NDArray metadataNDArray1,
+                          NDArray metadataNDArray2
+    ) throws TranslateException {
+
+        Predictor<NDList, NDList> predictor = model.newPredictor();
+        return predictor.predict(new NDList(spectrumNDArray1, spectrumNDArray2, metadataNDArray1, metadataNDArray2));
+    }
+
+    public NDList tensorizeSpectrum(Scan Spectrum){
```
robinschmid, line 46: instead of Scan use its parent interface MassSpectrum - it has all the same methods. 
The data is always sorted by mz ascending. 
Most efficient way is to just do a simple for loop over the data points by index and spectrum.getNumberOfDataPoints()

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
```diff
@@ -0,0 +1,62 @@
...
+//        todo write the method for tensorization, using the settings.
+        return new NDList();
+    }
+
+    public Double predictPair(Scan Spectrum1, Scan Spectrum2){
+//    Predict the similarity between two spectra
+//        todo write method
+        return 1.0;
+    }
+    public Double[][] predictMatrix(ArrayList<Scan> spectra1, ArrayList<Scan> spectra2){
```
robinschmid, line 57: List instead of ArrayList. You would always mention the parent types if your method does not depend on implementation details. ArrayList has just few special methods that you most likely dont use here.

mzmine-community/src/test/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModelTest.java
```diff
@@ -0,0 +1,116 @@
...
+        try (NDManager manager = NDManager.newBaseManager()) {
+            int nr_of_spectra = 3;
+            NDArray spectrumNDArray1 = manager.create(generateRandomList(990, nr_of_spectra));
+            NDArray spectrumNDArray2 = manager.create(generateRandomList(990, nr_of_spectra));
+            NDArray metadataNDArray1 = manager.create(generateRandomList(2, nr_of_spectra));
+            NDArray metadataNDArray2 = manager.create(generateRandomList(2, nr_of_spectra));
+
+
+            NDList predictions = model.predict(spectrumNDArray1, spectrumNDArray2, metadataNDArray1, metadataNDArray2);
+            assertEquals(predictions.size(), 1);
```
robinschmid, line 108: switch values its expected, actual so
assertEquals(1, predictions.size());

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public NDList predict(NDArray spectrumNDArray1,
+                          NDArray spectrumNDArray2,
+                          NDArray metadataNDArray1,
+                          NDArray metadataNDArray2
+    ) throws TranslateException {
+
+        Predictor<NDList, NDList> predictor = model.newPredictor();
+        return predictor.predict(new NDList(spectrumNDArray1, spectrumNDArray2, metadataNDArray1, metadataNDArray2));
```
robinschmid, line 43: interesting so thats how its defined in the model that you have spec1 spec2 meta1 meta2 in this order just in a single list and it wil combine the data internally?

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
robinschmid, line 57: We could use here List<SpectralTensor> a, ... b as input so that the precompute was already done somewhere else.

robinschmid (dismissed):
Great so far

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
niekdejonge: Creating a separate class for tensorization makes sense indeed. The reason I considered making it into a method here, is that the tensorization can be different per model and is therefore really a part of that particular model. Maybe an in-between solution: creating a separate class for tensorization, but creating a Tensorizer object in the constructor of MS2DeepscoreModel, based on the settings for that particular model.

niekdejonge: Alternatively we could also just hard code the settings for tensorization and just make it only compatible with the one version of MS2Deepscore model that we release. Which is in reality what I expect users to use anyway.

niekdejonge: Regarding Scan/MassSpectrum, thanks for the overview! This model needs precursor_mz and ionmode as well as input, so I think we will need to use Scan here.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
niekdejonge: Regarding the output: 
In python it is actually:
```python
def forward(self, spectra_tensors_1, spectra_tensors_2, metadata_1, metadata_2):
```
But in some examples it seemed that in java it expected this format, so that surprised me a bit as well, but it works like this.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
```diff
@@ -0,0 +1,62 @@
...
+                .setTypes(NDList.class, NDList.class)
+                .optModelPath(Paths.get(modelFilePath))
+                .optOption("mapLocation", "true") // this model requires mapLocation for GPU
+                .optProgress(new ProgressBar()).build();
+
+        this.model = criteria.loadModel();
+    }
+
+    public NDList predict(NDArray spectrumNDArray1,
+                          NDArray spectrumNDArray2,
```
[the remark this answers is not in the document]
niekdejonge: The input actually are multiple spectra against multiple spectra. So NDArray is 2D the rows are the vectors representing one spectrum per row. So the output also is a list of predictions currently. 

Ideally in the end we could actually get an embedding out instead of the scores, since this implementation would be a lot faster for matrix comparisons, since the embedding calculation is the "expensive" step. This is also how we do it in the python version, but in a slightly different way, which was not (yet) easy to convert to java. 

I will try to get this to work, but it also might require some rewriting on the python end. 

In the end I want to have a pair_prediction and matrix_prediction method. These should indeed just return floats or list of floats.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
robinschmid: you could create an enum class with constants representing each model that we support (also looking into the future. You can either add a constructor and pass in the clear name and link to download the model or you just create methods in the enum like this:
```java
public enum MS2DeepScoreModels {
POSITIVE_V1, BOTH_POLARITIES_V1;

public String toString() {
  return switch(this) {
    case POSITIVE_V1 -> "Positive (v1)";
    case POSITIVE_V1 -> "Both polarities (v1)";
  };
}
```

robinschmid: This enum could later be places in a combobox and can be used to decide which tensor to create

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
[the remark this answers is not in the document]
robinschmid: I think getting the embeddings would be great. And either we can then feed them to the model and get all pairs computed or just compute cosine in java. I got an easy parallel stream implementation to go thorugh all pairs

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
robinschmid: looks similar at least. 
So just giving a list of NDArrays in and the model will know how many to use etc

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
niekdejonge: Yes, that might be an easy solution. We also generate a settings json in python, which also contains the information about tensor sizes. We could also use that to use the right settings. But at first we can just support only 1 model.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
niekdejonge: Yes, seems like it.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/ms2deepscore/MS2DeepscoreTask.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * The task will be scheduled by the TaskController. Progress is calculated from the
+ * finishedItems/totalItems
+ */
+class MS2DeepscoreTask extends AbstractFeatureListTask {
+
+  private static final Logger logger = Logger.getLogger(MS2DeepscoreTask.class.getName());
+  private final @NotNull FeatureList[] featureLists;
+  private final int minSignals;
+  private final Double minScore;
```
robinschmid, line 61: this might be confusing at first and the auto create of variable will always make this a Double - just because it may be null at this point.
But we know that the score is never null - its not optional and need to be set. So you can make this a primitive double. This only has few implications but its always a nice way to also denote that it is not optional

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/ms2deepscore/MS2DeepscoreTask.java
```diff
@@ -0,0 +1,148 @@
...
+
+  @Override
+  protected @NotNull List<FeatureList> getProcessedFeatureLists() {
+    return List.of(featureLists);
+  }
+
+  record RowStats(int id, double mz, double rt, int numFragmentScans, int maxFragmentSignals,
+                  String annotation, long numSamplesDetected) {
+
+  }
```
robinschmid, line 146: Better remove this already its still here from copying the class

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class MS2DeepscoreModel {
+
+  /**
+   * Predicts the MS2Deepscore similarity
```
robinschmid, line 51: I guess this is old right. this just does the embedding not the similarity?

mzmine-community/src/test/resources/models/ms2deepscore_model_settings.json
```diff
@@ -0,0 +1,67 @@
...
+    "max_mz": 1000,
+    "mz_bin_width": 1.0,
+    "intensity_scaling": 0.5,
+    "additional_metadata": [
+        [
+            "StandardScaler",
+            {
+                "metadata_field": "precursor_mz",
+                "mean": 0.0,
+                "standard_deviation": 1000.0
```
robinschmid, line 30: Not sure where you defined this but shouldnt it be mean 500 standard deviation 500? or something similar?

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SettingsMS2Deepscore.java
```diff
@@ -0,0 +1,40 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.util.scans.similarity.impl.ms2deepscore;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record SettingsMS2Deepscore(@JsonProperty("embedding_dim") Integer embeddingDimension,
```
robinschmid, line 30: you can just add this above the class and that makes all names automatically use snake case in the json and camelCase in java.

@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)

Easiest way to adjust the names in this class is to hit rename (CTRL+ALT+R) on each variable and then max_mz would be maxMz in camelCase 
mzBinWidth...
Very optional - just something I thought to mention for future json works

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/ms2deepscore/MS2DeepscoreTask.java
niekdejonge: Yes, I maybe should have clarified a bit more, but I did not work on any of the Module/Task/Parameter implementation since we worked on it since Friday, but have been working from the other end up, adding the core functionality of predicting MS2Deepscore embeddings and tensorizing spectra in separate classes (MS2DeepscoreModel, SpectrumTensorizer etc.) I still have to do the integration of both later.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
niekdejonge: Yes indeed, the similarity scoring I have now (after your reviewed) added to a parent class called EmbeddingBasedSimilarity

mzmine-community/src/test/resources/models/ms2deepscore_model_settings.json
niekdejonge: Yes, we could also have used these settings, but we didn't use it for training the model, so we cannot change it here. But it does not matter too much. It is just important for initialization of weights in a neural network that all input data is scaled somewhat similar, to have the training go efficiently. So this just has to ensure that the precursor mz has somewhat the same order of magnitute as the intensities we input. So 0.0 and 1000.0, but also 500 and 500 would do the trick.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SettingsMS2Deepscore.java
niekdejonge: Ah thanks! Good to know

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/EmbeddingBasedSimilarity.java
```diff
@@ -0,0 +1,69 @@
...
+
+    return this.dotProduct(embeddings1, embeddings2);
+  }
+
+  public double[][] convertNDArrayToDoubleMatrix(NDArray ndArray) {
+    long[] shape = ndArray.getShape().getShape();
+    if (shape.length != 2) {
+      throw new AssertionError("The NDArray is not a 2D matrix");
+    }
+    double[][] result = new double[(int) shape[0]][(int) shape[1]];
```
robinschmid, line 63: Just to describe what is happening here. 
arrays (fixed size list or tuple in python) in Java have maximum size limited by integer thats why conversion to (int) is needed here

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/EmbeddingBasedSimilarity.java
```diff
@@ -0,0 +1,69 @@
...
+import ai.djl.translate.TranslateException;
+import io.github.mzmine.datamodel.Scan;
+
+public abstract class EmbeddingBasedSimilarity {
+
+  public abstract NDArray predictEmbedding(Scan[] scans) throws TranslateException;
+
+  public double[][] dotProduct(NDArray embedding1, NDArray embedding2) {
+    NDArray norm1 = embedding1.square().sum(new int[]{1}).sqrt();
+    NDArray norm2 = embedding2.square().sum(new int[]{1}).sqrt();
```
robinschmid, line 39: you can use embedding1.norm(new int[]{1}) instead.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/EmbeddingBasedSimilarity.java
```diff
@@ -0,0 +1,69 @@
...
+    System.out.println(norm1);
+    System.out.println(norm2);
+    for (long i = 0; i < embedding1.getShape().getShape()[0]; i++) {
+      embedding1.set(new NDIndex(i), embedding1.get(i).div(norm1.get(i)));
+    }
```
robinschmid, line 44: embedding1.transpose().div(norm1).transpose() does the same thing.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/EmbeddingBasedSimilarity.java
```diff
@@ -0,0 +1,69 @@
...
+    for (long i = 0; i < embedding2.getShape().getShape()[0]; i++) {
+      embedding2.set(new NDIndex(i), embedding2.get(i).div(norm2.get(i)));
+    }
+    return convertNDArrayToDoubleMatrix(embedding1.dot(embedding2.transpose()));
```
robinschmid, line 48: what I always wonder is - can we not do the whole matrix? Its super fast to calculate but one would only need to check A-B not B-A because the result is the same. 

Just a general thing

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
```diff
@@ -0,0 +1,103 @@
...
+    mapper.configure(DeserializationFeature.FAIL_ON_MISSING_CREATOR_PROPERTIES, true);
+
+    // JSON file to Java object
+    return mapper.readValue(new File(settingsFilePath), SettingsMS2Deepscore.class);
+  }
+
+  public NDArray predictEmbeddingFromTensors(TensorizedSpectra tensorizedSpectra)
+      throws TranslateException {
+//    Todo This is an autoclosable object. Using a try block is suggested, but in that case the output is not available outside this function ... I am not sure about how to fix this.
+    NDManager manager = NDManager.newBaseManager();
```
robinschmid, line 89: maybe we can pass the manager into the constructor of the MS2DeepScoreModel - then from outside the caller can control the lifetime of the manager?

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
```diff
@@ -0,0 +1,103 @@
...
+    Predictor<NDList, NDList> predictor = model.newPredictor();
```
robinschmid, line 90: This predictor piece also feels like it could be in the constructor so that its done once. Not sure how memory or CPU heavy it is... but could be better

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SettingsMS2Deepscore.java
```diff
@@ -0,0 +1,40 @@
...
+                                   @JsonProperty("ionisation_mode") String ionisationMode,
+                                   @JsonProperty("min_mz") int minimumMZ,
+                                   @JsonProperty("max_mz") int maximumMZ,
+                                   @JsonProperty("mz_bin_width") Double binWidth,
+                                   @JsonProperty("additional_metadata") Object[][] additionalMetadata,
+                                   @JsonProperty("intensity_scaling") float intensityScaling
```
robinschmid, line 36: just a more general comment. You can slab a naming convention on this and it will automatically translate from camelCase to snake_case.

Put this before the class - but then your names need to directly match, e.g., mzBinWidth --> mz_bin_width:
@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)

Also you can default ignore or fail on unknown properties
@JsonIgnoreProperties(ignoreUnknown = true)

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SpectrumTensorizer.java
```diff
@@ -0,0 +1,99 @@
...
+ * Bins spectra and returns two tensors, one for metadata and one for fragments
+ */
+public class SpectrumTensorizer {
+
+  private final SettingsMS2Deepscore settings;
+  private final int numBins;
+
+  public SpectrumTensorizer(SettingsMS2Deepscore settings) {
+    this.settings = settings;
+    this.numBins = (int) ((settings.maximumMZ() - settings.minimumMZ() / settings.binWidth()));
```
robinschmid, line 43: this will always round down - so maximum mz might not be captured by this. Just needs to be the same like in your python code but just to make sure. 
Otherwise. (int)Math.ceil( ... )

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SpectrumTensorizer.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  public float[] tensorizeFragments(Scan spectrum) {
+    float[] vector = new float[numBins];
+    int numberOfDataPoints = spectrum.getNumberOfDataPoints();
+    for (int i = 0; i < numberOfDataPoints; i++) {
+      float mz = (float) spectrum.getMzValue(i);
+      float intensity = (float) spectrum.getIntensityValue(i);
+      if (settings.minimumMZ() <= mz && mz < settings.maximumMZ()) {
+        int binIndex = (int) ((mz - settings.minimumMZ() / settings.binWidth()));
```
robinschmid, line 53: braces are wrong?
should be (mz-minMZ) / binWidth

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SpectrumTensorizer.java
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+    return vector;
+  }
+
+  private float binarizePolarity(Scan scan) {
+    @NotNull PolarityType polarity = scan.getPolarity();
+    if (!(polarity == PolarityType.POSITIVE || polarity == PolarityType.NEGATIVE)) {
+      throw new RuntimeException("The polarity has to be positive or negative");
+    }
```
robinschmid, line 65: sometimes the charge is not given. Maybe safer to default to positive mode or if the model could actually take the three options: postivie, negative, unknown like 1, -1, 0?

If the model cannot - then better to use POSITIVE as a plausible default

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SpectrumTensorizer.java
```diff
@@ -0,0 +1,99 @@
...
+    if (polarity == PolarityType.POSITIVE) {
+      return 1;
+    } else {
+      return 0;
+    }
+  }
+
+  private float scalePrecursorMZ(Scan scan, float mean, float standardDeviation) {
+//    @robin what happens if this is null, does that throw an exception?
+    @NotNull double precursorMZ = scan.getPrecursorMz();
```
robinschmid, line 75: Yeah. 
So you cannot put @NotNull on a double primitive because primitives cannot be null... 
Double objects (Double wraps a double) can be null and the return type of getPrecursorMz is Double so it could be null at this point.

I guess you never looked into GC-EI-MS data where you do not have precursor mz right? 
Is there an option to not provide the precursor mz? In case it is missing?
Usually we should have it for MS2 spectra.

mzmine-community/src/test/resources/models/ms2deepscore_model_settings.json
robinschmid: Ok I understand - this is not the mz 0 with SD 1000 but intensity...

mzmine-community/src/test/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModelTest.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+    //      todo add check that it is correct
+    System.out.println(Arrays.deepToString(similarityMatrix));
+  }
+
+  @Test
+  void testDotProduct() {
+    try (NDManager manager = NDManager.newBaseManager()) {
+      NDArray embedding1 = manager.create(
+          new double[][]{new double[]{1.0, 1.0, 0.0, 0.0}, new double[]{1.0, 0.0, 1.0, 1.0}});
```
robinschmid, line 153: just to make your life easier on this one. Here you can define the nested array just {}. Sometimes this is possible.
`new double[][]{{1.0, 1.0, 0.0, 0.0}, {1.0, 0.0, 1.0, 1.0}}`

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/EmbeddingBasedSimilarity.java
robinschmid: maybe they already abstract this away in the native code.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/EmbeddingBasedSimilarity.java
robinschmid: combined with below you could rewrite it to:
```java
  public float[][] dotProduct2(NDArray embedding1, NDArray embedding2) {
    NDArray norm1 = embedding1.norm(new int[]{1});
    NDArray norm2 = embedding2.norm(new int[]{1});
    embedding1 = embedding1.transpose().div(norm1).transpose();
    embedding2 = embedding2.transpose().div(norm2).transpose();
    return convertNDArrayToDoubleMatrix(embedding1.dot(embedding2.transpose()));
  }
```

mzmine-community/src/test/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModelTest.java
```diff
@@ -0,0 +1,172 @@
...
+    assertEquals(embeddings.get(0).getFloat(0), -0.06332766, 0.0001);
+//      Test that the first number in the embedding is correct for the second spectrum
+    assertEquals(embeddings.get(1).getFloat(0), -0.05749714, 0.0001);
+  }
+
+  @Test
+  void testPredictMatrix() {
+    double[][] similarityMatrix;
+    try {
+      similarityMatrix = model.predictMatrix(testSpectra, testSpectra);
```
robinschmid, line 141: Maybe we should also add a function for predictPairwise(testSpectra) with just one input where all spectra are compared pair wise. This is the case for molecular networking.
Then the embeddings only need to be created once.

robinschmid (changes requested):
Great so far. I think we are close of getting this in. Just need to finalize the API to feed in either 2 lists of scans (library matching) or 1 list of scans that is compared pairwise (molecular networking).

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
niekdejonge: I did move it to the constructor. I added a method closeNdManager to MS2DeepscoreModel. So we can call this to ensure the ndManager is closed, when not needed anymore. Is that what you had in mind?

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
niekdejonge: I have moved it to the constructor. I don't expect it to be super memory or CPU heave, but not fully sure. We can have a look at further optimization once we have everything up and running.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SpectrumTensorizer.java
niekdejonge: Good to check indeed! But it is the same as implemented in python.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SpectrumTensorizer.java
niekdejonge: Good catch. The tests did not capture it because bin_width was 1.0...

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SpectrumTensorizer.java
niekdejonge: The model will use the ion mode as input, so a different prediction is made for a positive mode spectrum or a negative mode spectrum. So guessing is a bit dangerous, since it might give the wrong predictions. Is the ionmode something that can be easily set in MZMine, or if it is missing in the metadata it will remain missing? We could maybe add an option to set the ionmode in the MS2Deepscore module?

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SpectrumTensorizer.java
niekdejonge: This relates also to the above comment. We take the precursor mz and ionmode as input for this model. So the prediction is different if these values are different. MS2Deepscore is therefore only compatible with acquisition types that return an precursor mz. So for GC-EI-MS data it won't work.

niekdejonge: For now having precursor mz is null or polarity is null will throw a runtime error.

mzmine-community/src/test/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModelTest.java
niekdejonge: Thanks! That makes everything a bit less verbose.

niekdejonge:
@robinschmid Thanks for the review! (and sorry it took so long before I had time again...) Really helpful, to have this quick feedback loop, to get myself familiar with the syntax. I have applied your changes, all your comments were clear. I will now have a look at integrating ms2deepscore with the MZMine modules. Once I have this working I will ask for a new review.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/SpectrumTensorizer.java
ansgarkorf: So what would currently happen if some has DIA data? I am not sure if precursor info is added based on the MS1 scan. Maybe @SteffenHeu can answer this?

For now I suggest to mention the restrictions in a java doc and if we add MS2DeepScore as an option to a parameter set, it should always be disabled for MS1 data. I think all modules using similarity scores have a MS level or scan filter parameter, which could be checked.

SteffenHeu: The precursor info is not added as scan info, but i guess you could take the row mz as a precursor in that case (for LC-DIA).

niekdejonge: We trained and tested MS2Deepscore only for LC MSMS data. I am not sure how similar DIA data becomes to DDA after processing. If that is very close we could consider using it for DIA as well, but otherwise I would just limit it to LC MSMS, since we have no clue how it will perform otherwise. 

We should of course ensure that this is handled well and does not result in any crashes.

ansgarkorf: Yes, I agree. We also have to make sure to add these information in the documentation, e.g. optimized for DDA, untested for DIA.

robinschmid:
In our documentation you should find a part in intellij where you can add a file filter for the copyright header. So formats that it should not apply to.

niekdejonge:
@robinschmid The current version has all basic functionality. MS/MS spectral networking (ms2deepscore) works. We can select a model file. I have added a test file to the resources/models so you can use this for testing (but the results will be meaningless). The larger models I tested myself also work fine. And visualization in interactive ion identity molecular networking is working. 

Could you let me know if everything is implemented in a correct way and what steps we should still do?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/ms2deepscore/MS2DeepscoreParameters.java
```diff
@@ -0,0 +1,77 @@
...
+  public static final IntegerParameter minSignals = new IntegerParameter("Minimum signals",
+      "The minimum number of fragments for using a spectrum (minimum = 3)", 3, 3, null);
+
+  public static final PercentParameter minScore = new PercentParameter("Min similarity",
+      "The minimum similarity score to store the MS2Deepscore prediction", 0.9, 0.0, 1.0);
+
+  public static final OptionalParameter<DirectoryParameter> downloadDirectory = new OptionalParameter<>(
+      new DirectoryParameter("Download model directory",
+          "The directory in which the model is stored",
+          Objects.requireNonNull(FileAndPathUtil.resolveInMzmineDir("models")).toString()));
```
niekdejonge, line 58: This should set the default to the users/.mzmine/models folder. However, I was not able to check this, since it seems to be overwritten by the last set directory. (which I did not no how to reset) So please double check if this is correct. @robinschmid

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/ms2deepscore/MS2DeepscoreParameters.java
```diff
@@ -0,0 +1,77 @@
...
+  public static final FileNameParameter ms2deepscoreModelFile = new FileNameParameter(
+      "MS2Deepscore model",
+      "The file location of the MS2Deepscore model, click download to download the model.",
+      FileSelectionType.OPEN, true);
```
niekdejonge, line 62: It would be ideal if this field is updated automatically when we set the download folder. Or at least that the last folder is updated to reflect the location of the downloaded files. Is something like that easily doable?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/ms2deepscore/MS2DeepscoreParameters.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public static final FileNameParameter ms2deepscoreSettingsFile = new FileNameParameter(
```
niekdejonge, line 64: I gave the user now has a lot of options. Like where to store the downloaded models, or to use a different model etc. An alternative is to remove that option altogether and just always downloading the model to a predefined location. What do you think is most user friendly? @robinschmid

SteffenHeu:
Thank you for all the work @niekdejonge! Did @robinschmid and you discuss why MS2Deepscore does not implement the SpectralSimilarityFunction interface so we can use it everywhere where we use cosine similarity?

robinschmid:
we have to check it. And see if we can come up with a good strategy for this. Not sure if the interface design is good. 
Definitely need to check how we can combine it in one module in the end. Very different parameters needed for cosine / modified cosine than for MS2Deepscore.

SteffenHeu:
The input takes a parameter set, so i think that should not be a problem, or?

robinschmid:
MS2Deepscore works so differently that it does not really fit in. No mz tolerance, no minimum matched signals. 
```java
public abstract SpectralSimilarity getSimilarity(ParameterSet parameters, MZTolerance mzTol,
      int minMatch, DataPoint[] library, DataPoint[] query);
```

Best is to send a batch of spectral data to calculate the embeddings - then the similarity is a simple vector comparison by cosine sim. So the spectral alignment step is not needed. I think it might be good to have two interfaces somehow - one for vector based and one for spectral based.

PR #1748, opened 2024-04-03
Users tab

robinschmid:


robinschmid:
We have a CSS style class icon-button and I added some code to it to accept FontIcon as graphic and style it on mouse over etc. 
I created classes for easier construction and styling of icon buttons:
`FxIconUtils.newIconButton` as a shortcut for the standard things a button should do and if it gets more complicated there is also a builder `FxIconButtonBuilder` if additional things are needed like disabling the button on condition etc. 

FxLayouts.newIconPane gives option for horizontal / vertical icon button bar with default insets etc. 

The new icon buttons could be useful for the landing page as well in PR #1749

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxLabels.java
```diff
@@ -0,0 +1,123 @@
...
+
+import io.github.mzmine.javafx.util.FxColorUtil;
+import javafx.beans.value.ObservableValue;
+import javafx.scene.control.Hyperlink;
+import javafx.scene.control.Label;
+import javafx.scene.paint.Color;
+import javafx.scene.text.TextAlignment;
+import org.jetbrains.annotations.Nullable;
+
+public class FxLabels {
```
SteffenHeu, line 36: with my pr we now have two classes of the same name, that's not really optimal. 

FxLabels.newLabel is pretty long. i think its clear that htis class generates labels, i think we should just go with the FxLabels.bold, FxLabels.boldTitle and so on

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFlows.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class FxTextFlows {
+
+  /**
+   * Centered text flow
+   *
+   * @param nodes typically {@link Text} or {@link Hyperlink}
+   */
+  public static TextFlow newTextFlow(Node... nodes) {
+    return newTextFlow(TextAlignment.LEFT, nodes);
```
SteffenHeu, line 42: You say centered in the javadoc but set the alignment to left

javafx-framework/src/main/java/io/github/mzmine/javafx/components/util/FxTabs.java
```diff
@@ -23,31 +23,16 @@
...
-import io.github.mzmine.main.MZmineCore;
-import io.mzio.users.user.CurrentUserService;
-import io.mzio.users.user.MZmineUser;
-import javafx.scene.control.Button;
-import javafx.scene.layout.BorderPane;
+import javafx.scene.Node;
+import javafx.scene.control.Tab;
 
-public class UsersButton extends BorderPane {
+public class FxTabs {
```
SteffenHeu, line 31: i dont know if we need a factory for the functionality that this class covers atm

javafx-framework/src/main/java/io/github/mzmine/javafx/util/FxIconUtil.java
```diff
@@ -119,11 +145,30 @@ public static FontIcon getFontIcon(String iconCode, int size, Color color) {
 
 
   public static FontIcon getCheckedIcon() {
-    return getFontIcon("bi-check2-circle", 12, ColorsFX.getPositiveColor(Vision.DEUTERANOPIA));
+    return getFontIcon(FxIcons.CHECK_CIRCLE, 12, ColorsFX.getPositiveColor(Vision.DEUTERANOPIA));
   }
 
   public static FontIcon getUncheckedIcon() {
-    return getFontIcon("bi-x-circle", 12, ColorsFX.getNegativeColor(Vision.DEUTERANOPIA));
+    return getFontIcon(FxIcons.X_CIRCLE, 12, ColorsFX.getNegativeColor(Vision.DEUTERANOPIA));
```
SteffenHeu, line 152: should use the color palette from the preferences to keep a uniform design, not default to this.

mzmine-community/src/main/resources/themes/jabref_light.css
```diff
@@ -440,6 +441,13 @@ TextFlow > .tooltip-text-monospaced {
     -fx-padding: 0.5em;
 }
 
+.icon-button .ikonli-font-icon {
```
SteffenHeu, line 444: as discussed move to additions

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxLabels.java
robinschmid: We can just merge the classes in the end. Did not see it before I had already started the fxlabels.  

I usually static import all the factory methods. The idea is that sometimes you already wrote new Label ... and then you just remove the space and have newLabel in case you actually need to differ too much from the factory you can switch between the two easily. 

We can also drop the new keyword for the special cases like title / boldTitle etc. might make it more concise.

javafx-framework/src/main/java/io/github/mzmine/javafx/components/util/FxTabs.java
robinschmid: it does not do much but my idea is to have all standard components from factories as much as possible to achieve convention over configuration. 
For the tabs, I always forget at first to disable the close option. And you need to do this for most static tabs usually.
This allows for code like this which makes the generation shorter and clear.
```java
new TabPane(
    newTab("Users", someNode),
    newTab("Sign in", someNode)
);
```

Otherwise you have to generate a new variable for the tab, set it non closable and then create the tabpane in an extra line.

javafx-framework/src/main/java/io/github/mzmine/javafx/util/FxIconUtil.java
robinschmid: this is in the fx-framework where we do not have access to the config.  
I will move just this method somewhere in mzmine community utils.

PR #1749, opened 2024-04-03
Landing page

SteffenHeu:
Reworked the landing page, removed standard themes


robinschmid:
- [ ] For smaller icons replace text as tooltip - people will learn what the icons mean
- [ ] documentation to small icons
- [ ] preferences to small icons


## bigger icons top

- [ ] Overview video showing key features
- [ ] Getting started video introducing the wizard
- [ ] mzwizard Workflows: Later we can add an animation there where the different instruments slide together: LC-Orbitrap, GC-TOF

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxLabels.java
```diff
@@ -0,0 +1,55 @@
...
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.javafx.components.factories;
+
+import javafx.scene.control.Label;
+
+public class FxLabels {
+
+  public static Label styled(String name, String styleClass) {
```
robinschmid, line 32: I like the styled!
For most factories I like the use of static import so FxLabels.styled would then be just styled. So maybe we could use styledLabel or styledLbl or something like that?

javafx-framework/src/main/java/io/github/mzmine/javafx/util/FxIconUtil.java
```diff
@@ -58,6 +60,25 @@ public static Image loadImageFromResources(final @NotNull String resourcePath) {
...
+    view.setImage(image);
+    view.setPreserveRatio(true);
+    view.setSmooth(true);
+    view.setCache(true);
+    view.setFitHeight(maxHeight);
+    view.setFitWidth(maxWidth);
+
+    HBox box = new HBox(view);
+    box.setMaxWidth(maxWidth);
+    box.setMaxHeight(maxHeight);
```
robinschmid, line 75: is the HBox needed to force the width? So setting maxWidth and maxHeight on imageView does not work?
Just asking because returing the image view might be nice in other places where the image might change.

mzmine-community/src/main/java/io/github/mzmine/gui/NewVersionCheck.java
```diff
@@ -27,23 +27,27 @@
...
-  private static final String newestVersionAddress = "http://mzmine.github.io/version.txt";
+  public static final String newestVersionAddress = "http://mzmine.github.io/version.txt";
 
   public enum CheckType {
     DESKTOP, MENU
   }
 
   private static final Logger logger = Logger.getLogger(NewVersionCheck.class.getName());
   private final CheckType checkType;
+  private final ObjectProperty<VersionCheckResult> result = new SimpleObjectProperty<>(null);
```
robinschmid, line 50: real nice with the property now

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/introductiontab/IntroductionTabBuilder.java
```diff
@@ -0,0 +1,202 @@
...
+  protected IntroductionTabBuilder(IntroductionTabModel model) {
+    super(model);
+  }
+
+  @Override
+  public Region build() {
+    final VBox main = new VBox(40);
+    main.setAlignment(Pos.CENTER);
+    model.isDarkModeProperty()
+        .bind(MZmineCore.getConfiguration().getPreferences().darkModeProperty());
```
robinschmid, line 72: you can use ConfigService and for the most used features you can pull up the method into the ConfigService. Like I already did for the number formats. 
So maybe ConfigService.darkModeProperty()

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/introductiontab/IntroductionTabBuilder.java
```diff
@@ -0,0 +1,202 @@
...
+        model.isIsDarkMode() ? mzmineImageDarkMode : mzmineImageLightMode);
+    model.isDarkModeProperty().addListener((_, _, isDarkMode) -> {
+      mzmineImageWrapper.getChildren().clear();
+      mzmineImageWrapper.getChildren().add(isDarkMode ? mzmineImageDarkMode : mzmineImageLightMode);
+    });
+
+    HBox title = new HBox(FxLayout.DEFAULT_SPACE,
+        /*FxLabels.styled("Welcome to ", "huge-title-label"),*/ mzmineImageWrapper /*,
+        FxLabels.styled("!", "huge-title-label")*/);
+    title.setAlignment(Pos.TOP_CENTER);
```
robinschmid, line 84: I also needed that use case and added new methods in my PR that let you do
newHBox(Pos.TOP_CENTER, nodes...)
At least something for later :)

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/introductiontab/IntroductionTabBuilder.java
```diff
@@ -0,0 +1,202 @@
...
+    scroll.setFitToHeight(true);
+    scroll.setCenterShape(true);
+    return scroll;
+  }
+
+  private Region createWizardRow() {
+    final GridPane pane = new GridPane(20, 5);
+
+    final HBox wizardImageWrapper = new HBox(
+        model.isIsDarkMode() ? wizardImageDarkMode : wizardImageLightMode);
```
robinschmid, line 111: might be nice to have a custom class that handles this. 
DarkLightImageView could take a single String as resource and like logo.png as light and then automatically derive the logo_dark.png from it?
And we could then bind the imageProperty in the ImageView by the isDarkMode property

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/introductiontab/IntroductionTabBuilder.java
```diff
@@ -0,0 +1,202 @@
...
+
+  private Pane createHowToCite() {
+    VBox pane = new VBox(FxLayout.DEFAULT_SPACE);
+    pane.setAlignment(Pos.CENTER);
+    pane.getChildren().add(FxLabels.styled("How to cite", "bold-title-label"));
+
+    final FlowPane box = new FlowPane(new Label("Schmid, R., Heuckeroth, S., Korf, A. "),
+        FxLabels.italic("et. al. "), //
+        new Label("Integrative analysis of multimodal mass spectrometry data in MZmine 3. "), //
+        FxLabels.italic("Nat Biotechnol "), FxLabels.bold("41"), new Label(", 447-449 (2023)."));
```
robinschmid, line 179: Just had the same issue and looked it up there is a thing called TextFlow and its best used with Text and Hyperlink components. Somehow it does not work well with labels as it disregards breaks then. 
I made a factory for that still in my PR

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/introductiontab/IntroductionTabBuilder.java
```diff
@@ -0,0 +1,202 @@
...
+    pane.add(lblWizard, 0, 0);
+    pane.add(btnWizard, 0, 1);
+
+    FxLayout.centerAllNodesHorizontally(pane);
+    pane.setAlignment(Pos.CENTER);
+
+    return pane;
+  }
+
+  private ColumnConstraints createColumnConstraints() {
```
robinschmid, line 132: unused

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/introductiontab/IntroductionTabBuilder.java
```diff
@@ -0,0 +1,202 @@
...
+  private Region createWizardRow() {
+    final GridPane pane = new GridPane(20, 5);
+
+    final HBox wizardImageWrapper = new HBox(
+        model.isIsDarkMode() ? wizardImageDarkMode : wizardImageLightMode);
+    model.isDarkModeProperty().addListener((_, _, isDarkMode) -> {
+      wizardImageWrapper.getChildren().clear();
+      wizardImageWrapper.getChildren().add(isDarkMode ? wizardImageDarkMode : wizardImageLightMode);
+    });
+    final Label lblWizard = FxLabels.styled("Easy workflow setup", "bold-title-label");
```
robinschmid, line 116: use boldTitle instead of styled?
also in the other places where styled is used

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/introductiontab/IntroductionTabBuilder.java
```diff
@@ -0,0 +1,202 @@
...
+
+  private static final Logger logger = Logger.getLogger(IntroductionTabBuilder.class.getName());
+  private final HBox wizardImageLightMode = FxIconUtil.resizeImage(
+      "icons/introductiontab/logos_mzio_mzwizard.png", 300, 220);
+  private final HBox wizardImageDarkMode = FxIconUtil.resizeImage(
+      "icons/introductiontab/logos_mzio_mzwizard_light.png", 300, 220);
+  private HBox mzmineImageLightMode = FxIconUtil.resizeImage(
+      "icons/introductiontab/logos_mzio_mzmine.png", 350, 200);
+  private HBox mzmineImageDarkMode = FxIconUtil.resizeImage(
+      "icons/introductiontab/logos_mzio_mzmine_light.png", 350, 200);
```
robinschmid, line 61: Sadly we cannot put things into the builder as the build method will create a new view each time it is called and this would mean that those components are placed in multiple parents at the same time. 
Image should work at this level - but could also be static final then. 

Best could be to move this into a sub class like DarkLightImageView that then just takes one string for the image and loads both light and dark mode images bound to a isDarkMode property. What do you think?

mzmine-community/src/main/java/io/github/mzmine/gui/preferences/Themes.java
```diff
@@ -29,17 +29,17 @@
...
   /*MZMINE_AKK("MZmine AKK", List.of("themes/MZmine_AKK.css"), true), *///
-  JABREF_LIGHT("JabRef light (https://github.com/JabRef/jabref)",
+  JABREF_LIGHT("Light (JabRef)",
       List.of("themes/jabref_light.css", "themes/jabref_additions_light.css"), false), //
-  JABREF_DARK("JabRef dark (https://github.com/JabRef/jabref)",
+  JABREF_DARK("Dark (JabRef)",
       List.of("themes/jabref_light.css", "themes/jabref_dark.css",
           "themes/jabref_additions_dark.css"), true), //
-  JABREF_DARK_CUSTOM("JabRef dark custom (https://github.com/JabRef/jabref)",
+  JABREF_DARK_CUSTOM("Dark 2 (JabRef)",
```
robinschmid, line 42: I was always using the dark custom - is there still going to be two dark themes? maybe we make it one?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/ComboParameter.java
```diff
@@ -54,7 +54,7 @@ public ComboParameter(String name, String description, ValueType[] choices,
   }
 
   public ComboParameter(String name, String description, List<ValueType> choices) {
-    this(name, description, choices, null);
+    this(name, description, choices, choices.getFirst());
```
robinschmid, line 57: potential out of range exception here. 
Maybe leave it at as null and then the other constructor will initialize with first value.

We should check where this constructor is called without default value because some modules were designed to force the user to choose instead of having something preselected

robinschmid (changes requested):
Almost done. Great progress

javafx-framework/src/main/java/io/github/mzmine/javafx/util/FxIconUtil.java
SteffenHeu: no, it was a test to make the image auto scale. but it did not work

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/introductiontab/IntroductionTabBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+            Color.web("3391C1")),
+        () -> MZmineCore.getDesktop()
+            .openWebPage("https://github.com/mzmine/mzmine3/releases/tag/v3.9.0"));
+    box.getChildren().addAll(label, downloadButton);
+//    box.visibleProperty().bindBidirectional(model.newVersionAvailableProperty());
+    box.setAlignment(Pos.CENTER);
+    return box;
+  }
+
+  void unsubsribe() {
```
robinschmid, line 195: public / private?

robinschmid (approved):
All Done

PR #1816, opened 2024-04-25
move thermo parser extraction path, add to tmp file cleanup

SteffenHeu:
Thermo parser was still extracted to system temp dir. Added to temp file cleanup.


robinschmid:
Looks good. 
There is also this try at improving it:
https://github.com/mzmine/mzmine3/pull/811

But there were issues with macOS

robinschmid (approved):
Good catch

PR #1831, opened 2024-05-08
Add RepeatingUnitSuggester

kyrasolaris01:


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotParameters.java
```diff
@@ -166,6 +169,15 @@ public ExitCode showSetupDialog(boolean valueCheckRequired) {
     return dialog.getExitCode();
   }
 
+  private void addSuggestedRepeatingUnits(ParameterSetupDialog dialog) {
+    VBox vbox = new VBox();
+    RepeatingUnitSuggester repeatingUnitSuggester = new RepeatingUnitSuggester(
+        featureList.getValue().getMatchingFeatureLists()[0]);
```
robinschmid, line 175: index out of range if no feature lists are selected like in batch mode

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotParameters.java
```diff
@@ -166,6 +169,15 @@ public ExitCode showSetupDialog(boolean valueCheckRequired) {
...
+    vbox.getChildren().add(repeatingUnitSuggester.getListView());
+    dialog.getParamsPane().addColumn(2);
+    dialog.getParamsPane().add(new Label("Suggested repeating units:"), 2, 0);
```
robinschmid, line 178: Why not just put this label into the VBox?

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,274 @@
...
+      }
+    }
+    return list;
+  }
+
+  /*This method takes into account the detected charge state of a feature list row.
+   If the charge was detected it will be considered when calculating the repeating unit*/
+  private double[] extractMzValues() {
+    double[] mzValues = featureList.getRows().stream()
+        .flatMapToDouble(row -> DoubleStream.of(row.getAverageMZ())).toArray();
```
robinschmid, line 123: ChatGPT? :D
just mapToDouble you have a single value that you want to map to

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,274 @@
...
+    int[] charges = featureList.getRows().stream()
+        .flatMapToInt(row -> IntStream.of(row.getRowCharge())).toArray();
```
robinschmid, line 125: same here mapToInt

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,274 @@
...
+        charge = charges[i];
+      }
+      products[i] = mzValues[i] * charge;
+    }
+    return products;
+  }
+
+  private Map<Double, Integer> calculateDeltaFrequencies(double[] masses) {
+    Map<Double, Integer> frequencyMap = new HashMap<>();
+    for (int i = 0; i < masses.length; i++) {
```
robinschmid, line 146: could be <masses.length-1

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,274 @@
...
+      for (int j = i + 1; j < masses.length; j++) {
+        double delta = Math.abs(
+            Math.round((masses[j] - masses[i]) * 1000) / 1000.0); // Round to 3 decimal places
```
robinschmid, line 149: I guess Math.round(Math.abs()) might be better. Rounding sometimes depend on negative/positive number

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,279 @@
...
+        .toArray();
+    int[] charges = featureList.getRows().stream().mapToInt(FeatureListRow::getRowCharge).toArray();
+
+    if (mzValues.length != charges.length) {
+      logger.log(Level.WARNING, "The length of mzValues and charge arrays must be the same");
+      throw new IllegalArgumentException(
+          "The length of mzValues and charge arrays must be the same");
+    }
+
+    double[] products = new double[mzValues.length];
```
robinschmid, line 137: rename to neutralMasses, more meaningful

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,279 @@
...
+      }
+      products[i] = mzValues[i] * charge;
+    }
+    return products;
+  }
+
+  private Map<Double, Integer> calculateDeltaFrequencies(double[] masses) {
+    Map<Double, Integer> frequencyMap = new HashMap<>();
+    for (int i = 0; i < masses.length - 1; i++) {
+      for (int j = i + 1; j < masses.length; j++) {
```
robinschmid, line 152: maybe we should put a limit here for the maximum number of features to check. and then use it in both loops. Feature detection could end up in 1M features and that would be hard in a 2D loop 

```java
int maxFeaturesToCheck = Math.min(10000, masses. length);
```

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,279 @@
...
+        charge = charges[i];
+      }
+      neutralMasses[i] = mzValues[i] * charge;
+    }
+    return neutralMasses;
+  }
+
+  private Map<Double, Integer> calculateDeltaFrequencies(double[] masses) {
+    Map<Double, Integer> frequencyMap = new HashMap<>();
+    for (int i = 0; i < masses.length - 1; i++) {
```
robinschmid, line 151: another thing you can do to limit the number of comparisons - sort the masses array and then stop the inner loop if the delta mass grows too big to detect repeating units like above 300 or so.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,279 @@
...
+      int charge = 1;
+      if (charges[i] != 0) {
+        charge = charges[i];
+      }
+      neutralMasses[i] = mzValues[i] * charge;
+    }
+    return neutralMasses;
+  }
+
+  private Map<Double, Integer> calculateDeltaFrequencies(double[] masses) {
```
robinschmid, line 149: you could add a second argument double massTolerance and maybe add a text input or combobox with limited options above the repeating unit component.  

You can use values like 0.5, 0.1, 0.01, 0.001, 0.0005 and then just reverse the math:
```java
double delta = 
            Math.round(Math.abs(masses[j] - masses[i]) / massTolerance) * massTolerance;
```

robinschmid: this would allow for different resolutions

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,279 @@
...
+      IIsotope f = ifac.getMajorIsotope("F");
+
+      MolecularFormulaRange mfRange = new MolecularFormulaRange();
+      mfRange.addIsotope(c, 0, 50);
+      mfRange.addIsotope(h, 0, 100);
+      mfRange.addIsotope(n, 0, 50);
+      mfRange.addIsotope(o, 0, 50);
+      mfRange.addIsotope(p, 0, 10);
+      mfRange.addIsotope(s, 0, 10);
+      mfRange.addIsotope(f, 0, 100);
```
robinschmid, line 191: maybe reduce the number of all the atoms - its  only for repeating units right?
So those should be relatively small units?

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
ansgarkorf: good idea. One could also use the mzTolerance extracted from the chromatogram building -> extractMzToleranceFromPreviousMethods()

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,279 @@
...
+
+      MolecularFormulaGenerator mfg = new MolecularFormulaGenerator(builder, minMass, maxMass,
+          mfRange);
+      IMolecularFormula bestFormula = null;
+      double bestMassError = Double.MAX_VALUE;
+
+      IMolecularFormula formula;
+      while ((formula = mfg.getNextFormula()) != null) {
+        double formulaMass = MolecularFormulaManipulator.getMass(formula);
```
robinschmid, line 200: It would be amazing if we finally create a method out of this. This code is already copied a couple of times in different places. 
Maybe MolecularFormulaPredictor or something similar. 
We could set all the values needed and we could activate/deactivate all checks and then predict formulas here in a single line of code.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
robinschmid: Could work. You need to be aware that the absolute tolerance might be 0 and only ppm. 
But then just give in a number like
```java
double tolerance = tol.getMzToleranceForMass(200);
```

robinschmid: One thing to consider is that the mz in the map is then quite low resolution....
Maybe instead of counting you could add all the values to a DoubleList as value:
Map<Double, DoubleList> from fastutil. 
This way you can recalculate the actual median or mean of the mzs and not the rounded values

kyrasolaris01: I like that idea! Using the median should result in much better formula prediction results.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
kyrasolaris01: yep, but another time

kyrasolaris01:
done

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/KendrickMassPlotSetupDialog.java
```diff
@@ -0,0 +1,105 @@
...
+        vbox.getChildren().setAll(newListView);
+      }
+    });
+
+  }
+
+
+  private boolean areListViewsEqual(ListView<String> oldListView, ListView<String> newListView) {
+    ObservableList<String> oldItems = oldListView.getItems();
+    ObservableList<String> newItems = newListView.getItems();
```
robinschmid, line 88: You can just check for equality:
Objects.equals is null safe
otherwise a.equals(b)

```java 
  @Test
  public void testLists() {
    var a = List.of("Hallo", "test", "5");
    var b = new ArrayList<>(a);
    var c = a.stream().toList();

    assertTrue(Objects.equals(a, b));
    assertTrue(Objects.equals(a, c));
    assertTrue(Objects.equals(b, c));
    assertTrue(ListUtils.isEqualList(a, b));
    assertTrue(ListUtils.isEqualList(a, c));
    assertTrue(ListUtils.isEqualList(b, c));
  }
```

robinschmid: so the whole method can be removed

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,325 @@
...
+    listView.setItems(itemList);
+    listView.setPrefHeight(300);
+    mzTolerance = extractMzToleranceFromPreviousMethods();
+    loadItems();
+  }
+
+  private void loadItems() {
+    loadTask = new Task<>() {
+      @Override
+      protected ObservableList<String> call() {
```
robinschmid, line 86: You should use List<String> here instead of ObservableList

kyrasolaris01:
done,
@robinschmid I reverted to my old findTopNDeltaMedians method. Yours gave weird result. I guess the sorting was wrong.

robinschmid:
Ok only issue was that the comparator needs to be reversed 
```java
  private List<Double> findTopNDeltaMedians(Map<Double, DoubleList> deltaMap, int topN) {
    // Get the top N entries based on the size of the DoubleList
    return deltaMap.values().stream()
        .sorted(Comparator.comparing(List::size, Comparator.reverseOrder()))
    // only topN deltas with most occurrence
    .limit(topN)
// calculate median of mz
        .map(DoubleCollection::toDoubleArray).map(MathUtils::calcMedian).toList();
```

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/kendrickmassplot/RepeatingUnitSuggester.java
```diff
@@ -0,0 +1,321 @@
...
+    // Calculate the frequency of all m/z deltas
+    Map<Double, DoubleList> deltaMap = calculateDeltaFrequencies(mzValues);
+
+    List<Double> topFiveDeltas = findTopNDeltaMedians(deltaMap, 5);
+    List<Double> filteredDeltas = filterMultimers(topFiveDeltas);
+
+    for (Double delta : filteredDeltas) {
+      List<String> predictedFormulas = predictFormula(delta);
+      for (String predictedFormula : predictedFormulas) {
+        if (!predictedFormula.isBlank() && !list.contains(predictedFormula)) {
```
robinschmid, line 123: still the same issue. list.contains(predictedFormula) will always be false because you are adding a composite string here.

Best would be to use a List<FormulaWithExactMz>

robinschmid (changes requested):
I forgot to send the review and it was pending sorry

PR #1832, opened 2024-05-08
fix showing structure cell if no structure is there

SteffenHeu:
the empty structure cell was showing even if there was no structure.


robinschmid (approved):
I thought empty would say true if the item is null but great if that works now

PR #1833, opened 2024-05-10
add utility methods for ion calc

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonModification.java
```diff
@@ -685,4 +687,34 @@ public PolarityType getPolarity() {
...
+   * @return The best fitting modification or null.
+   */
+  public static @Nullable IonModification getBestIonModification(double neutralMass, double mz,
+      @NotNull MZTolerance tol, @Nullable PolarityType polarity) {
+
+    // select the appropriate modifications to search
+    final Stream<IonModification> modifications = switch (polarity) {
+      case POSITIVE -> Stream.of(IonModification.DEFAULT_VALUES_POSITIVE);
+      case NEGATIVE -> Stream.of(IonModification.DEFAULT_VALUES_NEGATIVE);
+      case NEUTRAL -> null;
```
robinschmid, line 706: I think the easiest is to return an empty stream and then the null check is not needed.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonType.java
```diff
@@ -539,6 +540,20 @@ public IMolecularFormula addToFormula(IMolecularFormula formula)
     return result;
   }
 
+  /**
+   * Clones and ionizes the given formula. The resulting molecule may be neutral if the charge of
+   * the molecule and the charge of this adduct are opposite.
+   *
+   * @param formula The formula to ionize.
+   */
+  public IMolecularFormula ionize(IMolecularFormula formula) throws CloneNotSupportedException {
```
robinschmid, line 549: Maybe you could just add a boolean flag to the addToFormula method like applyCharge and overload the method and make it true by default?
This way it may be more visible if someone is searching for the method. 
It definitely is confusing that addFormula does not change the charge. May also be wrong for the other callers

robinschmid (dismissed):
Looks good just one optional comment to maybe improve visibility of the method

PR #1840, opened 2024-05-13
ScanPolarity to wizard and to raw data overview as quick control

### LINKED ISSUE #1836: Polarity filter in wizard
Add "any" option and set filter in chromatogram builder

robinschmid:


robinschmid:
Done from my side. 
One more thing for another PR would be how to tell people that chromatogram building on both polarities does not work

ansgarkorf:
> Done from my side. One more thing for another PR would be how to tell people that chromatogram building on both polarities does not work

We can validate the parameters before triggering the task and show a message box if polarity "ANY" was chosen for polarity switching data. The message should also point the user to polarity scan filter option

robinschmid:
Not sure its usually multiple tasks running in parallel - so we would need to have a mechanism to only show one dialog. 
Also I would say the dialog should not block the processing but just show up. 

I will create a new issue for this.
#1846

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/chromatogramandspectra/ChromatogramPlotControlPane.java
```diff
@@ -89,7 +95,15 @@ public ChromatogramPlotControlPane(final ParameterSet parameters) {
       }
     });
 
-    HBox controlsWrap = new HBox(5, cbXIC, btnUpdateXIC, btnParam);
+    polarityCombo = new ComboBox<>(FXCollections.observableList(
+        List.of(WizardMsPolarity.No_filter, WizardMsPolarity.Positive, WizardMsPolarity.Negative)));
```
SteffenHeu, line 99: stick to PolarityType

SteffenHeu: use string converter -> any -> no filter

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/custom_parameters/WizardMsPolarity.java
```diff
@@ -0,0 +1,58 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.tools.batchwizard.subparameters.custom_parameters;
+
+import io.github.mzmine.datamodel.PolarityType;
+
+public enum WizardMsPolarity {
```
SteffenHeu, line 30: make package private to wizard package

robinschmid: problem is that it is being used by the parameters and the builders that are in different subpackages. So cannot be package private

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/chromatogramandspectra/ChromatogramPlotControlPane.java
robinschmid: using PolarityType now and made a PolarityFilterComboBox to hide the StringConverter No filter is now Any polarity

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/custom_parameters/WizardMsPolarity.java
robinschmid: at least now its only used in wizard

robinschmid:
All done

PR #1858, opened 2024-05-16
Add NumberFormatType and change NumberType to T extends Number

robinschmid:
NubmerFromatType is now just a type that contaisn a number format. 

NumberType (maybe we rename it to SimpleNumberType) is the parent of DoubleType, IntType etc. This way we can get a number from
ModularDataModel.get(anyNumberType.class) --> is Number


SteffenHeu:
Is there a need to split the number format from the number type?

PR #1861, opened 2024-05-17
Add proxy settings to UsersView

robinschmid:
On connection issue we open the proxy settings now in the users view. There is also a message how to resolve this issue and potential sources

image


SteffenHeu:
can not compile, the path to the global events package seems to be wrong

PR #1865, opened 2024-05-18
User proxy and CLI user argument

robinschmid:
adds to PR #1861 I think it is easier to review together

Adds a -user argument which wants a user file as input.

also adds a -login argument that forces a user login + download of the user file.


mzmine-community/src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -188,35 +189,59 @@ public static void main(final String[] args) {
...
-                  """, ButtonType.YES, ButtonType.NO);
-              if (btn == ButtonType.YES) {
-                ConfigService.getConfiguration().getPreferences().showSetupDialog(true, "Proxy");
-              }
-            }
-          });
+        if (mzEvent instanceof ProxyChangedEvent pevent) {
+          ConfigService.getPreferences().setProxy(pevent.proxy());
         }
+//        if (mzEvent instanceof AuthServerNotReachedEvent) {
```
SteffenHeu, line 203: remove

mzmine-community/src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -188,35 +189,59 @@ public static void main(final String[] args) {
           if (DesktopService.isGUI()) {
             getDesktop().addTab(UsersTab.showTab());
           } else {
-            getDesktop().displayMessage("Requires user login. Open mzmine and login to a user");
+            try {
+              new UsersController().onLoginOrRegister(LoginOptions.CONSOLE);
```
SteffenHeu, line 193: should this be handled by the current user service?

mzmine-community/src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -228,6 +253,17 @@ public static void main(final String[] args) {
       // apply memory management option
       keepInMemory.enforceToMemoryMapping();
 
+      // login user if cli option
+      if (argsParser.isCliLogin()) {
+        try {
+          logger.info("CLI user login");
+          new UsersController().loginOrRegisterConsoleBlocking();
```
SteffenHeu, line 260: see above

SteffenHeu (commented):
looks good, some slight comments

mzmine-community/src/main/java/io/github/mzmine/main/MZmineCore.java
robinschmid: Yeah I think its the best way to do things its a single point of entry and the controller streamlines the whole user things

robinschmid: At least feels good to use

PR #1886, opened 2024-05-27
fix edit combo factory unreliable update on scroll

SteffenHeu:


robinschmid (approved):
Great improvement to the table

PR #1888, opened 2024-05-28
Enable any polarity in ion identity networking

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonModification.java
```diff
@@ -52,6 +52,9 @@
 public class IonModification extends NeutralMolecule implements Comparable<IonModification>,
     StringMapParser<IonModification> {
 
+  public static Comparator<IonModification> POLARITY_MASS_SORTER = Comparator.comparingInt(
```
SteffenHeu, line 55: maybe only sort by mass / charge? this leads to all the 2+ ions on top

robinschmid: done

PR #1891, opened 2024-05-29
Enhanced Isotope Peak Scanner

Sonja-analytik:
new isotope peak scanner
- Scanning and filtering for features with characteristic isotope patterns for multiple element combinations and charges.
- Identification of monoisotopic signals (major isotopes) by calculating isotope pattern similarity scores and comparing them (local maximum of the score = major isotope)
- optional reduction of features to those with the best isotope pattern scores within the same retention range and mass range of the isotope pattern

addition to the IsotopeFinderTask
- Integration of the possibility of using merged mobility scans

addition to the DatabaseIsotopeRefiner:
- Optional removal of the features of the isotope signals of the database matches, so that only the major isotopes of the found components are listed


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderTask.java
```diff
@@ -347,6 +363,48 @@ private Scan findBestScanOrMobilityScan(ScanDataAccess scans, MobilityScanDataAc
     return mobilityScan != null ? mobScans : scans;
   }
 
+  private Scan findBestScanOrMergedMobilityScan(ScanDataAccess scans,
+      Feature feature, MZTolerance mzTolerance) {
+
+
+    final Scan maxScan = feature.getRepresentativeScan();
+    final int scanIndex = scans.indexOf(maxScan);
+    scans.jumpToIndex(scanIndex);
```
robinschmid, line 372: indexOf is quite expansive. 
You could put the if on the top and reverse the condition 
``` java
if(!(mobility && featureData instanceof IonMobilogramTimeSeries imsData)) {
    return feature.getRepresentativeScan();
} 
// handle merging of IMS data
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_isotopepeakscanner/EnhancedIsotopePeakScannerTask.java
```diff
@@ -0,0 +1,549 @@
...
+    for (FeatureListRow row : rows) {
+      featureMap.addRow(row);
+    }
+
+    // Scanning for characteristic isotope patterns for each element combination and charge.
+    // Stored in "ResultMap" if the Similarity Score is higher than minScore.
+    //In case of multiple scores for different charges, only the isotope pattern with the highest
+    // score is stored as "detectedIsotopePattern.
+
+    HashMap<Integer, IsotopePattern> calculatedIsotopePattern = new HashMap<>();
```
robinschmid, line 150: Just a general thing to note. If you are using a map or list you typically define it by its interface. Standard list implementation is ArrayList but all your methods typically do not care if its an ArrayList or LinkedList or other... Thats why we use a parent interface/class. 
```java
List<String> lines = new ArrayList<>();
```

Same for maps. HashMap is only one implementation but the use usually does not depend on the implementation details.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_isotopepeakscanner/EnhancedIsotopePeakScannerTask.java
```diff
@@ -0,0 +1,549 @@
...
+    HashMap<Integer, IsotopePattern> detectedIsotopePattern = new HashMap<>();
+    ArrayList <FeatureListRow> rowsWithIPs = new ArrayList <>();
+    ArrayList <FeatureListRow> rowsWithBestIPs = new ArrayList <>();
+    HashMap<Integer, Double> scores = new HashMap<>();
+    HashMap<Integer, Double> resultingScores = new HashMap<>();
+    HashMap<Integer, Integer> numbersOfFoundIsotopes = new HashMap<>();
+    HashMap<Integer, Integer> bestCharges = new HashMap<>();
+    PeakListHandler resultMap = new PeakListHandler();
+    PeakListHandler finalMap = new PeakListHandler();
```
robinschmid, line 159: Defining so many lists and maps in the same context/method usually points to an issue. 
So the method is too long and applies to many steps. 
APplying the Single Responsibility (SRY) concept you would split this into multiple functions, each doing a single step - arguments to those methods and either saving the state globally in the task class - or returning the result.

Then one method orchestrates the other methods into a flow.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_isotopepeakscanner/EnhancedIsotopePeakScannerTask.java
```diff
@@ -0,0 +1,549 @@
...
+    int[] charges = new int[charge];
+    for (int i = 0; i < charges.length; i++) {
+      charges[i] = i + 1;
+    }
+    if (element.contains(",")) {
+      elements = element.split(",");
+    } else {
+      elements = new String[1];
+      elements[0] = element;
+    }
```
robinschmid, line 170: if else is not needed so element.split(",") will always give you an array even if the separator is not contained. then you just get an array with size 1 and the original string.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_isotopepeakscanner/EnhancedIsotopePeakScannerTask.java
```diff
@@ -0,0 +1,549 @@
...
+    for (String element : elements) {
+      for (int charge : charges) {
+
+        for (FeatureListRow row : rows) {
```
robinschmid, line 174: maybe you can extract methods and add more comments to describe what you are doing and why this way?

robinschmid (changes requested):
There is some additional files in mzmine/datamodel/impl and lib

robinschmid (commented):
ExtendedIsotopePatternDataSet new constructor uses multiple patterns but only one is set in the end. How does htis work?

robinschmid (commented):
Not sure why is suffix defined in IsotopePatternMatcherParameters

Sonja-analytik:
> ExtendedIsotopePatternDataSet new constructor uses multiple patterns but only one is set in the end. How does htis work?

the idea is to search for multiple patterns and at the end set the pattern with the highest similarity score

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_isotopefinder/IsotopeFinderParameters.java
```diff
@@ -63,8 +67,12 @@ public class IsotopeFinderParameters extends SimpleParameterSet {
           + " or within all scans in full-width at half maximum range.", ScanRange.values(),
       ScanRange.SINGLE_MOST_INTENSE);
 
+  public static final BooleanParameter mergedMobilityScan = new BooleanParameter(
+      "Use merged mobility resolved mass spectra", "Use of merged mobility scan.", false);
```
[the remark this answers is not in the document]
Sonja-analytik: it use resolved mobility but the merged spectrum from mobility FWHM 
I have modified the description and changed the name to ‘use mobility-resolved spectra’.

Sonja-analytik:
> Not sure why is suffix defined in IsotopePatternMatcherParameters

I added an optional function to the IsotopePatternMatcher that allows to remove isotopic signals except the monoisotopic signal of database matches from the feature list to reduce the resulting data
For this, a new feature list is generated as output, which requires a suffix
It might be less confusing to define the suffix as a subparameter of the isotope removing function?

PR #1904, opened 2024-06-09
add own parameters for wizard data

SteffenHeu:
after using the advanced import, the files show up the next time you open the wizard becuase we reused the same parameter. This just adds a new parameter


mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/DataImportWizardParameters.java
```diff
@@ -36,11 +41,16 @@
  */
 public final class DataImportWizardParameters extends WizardStepParameters {
 
+  public static final OptionalParameter<FileNameParameter> metadataFile = new OptionalParameter<>(
+      ProjectMetadataImportParameters.fileName);
```
robinschmid, line 45: just a feeling that maybe you would also have to create a new parameter here for the child parameter fileName?

robinschmid (approved):
Looks good to me just one question.

PR #1913, opened 2024-06-14
Import for Bruker .baf files

SteffenHeu:
Currently supports MS1 only and DDA, found no good way to extract the quad isolation ranges in case of swath or so.


SteffenHeu:
Ready @robinschmid

robinschmid:
Can you merge the master

robinschmid (approved):
Looks good to me

Issue #1734, opened 2024-03-20
Stats dashboard


robinschmid:
## Metadata

- [x] add sample_type column per default in metadata TEXT
- [x] add sample, blank, QC, calibration/quantifier? 
- [x] default: sample
- [x] extract from file name: _qc_ or _qc.  starts with qc_ or qc-   (same for blank)

## Stats dashboard

Add a quick filter for samples. The most flexible approach might be to add a list of RawDataFile to the model and pass it into the functions that extract the stats. 

- [x] CheckComboBox for sample_type column to select sample, qc, ...

PR #1914, opened 2024-06-14
Add sample type metadata, add to statistics dashboard

SteffenHeu:
Addresses #1734 
grafik

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/SampleType.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/SampleType.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2004-2024 The MZmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+ * OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.visualization.projectmetadata;
+
+import io.github.mzmine.datamodel.RawDataFile;
+import java.util.regex.Pattern;
+import org.jetbrains.annotations.NotNull;
+
+public enum SampleType {
+  BLANK, SAMPLE, QC, CALIBRATION;
+
+  @Override
+  public String toString() {
+    return switch (this) {
+      case BLANK -> "blank";
+      case SAMPLE -> "sample";
+      case QC -> "qc";
+      case CALIBRATION -> "calibration";
+    };
+  }
+
+  public static SampleType ofFile(@NotNull final RawDataFile file) {
+    final String name = file.getName().toLowerCase();
+    final Pattern qcPattern = Pattern.compile(".*(qc).*");
+    final Pattern blankPattern = Pattern.compile(".*(blank|media).*");
+    final Pattern calPattern = Pattern.compile(".*(cal).*");
+
+    if (qcPattern.matcher(name).matches()) {
+      return QC;
+    }
+    if (blankPattern.matcher(name).matches()) {
+      return BLANK;
+    }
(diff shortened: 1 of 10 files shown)


robinschmid:
Amazing!

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/SampleType.java
```diff
@@ -0,0 +1,67 @@
...
+import io.github.mzmine.datamodel.RawDataFile;
+import java.util.regex.Pattern;
+import org.jetbrains.annotations.NotNull;
+
+public enum SampleType {
+  BLANK, SAMPLE, QC, CALIBRATION;
+
+  public static SampleType ofString(String sampleType) {
+    final String name = sampleType.toLowerCase();
+    final Pattern qcPattern = Pattern.compile(".*(qc).*");
```
robinschmid, line 37: I would maybe restrict the matching a little more. 
typically I would expect _ or . before and after qc.
So the most general assumption would then be that either qc is the start / end or there is no a-z
`(?<![a-z])(qc)(?![a-z])`

https://regex101.com/r/JB6NhF/1

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/SampleType.java
```diff
@@ -0,0 +1,67 @@
...
+package io.github.mzmine.modules.visualization.projectmetadata;
+
+import io.github.mzmine.datamodel.RawDataFile;
+import java.util.regex.Pattern;
+import org.jetbrains.annotations.NotNull;
+
+public enum SampleType {
+  BLANK, SAMPLE, QC, CALIBRATION;
+
+  public static SampleType ofString(String sampleType) {
```
robinschmid, line 35: maybe add doc that default is SAMPLE?

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/SampleTypeFilter.java
```diff
@@ -0,0 +1,114 @@
...
+import io.github.mzmine.modules.visualization.projectmetadata.table.columns.MetadataColumn;
+import io.github.mzmine.project.ProjectService;
+import java.util.List;
+
+public class SampleTypeFilter {
+
+  private final List<SampleType> types;
+
+  public SampleTypeFilter(final List<SampleType> types) {
+    this.types = types.stream().toList(); // copy, list might change content
```
robinschmid, line 41: I think EnumSet is great in those applications. Very space efficient and optimized for contains queries. 
```java
this.types = EnumSet.copyOf(types);
```

List copy usually with:
List.copyOf(types)

Only condition is that types is not null and types cannot contain null values

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/SampleTypeFilter.java
```diff
@@ -0,0 +1,114 @@
...
+   * not exist, tries to parse the file name.
+   *
+   * @param file
+   * @return
+   */
+  public boolean matches(RawDataFile file) {
+    final MetadataTable metadata = ProjectService.getProjectManager().getCurrentProject()
+        .getProjectMetadata();
+    final MetadataColumn<String> metadataColumn = (MetadataColumn<String>) metadata.getColumnByName(
+        MetadataColumn.SAMPLE_TYPE_HEADER);
```
robinschmid, line 59: This would be a great convenience method in the MetadataTable.getSampleTypeColumn
Maybe we can make it @NotNull and create this column from the start of a new project

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/MetadataTable.java
```diff
@@ -109,11 +112,21 @@ public void addFile(RawDataFile newFile) {
...
     } catch (Exception ignored) {
       logger.warning("Cannot set date " + newFile.getStartTimeStamp().toString());
     }
   }
 
+  private void assignSampleType(RawDataFile newFile) {
+    MetadataColumn sampleTypeColumn = getColumnByName(SAMPLE_TYPE_HEADER);
+    if (sampleTypeColumn == null) {
+      sampleTypeColumn = new StringMetadataColumn(SAMPLE_TYPE_HEADER, "The type of the sample");
+    }
```
robinschmid, line 126: Great that you bring this up here. We still need to add proper synchronization here. Very unlikely to hit issues here but we should still add a CloseableReentrantReadWriteLock. 
I will open issue #1915

I think we should add the sample type and date in the constructor to always  have them

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/MetadataTable.java
```diff
@@ -221,7 +234,16 @@ public boolean exportMetadata(File file) {
    * @return was the import successful?
    */
   public boolean importMetadata(File file, final boolean skipColOnError) {
-    return tableIOUtils.importFrom(file, skipColOnError);
+    final boolean b = tableIOUtils.importFrom(file, skipColOnError);
+
+    if (b) {
+      if(getColumnByName(SAMPLE_TYPE_HEADER) == null) {
+        getData().values().stream().flatMap(m -> m.entrySet().stream()).map(Entry::getKey)
```
robinschmid, line 241: m.keySet().stream()
instead of entrySet.map to key

robinschmid (approved):
looks great. Just a few comments. 
And I opened an issue to synchonize the MetadataTable

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/SampleTypeFilter.java
SteffenHeu: Just creating from a new project does not work, because it is not always contained in the import.

robinschmid: Ok we will need to find a solution in the future.

robinschmid:
Great thanks @SteffenHeu

PR #1919, opened 2024-06-17
fix msms annotation in profile spectra

SteffenHeu:
The Annotation of fragment ions in spectra only worked for centroided data. I don't know if it ever actually worked for profile data, because the m/z of the annotation is taken from the mass list. If the mass detector uses anything but the local maximum, the annotation will never be found or shown, because it differs from the scan.


robinschmid:
Question is only if we should show annotations in profile mode or not

SteffenHeu:
Why would we not want to show annotations in profile spectra?

PR #1930, opened 2024-06-26
GC decon and alignment update v2

ansgarkorf:
Changes:

- Add alignment score for GC aligner
- Fix bug in Hierarchical Clustering algorithm selection
- Add threshold parameter for Shape Correlation algorithm
- add missing row filter in GC-MS

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_spectraldeconvolutiongc/rtgroupingandsharecorrelation/RtGroupingAndShapeCorrelationParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_spectraldeconvolutiongc/rtgroupingandsharecorrelation/RtGroupingAndShapeCorrelationParameters.java
@@ -27,6 +27,7 @@ package io.github.mzmine.modules.dataprocessing.featdet_spectraldeconvolutiongc.
 
 import io.github.mzmine.parameters.impl.SimpleParameterSet;
 import io.github.mzmine.parameters.parametertypes.IntegerParameter;
+import io.github.mzmine.parameters.parametertypes.PercentParameter;
 import io.github.mzmine.parameters.parametertypes.tolerances.RTTolerance;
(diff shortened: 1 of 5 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_spectraldeconvolutiongc/rtgroupingandsharecorrelation/RtGroupingAndShapeCorrelationParameters.java
```diff
@@ -39,9 +40,12 @@ public class RtGroupingAndShapeCorrelationParameters extends SimpleParameterSet
   public static final IntegerParameter MIN_NUMBER_OF_SIGNALS = new IntegerParameter(
       "Minimum signals in pseudo spectrum",
       "Minimum number of deconvoluted signals in pseudo spectrum", 10, true, 1, 5000);
+  public static final PercentParameter MIN_R = new PercentParameter("Minimum shape similarity",
```
SteffenHeu, line 43: Use a double parameter with values from 0 to 1, then it matches the correlation from meta corr

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderGcEiDeconvolution.java
```diff
@@ -323,4 +330,56 @@ private MZmineProcessingStep<SpectralDeconvolutionAlgorithm> createSpectralDecon
     return new MZmineProcessingStepImpl<>(detect, param);
   }
 
+  private void createRowFilterStep(final BatchQueue q) {
```
SteffenHeu, line 333: looks like copy paste to me. can you override the existing method, call super.makeAndAddRowFilterStep and then edit the ms2 parameter that you want to change? looks better to maintain to me. and please adhere to the naming scheme of makeAndAdd, this method does not return anything

PR #1939, opened 2024-06-28
fix all msms window only showing first or last file

### LINKED ISSUE #1908: All msms prev/next bug
only jumps firs/last file

SteffenHeu:
fixes #1908


ansgarkorf:
looks good and works fine

PR #1956, opened 2024-07-11
External tools 2.0

SteffenHeu:


SteffenHeu:
The macOS build is running now, but the resolving of the path to the raw file parser does not work on macOS. Unfortunately i do not have a mac to debug this.

@tomas-pluskal / @roman-bushuiev: Please look into the ExternalTool.java, escpecially into the resolveOsDependent method, so it works on mac. 

Steps to reproduce:
- start mzmine
- drag & drop a thermo raw file

The actual command line instruction to run the parser is failing (see ThermoRawImportTask.java, line 131).

tomas-pluskal:
I have not tested it, but it seems you are assuming that workingDir == mzmine app dir, which is probably not the case on most unix-like platforms, including macOS. It depends from which directory the app was started.

robinschmid:
Put on halt for now. 
Currently considering other options to save tools to user folder

PR #1977, opened 2024-07-18
Adjust local min edges

SteffenHeu:
add 1 data point on each side if the flanking data point is zero (after filtering for chromatographic threshold)


robinschmid (approved):
Looks good. It changes the outcome but I think the results are better this way

PR #1992, opened 2024-07-29
Handle original feature list in metaCorr grouping

robinschmid:


robinschmid:
Now set to keep original results because that was the state before

PR #2003, opened 2024-07-31
Feature preview dialog

SteffenHeu:
Added a parameter setup dialog that makes it easier to generate a preview for a selected feature. You now only have two methods to implement


mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/ParameterDialogWithFeaturePreview.java
```diff
@@ -0,0 +1,150 @@
...
+    updatePreview();
+  }
+
+  /**
+   * Called if the parameters or the selected feature change.
+   */
+  private void updatePreview() {
+    updateParameterSetFromComponents();
+
+    var task = new FxUpdateTask<>("update baseline preview", new Object()) {
```
robinschmid, line 83: baseline preview sounds like a special implementation so maybe something more generic here?

mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/ParameterDialogWithFeaturePreview.java
```diff
@@ -0,0 +1,150 @@
...
+import javafx.scene.layout.Region;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public abstract class ParameterDialogWithFeaturePreview extends ParameterSetupDialogWithPreview {
+
+  protected final NumberFormats formats = ConfigService.getGuiFormats();
+  private final SimpleXYChart<? extends PlotXYDataProvider> chart;
+  private SortableFeatureComboBox featureBox;
+  private final DialogController controller = new DialogController();
```
robinschmid, line 54: I guess you only need the LatestTaskScheduler?
So maybe you can change the controller here for this?

Another thing to think about is if we actually want to have a scheduler for each controller / interface. We could maybe have a global LatestTaskScheduler and then pass in multiple properties to track which object a task originates. We could use the memory address of an object or just the Object itself?

Like record TaskTracker(Object taskOrigin, String taskName)

robinschmid (commented):
Looks good. Do you have an implementation to see it in action? 
Maybe as a draft PR?

mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/ParameterDialogWithFeaturePreview.java
SteffenHeu: changed, added this as part of the baseline preview and noticed i'm doing the same stuff for every preview dialog.

mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/ParameterDialogWithPreviewPanes.java
```diff
@@ -0,0 +1,82 @@
...
+import io.github.mzmine.parameters.ParameterSet;
+import io.github.mzmine.parameters.dialogs.previewpane.FeaturePreviewPane;
+import java.util.function.Function;
+import javafx.geometry.Pos;
+import javafx.scene.control.ScrollPane;
+import javafx.scene.layout.Region;
+import javafx.scene.layout.VBox;
+import org.jetbrains.annotations.NotNull;
+
+public class ParameterDialogWithPreviewPanes extends ParameterSetupDialogWithPreview {
```
robinschmid, line 40: Maybe add Feature to the name:
ParameterDialogWithFeaturePreviewPanes

mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/ParameterDialogWithPreviewPanes.java
```diff
@@ -0,0 +1,82 @@
...
+
+  private final ScrollPane scroll = new ScrollPane();
+  private final VBox vbox = FxLayout.newVBox(Pos.TOP_CENTER, FxLayout.DEFAULT_PADDING_INSETS);
```
robinschmid, line 43: maybe add a factory method to produce a VBox TOP_CENTER with vbox.setFillWidth(true); This will be needed here right?

mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/ParameterDialogWithPreviewPanes.java
```diff
@@ -0,0 +1,82 @@
...
+    updatePreview();
+  }
+
+  /**
+   * Called if the parameters or the selected feature change.
+   */
+  private void updatePreview() {
+    updateParameterSetFromComponents();
+    vbox.getChildren().stream().filter(n -> n instanceof FeaturePreviewPane)
+        .map(n -> (FeaturePreviewPane) n).forEach(FeaturePreviewPane::updatePreview);
```
robinschmid, line 78: No preference just a comment because I was looking for it previously:
you can also use method references here on the class object:
`.filter(FeaturePreviewPane.class::isInstance).map(FeaturePreviewPane.class::cast)`

but both is good.

robinschmid (dismissed):
Looks great only one comment if the VBox might need FillWidth. Can be merged afterward

robinschmid (approved):
All done

PR #2004, opened 2024-07-31
adding scan number for merged spectra

corinnabrungs:


corinnabrungs:
not finished yet

SteffenHeu:
What is the scan number for? I think currently it is set to -1 so it is clear that the spectrum is not a real/raw spectrum contained in the data file. If we set the scan number to something this assumption is not clear anymore, especially for external tools.

corinnabrungs:
The idea is to have the source spectra. So if we merge two spectra, we know which spectra were used.

robinschmid:
SCANS can point to multiple scans to highlight the source

https://www.matrixscience.com/help/data_file_help.html#:~:text=SCANS%20Anything%20from%20a%20single%20scan%20number%20to%20a%20complex%20list%20of%20ranges%2C%20e.g.%20SCANS%3D1278%2C1280%2D1284%2C1290%2D1294%2C1298.%20When%20there%20are%20multiple%20raw%20files%2C%20there%20can%20be%20multiple%20SCANS%20entries%20in%20a%20single%20query%2C%20each%20with%20a%20zero%2Dbased%20index%20that%20relates%20to%20a%20specific%20raw%20file%2C%20e.g.%20SCANS%5B3%5D.

SteffenHeu:
Ok, but is your plan to alter the Scan.getScanNumber() method to return a list/array? I don't know if that's good. 
How about setting a list of USIs to the merged spectra?

robinschmid:
Oh no its only about the export to libraries. Like setting the scan numbers to the mgf, json, etc. 
Also setting USIs as well.
So Scan is untouched

robinschmid:
I added a few things but all seems good from my side.
Maybe @SteffenHeu can check again.

robinschmid:
Some example files. Had to change mgf format to .txt because thats what GH accepts
[nih_lib.json](https://github.com/user-attachments/files/17233134/nih_lib.json)
[nih_lib.mgf.txt](https://github.com/user-attachments/files/17233138/nih_lib.mgf.txt)

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralLibraryEntry.java
```diff
@@ -145,20 +163,40 @@ static SpectralLibraryEntry create(final FeatureListRow row, @Nullable MemoryMap
...
         entry.putIfNotNull(DBEntryField.ISOLATION_WINDOW, RangeUtils.rangeLength(window));
       }
       entry.putIfNotNull(DBEntryField.MS_LEVEL, msMsInfo.getMsLevel());
     }
+    List<Float> energies = ScanUtils.extractCollisionEnergies(scan);
+    if (energies.size() == 1) {
+      entry.putIfNotNull(DBEntryField.COLLISION_ENERGY, energies.getFirst());
+    } else if (energies.size() > 1) {
+      entry.putIfNotNull(DBEntryField.COLLISION_ENERGY, energies);
+    }
```
SteffenHeu, line 178: is the special handling of this case necessary? either it would say [[60, 60, 60]] or just [60, 60, 60]. 

wouldnt it make it simpler overall if the entries follow the same pattern all the time?

SteffenHeu (commented):
nice, only a small comment

PR #2007, opened 2024-08-01
Frequency mass detector

robinschmid:
In the future we could use the frequency of static noise signals being the same height to denoise spectra.

There is a preprint with a similar idea. Crucial will be to detect the binning with properly to detect the 1 and 2 "counts" static noise signals 

https://www.researchsquare.com/article/rs-4758843/v1

https://github.com/FanzhouKong/spectral_denoising/blob/174930c6ecd28363b1a85e0800fa2d7f5f43fbd3/toolsets/denoising_related_functions.py#L98


SteffenHeu:
Sounds similar to the the factor of lowest that we already have, or?

PR #2014, opened 2024-08-05
Add docstrings for ms2deepscore module

niekdejonge:


mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/ms2deepscore/MS2DeepscoreModel.java
```diff
@@ -91,6 +93,9 @@ public NDArray predictEmbeddingFromTensors(TensorizedSpectra tensorizedSpectra)
     return predictions.getFirst();
   }
 
+  /*
```
robinschmid, line 96: javadoc is started with `/**` and block comment with just one star 
I will just address this. Thanks for adding comments

robinschmid (approved):
Thanks for more comments

PR #2015, opened 2024-08-05
fix resolver not using memory mapping for intensity + mz

SteffenHeu:


robinschmid (approved):
Looks great. Thanks for also profiling

PR #2027, opened 2024-08-09
Import for UV spectra and traces from native Bruker files

SteffenHeu:


SteffenHeu:
Added a test

robinschmid (commented):
Looks good to me. Just minor comments

PR #2030, opened 2024-08-12
add missing FxCombobox method

SteffenHeu:
someone told me to delete a method, but it was used in the users tab :/


robinschmid:
Who was that?

PR #2031, opened 2024-08-12
option use msconvert for thermo import

SteffenHeu:
- add option to use msconvert for thermo import
- allows import of uv traces and spectra
- checks if it's an academic license (then allows usage of thermo raw file parser and looks for msconvert otherwise)

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
@@ -57,8 +57,11 @@ import io.github.mzmine.util.web.Proxy;
 import io.github.mzmine.util.web.ProxyType;
 import io.github.mzmine.util.web.ProxyUtils;
 import io.mzio.users.gui.fx.UsersController;
+import io.mzio.users.service.UserType;
+import io.mzio.users.user.CurrentUserService;
 import java.io.File;
 import java.text.DecimalFormat;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import javafx.application.Platform;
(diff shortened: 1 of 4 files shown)


mzmine-community/src/main/java/io/github/mzmine/gui/preferences/ThermoImportOptions.java
```diff
@@ -0,0 +1,49 @@
...
+      return new ThermoImportOptions[]{THERMO_RAW_FILE_PARSER, MSCONVERT};
+    } else {
+      return new ThermoImportOptions[]{THERMO_RAW_FILE_PARSER};
+    }
+  }
+
+  @Override
+  public String toString() {
+    return switch (this) {
+      case THERMO_RAW_FILE_PARSER -> "Thermo raw file parser (academia only)";
```
robinschmid, line 45: I think I would remove academia only. 
Only thing we cannot do is distribute this parser. 
It does not matter what other people do on their side. 
Also people can use the thermo raw file parser for commercial projects.... just not distribute.

mzmine-community/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
```diff
@@ -489,4 +499,21 @@ public void setProxy(final Proxy proxy) {
...
+      boolean skipRawDataAndFeatureListParameters) {
+    final boolean superCheck = super.checkParameterValues(errorMessages,
+        skipRawDataAndFeatureListParameters);
+
+    if (getValue(MZminePreferences.thermoImportChoice) == ThermoImportOptions.THERMO_RAW_FILE_PARSER
+        && CurrentUserService.getUser() != null
+        && CurrentUserService.getUser().getUserType() != UserType.ACADEMIC) {
+      errorMessages.add(
+          "Importing Thermo raw files via the Thermo raw file parser is only available for academic "
+              + "users. Please select and install MSConvert.");
```
robinschmid, line 514: why should we limit this if this is not a requirement from the thermo raw file parser?
We cannot redistribute the parser but people can use it commerially like msconvert.

I would remove this check.

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
```diff
@@ -53,23 +63,21 @@
  */
 public class ThermoRawImportTask extends AbstractTask {
 
+  public static final String THERMO_RAW_PARSER_DIR = "mzmine_thermo_raw_parser";
+
   private static final Logger logger = Logger.getLogger(ThermoRawImportTask.class.getName());
+  private static final org.slf4j.Logger log = LoggerFactory.getLogger(ThermoRawImportTask.class);
```
robinschmid, line 69: remove

robinschmid (changes requested):
looks good to me. I just think we can remove the restrictions on thermo raw file parser use. Or is there a specific reason for this?

PR #2074, opened 2024-09-12
fix npe in image visualizer

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/image/RawImageProvider.java
```diff
@@ -97,7 +97,9 @@ public RawImageProvider(ImagingRawDataFile raw, ParameterSet parameters) {
     this.normalize = parameters.getValue(ImageVisualizerParameters.imageNormalization);
     this.transformation = parameters.getValue(ImageVisualizerParameters.imageTransformation);
     this.scanSelection = parameters.getValue(ImageVisualizerParameters.scanSelection);
-    this.mzRange = parameters.getValue(ImageVisualizerParameters.mzRange);
+    this.mzRange =
+        parameters.getValue(ImageVisualizerParameters.mzRange) != null ? parameters.getValue(
```
robinschmid, line 101: good place for requreNonNullElse this allows you to skip the second parameters.getValue and makes it more readable. 

and default could be Range.all() right?

But can be merged like this.

PR #2083, opened 2024-09-16
Update jpackage plugin, build, actions

robinschmid:


robinschmid:
first testing this on another branch

PR #2084, opened 2024-09-16
Enable thermo import via self-downloaded parser

SteffenHeu:
- Add parameter to settings for raw file parser location
- Add parameter for thermo import option (msconvert or raw file parser)
- Thread-blocking check if the dir is set correctly, show an error message and open preferences otherwise
- If a zip is selected, unzip to external_resources folder
- also check external_resources folder for already-unzipped parsers if the current path is invalid (auto discover if the config was broken)

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
@@ -28,6 +28,7 @@ package io.github.mzmine.modules.io.import_rawdata_thermo_raw;
 import com.sun.jna.Platform;
 import io.github.mzmine.datamodel.MZmineProject;
 import io.github.mzmine.datamodel.RawDataFile;
+import io.github.mzmine.gui.DesktopService;
 import io.github.mzmine.gui.preferences.MZminePreferences;
 import io.github.mzmine.gui.preferences.ThermoImportOptions;
 import io.github.mzmine.javafx.concurrent.threading.FxThread;
@@ -35,6 +36,7 @@ import io.github.mzmine.javafx.dialogs.DialogLoggerUtil;
 import io.github.mzmine.main.ConfigService;
 import io.github.mzmine.main.MZmineCore;
 import io.github.mzmine.modules.MZmineModule;
+import io.github.mzmine.modules.io.download.ExternalAsset;
 import io.github.mzmine.modules.io.import_rawdata_all.spectral_processor.ScanImportProcessorConfig;
 import io.github.mzmine.modules.io.import_rawdata_msconvert.MSConvert;
 import io.github.mzmine.modules.io.import_rawdata_msconvert.MSConvertImportTask;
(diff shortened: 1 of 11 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
```diff
@@ -66,6 +75,9 @@ public class ThermoRawImportTask extends AbstractTask {
   public static final String THERMO_RAW_PARSER_DIR = "mzmine_thermo_raw_parser";
 
   private static final Logger logger = Logger.getLogger(ThermoRawImportTask.class.getName());
+  private static final CloseableReentrantReadWriteLock unzipLock = new CloseableReentrantReadWriteLock();
```
robinschmid, line 78: maybe this should already be unziped by user?

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
```diff
@@ -188,34 +202,26 @@ public void run() {
...
+
+    if (!checkParserExistenceAndUnzipBlocking()) {
+      return null;
     }
 
-    final File thermoRawFileParserDir = unzipThermoRawFileParser();
-    taskDescription = "Opening file " + fileToOpen;
-    String thermoRawFileParserCommand;
+    String thermoRawFileParserCommand = ConfigService.getPreference(
+        MZminePreferences.thermoRawFileParserPath).getAbsolutePath();
```
robinschmid, line 212: could this parameter be null? maybe add check?

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
```diff
@@ -233,56 +239,106 @@ public void run() {
...
 
+  /**
+   * Checks if the raw file parser is already unzipped. If not, the parser is either unzipped or the
+   * user is prompted to download the parser. If a zip is selected, the zip is unpacked to the raw
+   * file parser dir in the mzmine folder.
+   *
+   * @return
+   * @throws IOException
+   */
+  private boolean checkParserExistenceAndUnzipBlocking() throws IOException {
```
robinschmid, line 255: Best to have the unzipping done by the user... but otherwise I guess this would also need to be a static synchronized method so that all threads are waiting for one to unzip the file - right?

SteffenHeu: the lock itself is static

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
SteffenHeu: not at this point, the checkParserExistenceAndUnzipBlocking ensures it is a valid path or returns false

SteffenHeu: added a check anyway

PR #2088, opened 2024-09-18
Rework Ion Identity types

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/IonPart.java
```diff
@@ -0,0 +1,424 @@
...
+
+  public IonPart withCount(final int count) {
+    if (count == this.count) {
+      return this;
+    }
+    return new IonPart(name, singleFormula, absSingleMass, singleCharge, count);
+  }
+
+
+  public int totalCharge() {
```
SteffenHeu, line 248: do we specifically need to differentiate between a single charge + count and the total charge? Looks like more complexity to me. a single int for the total charge would be simpler. if they are added to a list of ion parts the iontype could do the merging/de-merging on demand

robinschmid: The idea was to define what a part is:
Fe+2 --> formula: Fe, singleCharge +2, singleAbsMass: ... and then easily define that we have n of this IonPart like +2Fe or -Fe ... 
I could also change this into the totalCharge instead of singleCharge. 
Initially I thought it is easier to define with the actual charge of the part instead of the total like
-H+ would then be defined as H, totalCharge -1, count -1.
That gets easier using a wither: 
new IonPart("H", totalcharge=1).withCount(-1) this would then flip the charge

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/IonType.java
```diff
@@ -0,0 +1,300 @@
...
+  public static final Comparator<IonType> DEFAULT_ION_ADDUCT_SORTER = Comparator.comparingDouble(
+      IonType::totalCharge).thenComparingDouble(IonType::totalMass);
+
+  public static IonType create(@NotNull IonPart... parts) {
+    return create(List.of(parts), 1);
+  }
+
+  public static IonType create(@NotNull List<@NotNull IonPart> parts, int molecules) {
+    // requires to merge all the same IonParts into single objects by adding up their count multiplier
+    HashMap<IonPart, IonPart> unique = HashMap.newHashMap(parts.size());
```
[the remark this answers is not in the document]
robinschmid: The great thing with this approach was that the IonPart is uniquely defined by: Formula, name, mass, charge
The count is just extra baggage. 
And this hashmap makes it easy to collapse things like +H +H -H --> +1H because the counts are just added up. 

Otherwise, I would need to create another representation for the hashmap key. 

I think best would maybe be to just document the equals and hash method better with javadoc

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/IonPart.java
```diff
@@ -0,0 +1,424 @@
...
+
+
+  /**
+   * Exclude count from equals and hash so that duplicate elements can be more easily merged
+   *
+   * @param o the reference object with which to compare.
+   * @return true if equals
+   */
+  @Override
+  public boolean equals(final Object o) {
```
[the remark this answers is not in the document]
robinschmid: Added count to hash and equals

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/IonPart.java
robinschmid: issue here was mostly for the matching if parts are the same so for example if we compare two ion types and both have something like +2H and the other -H then we want to compare and find out that both H are the same IonPart definition (same single charge, same single mass). If we now start calculating the single mass from a total mass we might introduce more issues than necessary. We are already checking within double significance but I think it is better to keep the single mass and single charge instead of the total charge. 
The total is not needed so often actually.

PR #2093, opened 2024-09-20
Fix assertion error in file name pattern

SteffenHeu:
StringParameter is nonnull, but the pattern may be null if nothing has been set yet


robinschmid (approved):
Good catch

PR #2100, opened 2024-09-27
Fix old projects not savable with mzmine 4.3

SteffenHeu:
in case a module was renamed and an old project was opened with a new mzmine version, the project cout not be saved anymore. this replaces missing modules with a placeholder and a xml string covering all the parameter settings

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/SimpleFeatureListAppliedMethod.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/SimpleFeatureListAppliedMethod.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2022 The MZmine Development Team
+ * Copyright (c) 2004-2024 The mzmine Development Team
  *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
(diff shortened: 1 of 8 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/SimpleFeatureListAppliedMethod.java
```diff
@@ -147,7 +154,8 @@ public static SimpleFeatureListAppliedMethod loadValueFromXML(Element element) {
       moduleParameters.loadValuesFromXML(parametersElement);
     } catch (Exception | NoClassDefFoundError e) {
       logger.log(Level.SEVERE, "Cannot parse module parameters", e);
```
robinschmid, line 156: maybe change from SEVERE to FINE or INFO and remove the stack trace from the error and only include the e.getMessage?
As this is expected in some cases

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
```diff
@@ -283,4 +283,9 @@ public void setOutputBaseFile(final String overrideOutBaseFile) {
     ChangeOutputFilesUtils.applyTo(this, baseFile);
     logger.info("Done changing output file paths.");
   }
+
+  @Override
+  public void add(int index, MZmineProcessingStep<MZmineProcessingModule> element) {
+    super.add(index, element);
+  }
```
robinschmid, line 290: can be removed?

robinschmid (approved):
Only small comment

PR #2101, opened 2024-09-27
Sort raw data files in FeatureList by name

robinschmid:
- sorting by name will  harmonize the FeatureTableFx and CSV export and other things


robinschmid:
image

PR #2114, opened 2024-10-02
fix ElementsParameter with duplicate values in summary

### LINKED ISSUE #2104: Feature list rows filter
### Basic information

* My operating system and version:
* My mzmine version: 4.3.0

### What happened

When opening the feature list rows filter, each time the selected elements are added (3x open, 3 times included). Also if deselected under setup, it still appears green, but is deselected. 
image

SteffenHeu:
fixes #2104


robinschmid (approved):
Easy fix

PR #2117, opened 2024-10-04
add types for uv support

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/otherdectectors/WavelengthType.java
```diff
@@ -50,7 +50,7 @@ public NumberFormat getExportFormat() {
 
   @Override
   public @NotNull String getUniqueID() {
-    return "wavelength";
+    return "otherdetectors_wavelength";
```
robinschmid, line 53: I think wavelength would be enough here

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchModeModulesList.java
```diff
@@ -183,273 +183,276 @@
...
-  public static final List<Class<? extends MZmineProcessingModule>> MODULES = List.of(
-      /*
-       * {@link io.github.mzmine.modules.MZmineModuleCategory.MainCategory#PROJECT}
-       */
-      ProjectLoadModule.class, //
-      ProjectSaveModule.class, //
-      ProjectSaveAsModule.class, //
-      ClearProjectModule.class, //
+  public static final List<Class<? extends MZmineProcessingModule>> MODULES = new ArrayList<>(
+      List.of(
```
robinschmid, line 192: maybe add a comment to code why modifiable list is required here.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/UnivariateBaselineCorrector.java
```diff
@@ -61,8 +61,13 @@ protected <T extends IntensityTimeSeries> T subSampleAndCorrect(T timeSeries, in
...
     for (int i = 0; i < numValues; i++) {
-      // must be above zero, but not bigger than the original value.
-      yBuffer[i] = Math.min(Math.max(yBuffer[i] - splineFunction.value(xBuffer[i]), 0), yBuffer[i]);
+      final double baseline = splineFunction.value(xBuffer[i]);
+      if(Double.isNaN(baseline)) {
+        continue;
+      }
+      // corrected value must be above zero. the baseline itself may be below zero, e.g. for signal
+      // drifts due to solvent composition change
+      yBuffer[i] = Math.max(yBuffer[i] - baseline, 0);
```
robinschmid, line 70: Why change from from not <0 and not >original intensity?
Previously it was this:
```java
yBuffer[i] = Math.min(Math.max(yBuffer[i] - splineFunction.value(xBuffer[i]), 0), yBuffer[i]);
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/polynomial/PolynomialBaselineCorrection.java
```diff
@@ -110,8 +112,10 @@ public <T extends IntensityTimeSeries> T correctBaseline(T timeSeries) {
             i -> function.value(xBuffer[i])));
       }
       for (int i = 0; i < numValues; i++) {
-        // must be above zero, but not bigger than the original value.
-        yBuffer[i] = Math.min(Math.max(yBuffer[i] - function.value(xBuffer[i]), 0), yBuffer[i]);
+        final double baseline = function.value(xBuffer[i]);
+        // corrected value must be above zero. the baseline itself may be below zero, e.g. for signal
+        // drifts due to solvent composition change
+        yBuffer[i] = Math.max(yBuffer[i] - baseline, 0);
```
robinschmid, line 118: same here

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/DoubleComponent.java
```diff
@@ -45,6 +51,7 @@ public DoubleComponent(int inputsize, Double minimum, Double maximum, NumberForm
     textField.setTextFormatter(new TextFormatter<>(new NumberStringConverter(format)));
     textField.setPrefWidth(inputsize);
     textField.setText(String.valueOf(defvalue)); // why not format.format(defValue)?
+    textField.textProperty().bindBidirectional(value, new DoubleStringConverter());
```
robinschmid, line 54: I found out recently that the TextFormatter that is set above already contains a valueProperty. Currently Property<Number>
You can change it to Property<Double> with this converter:
```java
// make formatter a field in the class
textFormatter = new TextFormatter<>(new FormatStringConverter<>(format));
// textFormatter.valueProperty() later
```

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/DoubleComponent.java
```diff
@@ -99,6 +106,11 @@ public TextField getTextField() {
     return textField;
   }
 
+  @Override
+  public Property<Double> valueProperty() {
+    return value;
```
robinschmid, line 111: textFormatter.valueProperty()

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/DoubleComponent.java
```diff
@@ -24,18 +24,24 @@
...
+import javafx.util.converter.DoubleStringConverter;
 import javafx.util.converter.NumberStringConverter;
 
-public class DoubleComponent extends FlowPane {
+public class DoubleComponent extends FlowPane implements ValuePropertyComponent<Double> {
 
   private final Double minimum;
   private final Double maximum;
   private final TextField textField;
+  private final ObjectProperty<Double> value = new SimpleObjectProperty<>();
```
robinschmid, line 44: not necessary, see comment with TextFormatter<Double>

robinschmid (dismissed):
Looks good only small comments

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/UnivariateBaselineCorrector.java
SteffenHeu: because the numerival value may be below zero for uv traces, leading to no correction for these traces.
With the new implementation:

grafik

robinschmid: oh ok. Maybe we can write that down as a comment there. 
I was wondering why it would be -

robinschmid: could one say that after the correction it needs to be >=0?
Or maybe for IonTimeSeries it needs to be >=0?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/DoubleComponent.java
SteffenHeu: makes it a ObjectProperty<Number>, not double

robinschmid: what do you mean?

robinschmid: You can also easily map between primitive and object properties:
```java
    DoubleProperty primProp = DoubleProperty.doubleProperty(value);
    ObjectProperty<Double> obProp = primProp.asObject();
```

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/DoubleComponent.java
SteffenHeu: does not work in the intended way for this application sadly

mzmine-community/src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
```diff
@@ -418,7 +418,7 @@ public String getAbsolutePath() {
   }
 
   @Override
-  public @NotNull List<@NotNull OtherDataFile> getOtherDataFiles() {
+  public @NotNull List<OtherDataFile> getOtherDataFiles() {
```
robinschmid, line 421: always wonder if we should just state this in the javadocs or even name the function like getReadOnlyOtherDataFiles. But this is also kind of ugly

mzmine-community/src/main/java/io/github/mzmine/util/javafx/SortableOtherFeatureComboBox.java
```diff
@@ -69,7 +70,21 @@ public SortableOtherFeatureComboBox() {
...
+      public String toString(OtherFeature f) {
+        if (f != null && f.getFeatureData() instanceof OtherTimeSeries series) {
+          return series.getName();
+        }
+        return f != null ? f.toString() : "";
+      }
+
+      @Override
+      public OtherFeature fromString(String string) {
+        return null;
```
robinschmid, line 84: Was just wondering when this is actually used and in combo box it is only used when commiting a change so if the combobox is editable. 
Kind of makes sense. The default implementation is kind of wild - when can you cast String to T? only if T is a String right?

Just a comment nothing to do here.

robinschmid (approved):
Looks good. Thanks Steffen

PR #2120, opened 2024-10-08
Zenodo download: Allow full archive download or regex file filter

robinschmid:
Zenodo records have a record ID that points to the latest record. But this is a redirct. This PR replaces #2119

- enables use of such latest record ID
- regex file filter to download only selected files in zenodo archive
- download whole archive and unzip


mzmine-community/src/main/java/io/github/mzmine/modules/io/download/ExternalAsset.java
```diff
@@ -73,8 +73,11 @@ public String getDownloadInfoPage() {
           "https://github.com/pluskal-lab/ThermoRawFileParserMacLinux/releases";
       case MSCONVERT -> "https://proteowizard.sourceforge.io/download.html";
       // libraries
-      case MSnLib -> "https://zenodo.org/records/11163381";
-      case MS2DEEPSCORE -> "https://zenodo.org/records/12628369";
+      case MSnLib -> "https://zenodo.org/records/11163380";
+      case GNPS_LIB -> "https://external.gnps2.org/gnpslibrary";
+      case MONA_LIB -> "https://mona.fiehnlab.ucdavis.edu/downloads";
+      case MASSBANK_EU -> "https://github.com/MassBank/MassBank-data/releases/latest";
```
SteffenHeu, line 79: there is no download form mzmine for this, right? What the license of the libraries? Could we provide a direct download? would we want to do that?

PR #2122, opened 2024-10-11
Handle empty json spectral libraries during import

robinschmid:


robinschmid:
tested and works for empty libraries

PR #2124, opened 2024-10-11
Fix NPE in StringParameter

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/StringParameter.java
```diff
@@ -57,20 +57,20 @@ public StringParameter(String name, String description, int inputsize) {
...
   }
 
-  public StringParameter(String name, String description, @NotNull String defaultValue,
+  public StringParameter(String name, String description, String defaultValue,
       boolean valueRequired) {
     this(name, description, defaultValue, valueRequired, false);
   }
 
-  public StringParameter(String name, String description, @NotNull String defaultValue,
+  public StringParameter(String name, String description, String defaultValue,
```
SteffenHeu, line 69: why remove the annotation? wouldnt it be even better to remove all calls that provide a null there?

robinschmid: getValue() should be NotNull and it is. 
But it is totally fine to pass a null value here or in setValue and the default behavior is to replace it with an empty string. This is also done in setValue

PR #2126, opened 2024-10-11
fix baseline correct unsupported operation in FeatureDataAccess

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/AbstractResolver.java
```diff
@@ -211,6 +211,7 @@ protected BinningMobilogramDataAccess getMobilogramDataAccess() {
       Arrays.fill(yBuffer, 0d);
       yBuffer = series.getIntensityValues(yBuffer);
 
+      // TODO this might be an issue x y buffer may contain empty values at the end - this might distort the results?
```
SteffenHeu, line 214: The chromatograms are mapped to all scans used for chromatogram building, which defines the length of the buffer. So there should be no zeros unless the ion was not detected. In that case, it should have zeros at the end.
Also, in local minimum the values are also removed by the threshold filter, i dont see how that would bee different.

mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/IonSpectrumSeries.java
```diff
@@ -107,7 +108,23 @@ default double getMzForSpectrum(T spectrum) {
...
+   * @param storage            May be null if the new series shall be stored in ram.
+   * @param newMzValues
+   * @param newIntensityValues
+   * @param scans              new list of spectra
+   * @return
+   */
   IonSpectrumSeries<T> copyAndReplace(@Nullable MemoryMapStorage storage,
-      @NotNull double[] newMzValues, @NotNull double[] newIntensityValues);
+      @NotNull double[] newMzValues, @NotNull double[] newIntensityValues,
+      final @NotNull List<@NotNull T> scans);
```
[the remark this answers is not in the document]
robinschmid: True this would be more like newInstance.
I introduced this in my first solution and I think later did not need it anymore. 
Or it was for one of the PR relying on this PR. 

With so many different classes this type could be - what is the current way of creating a new instance of the same type? 
I think this method might be as useful as the other copyAndReplace - maybe we can rename this one to newInstance though. 
It is the template builder pattern.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/AbstractResolver.java
robinschmid: true should not be an issue. When I found this I was just wondering if this makes a difference. 
For baseline correction it would make a difference but for feature detection it should not usually

robinschmid:
Should I remove the method or rename it to newInstance? 
Might be nice to have a template method to create new matching instances because otherwise there is no way to create those right?

SteffenHeu:
new instance may be ok from a naming perspective. But i dont know if it is necessary, it just increases the workload for new implementations. New instances can always be created using the constructor or the factory (in case of IonMobilogramTimeSeries)

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/AbstractResolver.java
SteffenHeu: The same mapping is done for baseline correction

mzmine-community/src/main/java/io/github/mzmine/datamodel/data_access/FeatureDataAccess.java
```diff
@@ -251,11 +250,26 @@ public IonTimeSeries<Scan> subSeries(@Nullable MemoryMapStorage storage,
...
 
+  /**
+   * Keeps mzs and scans and replaces intensities
+   *
+   * @param storage
+   * @param newIntensityValues
+   * @return
+   */
+  public IonSpectrumSeries<Scan> copyAndReplace(@Nullable MemoryMapStorage storage,
+      @NotNull double[] newIntensityValues) {
```
robinschmid, line 261: this method might be interesting to pull up into the interface IonSpectrumSeries.
This seems to be one the more common use cases to keep the mz and scans constant and just change the intensities. This method could then reuse the buffers or arrays in the different implementations.
What do you think @SteffenHeu

robinschmid:
This would be a Prototype pattern and makes it easier to construct the same type of object without requirement to make instance of checks and then build the object somewhere else. 
Other option would be a static factory that does instanceof checks. but I think as we already have the copyAndReplace method we are already using the prototype here.

I checked the code and as far as I can tell implementors had to implement the copy method and now they do not need to implement the copy and replace method but instead the newInstance method - so workload is the same. The old copyAndReplace method has a default that should work for all:

```java
  default IonSpectrumSeries<T> copyAndReplace(@Nullable MemoryMapStorage storage, @NotNull double[] newMzValues, @NotNull double[] newIntensityValues) {
    return newInstance(storage, newMzValues, newIntensityValues, getSpectra());
  }
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/AbstractResolver.java
robinschmid: Checked and all good

mzmine-community/src/main/java/io/github/mzmine/datamodel/data_access/MobilogramDataAccess.java
```diff
@@ -219,7 +222,21 @@ public IonSpectrumSeries<MobilityScan> copyAndReplace(@Nullable MemoryMapStorage
       @NotNull double[] newMzValues, @NotNull double[] newIntensityValues) {
     // depending on the type of data access, we can have more scans in currentSpectra than data points. (if 0s are included)
     // hence, this method is unsupported. A new SimpleIonMobilitySeries should be created with the respective spectra instead.
-    throw new IllegalArgumentException(
-        "MobilogramDataAccess shall be used to iterate over the mzs and intensities.");
+    // Will generate new series with zeroes included if that was the source for this data access
```
SteffenHeu, line 225: Is this method used somewhere? Having an implementation that is not optimised just for the sake of having an implementation is not ideal, as it may be a pain to debug in the future.

SteffenHeu (commented):
Got some comments

mzmine-community/src/main/java/io/github/mzmine/datamodel/data_access/MobilogramDataAccess.java
robinschmid: This was exactly the issue in AbstractBaselineCorrector.createNewTimeSeries. 
At least I hit the issue for FeatureDataAccess that did not implement the method. Therefore I thought it would be best to directly implement it for MobilogramDataAccess as well... 
I can remove it and make an issue to remember or I can change it in some other way, what do you think?

I thought the intended behavior would be to get a copy of the current mobilogram.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/AbstractBaselineCorrector.java
```diff
@@ -137,13 +139,15 @@ public static int removeRangesFromArray(List<IndexRange> indices, int numValues,
    */
   protected <T extends IntensityTimeSeries> T createNewTimeSeries(T timeSeries, int numValues,
       double[] newIntensities) {
+    final double[] newIntensityValues = Arrays.copyOfRange(newIntensities, 0, numValues);
     return switch (timeSeries) {
-      case IonSpectrumSeries<?> s -> (T) s.copyAndReplace(storage, s.getMzValues(new double[0]),
-          Arrays.copyOfRange(newIntensities, 0, numValues));
+      case FeatureDataAccess da -> (T) da.copyAndReplace(storage, newIntensityValues);
+      case MobilogramDataAccess da -> (T) da.copyAndReplace(storage, newIntensityValues);
```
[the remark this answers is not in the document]
robinschmid: Ok I think I added it because the initial cases captured was IonSpectrumSeries and I looked up what other classes extend this.

PR #2151, opened 2024-10-20
Optimize panama: auto delete tmp files as soon as possible

robinschmid:


utils/src/main/java/io/github/mzmine/util/files/FileAndPathUtil.java
```diff
@@ -744,6 +762,21 @@ public static FileChannel openTempFileChannel(String prefix, String suffix, Path
...
+        try {
+          // channel keeps file alive and we can still memory map and write, read to memory mapped file
+          // this will cause the file to disappear directly - tmp folder is empty
+          // tested on Windows 11:
+          // space is still taken from the disk of temp directory, but temp directory is empty
+          // GC will remove MemoryMapStorage, Arena, MemorySegment and with this automatically call
+          // munmap and unmap the file finally clearing the space on disk after GC
+          // TODO test on different platforms
+          // TODO macOS
+          // TODO linux (online read it should work)
```
SteffenHeu, line 774: can you share the sources?

SteffenHeu: How sure are you that is this not only working at the moment but is also designed and supposed to work like this? sounds more like a bug and an option to generate illegal memory accesses. 

Since it's memory mapping, the data is in RAM as long as the system decides to unload it. So while there is enough ram, the unloading may not be happening and all the current accesses ad read tests may just use the ram instead of the disk storage.

SteffenHeu (commented):
Im not so sure about this. sounds really experimental

utils/src/main/java/io/github/mzmine/util/files/FileAndPathUtil.java
robinschmid: We tried it on linux and windows. 
The files are always 0 MB as long as they are memory mapped. I tried a couple of things and they are only flushed to the file once GC unmaps the memory. 

I think what we are really deleting with file.delete() is the entry that is shown in the file explorer, the file itself is still there as long as there are references to the file like open channels, memory mapping, etc.
So its rather like hiding the file and setting a deleted flag on it. 

I read this in many posts. But cannot seem to find good sources at the moment. One is below: 
https://narkive.com/axMDuxWU:2.447.278

Idea is that the OS should be smart enough to secure your application from someone deleting a file that has open handles

robinschmid: Copilot also states its this way but the sources are off
image

robinschmid:
TODO add option to disable file delete

robinschmid:
All done from my side @SteffenHeu tests show that we save disk space faster with this approach.

PR #2155, opened 2024-10-21
Add message if specific spectral library is not imported

robinschmid:
Adds message and option to import now - only in GUI. Let batch fail in CLI mode

image


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/SpectralLibrarySelection.java
```diff
@@ -74,6 +88,92 @@ private List<SpectralLibrary> getMatchingSpecificFiles() {
...
+   */
+  public List<SpectralLibrary> getMatchingLibrariesAndCheckAvailability()
+      throws SpectralLibrarySelectionException {
+    if (this.getSelectionType() == SpectralLibrarySelectionType.SPECIFIC) {
+      List<File> missing = this.getMissingSpecificFiles();
+      if (!missing.isEmpty()) {
+        if (DesktopService.isGUI() && DialogLoggerUtil.showDialogYesNo(
+            "Import missing spectral libraries?", """
+                Some library files were not imported before spectral library matching - should mzmine import them now?
+                However, it is recommended to import spectral libraries during the initial data import with the MS data import or spectral library import. And maybe set the library selection to use all imported libraries.
```
SteffenHeu, line 127: Delete "However, "
The import module is called "Import MS data", better use the exact name here. Maybe just refert to the constant. 
"Then set the library selection to use all imported libraries."

robinschmid: done

PR #2158, opened 2024-10-22
New precursor mz search for spectral libraries

robinschmid:
Use the spectral library selection parameter and use the CompoundDBAnnotation for precursor matches against spectral libs

Inlcudes #2155 required changes to SpectralLibrarySelection


mzmine-community/src/main/java/io/github/mzmine/datamodel/structures/ComplexMolecularStructure.java
```diff
@@ -34,17 +34,11 @@
...
-public record ComplexMolecularStructure(
-     @NotNull IAtomContainer structure,
-     IMolecularFormula formula,
-     String inChIKey,
-     double monoIsotopicMass,
-     double mostAbundantMass,
-     int totalFormalCharge
-) implements MolecularStructure {
+public record ComplexMolecularStructure(@NotNull IAtomContainer structure,
+                                        IMolecularFormula formula, @Nullable String canonicalSmiles,
```
SteffenHeu, line 38: is the formula nullable or not?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
```diff
@@ -25,205 +25,140 @@
...
+    try {
+      final List<SpectralLibraryEntry> entries = parameters.getValue(
+          PrecursorDBSearchParameters.libraries).getMatchingLibraryEntriesAndCheckAvailability();
+
+      totalItems = entries.size();
+      finishedItems.set(0);
+
+      List<CompoundDBAnnotation> dbEntries = entries.stream().map(spec -> {
+            finishedItems.incrementAndGet();
+            return CompoundAnnotationUtils.convertSpectralToCompoundDb(spec);
```
SteffenHeu, line 144: Is it necessary to convert all library entries to a compound db annotation before it was matched? Is the common FeatureAnnotation interface not enough to handle mz and rt comparisons?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
```diff
@@ -25,205 +25,140 @@
...
 
-    // return tasks
-    parser.parse(this, dataBaseFile, null);
-    return tasks;
+  @Override
+  protected @NotNull List<FeatureList> getProcessedFeatureLists() {
+    return featureLists;
   }
 
   protected boolean checkMZ(FeatureListRow row, Double mz) {
```
SteffenHeu, line 164: not used anymore i guess

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
```diff
@@ -25,205 +25,140 @@
...
-    // Add task description to peakList
-    peakList.addDescriptionOfAppliedTask(new SimpleFeatureListAppliedMethod(
-        "Possible precursor identification using MS/MS spectral libraries " + dataBaseFile,
-        PrecursorDBSearchModule.class, parameters, getModuleCallDate()));
+      entryIndexStart = indexRange.min(); // skip a few entries next time
 
-    setStatus(TaskStatus.FINISHED);
+      indexRange.forEach(index -> {
+        CompoundDBAnnotation db = mzSortedEntries.get(index);
+        if (checkRT(row, db.getRT())) {
```
SteffenHeu, line 118: add ccs and mobility checks if available in the library and set in the row

mzmine-community/src/main/java/io/github/mzmine/util/annotations/CompoundAnnotationUtils.java
```diff
@@ -191,4 +198,37 @@ public static Stream<FeatureAnnotation> streamFeatureAnnotations(final FeatureLi
...
+        try {
+          logger.finer(
+              "Cannot convert value from spectral library to compound DB entry: %s = %s".formatted(
+                  dataType.getName(), value));
+        } catch (Exception ex) {
+        }
+      }
+    });
+
+    // currently spec library entry uses MZType and not PrecursorMZType for mz so enter it in both
```
SteffenHeu, line 226: there should be no mz type set to compound dbs to avoid confusion in future downstream modules.

mzmine-community/src/main/java/io/github/mzmine/datamodel/structures/ComplexMolecularStructure.java
```diff
@@ -34,17 +34,11 @@
...
  */
-public record ComplexMolecularStructure(
-     @NotNull IAtomContainer structure,
-     IMolecularFormula formula,
-     String inChIKey,
-     double monoIsotopicMass,
-     double mostAbundantMass,
-     int totalFormalCharge
-) implements MolecularStructure {
+public record ComplexMolecularStructure(@NotNull IAtomContainer structure,
```
SteffenHeu, line 37: If the point of this class is that the values are precomputed in comparison to the SimpleMolecularStructureType then this one should be called PrecomputedMolecularStructure to make it clearer

SteffenHeu (commented):
Thanks, got some comments

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
robinschmid: I think it may be better to convert beforehand - or I could build up a HashMap as a cache of converted db entries. I expect that often there are multiple m/z matching and this way we only have one instance that we retain. So upfront more memory used but then later less memory retained. 

Should I try to implement the cache instead of upfront?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
robinschmid: what do you mean with set in the row?

mzmine-community/src/main/java/io/github/mzmine/util/annotations/CompoundAnnotationUtils.java
robinschmid: Is MZType used for anything then?

mzmine-community/src/main/java/io/github/mzmine/datamodel/structures/ComplexMolecularStructure.java
robinschmid: just thought the same that the name is confusing. 
Precomputed is much better

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
SteffenHeu: only compare ccs and mobility if the row also has these values set. So if library has a ccs, but row does not, don't compare those. Only if there is a ccs set in both.

mzmine-community/src/main/java/io/github/mzmine/util/annotations/CompoundAnnotationUtils.java
SteffenHeu: not in the compound db. only in the features and rows.

mzmine-community/src/main/java/io/github/mzmine/datamodel/structures/ComplexMolecularStructure.java
robinschmid: not sure why this was missing. 
@Nullable

robinschmid: Changed it to NotNull. At least the CDK methods look like they dont throw exceptions or return anything else than a formula

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
robinschmid: removed

mzmine-community/src/main/java/io/github/mzmine/util/annotations/CompoundAnnotationUtils.java
robinschmid: I will just remove the mapping for MZType. First thought about skipping the addition of MZType in the loop above but now thought it is the easiest this way.
I added javadoc to CompoundDBAnnotation.put and added a if(value==null) map.remove(key); 
This is similar to the ModularDataModel behavior.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
robinschmid: Actually now uses the CompoundDBAnnotation.checkMatchAndCalculateDeviation for the matching to make it consistent with the Local DB matching. 
So we need to convert this in advance.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
robinschmid: I am using the CompoundDBAnnotation.checkMatchAndCalculateDeviation method now to make it work exactly like local csv database.

mzmine-community/src/main/java/io/github/mzmine/datamodel/structures/StructureUtils.java
```diff
@@ -0,0 +1,190 @@
...
+import org.jetbrains.annotations.Nullable;
+import org.openscience.cdk.exception.CDKException;
+import org.openscience.cdk.inchi.InChIGenerator;
+import org.openscience.cdk.interfaces.IAtomContainer;
+import org.openscience.cdk.interfaces.IMolecularFormula;
+import org.openscience.cdk.smiles.SmilesGenerator;
+import org.openscience.cdk.tools.manipulator.AtomContainerManipulator;
+import org.openscience.cdk.tools.manipulator.MolecularFormulaManipulator;
+
+public class StructureUtils {
```
SteffenHeu, line 42: Looks like a class we should have tests for, I'll open a new issue so we do it in the future.

PR #2172, opened 2024-10-28
add import for mbi files via msconvert

SteffenHeu:
for testing only

- need to add version check for msconvert if mbi file is imported
- show dialog if non supported version is installed


robinschmid:
Waiting for msconvert version

robinschmid (approved):

PR #2180, opened 2024-10-30
Image correlation avoid GC and precompute data, improve performance

robinschmid:
- Overall improvements to performance, results, and setup.

- Removes uneeded parameters


robinschmid:
all updated

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingParameters.java
```diff
@@ -56,40 +56,48 @@ public class ImageCorrelateGroupingParameters extends SimpleParameterSet {
...
+  @Override
+  public int getVersion() {
+    return 2;
+  }
+
+  @Override
+  public Map<String, Parameter<?>> getNameParameterMap() {
+    // parameters were renamed but stayed the same type
+    var nameParameterMap = super.getNameParameterMap();
+    // we use the same parameters here so no need to increment the version. Loading will work fine
```
SteffenHeu, line 99: remove this comment, as the version was bumped due to the removed parameter

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
```diff
@@ -175,27 +186,98 @@ public void checkAllFeatures(R2RMap<RowsRelationship> mapSimilarity)
...
+   */
+  private record FilteredRowData(double[] intensities, double noiseLevelOrLowerPercentile,
+                                 double upperPercentile) {
+
+
+    private static FilteredRowData create(final double[] intensities, final double lowerQuantile,
+        final double upperQuantile, double noiseLevel, final Transform transform) {
+      if (transform != null) {
+        for (int i = 0; i < intensities.length; i++) {
+          intensities[i] = transform.transform(intensities[i]);
```
SteffenHeu, line 214: the noise level also needs to be transformed if it is an absulute intensity. Also the javadoc does not match the parameters

SteffenHeu: Also a transform function may be Math.log (not in this case, but the method would mermit it), which cannot handle 0 as input.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
```diff
@@ -175,27 +186,98 @@ public void checkAllFeatures(R2RMap<RowsRelationship> mapSimilarity)
...
+              lowerQuantile);
+        }
+        if (useUpperQuantile && nonZeroIndex >= 0) { // otherwise filter off
+          upperPercentile = MathUtils.calcQuantileSorted(sorted, nonZeroIndex, sorted.length,
+              upperQuantile);
+        }
+      }
+
+      // minimum is 1
+      noiseLevel = Math.max(1, Math.max(noiseLevel, lowerPercentile));
```
SteffenHeu, line 240: transform noise level

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
```diff
@@ -175,27 +186,98 @@ public void checkAllFeatures(R2RMap<RowsRelationship> mapSimilarity)
...
+      }
+      // percentiles are optional
+      // use 1 as minimum to cut out 0 intensity
+      double lowerPercentile = 1;
+      double upperPercentile = 0;
+      boolean useUpperQuantile = upperQuantile < 1;
+      boolean useLowerQuantile = lowerQuantile > 0;
+      if (useLowerQuantile || useUpperQuantile) {
+        double[] sorted = Array.copy(intensities);
+        Arrays.sort(sorted);
```
SteffenHeu, line 225: Not sure how Google's Quantiles is implemented, but it does not require sorting. So unless the sorting is not required elsewhere, it may be more efficient to use that method rather than sorting thousands of values for every feature.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderImagingDda.java
```diff
@@ -122,29 +120,21 @@ public BatchQueue createQueue() {
...
-      Boolean applyMedianFilter, Boolean applyQuantileFilter, Boolean applyHotspotRemoval) {
+      Boolean applyQuantileFilter, Boolean applyHotspotRemoval) {
 
     ParameterSet param = MZmineCore.getConfiguration()
         .getModuleParameters(ImageCorrelateGroupingModule.class).cloneParameterSet();
 
     param.setParameter(ImageCorrelateGroupingParameters.FEATURE_LISTS,
         new FeatureListsSelection(FeatureListsSelectionType.BATCH_LAST_FEATURELISTS));
-    param.getParameter(ImageCorrelateGroupingParameters.NOISE_LEVEL).setValue(1E2);
+    param.setParameter(ImageCorrelateGroupingParameters.NOISE_LEVEL, 1d);
```
SteffenHeu, line 130: 1? so just not zero? Would it make sense to add this limitation to the noise level parameter then? Or at least set the default to 1 there as well, since all zeros are removed anyway?

SteffenHeu (commented):
Some comments

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
robinschmid: good catch. difficult how to handle 0 values. I think for intensities it may be ok to just keep 0 as 0 after the transformation. 
Especially in this case the 0 values will not be used either way.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
robinschmid: good catch. done now above together with intensities

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
robinschmid: ran some tests and our implementation seems faster. 
They create many different objects, a copy of the data, and also apply sorting to parts of the array. 
jmhSortQuant is our implementation
```
Benchmark                        (arraySize)  Mode  Cnt       Score       Error  Units
QuantilesBenchmark.jmhGuava             1000  avgt    5    3028.518 ±  1231.455  ns/op
QuantilesBenchmark.jmhGuava            10000  avgt    5   25999.120 ±   213.814  ns/op
QuantilesBenchmark.jmhGuava           100000  avgt    5  286787.598 ± 44183.272  ns/op
QuantilesBenchmark.jmhSortQuant         1000  avgt    5    1420.291 ±    15.484  ns/op
QuantilesBenchmark.jmhSortQuant        10000  avgt    5   15339.470 ±   207.421  ns/op
QuantilesBenchmark.jmhSortQuant       100000  avgt    5  183918.745 ±  1124.522  ns/op
```

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderImagingDda.java
robinschmid: added a constraint to a minimum intensity 0.01 for noise level. this should work with any type of number formatting

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
robinschmid: Another run with less other tools open. 
```
Benchmark                        (arraySize)  Mode  Cnt       Score       Error  Units
QuantilesBenchmark.jmhGuava             1000  avgt    5    2795.275 ±    16.135  ns/op
QuantilesBenchmark.jmhGuava            10000  avgt    5   26143.266 ±   645.275  ns/op
QuantilesBenchmark.jmhGuava           100000  avgt    5  282646.764 ± 50927.023  ns/op
QuantilesBenchmark.jmhSortQuant         1000  avgt    5    1476.845 ±    36.310  ns/op
QuantilesBenchmark.jmhSortQuant        10000  avgt    5   15251.086 ±   113.397  ns/op
QuantilesBenchmark.jmhSortQuant       100000  avgt    5  184097.835 ±   693.665  ns/op
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
SteffenHeu: How about just putting this in a try catch and adding a throws to the Transform interface. Thereby we can keep the simplicity of the interface.

I think we should already have a TryCatch utils class somewhere that can handle this nicely

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
robinschmid: Here another with 1M added as last row but first row with 1000 removed:
```
Benchmark                        (arraySize)  Mode  Cnt        Score        Error  Units
QuantilesBenchmark.jmhGuava            10000  avgt    5    27303.483 ±   3923.195  ns/op
QuantilesBenchmark.jmhGuava           100000  avgt    5   286652.349 ±  49363.678  ns/op
QuantilesBenchmark.jmhGuava          1000000  avgt    5  3492558.918 ±  55483.285  ns/op
QuantilesBenchmark.jmhSortQuant        10000  avgt    5    15298.276 ±    776.445  ns/op
QuantilesBenchmark.jmhSortQuant       100000  avgt    5   185344.906 ±  11798.365  ns/op
QuantilesBenchmark.jmhSortQuant      1000000  avgt    5  2373608.253 ± 594322.591  ns/op
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
robinschmid: I think throwing exceptions in a loop like this is a huge performance impact.
I think it could be better to add methods that handle 0 values like the one I added. 

Or force all transforms to be 0 safe

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_imagecorrelate/ImageCorrelateGroupingTask.java
SteffenHeu: 👍

PR #2182, opened 2024-10-30
handling empty spectra in imported libraries, only problem if datapoints are: "peaks":[] not "peaks":[[]]

corinnabrungs:


robinschmid:
looks good to me. @SteffenHeu what do you think?

PR #2183, opened 2024-10-30
changing software name to mzmine instead of mzmine#

corinnabrungs:


SteffenHeu:
I think the version of mzmine should be included since the spectral library format may change at some point.
You can use `SemverVersionReader.getMZmineVersion()` for that

PR #2199, opened 2024-11-07
Import mrm traces from mzml

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/otherdectectors/MsChromatogramPolarityType.java
```diff
@@ -0,0 +1,69 @@
...
+import io.github.mzmine.util.ParsingUtils;
+import javafx.beans.property.Property;
+import javafx.beans.property.SimpleObjectProperty;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import javax.xml.stream.XMLStreamWriter;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class MsChromatogramPolarityType extends DataType<PolarityType> implements NullColumnType,
```
robinschmid, line 22: could this be just PolarityType (which we also do not have yet but maybe useful in MS feature list if we ever get to merge polarities in one list)
Or is this required to be more specific by its name?
Also it could be a column that just shows + or - as text?

mzmine-community/src/main/java/io/github/mzmine/datamodel/msms/ActivationMethod.java
```diff
@@ -54,4 +58,17 @@ public String getName() {
   public String getUnit() {
     return unit;
   }
+
+  public static ActivationMethod fromActivationType(ActivationType type) {
```
robinschmid, line 62: could annotate argument as @Nullable
And could then simplify the caller and remove the null check there. But optional.

robinschmid (dismissed):
Just one question if we could just rename the new class to PolarityTypeType. 
Could be useful to have a single PolarityType data type that can be used in CompoundDBAnnotation, SpectralLibraryEntry, FeatureListRow

Can be merged like this or after rename

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/otherdectectors/MsChromatogramPolarityType.java
SteffenHeu: I think for feature lists we should rely on the feature.getRepresentativeSpectrum.getPolarity, because it is directly connected to the raw data. That's why the name is so specific. For libraries we usually know what adduct it is and can rely on that. 

If we want to expand this, we should add setting of the polarity in the feature data utils.

robinschmid: This could make sense. For the feature it could be a linked data type.
One issue is that sometimes we dont have a representative scan so this may be null. 
So could be still valid to put a simple PolarityTypeType. 
But we can do this once we merge across polarity

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_mzml/ConversionUtils.java
```diff
@@ -460,8 +460,8 @@ private static void extractAndSetMsMsInfoToChromatogram(MzMLChromatogram chrom,
       final Float energy =
           activationInfo != null ? Objects.requireNonNullElse(activationInfo.getActivationEnergy(),
               0d).floatValue() : null;
-      final ActivationMethod method = activationInfo != null ? ActivationMethod.fromActivationType(
-          activationInfo.getActivationType()) : null;
+      final ActivationMethod method = ActivationMethod.fromActivationType(
+          activationInfo != null ? activationInfo.getActivationType() : null);
```
robinschmid, line 464: true still required the null check, sorry thought it would make it easier but its only a little bit easier to call

PR #2200, opened 2024-11-07
Enable merged spectra import from mzml

SteffenHeu:
- disable centroiding for agilent & waters ims files (not supported by msconvert)
- add min detections to mob scan merging
- fix error for base peak offsets from mzml
- Enable merged spectra import from mzml (default)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_mobilityscanmerger/MobilityScanMergerParameters.java
```diff
@@ -60,11 +61,17 @@ public class MobilityScanMergerParameters extends SimpleParameterSet {
   public static final ScanSelectionParameter scanSelection = new ScanSelectionParameter();
 
   public static final MZToleranceParameter mzTolerance = new MZToleranceParameter(
-      ToleranceType.SCAN_TO_SCAN, 0.003, 15, false);
+      ToleranceType.SCAN_TO_SCAN, 0.005, 15, false);
+
+  public static final IntegerParameter minNumberOfDetections = new IntegerParameter(
+      "Minimum detections per m/z", """
+      Specify in how many mobility scans a summed m/z signal has to appear to be included in the frame.
```
robinschmid, line 68: Looks like a good addition. 
Maybe this could be an AbsoluteAndRelativeIntParameter?
I could imagine that setting a percentage could work better for this?
Considering that many mobility scans may be empty we could maybe use the maximum number of detections as the 100% instead of the total number of scans. 

Ok I see the method used at the end uses Integer minSignals as input. Could be an idea for the future. Just a question as this would be a breaking change to this parameterset in the future.

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_mzml/msdk/data/BuildingMzMLMsScan.java
```diff
@@ -645,6 +657,66 @@ public boolean loadProcessMemMapMzData(final MemoryMapStorage storage,
     return true;
   }
 
+  public BuildingMobilityScanStorage loadProccessMemMapMzDataForMergedMobilityScan(
```
robinschmid, line 660: Maybe better to add a comment here that this refers to the msconvert option "Combine ion mobility scans" that merges writes all mobility scans of a frame into a single scan object. 
Optionally this could also be added then to the corresponding BuildingMobilityScanStorage constructor.

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_mzml/msdk/data/MzMLParser.java
```diff
@@ -210,6 +209,11 @@ public void processOpeningTag(XMLStreamReader xmlStreamReader, String openingTag
             manageCompression(vars.binaryDataInfo, accession);
           } else if (arrayTypeMap.get(accession) != null) {
             vars.binaryDataInfo.setArrayType(arrayTypeMap.get(accession));
+            final String unitAccession = xmlStreamReader.getAttributeValue(null, "unitAccession");
+            vars.binaryDataInfo.setUnitAccession(unitAccession);
+          } else if (accession.equals("MS:1003006")) {
+            throw new IllegalStateException(
+                "Importing timsTOF files from mzML with the --combineIonMobilitySpectra option is not supported");
```
robinschmid, line 216: maybe mention that
"It is recommended to load timsTOF files from the native format."

mzmine-community/src/main/java/io/github/mzmine/util/RawDataFileTypeDetector.java
```diff
@@ -252,28 +259,46 @@ public static WatersAcquisitionType detectWatersAcquisitionType(File watersFolde
...
+        return new WatersAcquisitionInfo(WatersAcquisitionType.MSE, mobilityCounter.matches > 0,
+            polarity);
       } else if (parentCounter.getMatches() == 1) {
-        return WatersAcquisitionType.MS_ONLY;
+        return new WatersAcquisitionInfo(WatersAcquisitionType.MS_ONLY, mobilityCounter.matches > 0,
+            polarity);
       }
-      return WatersAcquisitionType.MSE;
+      return new WatersAcquisitionInfo(WatersAcquisitionType.MSE, mobilityCounter.matches > 0,
+          polarity);
```
robinschmid, line 301: same return value as else if (parentCounter.getMatches() > 1 
maybe remove this else if clause to make it clearer that there are 3 different return types. MSE, DDA, MS_ONLY

robinschmid (commented):
Minor comments looks good

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_mobilityscanmerger/MobilityScanMergerParameters.java
SteffenHeu: I think having a percentage does not make sense. It's to rule out electrical noise, that only appears in a few signals. While the number of mobility scans changes based on the instrument settings, the occurrence of these noise signals does not change, as the are random.

mzmine-community/src/main/java/io/github/mzmine/util/RawDataFileTypeDetector.java
SteffenHeu: it's more about making clear that the file did not conform to what we expect. added a logging message instead.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_mobilityscanmerger/MobilityScanMergerParameters.java
robinschmid: Still if you have only 2 mobility scans or 20 or 200 that makes a difference. Similar to the minimum detections in samples it may be better to just say require in 2 mobility scans and 30% or something like that.
Relative percentages may be even safer because people dont have to change it much to adjust to larger datasets or other acquisition methods.

Issue #2197, opened 2024-11-06
Export data from Volcano Plot


AntoineEtourneau:
After creating a Volcano Plot on mzmine 4.1.0 from a feature list, when exporting the data to Excel, we get a table with the columns log2(fold change) and -log10(p-Value) but not associated with the feature ID, mass or retention time. This association would be useful.

The information exists since we can already click on each point of the volcano blot to obtain the bloxplot with the corresponding feature ID.

So my question is: would it be possible when exporting to add information on the feature ID at a minimum, or even the m/z and RT?

SteffenHeu:
Good idea, will add the feature id as a z-axis, so it's in the exported data, which will also include the tooltip.

PR #2206, opened 2024-11-08
Chart data export - include tooltips

SteffenHeu:
Tooltips of datasets are now included when exporting the data from a chart. May be handy for generating some plots.

Adresses #2197

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAViewBuilder.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAViewBuilder.java
@@ -183,7 +183,6 @@ public class PCAViewBuilder extends FxViewBuilder<PCAModel> {
     if (newValue == null || newValue.isEmpty()) {
       return;
     }
-//    newValue.forEach(d -> plot.addDataset(d.dataset(), d.renderer()));
 
     LegendItemCollection collection = new LegendItemCollection();
     newValue.forEach(d -> {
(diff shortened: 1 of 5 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAViewBuilder.java
```diff
@@ -183,7 +183,7 @@ private void setDatasets(final SimpleXYChart<?> plot, final List<DatasetAndRende
     if (newValue == null || newValue.isEmpty()) {
       return;
     }
-    newValue.forEach(d -> plot.addDataset(d.dataset(), d.renderer()));
+//    newValue.forEach(d -> plot.addDataset(d.dataset(), d.renderer()));
```
robinschmid, line 186: remove? or why comment?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoDatasetProvider.java
```diff
@@ -97,4 +99,23 @@ public void computeValues(Property<TaskStatus> status) {
...
   }
 
+  @Override
+  public @Nullable PaintScale getPaintScale() {
+    return null;
+  }
+
+  @Override
+  public double getZValue(int index) {
+    return results.get(index).row().getID();
```
robinschmid, line 109: is this just for adding it to the tooltip or what happens now that z is the row.id? 

Not sure why does XYZValueProvider require getPaintScale, getBoxHeight, getBoxWidth? Maybe there should be a simple version like the XYValueProvider that does not require any additional methods - just the getZValue()?

SteffenHeu: z is the row id and the tool tip contains additional information.

XYZValueProvider is mainly used for the ColoredXYZDataset, which requires getPaintScale, getBoxHeight, getBoxWidth. The ColoredXYShapeRenderer is just also compatible with it.

robinschmid (approved):

PR #2208, opened 2024-11-10
Optimize gc decon preview

ansgarkorf:
Spectral deconvolution preview now displays feature rt range. Spectra and EICs now have some color as grouped features in scatter plot.
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_spectraldeconvolutiongc/SpectralDeconvolutionPreviewPlot.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_spectraldeconvolutiongc/SpectralDeconvolutionPreviewPlot.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2024 The MZmine Development Team
+ * Copyright (c) 2004-2024 The mzmine Development Team
  *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
@@ -25,11 +25,16 @@
 package io.github.mzmine.modules.dataprocessing.featdet_spectraldeconvolutiongc;
 
 import com.google.common.collect.Range;
+import io.github.mzmine.datamodel.features.ModularFeature;
 import io.github.mzmine.gui.chartbasics.gui.javafx.EChartViewer;
+import java.awt.BasicStroke;
 import java.awt.Color;
 import java.awt.Rectangle;
+import java.awt.geom.Rectangle2D;
+import java.util.List;
 import org.jfree.chart.ChartFactory;
 import org.jfree.chart.JFreeChart;
+import org.jfree.chart.annotations.XYShapeAnnotation;
 import org.jfree.chart.axis.NumberAxis;
 import org.jfree.chart.plot.DatasetRenderingOrder;
 import org.jfree.chart.plot.IntervalMarker;
@@ -45,6 +50,7 @@ public class SpectralDeconvolutionPreviewPlot extends EChartViewer {
 
   private final XYPlot plot;
   private int datasetIndex;
+  private final XYLineAndShapeRenderer renderer;
 
   public SpectralDeconvolutionPreviewPlot(String title, String xAxisLabel, String yAxisLabel) {
     super(null);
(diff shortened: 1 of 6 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/pseudospectrumvisualizer/PseudoSpectrumFeatureDataSetCalculationTask.java
```diff
@@ -140,10 +150,17 @@ public void run() {
...
+        chromPlot.addFeatureDataSetSpecificColor(new FeatureDataSet(feature),
+            FxColorUtil.fxColorToAWT(featureColor));
+        for (var dataset : datasets) {
+          chromPlot.addMzRangeEicDataSet(dataset, FxColorUtil.fxColorToAWT(featureColor));
+        }
+      } else {
+        chromPlot.addFeatureDataSetRandomColor(new FeatureDataSet(feature));
+        for (var dataset : datasets) {
+          chromPlot.addDataSet(dataset);
+        }
```
SteffenHeu, line 163: Can't you simply use one of the Colored renderes? then they would just use the color you set in the dataset and we don't need 2 new methods in the tic plot

SteffenHeu: The ColoredAreaShapeRenderer or the ColoredXYLineRenderer should be the correct one. And then use addDatasetAndRenderer in a chart.applyWithNotifyChanges

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/pseudospectrumvisualizer/PseudoSpectrumVisualizerPane.java
```diff
@@ -49,6 +50,8 @@
 import javafx.geometry.Orientation;
 import javafx.scene.control.SplitPane;
 import javafx.scene.layout.BorderPane;
+import javafx.scene.paint.Color;
+import javax.annotation.Nullable;
```
SteffenHeu, line 54: import correct annotation

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_spectraldeconvolutiongc/SpectralDeconvolutionPreviewPlot.java
```diff
@@ -63,19 +69,41 @@ public SpectralDeconvolutionPreviewPlot(String title, String xAxisLabel, String
...
+
+      // Calculate box width and height
+      double boxWidth = rtEnd - rtStart;  // Width based on RT range
+      double boxHeight = 0.5;  // Fixed height; adjust as needed
+
+      // Create the rectangle centered at the feature's RT and m/z values
+      Rectangle2D box = new Rectangle2D.Double(rtStart, mz - boxHeight / 2, boxWidth, boxHeight);
+
+      // Create the annotation with the rectangle shape and color
+      XYShapeAnnotation annotation = new XYShapeAnnotation(box, new BasicStroke(1.0f), color,
```
SteffenHeu, line 99: so the size just comes from the annotation and not from the dataset? not sure if this is the best solution. Maybe the renderer and the dataset should solve that problem?

SteffenHeu (commented):
Got some questions.

Also: I noticed the SpectralDeconvolutionGCDialog  does not use the ParameterSetupDialogWithPreview. So it doesnt have the show preview checkbox as all the other previews. That would be nice to change.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_spectraldeconvolutiongc/SpectralDeconvolutionPreviewPlot.java
ansgarkorf: Yes, you are right that would be an option. I checked the performance differences. When deactivating chart notifications during plot population performance is fine.

ansgarkorf:
> Got some questions.
> 
> Also: I noticed the SpectralDeconvolutionGCDialog does not use the ParameterSetupDialogWithPreview. So it doesnt have the show preview checkbox as all the other previews. That would be nice to change.

done

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/pseudospectrumvisualizer/PseudoSpectrumFeatureDataSetCalculationTask.java
```diff
@@ -124,29 +133,46 @@ public void run() {
 
     List<ColoredXYDataset> datasets = new ArrayList<>();
 
+    var nextColor = featureColor != null ? featureColor
+        : (feature.getRawDataFile() != null && feature.getRawDataFile().getColor() != null)
+            ? feature.getRawDataFile().getColor()
+            : MZmineCore.getConfiguration().getDefaultColorPalette().getNextColor();
```
SteffenHeu, line 139: ConfigService over MzmineCore

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/pseudospectrumvisualizer/PseudoSpectrumFeatureDataSetCalculationTask.java
```diff
@@ -124,29 +133,46 @@ public void run() {
...
+      chromPlot.applyWithNotifyChanges(false, () -> {
 
-      for (var dataset : datasets) {
-        chromPlot.addDataSet(dataset);
-      }
-    });
+        chromPlot.removeAllFeatureDataSets(false);
 
+        ColoredAreaShapeRenderer coloredAreaShapeRenderer = new ColoredAreaShapeRenderer();
+        coloredAreaShapeRenderer.setDefaultFillPaint(FxColorUtil.fxColorToAWT(nextColor));
```
SteffenHeu, line 160: this should not be necessary as the color is read from the ColoredXYDataset.

ansgarkorf: correct

PR #2209, opened 2024-11-10
Add graphical tooltip to 4D feature plot

ansgarkorf:
https://github.com/user-attachments/assets/b3648a91-d974-46a0-850f-9562fd6134ee


ansgarkorf:
> Great addition. is there any way to place the tooltip at a corner of the data point?

Should be possible. Requires some calculation of data points and pixels and between AWT and javaFX.

PR #2223, opened 2024-11-14
update imaging wizard

ansgarkorf:


mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/BaseWizardBatchBuilder.java
```diff
@@ -1277,10 +1283,10 @@ protected void makeAndAddLipidAnnotationStep(final BatchQueue q, boolean checkMS
     param.setParameter(LipidAnnotationParameters.lipidChainParameters,
         new LipidAnnotationChainParameters());
     param.setParameter(LipidAnnotationParameters.mzTolerance, mzTolInterSample);
-    param.setParameter(LipidAnnotationParameters.searchForMSMSFragments, checkMS2);
-    if (checkMS2) {
+    param.setParameter(LipidAnnotationParameters.searchForMSMSFragments, (checkMS2 || isImaging));
```
robinschmid, line 1286: this one is kind of confusing to me. 
checkMS2 is always true so far - which is ok. 
But why should isImaging then set this to true if checkMS2 is off... 

I think this should just rely on checkMS2 and also the if below should only rely on checkMS2 or even better the if should be removed to always set all the embedded parameters.

isImaging should only change the keepUnconfirmedAnnotations I would think

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderImagingDda.java
```diff
@@ -116,6 +116,8 @@ public BatchQueue createQueue() {
 
     // annotation
     makeAndAddLibrarySearchStep(q, false);
+    makeAndAddLocalCsvDatabaseSearchStep(q, null);
+    makeAndAddLipidAnnotationStep(q, false);
```
robinschmid, line 120: argument should be checkMS2=true
Or even better just remove the checkMS2 argument as its always true. 

I thought the idea was to always check MS2 if present (which we only know at runtime) but then for imaging we will allow unconfirmed matches.

robinschmid (changes requested):

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderImagingDda.java
ansgarkorf: done

PR #2231, opened 2024-11-18
Add online help to raw data import

ansgarkorf:


mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_all/AllSpectralDataImportParameters.java
```diff
@@ -67,12 +68,13 @@ public class AllSpectralDataImportParameters extends SimpleParameterSet {
...
   public AllSpectralDataImportParameters() {
-    super(fileNames, //
+    super(new Parameter[]{fileNames, //
         advancedImport, // directly process masslists
         metadataFile, // metadata import
         sortAndRecolor, // sort and recolor
         // allow import of spectral libraries
-        SpectralLibraryImportParameters.dataBaseFiles);
+            SpectralLibraryImportParameters.dataBaseFiles},
+        "https://mzmine.github.io/mzmine_documentation/module_docs/io/data-import.html");
```
robinschmid, line 77: you can put this as the first argument and then add Parameter... This way you do not need to create a new Parameter array

PR #2244, opened 2024-11-19
Improve graphics export

robinschmid:
- Maximum size
- PLOT sizing
- Layout improvements of dialog

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/graphicsexport/ChartExportUtil.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/graphicsexport/ChartExportUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2024 The MZmine Development Team
+ * Copyright (c) 2004-2024 The mzmine Development Team
  *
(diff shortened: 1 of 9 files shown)


mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/graphicsexport/ChartExportUtil.java
```diff
@@ -127,35 +147,31 @@ public static void writeChartToImage(ChartPanel chart, GraphicsExportParameters
...
+    }
     // resize
     chart.setPreferredSize(size);
     chart.setMaximumSize(size);
     chart.setMinimumSize(size);
-    // repaint
-    if (repaint) {
-      chart.revalidate();
-      chart.repaint();
-    }
```
SteffenHeu, line 138: are you sure this is not necessary? i remember something along the lines of that the chart is not actually resized if it's not repainted. not sure if that was only on swing

robinschmid: was somehow always false... so I removed it

PR #2248, opened 2024-11-20
Add warning large dataset align and gap fill

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/align_common/BaseFeatureListAligner.java
```diff
@@ -218,6 +227,36 @@ public ModularFeatureList alignFeatureLists() {
...
+    double gbMemoryPerMillionFeatures = 3.74; // this is from 15 GB per 4M features
+    double maxMemoryGB = ConfigService.getConfiguration().getMaxMemoryGB();
+
+    logger.info("""
+        Alignment started on a total of %d rows across %d samples (mean %d rows).
+        Max memory available: %.1f GB""".formatted(totalRows, featureLists.size(), rowsPerList,
+        maxMemoryGB));
+
+    // estimate if there might be an issue with this size and memory
+    if (gbMemoryPerMillionFeatures / 1_000_000 * totalRows > maxMemoryGB) {
```
SteffenHeu, line 247: you are going straight from your calculation. I think a safety factor of maybe 0.8 might be adequate.

if (gbMemoryPerMillionFeatures / 1_000_000 * totalRows > maxMemoryGB * 0.8)

robinschmid: I thought only to show it when it's clearly on the edge but we can also reduce the trigger. 
I added 0.85 x maxMemory now.

PR #2259, opened 2024-11-26
refactor MZmineCore startup

SteffenHeu:
refactored mzmine core startup. 
moved all the args parsing to ConfigService and refactored all that into individual methods.


mzmine-community/src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -105,302 +98,165 @@ private MZmineCore() {
...
+    // add event listener
+    EventService.subscribe(mzEvent -> {
+      if (mzEvent instanceof AuthRequiredEvent) {
+        if (DesktopService.isGUI()) {
+          getDesktop().addTab(UsersTab.showTab());
+        } else {
+          try {
+            if (DesktopService.hasTerminalInput()) {
               UsersController.getInstance()
                   .loginOrRegisterConsoleBlocking(LoginOptions.CONSOLE_ENTER_CREDENTIALS);
```
robinschmid, line 147: I see I forgot somthing here. No impact so far as this should not make a difference for now. 
Can you add this for me:
if(CurrentUserService.isValid()) return; 
After the loginOrRegisterConsoleBlocking. No need to System.exit if login was successful.

robinschmid (changes requested):

utils/src/main/java/io/github/mzmine/util/io/SemverVersionReader.java
```diff
@@ -39,17 +39,17 @@ public static Semver getMZmineVersion() {
       ClassLoader myClassLoader = SemverVersionReader.class.getClassLoader();
       InputStream inStream = myClassLoader.getResourceAsStream("mzmineversion.properties");
       if (inStream == null) {
-        return new Semver("3-SNAPSHOT", SemverType.LOOSE);
+        return new Semver("4-SNAPSHOT", SemverType.LOOSE);
```
robinschmid, line 42: good catch

robinschmid (approved):

PR #2267, opened 2024-11-28
Add logs to temp sparse file creation

robinschmid:


utils/src/main/java/io/github/mzmine/util/files/FileAndPathUtil.java
```diff
@@ -783,9 +788,15 @@ public static FileChannel openTempFileChannel(String prefix, String suffix, Path
           }
         } catch (Exception e) {
         }
+        logger.fine("Open file channel to: " + f.toFile().getAbsolutePath());
         return channel;
       } catch (FileAlreadyExistsException e) {
         // ignore and try next file name
+      } catch (AccessDeniedException e) {
+        logger.log(Level.WARNING,
+            "Access denied - make sure to use temporary directory with write access. "
```
SteffenHeu, line 797: "You don't have write access in <path>. Please choose a temp directory with write access."

better to have the path in the log i think

robinschmid: pushed the No write access plus file path into the next line as this might be quite long

PR #2270, opened 2024-11-29
Harmonize intensity in mgf export

robinschmid:
Adds options for formatting of intensity values.

SIRIUS used to normalize max signal to 100 now uses scientific notation 1.0005E4. 

GNPS still uses the original intensity value without normalization. Scientific notation is untested so I removed this option. 

image

image


mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_gnps/fbmn/GnpsFbmnExportAndSubmitParameters.java
```diff
@@ -75,6 +76,10 @@ public class GnpsFbmnExportAndSubmitParameters extends SimpleParameterSet {
       "Merge MS/MS (experimental)",
       "Merge high-quality MS/MS instead of exporting just the most intense one.",
       new MsMsSpectraMergeParameters(), true);
+
+  // scientific format untested on GNPS FBMN
+  public static final NormalizeIntensityComboParameter NORMALIZER = NormalizeIntensityComboParameter.createWithoutScientific();
```
SteffenHeu, line 81: increment parameter set version

SteffenHeu: should we even add this parameter if we know it does not work for gnps? why not simply use the correct format for the gnps export out of the box?

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
```diff
@@ -87,27 +93,16 @@ public class SiriusExportParameters extends SimpleParameterSet {
...
-      become "blahSourceFeatureListNameblah.mgf"). If the file already exists, it will be overwritten.""",
+      to substitute with feature list name. (i.e. "prefix_\{SiriusExportTask.MULTI_NAME_PATTERN}_suffix.mgf" would
+      become "prefix_SourceFeatureListName_suffix.mgf"). If the file already exists, it will be overwritten.""",
       extensions, "sirius" // suffix
   );
 
   public SiriusExportParameters() {
-    super(new Parameter[]{FEATURE_LISTS, FILENAME, MERGE_PARAMETER, MZ_TOL, NEED_ANNOTATION,
-        EXCLUDE_MULTICHARGE, EXCLUDE_MULTIMERS});
+    super(FEATURE_LISTS, FILENAME, NORMALIZE, MERGE_PARAMETER, MZ_TOL, NEED_ANNOTATION,
```
SteffenHeu, line 102: increment parameter set version

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTask.java
```diff
@@ -266,16 +263,45 @@ public boolean exportRow(BufferedWriter writer, FeatureListRow row) throws IOExc
...
+    return actuallyExported > 0;
+  }
+
+
+  private @Nullable SpectralLibraryEntry getCorrelatedOrBestMS1Spectrum(final FeatureListRow row) {
+    final Feature bestFeature = row.getBestFeature();
+    if (bestFeature == null) {
+      // maybe no MS1 data?
+      logger.warning(
+          "Cannot export MS1 data for this feature list. This maybe due to missing MS1 data or unsupported workflow. mzmine will skip this MS1 scan.");
```
SteffenHeu, line 288: may be

SteffenHeu: `skip ms1 scan of row %s".formatted(FeatureUtils.rowToString(row))`

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTask.java
```diff
@@ -266,16 +263,45 @@ public boolean exportRow(BufferedWriter writer, FeatureListRow row) throws IOExc
...
+
+    final MassSpectrum correlated = generateCorrelationSpectrum(row, null);
+    if (correlated != null && correlated.getNumberOfDataPoints() > 1) {
+      return spectrumToEntry(MsType.CORRELATED, correlated, bestFeature);
+    } else {
+      // export best MS1
+      var ms1Scan = bestFeature.getRepresentativeScan();
+      if (ms1Scan == null) {
+        logger.fine(
+            "Best feature has no representative scan. This may be due to missing MS1 data or unsupported workflow. mzmine will skip this MS1 scan.");
```
SteffenHeu, line 300: add `FeatureUtils.featureToString(bestFeature)`

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_library_gnps_batch/GNPSLibraryBatchExportParameters.java
```diff
@@ -45,9 +46,12 @@ public class GNPSLibraryBatchExportParameters extends SimpleParameterSet {
       "Name of the output tsv file. " + "If the file already exists, it will be overwritten.",
       extensions, "gnps_lib");
 
+  // scientific format untested on GNPS library submission
+  public static final NormalizeIntensityComboParameter normalizer = NormalizeIntensityComboParameter.createWithoutScientific();
```
SteffenHeu, line 50: increment parameter set version

mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationParameters.java
```diff
@@ -85,8 +88,8 @@ public class LibraryBatchGenerationParameters extends SimpleParameterSet {
       new LibraryExportQualityParameters());
 
   public LibraryBatchGenerationParameters() {
-    super(flists, file, exportFormat, postMergingMsLevelFilter, metadata, mergeMzTolerance,
-        handleChimerics, quality);
+    super(flists, file, exportFormat, postMergingMsLevelFilter, metadata, normalizer,
+        mergeMzTolerance, handleChimerics, quality);
```
SteffenHeu, line 92: increment parameter set version

mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/formats/MGFEntryGenerator.java
```diff
@@ -124,44 +126,66 @@ public static String createMGFEntry(LibrarySubmitIonParameters param, DataPoint[
...
+        continue; // skip data point if formatted intensity value is 0
       }
+
+      dataBuilder.append(mzForm.format(dp.getMZ())).append(" ").append(formattedIntensity)
+          .append("\n");
+      numExportedValues++;
     }
-    s.append("END IONS").append(br);
-    return s.toString();
+    String spectralData = createValueLine(DBEntryField.NUM_PEAKS, numExportedValues) + dataBuilder;
```
SteffenHeu, line 184: i'm not a fan that this is nested in this method. makes it harder to maintain i think. would be clearer if the num peaks string was added in the `createMGFEntry` method below adding the other identifiers. you could generate the string for the data points before writing the entry fields, then write the regular fields, then the num peaks, and then append the spectrum

SteffenHeu: The SpectrumString already has the numPeaks in there, so should be simple to change and better to understand in the future

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/NormalizeIntensityOptions.java
```diff
@@ -0,0 +1,188 @@
...
+  public static NormalizeIntensityOptions[] valuesNoScientific() {
+    return Arrays.stream(values()).filter(
+            v -> v != ORIGINAL_SCIENTIFIC_FORMAT && !v.name().toLowerCase().contains("scientific"))
+        .toArray(NormalizeIntensityOptions[]::new);
+  }
+
+  @Override
+  public String toString() {
+    return switch (this) {
+      case ORIGINAL -> "Original";
```
SteffenHeu, line 53: Original is a bit unclear. Does it mean as it was done before in the module or does it retain the values as they are?

maybe "No normalization (decimal)" and "No normalization (scientific)"?

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_gnps/fbmn/GnpsFbmnExportAndSubmitParameters.java
SteffenHeu: also this parameter needs to be set from the wizard now.

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
SteffenHeu: also this parameter needs to be set from the wizard now.

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_library_gnps_batch/GNPSLibraryBatchExportParameters.java
SteffenHeu: not i nwizard right?

mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationParameters.java
SteffenHeu: also this parameter needs to be set from the wizard now.

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_gnps/fbmn/GnpsFbmnExportAndSubmitParameters.java
robinschmid: True forgot the wizard. 
Here it is the same behavior as before. So no need to up the version of the parameter. Basically just added choice.
People use this export for many other tools as well and also as a general export of MS2 data paired with CSV table, therefore I think the normalization options are great. Also makes it more comparable to SIRIUS export in case people want to export with the same normalization

robinschmid: wizard is done

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
robinschmid: I guess upping the version number is ok here because we actually changed the behavior. Now its not normalized but scientific notation - so may break compatibility with other tools. 
SIRIUS works fine.
It would be great if we could add a message as well what has been changed.

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportTask.java
robinschmid: great idea

robinschmid: logger.fine(
            "Best feature has no representative scan. This may be due to missing MS1 data or unsupported workflow. mzmine will skip MS1 scan of row "
            + FeatureUtils.rowToString(row));

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
robinschmid: Added version message to ParameterSet and BatchQueue is handling this now.

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_library_gnps_batch/GNPSLibraryBatchExportParameters.java
robinschmid: Not in wizard yet. 
Default has not been changed. I thought upping the version is only when there is some major change  or even breaking change.
We could add a message now - but even that seems not needed?

robinschmid: better to keep the messages for important changes

mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationParameters.java
robinschmid: Set from wizard now. Good catch. 
I would keep version number

mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/formats/MGFEntryGenerator.java
robinschmid: Makes sense. So just remove num peaks string here and write it in the calling method for more clarity.
Changed

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/NormalizeIntensityOptions.java
robinschmid: maybe in the context of the parameter name: Intensity normalization 
image

robinschmid:
Maybe another question @SteffenHeu should we change the scientific option to a checkbox next to the combobox?
This way we could combine the format and the normalization more freely. So we could have normalized to 100 but still benefit of scientific number format - like no issues with multiple magnitudes.
What do you think?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/NormalizeIntensityOptions.java
SteffenHeu: I think it does not have to sound "positive". No normalization is just clearer in my opinion.

robinschmid: changed to No normalization and removed the second original option by adding the scientific format combo

robinschmid:
All done

image

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
```diff
@@ -145,4 +141,21 @@ public boolean checkParameterValues(Collection<String> errorMessages) {
...
+  public int getVersion() {
+    return 2;
+  }
+
+  @Override
+  public @Nullable String getVersionMessage(final int version) {
+    return switch (version) {
+      case 2 -> """
+          Up to mzmine 3.4.3 the intensities were exported normalized to the highest signal as 100%. \
+          This mzmine version adds options to control normalization. The default changed to original intensities exported in scientific notation (e.g., 1.05E5).""";
```
SteffenHeu, line 155: mzmine 4.5 added... "this" makes no sense in the future.

mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationTask.java
```diff
@@ -369,10 +373,12 @@ private Map<Scan, ChimericPrecursorResults> handleChimericsAndFilterScansIfSelec
 
   private void exportEntry(final BufferedWriter writer, final SpectralLibraryEntry entry)
       throws IOException {
+    // TODO maybe skip empty spectra. After formatting the number of signals may be smaller than before
+    // if intensity is 0 after formatting
     String stringEntry = switch (format) {
       case msp -> MSPEntryGenerator.createMSPEntry(entry);
       case json -> MZmineJsonGenerator.generateJSON(entry);
-      case mgf -> MGFEntryGenerator.createMGFEntry(entry);
+      case mgf -> MGFEntryGenerator.createMGFEntry(entry, normalizer).spectrum();
```
SteffenHeu, line 381: are the spectra only normalised for mgf entries?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/IntensityNormalizer.java
```diff
@@ -0,0 +1,96 @@
...
+package io.github.mzmine.parameters.parametertypes;
+
+import io.github.mzmine.datamodel.DataPoint;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
+import org.jetbrains.annotations.NotNull;
+
+public record IntensityNormalizer(IntensityNormalizerOptions option, boolean scientificFormat) {
+
+  public static IntensityNormalizer createDefault() {
```
SteffenHeu, line 35: add javadoc for that the default is

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/IntensityNormalizerOptions.java
```diff
@@ -0,0 +1,173 @@
...
+import org.jetbrains.annotations.Nullable;
+
+public enum IntensityNormalizerOptions {
+  ORIGINAL, HIGHEST_SIGNAL_AS_100, HIGHEST_SIGNAL_AS_1, SUM_AS_100, SUM_AS_1;
+
+
+  @Override
+  public String toString() {
+    return switch (this) {
+      case ORIGINAL -> "No normalization";
```
SteffenHeu, line 43: rename enum constant

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_sirius/SiriusExportParameters.java
robinschmid: done

mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/batch/LibraryBatchGenerationTask.java
robinschmid: true, I added the normalzier to the other formats as well.

PR #2281, opened 2024-12-05
Use workspace

SteffenHeu:
Use the workspace menu instead of the menu fxml.
Went through the main menu again with git changelog to see which modules we recently added. Hopefully should be all of them.


mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/MainWindowController.java
```diff
@@ -330,6 +337,23 @@ public void initialize() {
       memoryBarLabel.setText("%.1f/%.1f GB used".formatted(usedMemGB, totalMemGB));
     }));
     memoryUpdater.play();
+
+    final AcademicWorkspace academicWorkspace = new AcademicWorkspace();
+    WorkspaceMenuHelper.addWorkspace(academicWorkspace);
+    if (WorkspaceMenuHelper.getDefaultWorkspaceId() == null) {
+      WorkspaceMenuHelper.setDefaultWorkspace(academicWorkspace);
```
robinschmid, line 344: This was a bit confusing WorkspaceMenuHelper seems to both work like a WorkspaceService with static behavior and like an implementation detail to a workspace like all the default methods for users etc.

Just an option but maybe we could add a WorkspaceService like the ConfigService etc?
Can be done in next PR

robinschmid (approved):

robinschmid:
Looks good to me. 
Will open a PR on this branch with minor adjustments fixes and renames

PR #2285, opened 2024-12-09
Action to publish packages to mzio

SteffenHeu:
Could not figure out yet how to publish the convention plugins as they are not included in the standard build logic.


.github/workflows/push_stable_packages.yml
```diff
@@ -0,0 +1,38 @@
...
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          java-version: '22.0.1'
+          distribution: 'temurin'
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@af1da67850ed9a4cedd57bfd976089dd991e2582 # v4.0.0
```
robinschmid, line 27: just uses: gradle/actions/setup-gradle@v4?

build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+                        url = uri("https://maven.pkg.github.com/mzio-gmbh/mzio_mzmine")
+                        credentials {
+                            username = System.getenv("PUBLISH_PACKAGE_USERNAME")
+                            password = System.getenv("PUBLISH_PACKAGE_TOKEN")
+                        }
+                    }
+                } else {
+                    maven {
+                        url = uri(layout.projectDirectory.dir("../local-repo/"))
+                    }
```
robinschmid, line 35: Is the local repo needed for the convention pugins to be always shared through artifactory?
Or does this also work without publishing it to local?

settings.gradle.kts
```diff
@@ -20,4 +20,4 @@ include(
     "javafx-framework",
     "config",
 )
-//includeBuild("convention-plugins")
\ No newline at end of file
+//includeBuild("convention-plugins")
```
robinschmid, line 23: so it seems this is not needed? Did you try if this links up the convention plugins or does it break something?

robinschmid (commented):
Looks good. Now just a question if it works to access the packages

.github/workflows/push_stable_packages.yml
SteffenHeu:
```
# This workflow uses actions that are not certified by GitHub.
# They are provided by a third-party and are governed by
# separate terms of service, privacy policy, and support
# documentation.

# GitHub recommends pinning actions to a commit SHA.
# To get a newer version, you will need to update the SHA.
# You can also reference a tag or branch, but the action may change without warning.
```

took the workflow from the github docs, had this comment

settings.gradle.kts
SteffenHeu: i does not cause the plugins to be published on gradlew publish

build.gradle.kts
SteffenHeu: i just thought it would be nice to be able to test the build locally as well in the future. that's why i added this.

robinschmid (approved):
only some comments

PR #2292, opened 2024-12-16
Non sealed wizard workflow

SteffenHeu:
Refactored the mzwizard to allow addition of more workflows.
The workflow now decides which WizardBatchBuilder is used.
The workflow is able to check the currently selected steps and thereby only show in the combobox if it is possible to build the workflow with the current settings. (through a `WizardPartFilter`)
The ims filter (due to selected ion source) and ms filter (due to selected ims) still use the previously used enums to guarantee exhaustiveness. (but apply the `WizardPartFilter` for it)


mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/WizardPartFilter.java
```diff
@@ -0,0 +1,53 @@
...
+   * @return true or false
+   */
+  boolean accept(@NotNull WizardParameterFactory part);
+
+  /**
+   *
+   * @param allowed The allowed steps
+   */
+  static WizardPartFilter allow(Collection<WizardParameterFactory> allowed) {
+    return part -> new HashSet<>(allowed).contains(part);
```
robinschmid, line 27: move creation of new HashSet before the predicate lambda.
so final var set = new HashSet<>(allowed)

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/WizardPartFilter.java
```diff
@@ -0,0 +1,53 @@
...
+  static WizardPartFilter allow(WizardParameterFactory... allowed) {
+    return allow(Arrays.asList(allowed));
+  }
+
+  /**
+   *
+   * @param denied The denied steps
+   */
+  static WizardPartFilter deny(Collection<WizardParameterFactory> denied) {
+    return part -> !new HashSet<>(denied).contains(part);
```
robinschmid, line 43: same here hashset creation outside the anonymous lambda

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/BatchWizardTab.java
```diff
@@ -179,28 +179,45 @@ private void evaluateWizardSequenceLimitChoices() {
         .orElse(IonInterfaceWizardParameterFactory.HPLC);
 
     // apply filters
-    filterComboBox(WizardPart.IMS, ionization.getMatchingImsPresets());
-    filterComboBox(WizardPart.WORKFLOW, ionization.getMatchingWorkflowPresets());
+    filterComboBox(WizardPart.IMS, WizardPartFilter.allow(ionization.getMatchingImsPresets()));
```
robinschmid, line 182: I like this allow or deny

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/WorkflowWizardParameters.java
```diff
@@ -29,10 +29,10 @@
 import io.github.mzmine.modules.tools.batchwizard.subparameters.factories.WorkflowWizardParameterFactory;
 import io.github.mzmine.parameters.Parameter;
 
-public sealed class WorkflowWizardParameters extends WizardStepParameters permits
+public non-sealed class WorkflowWizardParameters extends WizardStepParameters /*permits
     WorkflowDdaWizardParameters, WorkflowGcElectronImpactWizardParameters,
     WorkflowImagingWizardParameters, WorkflowLibraryGenerationWizardParameters,
-    WorkflowDiaWizardParameters, WorkflowTargetPlateWizardParameters {
+    WorkflowDiaWizardParameters, WorkflowTargetPlateWizardParameters*/ {
```
robinschmid, line 35: this comment can be removed then?

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/factories/IonInterfaceWizardParameterFactory.java
```diff
@@ -154,17 +160,16 @@ public IonMobilityWizardParameterFactory[] getMatchingImsPresets() {
   public WorkflowWizardParameterFactory[] getMatchingWorkflowPresets() {
     return switch (this) {
       case DIRECT_INFUSION, FLOW_INJECT, GC_CI ->
-          new WorkflowWizardParameterFactory[]{WorkflowWizardParameterFactory.DDA,
-              WorkflowWizardParameterFactory.LIBRARY_GENERATION};
+          new WorkflowWizardParameterFactory[]{new WorkflowDDA(),
+              new WorkflowLibraryGeneration()};
```
robinschmid, line 164: this would need to change in the future right?
Either making it a static map or even better pushing this knowledge to the workflow?
Like the workflow would then know if it is applicable to LC, GC, IMS, ... ?

This way only one class needs to be registered from the outside.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/factories/WorkflowWizardParameterFactory.java
```diff
@@ -25,77 +25,69 @@
...
-      case LIBRARY_GENERATION -> "Library generation";
-      case IMAGING -> "Imaging";
-      case TARGET_PLATE -> "Target plate";
-    };
+  private static final Set<WorkflowWizardParameterFactory> values = new LinkedHashSet<>();
+
+  static {
+    values.addAll(List.of(new WorkflowDDA(), new WorkflowDIA(), new WorkflowDeconvolution(),
+        new WorkflowLibraryGeneration(), new WorkflowImaging(), new WorkflowTargetPlate(),
+        new WorkflowMs1Only()));
```
robinschmid, line 58: I get a warning that:
Referencing subclass WorkflowMs1Only from superclass WorkflowWizardParameterFactory initializer might lead to class loading deadlock 

Maybe best to just separate this logic and create a WizardWorkflows class in workflows package with all the values related things:

```java
  private static final Set<WorkflowWizardParameterFactory> values = new LinkedHashSet<>(
      // default workflows
      List.of(new WorkflowDDA(), new WorkflowDIA(), new WorkflowDeconvolution(),
          new WorkflowLibraryGeneration(), new WorkflowImaging(), new WorkflowTargetPlate(),
          new WorkflowMs1Only()));

public static WorkflowWizardParameterFactory[] values() {
  return values.toArray(WorkflowWizardParameterFactory[]::new);
}

public static void addWorkflow(WorkflowWizardParameterFactory workflow) {
  values.add(workflow);
}
```

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/factories/workflows/WorkflowDDA.java
```diff
@@ -0,0 +1,64 @@
...
+    return new WorkflowDdaWizardParameters(true, true, null, true, true, false);
+  }
+
+  @Override
+  public String toString() {
+    return "DDA";
+  }
+
+  @Override
+  public Map<WizardPart, List<WizardPartFilter>> getStepFilters() {
```
robinschmid, line 40: not super important here but I had a similar problem in the scan selection. 
I think this will always be one filter that just allows or denies specific values for each part. 
So Map<WizardPart, WizardPartFilter> would make it much simpler also in the application of the filter (more readable - one loop less).
And if we need a list of filters we could do something like: WizardPartFilter.combine(WizardPartFilter... filters) and apply all filters in a new filter.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/factories/workflows/WorkflowDDA.java
```diff
@@ -0,0 +1,64 @@
...
+            IonInterfaceWizardParameterFactory.GC_CI, IonInterfaceWizardParameterFactory.HPLC,
+            IonInterfaceWizardParameterFactory.UHPLC, IonInterfaceWizardParameterFactory.HILIC))));
+  }
+
+  @Override
+  public @NotNull WizardBatchBuilder getBatchBuilder(final @NotNull WizardSequence steps) {
+    // throw in case we hit unsupported workflow
+    // those combinations should be filtered out previously though
+    var unsupportedException = new UnsupportedOperationException(
+        "Currently not implemented workflow " + this);
```
robinschmid, line 53: better to create this exception right at the throw. 
This already prepopulates the stack trace and everything so best pattern is to always just do it when its needed.

robinschmid: Maybe just create a new WorkflowUnsupportedException extends RuntimeException? 
This way we can have the same message in all workflows and easily add more info if it is needed later

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/WizardService.java
```diff
@@ -0,0 +1,12 @@
+package io.github.mzmine.modules.tools.batchwizard;
+
+import io.github.mzmine.modules.tools.batchwizard.builders.WizardBatchBuilder;
+import java.util.ArrayList;
+import java.util.List;
+
+public class WizardService {
```
robinschmid, line 7: remove?

robinschmid (approved):
Looks good to me. Only minor comments and ideas

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/factories/IonInterfaceWizardParameterFactory.java
SteffenHeu: Already not used anymore, this is already applied in the respective workflows

robinschmid: ok nice

PR #2293, opened 2024-12-16
Types for MRM processing

SteffenHeu:
This PR adds the feature types for the MRM processing.

Adds the types:
MrmTransition: contains q1 and q3 mass and the respective chromatogram (chromatograms have the q1 mass as mz)
MrmTransitionList: Combines multiple MrmTransitions, defines the quantifier
MrmTransitionListType: Carrier to store an MrmTransitionList in a feature

One thing to discuss: We could not process the MrmTransition in the MrmTransitionList (e.g. smoothing, resolving, baseline, maybe just not resolving) and in case of updating the quant transition, we just take a sub-series from current rt start to rt end. this would make re-integration easier, but we would have to calc the area of all qualifier ions in the rt of the feature (in case we do not resolve)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/otherdectectors/MrmTransitionListType.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.stream.Collectors;
+import javafx.beans.property.Property;
+import javafx.beans.property.SimpleObjectProperty;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import javax.xml.stream.XMLStreamWriter;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class MrmTransitionListType extends DataType<MrmTransitionList> {
```
robinschmid, line 22: does this have to be DataType<MrmTransitionList<?>> ?

mzmine-community/src/main/java/io/github/mzmine/datamodel/otherdetectors/MrmTransitionList.java
```diff
@@ -0,0 +1,221 @@
...
+  /**
+   * @param transitions The transitions of this transition list. The
+   *                    {@link MrmTransitionList#quantifier()} is set by determining the most
+   *                    intense {@link MrmTransition#chromatogram()}. All transitions must have the
+   *                    same q1 mass.
+   */
+  public MrmTransitionList(@NotNull final List<@NotNull MrmTransition<T>> transitions) {
+    final Optional<@NotNull MrmTransition<T>> max = transitions.stream()
+        .max(Comparator.comparingDouble(mrm -> FeatureDataUtils.getHeight(mrm.chromatogram())));
+    final double q1Mass = transitions.getFirst().q1mass();
```
robinschmid, line 56: finally we have code before super calls. So nice

mzmine-community/src/main/java/io/github/mzmine/datamodel/otherdetectors/MrmTransitionList.java
```diff
@@ -0,0 +1,221 @@
...
+import javax.xml.stream.XMLStreamReader;
+import javax.xml.stream.XMLStreamWriter;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Groups several {@link MrmTransition}s in the
+ * {@link io.github.mzmine.datamodel.features.ModularDataModel} for {@link ModularFeature}s.
+ */
+public final class MrmTransitionList<T extends Scan> {
```
robinschmid, line 27: does the <T> add value here? 
All constructors currently do not specify the type and avoid the <> all together. 
Maybe we could just make it Scan? because downstream in most/all places this will be undefined either way and we would need to do the checks for Frames or what ever scan type. 
ScanUtils really helps here in my experience to get the info from all types of scans

mzmine-community/src/main/java/io/github/mzmine/datamodel/otherdetectors/MrmTransition.java
```diff
@@ -0,0 +1,117 @@
...
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import javax.xml.stream.XMLStreamWriter;
+import org.math.io.files.DataFile;
+
+/**
+ * @param chromatogram The individidual transitions. Important note: the masses in the
+ *                     {@link IonTimeSeries} must be the q1 mass.
+ */
+public record MrmTransition<T extends Scan>(double q1mass, double q3mass,
```
robinschmid, line 25: same here - maybe remove the <T> and replace with scan and all the warning will be gone. This might make it easier in the code as well

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/BaselineCorrectionTask.java
```diff
@@ -100,6 +105,8 @@ protected void process() {
 
       final IonTimeSeries<? extends Scan> its = corrector.correctBaseline(access);
 
+      handleMrmFeature(feature);
```
robinschmid, line 108: maybe add some documentation here. 
Basically we are correcting the main trace already and then we are also correcting Mrm if available

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/BaselineCorrectionTask.java
```diff
@@ -113,6 +120,31 @@ protected void process() {
     handleOriginal.reflectNewFeatureListToProject(suffix, project, newFlist, originalFlist);
   }
 
+  private void handleMrmFeature(Feature feature) {
+    final ModularFeature f = (ModularFeature) feature;
+    if (f.get(MrmTransitionListType.class) instanceof MrmTransitionList<?> transitions) {
```
robinschmid, line 125: thats just a null check or is this to cast MrmTransitionList --> MrmTransitionList<?>?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/BaselineCorrectionTask.java
```diff
@@ -113,6 +120,31 @@ protected void process() {
...
+      final List<? extends Scan> allScans = newFlist.getSeletedScans(f.getRawDataFile());
+      final List<MrmTransition<?>> correctedTransitions = new ArrayList<>();
+      for (MrmTransition<?> transition : transitions.transitions()) {
+        // todo this may be optimised by an MrmDataAccess, similar to the feature data access,
+        //  if this limits performance
+        // remap so we get the same behaviour
+        final IonTimeSeries<Scan> remapped = IonTimeSeriesUtils.remapRtAxis(
+            transition.chromatogram(), allScans);
```
robinschmid, line 133: So before the baseline correction the RT axis might be different but then after the correction we also keep the corrected RT axis right?
Maybe we can just always remap the retention time? Does this make sense?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/FeatureResolverTask.java
```diff
@@ -39,31 +40,52 @@
...
+import java.util.stream.Collectors;
 import org.jetbrains.annotations.NotNull;
 
 public class FeatureResolverTask extends AbstractTask {
 
   // Logger.
   private static final Logger logger = Logger.getLogger(FeatureResolverTask.class.getName());
 
+  // These types will not be copied to a new feature
+  private final Set<DataType<?>> excludedTypes = DataTypes.getInstances().stream()
```
robinschmid, line 77: maybe give it a more descriptive name?
featureCopyExcludedTypes?

I was first confused what this is doing because of the mapMulti but its basically just a stream.filter

Maybe AnnotationType could be copied over? Seems fine if someone really decides to annotate chromatograms but not sure.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/BaselineCorrectionTask.java
SteffenHeu: it combines null check and assigning a variable. just less code

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/FeatureResolverTask.java
```diff
@@ -241,6 +259,26 @@ private void dimensionIndependentResolve(ModularFeatureList originalFeatureList)
     newPeakList = resolvedFeatureList;
   }
 
+  private void handleMrmTraces(ModularFeature f) {
```
robinschmid, line 262: maybe add some javadoc to describe the strategy. 
So basically just use the features resolved RT range and the remapping that is doen in baseline correct is not done here because its always applied before?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/BaselineCorrectionTask.java
SteffenHeu: Why would the rt axis be different? the rt is the ms1 rt and they are all uniform from the mrmtoscans module. this is just about having the full range of the rt axis.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/FeatureResolverTask.java
SteffenHeu: No, annotations will not be copied. you can always annotate with no rt filter if you really want to.

SteffenHeu: I thought the comment would make it clear

mzmine-community/src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
```diff
@@ -384,6 +384,15 @@ public String toString() {
     return scans;
   }
 
+  public void clearScans() {
```
robinschmid, line 387: Interesting - this is used by the new MRM workflow?

mzmine-community/src/main/java/io/github/mzmine/datamodel/otherdetectors/MrmTransitionList.java
SteffenHeu: Because a List< Frame> ist not a List< Scan>. Then you always have to cast in that case. this jsut accepts it.

mzmine-community/src/test/java/datamodel/OtherDataTest.java
```diff
@@ -131,11 +136,55 @@ void testMsOtherCorrelationResultType() {
...
-        PolarityType.POSITIVE);
-    DataTypeTestUtils.simpleDataTypeSaveLoadTest(new PolarityTypeType(),
-        PolarityType.UNKNOWN);
+    DataTypeTestUtils.simpleDataTypeSaveLoadTest(new PolarityTypeType(), PolarityType.NEGATIVE);
+    DataTypeTestUtils.simpleDataTypeSaveLoadTest(new PolarityTypeType(), PolarityType.POSITIVE);
+    DataTypeTestUtils.simpleDataTypeSaveLoadTest(new PolarityTypeType(), PolarityType.UNKNOWN);
+  }
+
+  @Test
+  void textMrmLoadSave() {
```
robinschmid, line 145: text too test typo

mzmine-community/src/main/java/io/github/mzmine/project/impl/RawDataFileImpl.java
SteffenHeu: yes, otherwise they would be re-added

mzmine-community/src/test/java/datamodel/OtherDataTest.java
SteffenHeu: too to to typo :D

mzmine-community/src/main/java/io/github/mzmine/datamodel/otherdetectors/MrmTransitionList.java
robinschmid: Where is List<Frame> or List<Scan> used?
I can only see that MrmTransition has IonTimeSeries<T> chromatogram, which in turn may has a list of scans or frames somewhere... But this does not affect the MrmTransition right?

All constructor calls pass in ? into the `MrmTransition<T>`  
Also all places that use MrmTransition will never really know if its Frames or Scans.

I think this might make the whole design and usage easier. Better if we can limit the number of warnings we get for "Raw use of parameterized class 'MrmTransition' "

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/BaselineCorrectionTask.java
robinschmid: not sure if its always better to have less code. 
instanceof implies that there may be multiple subclasses that you want to check and new people coming to the project might misread this. 
I think a standard null check is clearer but you can leave it if you want.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/BaselineCorrectionTask.java
robinschmid: oh ok. Reading this I directly thought about the different retention times of different MRM traces. 
So my brain auto completed the logic - but wrong. Was not checking which set of scans were used to initially build the traces and then remap here. 
Good then.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/FeatureResolverTask.java
```diff
@@ -238,7 +239,7 @@ private void dimensionIndependentResolve(ModularFeatureList originalFeatureList)
 
         f.set(FeatureDataType.class, resolved);
         FeatureDataUtils.recalculateIonSeriesDependingTypes(f);
-        handleMrmTraces(f);
+//        handleMrmTraces(f);
```
robinschmid, line 242: Maybe delete the method?

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/otherdetectors/integrationplot/IntegrationPlotController.java
```diff
@@ -0,0 +1,154 @@
...
+  protected IntegrationPlotController() {
+    super(new IntegrationPlotModel());
+
+    model.currentIntegrationValidProperty().bind(Bindings.createBooleanBinding(
+        () -> model.getCurrentStartTime() != null && model.getCurrentEndTime() != null
+            && model.getCurrentStartTime() < model.getCurrentEndTime(),
+        model.currentStartTimeProperty(), model.currentEndTimeProperty()));
+
+    model.currentStartTimeProperty().addListener((_, _, value) -> model.setCurrentStartMarker(
+        value != null ? new ValueMarker(model.getCurrentStartTime(),
```
robinschmid, line 57: potential NPE @Nullable Double -> double
Also below for end time

robinschmid (dismissed):

PR #2296, opened 2024-12-17
fix import of wiff2 files

SteffenHeu:
circumvents: https://github.com/ProteoWizard/pwiz/issues/3281


robinschmid (approved):
Interesting finding

PR #2298, opened 2024-12-18
Darkgrey

SteffenHeu:
grafik
grafik
grafik


robinschmid:
Seems quite nice... Maybe change gridline color in the table
They are invisible on the dark - visible on the mouse over though
image

SteffenHeu:
grafik

PR #2299, opened 2024-12-18
drop duplicate data points from other traces

SteffenHeu:
Apparently it is not guaranteed that data points for chromatograms are unique or sorted. Occurred in bruker chromatography files. I put the same changes into the mzml import to be safe, as the bruker files may be converted to mzml.

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_uv/BrukerUvReader.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_uv/BrukerUvReader.java
@@ -27,9 +27,11 @@ package io.github.mzmine.modules.io.import_rawdata_bruker_uv;
 
 import gnu.trove.list.array.TDoubleArrayList;
 import gnu.trove.list.array.TFloatArrayList;
+import io.github.mzmine.datamodel.DataPoint;
 import io.github.mzmine.datamodel.MassSpectrumType;
 import io.github.mzmine.datamodel.RawDataFile;
 import io.github.mzmine.datamodel.featuredata.impl.StorageUtils;
+import io.github.mzmine.datamodel.impl.SimpleDataPoint;
 import io.github.mzmine.datamodel.otherdetectors.OtherDataFile;
 import io.github.mzmine.datamodel.otherdetectors.OtherDataFileImpl;
 import io.github.mzmine.datamodel.otherdetectors.OtherFeature;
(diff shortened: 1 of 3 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_bruker_uv/BrukerUvReader.java
```diff
@@ -316,14 +319,39 @@ private OtherTimeSeries loadTimeSeriesData(MemoryMapStorage storage,
...
+   */
+  private @NotNull List<DataPoint> dropDuplicateRtDataPoints(Trace trace,
+      TFloatArrayList intensitiesList, TDoubleArrayList timesList) {
+    List<DataPoint> dps = new ArrayList<>();
+    for (int i = 0; i < intensitiesList.size(); i++) {
+      final SimpleDataPoint dp = new SimpleDataPoint(timesList.get(i), intensitiesList.get(i));
+      dps.add(dp);
+    }
+
+    // intermediate filtering to remove duplicates from traces. Why is this even necessary?
```
robinschmid, line 343: maybe the detector is super fast and the rounding is too high? Not sure.

SteffenHeu: No it's repeated values

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_mzml/ConversionUtils.java
```diff
@@ -436,6 +444,28 @@ public static List<OtherDataFile> convertOtherTraces(RawDataFile file,
...
+      MzMLChromatogram chrom) {
+    final List<DataPoint> dps = new ArrayList<>();
+    final float[] rts = chrom.getRetentionTimes();
+    final double[] intensities = chrom.getIntensities();
+    for (int i = 0; i < chrom.getNumberOfDataPoints(); i++) {
+      dps.add(new SimpleDataPoint(rts[i], intensities[i]));
+    }
+    dps.sort(DataPointSorter.DEFAULT_MZ_ASCENDING);
+    final int before = dps.size();
+    CollectionUtils.dropDuplicatesRetainOrder(dps);
```
robinschmid, line 460: same here - drop solely based on RT?

SteffenHeu: No it's repeated value pairs

PR #2313, opened 2025-01-02
allow import of chromatograms with same x values but different y values

SteffenHeu:
I think some vendors put the MS1 and MS2 TIC/BPC chromatograms at the same rt values -> not removed by the current filters. So we need to allow same rt but different intensity in the chromatograms


SteffenHeu:
grafik

robinschmid:
So they export one chromatogram with both MS1 and MS2 BPC? 
Thats weird behavior

SteffenHeu:
i think so

PR #2317, opened 2025-01-03
DreaMS molecular networking

roman-bushuiev:
This PR extends the `Feature list methods -> Feature grouping -> Spectral / Molecular Networking` module with a new option: `DreaMS`. When this option is selected, the module first loads a pre-trained DreaMS model and computes DreaMS embeddings for the fragmentation mass spectra in the selected feature lists. Next, it constructs a molecular network as a cosine similarity-based kNN graph using the computed embeddings.


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/dreams/DreaMSNetworkingTask.java
```diff
@@ -0,0 +1,259 @@
...
+
+            scanList.add(scan);
+            featureListRows.add(row);
+        }
+
+        // Predict the matrix of pairwise DreaMS similarities
+        float[][] similarityMatrix;
+        try {
+            // Pre-process mass spectra
+            float[][][] tensorizedSpectra = model.spectrumTensorizer.tensorizeSpectra(scanList);
```
robinschmid, line 146: spectrumTensorizer should be private you never provide public access to objects. 
You could either do getSpectrumTensorizer or create a method in model that forwards to tensorizeSpectra

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/dreams/DreaMSNetworkingTask.java
```diff
@@ -0,0 +1,259 @@
...
+                similarityMatrix);
+        R2RNetworkingMaps rowMaps = featureList.getRowMaps();
+        rowMaps.addAllRowsRelationships(relationsMap, Type.DREAMS);
+
+        // stats are currently only available for modified cosine
+        // addNetworkStatisticsToRows(featureList, rowMaps);
+    }
+
+    public R2RMap<R2RSimpleSimilarity> convertMatrixToR2RMap(List<FeatureListRow> featureListRow,
+                                                             float[][] similarityMatrix) {
```
robinschmid, line 193: direct copy of MS2DeepscoreNetworkingTask.convertMatrixToR2RMap 
We can make the original static by pushing the Type.DREAMS and minScore into the method arguments. With static it is visible here.

utils/src/main/java/io/github/mzmine/util/collections/CollectionUtils.java
```diff
@@ -315,4 +315,17 @@ public static <T> List<T> combine(List<? extends T>... lists) {
     }
     return value;
   }
+
+  public static int[] argsort(float[] array) {
+    Integer[] indices = new Integer[array.length];
```
robinschmid, line 320: this should be int[]. Integer only if value is @Nullable

utils/src/main/java/io/github/mzmine/util/collections/CollectionUtils.java
```diff
@@ -315,4 +315,17 @@ public static <T> List<T> combine(List<? extends T>... lists) {
     }
     return value;
   }
+
+  public static int[] argsort(float[] array) {
```
robinschmid, line 319: name is more accurate as argsortReversed as it sorts descending order

utils/src/main/java/io/github/mzmine/util/collections/CollectionUtils.java
```diff
@@ -315,4 +315,17 @@ public static <T> List<T> combine(List<? extends T>... lists) {
...
+    Integer[] indices = new Integer[array.length];
+    for (int i = 0; i < array.length; i++) {
+      indices[i] = i; // Initialize with 0, 1, 2, ...
+    }
+
+    // Sort indices based on the values in the original array
+    Arrays.sort(indices, (a, b) -> Float.compare(array[b], array[a]));
+
+    // Convert Integer[] to int[]
+    return Arrays.stream(indices).mapToInt(Integer::intValue).toArray();
```
robinschmid, line 329: can directly return indices if it was initialized as int[]

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSSpectrumTensorizer.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    // If there are fewer peaks than nHighestPeaks, the remaining rows are already zero-filled
+    return result;
+  }
+
+  /**
+   * Only works on scans or {@link SpectralLibraryEntry} with precursor mz.
+   *
+   * @return A float[][][] where each spectrum is represented as (nHighestPeaks + 1, 2).
```
robinschmid, line 108: fully describe return type like what are the different arrays

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSSpectrumTensorizer.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  /**
+   * Tensorizes a mass spectrum. TODO: docstring
+   */
+
+  public float[][] tensorizeFragments(@NotNull MassSpectrum spectrum, @NotNull Double precMz) {
+
+    int nHighestPeaks = settings.nHighestPeaks();
+    double precIntensity = settings.precIntensity();
```
robinschmid, line 62: add comment to describe that precursor intensity is constant. 
Why is it actually constant? The spectrum may also contain the precursor mz, does this interfer with the predictions?

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSSpectrumTensorizer.java
```diff
@@ -0,0 +1,147 @@
...
+    int totalPeaks = spectrum.getNumberOfDataPoints();
+
+    for (int i = 0; i < totalPeaks; i++) {
+      double mz = spectrum.getMzValue(i);
+      double intensity = spectrum.getIntensityValue(i);
+      peaks.add(new double[]{mz, intensity});
+    }
+
+    // Sort peaks by intensity in descending order
+    peaks.sort((a, b) -> Double.compare(b[1], a[1])); // Sort by intensity (index 1)
```
robinschmid, line 82: could also do 
```java
    DataPoint[] dps = ScanUtils.extractDataPoints(spectrum);
    Arrays.sort(dps, DataPointSorter.DEFAULT_INTENSITY);
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/dreams/DreaMSNetworkingTask.java
roman-bushuiev: Resolved: [48893f9](https://github.com/mzmine/mzmine/pull/2317/commits/48893f92b271adabdb3f8784692d40d2f1eb53d0).

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/dreams/DreaMSNetworkingTask.java
roman-bushuiev: Resolved: [0f073ac](https://github.com/mzmine/mzmine/pull/2317/commits/0f073acc15c40450cd31472f1c0f37e94d25abbb).

utils/src/main/java/io/github/mzmine/util/collections/CollectionUtils.java
roman-bushuiev: Resolved: [315f318](https://github.com/mzmine/mzmine/pull/2317/commits/315f318b09faa8fef5873df5f79b919c6df1da82).

utils/src/main/java/io/github/mzmine/util/collections/CollectionUtils.java
roman-bushuiev: Is it ok to keep `Integer[]`? With `int[]`, `Float.compare` below does not work.

robinschmid:
For network style you can below to this fil: graph_network_style.css

```java
edge.DreaMS {
  size-mode: dyn-size;
  fill-mode: dyn-plain;
  fill-color: #ffbf8b, #ff8800, #915c20, #49310d;
}
```

Orange edges are DreaMS (without kNN)
image

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSSpectrumTensorizer.java
roman-bushuiev: Resolved: [873feb7](https://github.com/mzmine/mzmine/pull/2317/commits/873feb762fdb28ebd50d255077b534c1055c3fa6).

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSSpectrumTensorizer.java
roman-bushuiev: Resolved: [6eeb853](https://github.com/mzmine/mzmine/pull/2317/commits/6eeb853157079efaa9c4b1019a279d059776179b). Precursor m/z should be set to `ScanUtils.getPrecursorMz(scan)`. We just do not use MS1 precursor intensities and use the artificial intensity of 1.1 to distinguish this "artificial MS2 signal" from the rest of the signals.

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSSpectrumTensorizer.java
roman-bushuiev: I would prefer to keep the `peaks.sort` approach. It seems to be more readable.

utils/src/main/java/io/github/mzmine/util/collections/CollectionUtils.java
robinschmid: Oh ok makes sense the primitive does not allow a custom comparator... 
Then you could just skip all the initial array creation this way and work with the stream right away:
```java
    return IntStream.range(0, array.length).boxed()
        .sorted((a, b) -> Float.compare(array[b], array[a])).mapToInt(Integer::intValue).toArray();
```

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSSpectrumTensorizer.java
robinschmid: You can keep it this way. I just have to disagree on the readability ;) 

Double.compare(b[1], a[1]) uses array index 1 for intensities (which one needs to know at this point) instead of a clear method name. Also compare b,a kind of hides that it is a reversed sorting. 

I know the DEFAULT_INTENSITY could be renamed to INTENSITY_DESCENDING but you could even use your own comparator to make it more clear. To me this is much more clear than sorting a 2D array by its 2nd element. 

```java
    var dps
        = ScanUtils.extractDataPoints(spectrum);
    Arrays.sort(dps, Comparator.comparingDouble(DataPoint::getIntensity).reversed());
```

robinschmid:
This looks great already. 
Maybe we could meet and discuss if there is anything else to do.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/dreams/DreaMSNetworkingParameters.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Define any parameters here (see io.github.mzmine.parameters for parameter types) static is needed
+ * here to use this parameter as a key to lookup values
+ * <p>
+ * var flists = parameters.getValue(EmptyFeatureListParameters.featureLists);
+ */
+public class DreaMSNetworkingParameters extends SimpleParameterSet {
+
+    public static final PercentParameter minScore = new PercentParameter("Min similarity",
+            "The minimum similarity score to store the DreaMS prediction", 0.75, 0.0, 1.0);
```
robinschmid, line 57: Did you check what works in most cases?
Is 0.75 a good threshold?

I know it is super hard to pin point the similarity thrshold and it may be dependent on the analyzed compound classes or maybe instruments used...

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/dreams/DreaMSNetworkingTask.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private void processFeatureList(FeatureList featureList, DreaMSModel model) {
+        description = "Calculating DreaMS similarities";
+        List<Scan> scanList = new ArrayList<>();
+        List<FeatureListRow> featureListRows = new ArrayList<>();
+
+        for (FeatureListRow row : featureList.getRows()) {
+            Scan scan = row.getMostIntenseFragmentScan();
+            if (scan == null) {
```
robinschmid, line 132: maybe best to add this check here as well. Only scans with precursor mz are valid
if (scan == null || scan.getPrecursorMz()==null)

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSSpectrumTensorizer.java
```diff
@@ -0,0 +1,164 @@
...
+    // Initialize the result array with nHighestPeaks + 1 rows and 2 columns
+    float[][] result = new float[nHighestPeaks + 1][2];
+
+    // Set the first row: [precMz, precIntensity]
+    result[0][0] = precMz.floatValue();
+    result[0][1] = (float) precIntensity;
+
+    // Collect all peaks into a list of pairs (m/z, intensity)
+    List<double[]> peaks = new ArrayList<>();
+    int totalPeaks = spectrum.getNumberOfDataPoints();
```
robinschmid, line 88: add check if spectrum is empty:
```java
 if(totalPeaks == 0) {
    return result;
}
```

utils/src/main/java/io/github/mzmine/util/collections/CollectionUtils.java
roman-bushuiev: True. Your one-liner is looking much more elegant. I have changed it: [a917220](https://github.com/mzmine/mzmine/pull/2317/commits/a9172202a2021793fb5918985212f444380df24e).

mzmine-community/src/main/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSSpectrumTensorizer.java
roman-bushuiev: 👍

roman-bushuiev:
> For network style you can below to this fil: graph_network_style.css
> 
> ```java
> edge.DreaMS {
>   size-mode: dyn-size;
>   fill-mode: dyn-plain;
>   fill-color: #ffbf8b, #ff8800, #915c20, #49310d;
> }
> ```

I set a fancy color ([f95db0e](https://github.com/mzmine/mzmine/pull/2317/commits/f95db0e5f86c300044167331524f2d0ce1d8053c)), is it ok? What are the other three colors in the list? Should I change them accordingly?

mzmine-community/src/test/java/io/github/mzmine/util/scans/similarity/impl/DreaMS/DreaMSModelTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class DreaMSModelTest {
+
+    private static DreaMSModel model;
+    private static List<SimpleScan> testSpectra;
+    private static DreaMSSettings settings;
+    private static final String MODEL_URL = "https://huggingface.co/roman-bushuiev/DreaMS/resolve/main/DreaMS_embedding_model_torchscript.pt";
+    private static final String SETTINGS_URL = "https://huggingface.co/roman-bushuiev/DreaMS/resolve/main/DreaMS_embedding_model_torchscript_settings.json";
```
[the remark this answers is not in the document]
roman-bushuiev: I don't see any straightforward way to create a minimal model. I think the best solution is to make it disabled. It is quite separated from the rest of the code anyway.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/dreams/DreaMSNetworkingParameters.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public static final IntegerParameter numNeighbors = new IntegerParameter("Num. neighbors",
+            "The number of nearest neighbors to display for each node in the network. If left empty, all edges will be displayed.",
+            3, false);
```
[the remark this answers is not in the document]
roman-bushuiev: Resolved: [a42ee5a](https://github.com/mzmine/mzmine/pull/2317/commits/a42ee5a4d2681e1a7829869c992a42dfdefe088d).

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/dreams/DreaMSNetworkingParameters.java
```diff
@@ -0,0 +1,137 @@
...
+    public static final FileNameWithDownloadParameter dreaMSModelFile = new FileNameWithDownloadParameter(
+            "DreaMS model",
+            "The file location of the DreaMS model, click download to download the model.",
+            List.of(ExtensionFilters.PT), AssetGroup.DREAMS);
+
+    public static final PercentParameter minScore = new PercentParameter("Min similarity",
+            "The minimum similarity score to store the DreaMS prediction", 0.75, 0.0, 1.0);
+
+    public static final OptionalModuleParameter<DreaMSNetworkingKNNParameter> kNN = new OptionalModuleParameter<>(
+            "k-NN parameters", "Additional parameters constraining the network to have a k-nearest neighbors " +
```
robinschmid, line 66: change name to k-Nearest neighbors

robinschmid:
> > For network style you can below to this fil: graph_network_style.css
> > ```java
> > edge.DreaMS {
> >   size-mode: dyn-size;
> >   fill-mode: dyn-plain;
> >   fill-color: #ffbf8b, #ff8800, #915c20, #49310d;
> > }
> > ```
> 
> I set a fancy color ([f95db0e](https://github.com/mzmine/mzmine/pull/2317/commits/f95db0e5f86c300044167331524f2d0ce1d8053c)), is it ok? What are the other three colors in the list? Should I change them accordingly?

The other colors are meant to scale the coloring in steps you can add any number of color steps. Currently it is like below. The colors should scale from light to dark. Later we can find a way to set stronger color if the coloring is not selected. 
image

robinschmid:
Some nice examples. Here the precursor mz filter in modified cosine only allows for single modifications but the modifications are split on two parts of the structure, this is why some nodes have no modified cosine score. But DreaMS worked
image

robinschmid:
Here it is flipped:
DreaMS has no similarity between the small and the larger structure top two rows in the table but modified cosine is perfect and all the signals align:
image

robinschmid:
Those are from NIH NP POS from MSnLib

roman-bushuiev:
Yeah, unfortunately the "flipped examples" are also possible but according to our estimates it should much less of them compared to the "nice examples". Anyway, people can just use DreaMS as an orthogonal source of information since it is much more conceptually easy to find spectra of similar compounds by simply matching peaks.

robinschmid:
This is expected that both tools have their strengths and weaknesses.

robinschmid:
This PR was merged. Somehow, it shows as open. Will close it now.

PR #2328, opened 2025-01-14
Mass calibration update

SteffenHeu:
Use the auto separator parsing, allow mz in column 4 of csv lists.


robinschmid (approved):
Looks good

PR #2334, opened 2025-01-16
OptionalModuleParameter better layout

robinschmid:


SteffenHeu:
Looks the same in most cases to me. We can merge it, but im interested: where did problems occur?

PR #2335, opened 2025-01-16
Integration dashboard

SteffenHeu:
grafik


ansgarkorf:
•	Add visualization menu entry
•	Deactivate for imaging data
•	When selecting edit button, also show left integration side button as active
•	When selecting too many columns or rows the view breaks and you cannot go back. Define a limit of minimum space per plot and dont allow more rows/columns if there is not enough space OR make sure the controls above the plots never disappear. 
image

robinschmid:
Maybe set the minHeight of the top component to Region.USE_PREF_SIZE this worked in the gridpane of the parameter pane.

robinschmid:
Maybe we can auto zoom into the height / ranges of a feature?
Here the top one I zoomed manually - the lower one has its original zoom but also requires manual zoom:

image

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/FeatureDataEntryTask.java
```diff
@@ -0,0 +1,149 @@
...
+import io.github.mzmine.util.RangeUtils;
+import io.github.mzmine.util.color.SimpleColorPalette;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+class FeatureDataEntryTask extends FxUpdateTask<IntegrationDashboardModel> {
```
robinschmid, line 36: maybe add some jdoc to describe what this class does. Not clear from name and the function is named process

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardViewBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+            BinningMobilogramDataAccess.getPreviousBinningWith(model.getFeatureList(),
+                ims.getMobilityType())));
+      }
+//        plot.setChartGroup(chartGroup);
+      // auto update on change of the feature data entry. must be subscribed in here because we don't want to subscribe multiple times
+      model.featureDataEntriesProperty()
+          .addListener((MapChangeListener<RawDataFile, FeatureDataEntry>) change -> {
+            if (change.getKey() == file) {
+              plot.setTitle(file.getName());
+              plot.setFeatureDataEntry(change.getValueAdded());
```
robinschmid, line 174: maybe we can either auto zoom both axes of plot (if the user zoomed and then selects a different row the zoom stays fixed and appears empty sometimes). 
Or we could force zoom onto the actual feature like the height + RT range +-

mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/IonTimeSeriesUtils.java
```diff
@@ -238,6 +244,60 @@ public static IonTimeSeries<Scan> extractIonTimeSeries(@NotNull ScanDataAccess a
         scans);
   }
 
+  public static IonMobilogramTimeSeries extractIonMobilogramTimeSeries(
```
robinschmid, line 247: add some javadoc like for the other methods in this class

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/chartthemes/EStandardChartTheme.java
```diff
@@ -372,6 +372,7 @@ public void applyToTitles(@NotNull JFreeChart chart) {
     TextTitle title = chart.getTitle();
     if (title != null) {
       title.setVisible(isShowTitle());
+      super.applyToTitle(title);
```
robinschmid, line 375: without super. 
otherwise if we override this method in the future we will call super but not this.

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/series/IntensityTimeSeriesToXYProvider.java
```diff
@@ -40,20 +41,43 @@
...
-  public IntensityTimeSeriesToXYProvider(IntensityTimeSeries series, Color colorAwt) {
+  public IntensityTimeSeriesToXYProvider(IntensityTimeSeries series, @NotNull Color colorAwt) {
+    this(series, colorAwt, null);
+  }
+
+  public IntensityTimeSeriesToXYProvider(@NotNull MrmTransition transition,
+      @NotNull Color colorAwt) {
+    final NumberFormats formats = ConfigService.getGuiFormats();
+    this(transition.chromatogram(), colorAwt,
+        "%s -> %s".formatted(formats.mz(transition.q1mass()), formats.mz(transition.q3mass())));
```
robinschmid, line 62: use arrow symbol instead of -> →

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_smoothing/loess/LoessSmoothingParameters.java
```diff
@@ -25,22 +25,22 @@
...
+import io.github.mzmine.modules.dataprocessing.featdet_smoothing.savitzkygolay.SavitzkyGolayParameters;
 import io.github.mzmine.parameters.impl.SimpleParameterSet;
 import io.github.mzmine.parameters.parametertypes.IntegerParameter;
 import io.github.mzmine.parameters.parametertypes.OptionalParameter;
 
 public class LoessSmoothingParameters extends SimpleParameterSet {
 
   public static final OptionalParameter<IntegerParameter> rtSmoothing = new OptionalParameter<>(
-      new IntegerParameter("Retention time width (scans)",
+      new IntegerParameter(SavitzkyGolayParameters.rtSmoothingName,
```
robinschmid, line 36: the parameter names were changed for both rtSmoothingName and mobilitySmoothingName. 
either change back to original name or add to name map.

Original names were:
Retention time width (scans)
Mobility width (scans)

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/FeatureDataEntry.java
```diff
@@ -0,0 +1,21 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * @param file           The raw data file
+ * @param feature        the integrated feature as shown in the feature table. Null if the feature
+ *                       was not detected.
+ * @param chromatogram   the chromatogram within 2x the rt range of the row
+ * @param additionalData additional chromatograms, e.g. from mrm traces. May be empty
+ */
+public record FeatureDataEntry(@NotNull RawDataFile file, @Nullable IntensityTimeSeries feature,
```
robinschmid, line 17: maybe rename to FeatureIntegrationData to make it clear that this class is mostly / only used in this context.
I checked if this could be package private but it is used in IntegrationPlotController so not that easy.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/FeatureDataEntryTask.java
```diff
@@ -0,0 +1,149 @@
...
+
+      final List<IntensityTimeSeriesToXYProvider> additionalData;
+      if (feature != null && feature.get(MrmTransitionListType.class) != null) {
+        final MrmTransitionList mrms = feature.get(MrmTransitionListType.class);
+        AtomicInteger clr = new AtomicInteger(0);
+        additionalData = mrms.transitions().stream().map(t -> new IntensityTimeSeriesToXYProvider(
+            t.chromatogram()
+                .subSeries(null, extendedRtRange.lowerEndpoint(), extendedRtRange.upperEndpoint()),
+            colors.getAWT(clr.getAndIncrement()),
+            "%s -> %s".formatted(formats.mz(t.q1mass()), formats.mz(t.q3mass())))).toList();
```
robinschmid, line 89: change arrow -> to →

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/FeatureDataEntryTask.java
```diff
@@ -0,0 +1,149 @@
...
+      processed++;
+    }
+  }
+
+  private IonTimeSeries<? extends Scan> extractChromatogram(RawDataFile file,
+      @Nullable ModularFeature feature, @NotNull ModularFeatureList flist, Range<Double> mzRange,
+      Range<Float> extendedRtRange, FeatureListRow row) {
+    final IonTimeSeries<? extends Scan> chromatogram;
+
+    if (feature != null && feature.get(MrmTransitionListType.class) != null) {
```
robinschmid, line 105: how about adding a function for this just to enhace readability. Something like feature.isMRM or feature isReactionMonitoring or adding it to FeatureUtils.isMRM(feature)
I think this makes it much easier to directly see what is happening inside the if

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardController.java
```diff
@@ -0,0 +1,135 @@
...
+    for (ParameterSet ps : smoothingParams) {
+      // find the smoothing step that smoothed RT
+      final ModuleOptionsEnumComboParameter<FeatureSmoothingOptions> smoothingParam = ps.getParameter(
+          SmoothingParameters.smoothingAlgorithm);
+      OptionalParameter<IntegerParameter> rtSmoothing = smoothingParam.getEmbeddedParameters()
+          .getParameter(LoessSmoothingParameters.rtSmoothing);
+      if (!rtSmoothing.getValue()) {
+        continue;
+      }
+      return FeatureSmoothingOptions.createSmoother(ps);
```
robinschmid, line 76: Not related to code line.
If I select and deselect *apply latest smoothing* check box I get more and datasets added to the same charts. The old datasets are not removed.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardController.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+    return null;
+  }
+
+  private static @Nullable BaselineCorrector extractBaselineCorrector(ModularFeatureList flist) {
```
robinschmid, line 81: Unused for now. Just checking if this is for future maybe tag it with TODO?

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardController.java
```diff
@@ -0,0 +1,135 @@
...
+//    final BaselineCorrector corrector = extractBaselineCorrector(flist);
+
+    return its -> {
+      if (!model.isApplyPostProcessing()) {
+        return its;
+      }
+
+      final T smoothedRt;
+      if (smoother != null) {
+        final @Nullable double[] rtSmoothedIntensities = smoother.smoothRt(its);
```
robinschmid, line 120: seems dumb but has to be double @Nullable [] to avoid warnings that primitive cannot be null.

For objects this actually works similar to lists:
@Nullable Object @Nullable []

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardModel.java
```diff
@@ -0,0 +1,208 @@
...
+      FXCollections.observableArrayList());
+  private final ObjectProperty<@NotNull MetadataColumn<?>> rawFileSortingColumn = new SimpleObjectProperty<>(
+      ProjectService.getMetadata().getSampleTypeColumn());
+  private final ObjectProperty<IntegrationTransfer> syncReIntegration = new SimpleObjectProperty<>(
+      IntegrationTransfer.NONE);
+  private final ObjectProperty<@NotNull MZTolerance> integrationTolerance = new SimpleObjectProperty<>(
+      new MZTolerance(0.005, 10));
+  private final MapProperty<RawDataFile, FeatureDataEntry> featureDataEntries = new SimpleMapProperty<>(
+      FXCollections.observableHashMap());
+  private final ObjectProperty<@NotNull Function> postProcessingMethod = new SimpleObjectProperty<>(
```
robinschmid, line 53: Maybe define the types for Function - this also avoids some casts where it is applied
`ObjectProperty<@NotNull Function<IonTimeSeries<? extends Scan>, IonTimeSeries<? extends Scan>>>`

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardParameters.java
```diff
@@ -0,0 +1,14 @@
+package io.github.mzmine.modules.visualization.dash_integration;
+
+import io.github.mzmine.parameters.impl.SimpleParameterSet;
+import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsParameter;
+
+public class IntegrationDashboardParameters extends SimpleParameterSet {
```
robinschmid, line 6: change to extends CurrentProjectNoDialogParameterSet
This will open the dashboard directly with the currently selected feature list
Similar to stats dashboard

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardViewBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+    BorderPane mainBorder = new BorderPane();
+    BorderPane ftableControlsPane = new BorderPane();
+    BorderPane gridWrapper = new BorderPane();
+    final SplitPane mainSplit = new SplitPane(gridWrapper, ftableControlsPane);
+    mainBorder.setCenter(mainSplit);
+    mainSplit.setOrientation(Orientation.HORIZONTAL);
+    mainSplit.setDividerPositions(0.7);
+
+    model.featureTableFxProperty().get().getSelectionModel().selectedItemProperty()
+        .addListener((_, _, row) -> model.setRow(row == null ? null : row.getValue()));
```
robinschmid, line 73: No need to change just another option to unwrap the null within the property. Might be useful in other cases:
```java
.....selectedItemProperty().map(TreeItem::getValue)
```

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardViewBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+    final Label lblCols = FxLabels.newLabel("Columns");
+    final Label lblRows = FxLabels.newLabel("Rows");
+
+    final Spinner<Integer> spCols = new Spinner<>(1, 10, model.getGridNumColumns());
+    final Spinner<Integer> spRows = new Spinner<>(1, 10, model.getGridNumRows());
+
+//    spCols.getValueFactory().valueProperty()
+//        .bindBidirectional(model.gridNumColumnsProperty().asObject());
+    // binding in this fashion did not work, somehow it stopped updating after a few changes.
+    // maybe the asObject() was gc'ed at some point.
```
robinschmid, line 248: So convenient the mapping but also so easy to forget this. Also had to look it up again and you are right. They use WeakReference in TypedGenericBidirectionalBinding.

I guess the best option is to either just change the IntegerProperty to ObjectProperty<Integer> (I think most javafx components do not use the primitive versions right?)

Or to just create a FxSpinners.newSpinner that accepts the IntegerProperty and uses the listeners like below.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardViewBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+        .addListener((_, _, i) -> model.setGridNumRows(i != null ? i : model.getGridNumRows()));
+    model.gridNumRowsProperty()
+        .addListener((_, _, i) -> spRows.getValueFactory().setValue(i.intValue()));
+
+    spCols.getValueFactory().valueProperty().addListener((_, _, c) -> {
+      logger.finest(c.toString());
+    });
+    spRows.getValueFactory().valueProperty().addListener((_, _, r) -> {
+      logger.finest(r.toString());
+    });
```
robinschmid, line 264: can be removed

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardViewBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+
+    final Button previousPage = FxButtons.createButton(null, FxIcons.ARROW_LEFT, "Previous page",
+        () -> {
+          final int currentOffset = model.getGridPaneFileOffset();
+          model.setGridPaneFileOffset(
+              Math.max(0, currentOffset - (model.getGridNumRows() * model.getGridNumColumns())));
+        });
+    final Button nextPage = FxButtons.createButton(null, FxIcons.ARROW_RIGHT, "Next page", () -> {
```
robinschmid, line 272: the next page button allows to jump to a page that does not exist - only displaying the last entry eventhough it was on the previous page.

Maybe add a page counter to the model and always just evaluate this page counter once nextPage / previousPage / change of rows+cols

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardViewBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+      final int currentOffset = model.getGridPaneFileOffset();
+      final int step = model.getGridNumRows() * model.getGridNumColumns();
```
robinschmid, line 274: this is used in many functions / listeners here. 
Maybe add a model.numCells or similar.
gridNumColumns.multiply(gridNumRows) sadly provides a NumberBinding isntead of IntegerProperty but then you can at least bind it to a new property. 

This might help clear up some of the calculation of page size and number. 
It tells me there are two pages when I have 9 samples and 9 cells.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardViewBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  private Region buildMetadataColSelectionForSorting() {
+    final List<MetadataColumn<?>> columns = ProjectService.getMetadata().getColumns().stream()
+        .sorted(Comparator.comparing(MetadataColumn::getTitle)).toList();
+
+    final Label lblSortBy = FxLabels.newLabel("Sort files by:");
+    ComboBox<MetadataColumn<?>> cmbMetadataCol = new ComboBox<>(
+        FXCollections.observableArrayList(columns));
+    cmbMetadataCol.setConverter(new StringConverter<>() {
```
robinschmid, line 308: maybe easier to override the MetadataColumn.toString and return the title there. Definitely makes sense to me

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationTransfer.java
```diff
@@ -0,0 +1,75 @@
...
+    };
+  }
+
+  public String getToolTip() {
+    return switch (this) {
+      case ALL ->
+          "%s: Transfers a single re-integration to to all other files, overriding current integrations.".formatted(
+              this.toString());
+      case ONLY_MISSING ->
+          "%s: Transfers a reintegration to all files in which no feature was detected.".formatted(
```
robinschmid, line 51: harmonize re-integration in the other options reintegration here.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/RegionController.java
```diff
@@ -0,0 +1,9 @@
+package io.github.mzmine.modules.visualization.dash_integration;
+
+import io.github.mzmine.modules.visualization.otherdetectors.integrationplot.IntegrationPlotController;
+import io.github.mzmine.modules.visualization.otherdetectors.integrationplot.IntegrationPlotViewBuilder;
+import javafx.scene.layout.Region;
+
+public record RegionController(IntegrationPlotController controller, Region region) {
```
robinschmid, line 7: Can stay like this but there is also the FxCachedViewController that may serve a similar purpose. Builds the view once and then on next buildView calls it returns the cached view.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/otherdetectors/chromatogramplot/ChromatogramPlotController.java
```diff
@@ -203,4 +210,8 @@ public void setChartGroup(ChartGroup chartGroup) {
   public void setRangeAxisStickyZero(boolean stickyZero) {
     model.getChart().setStickyZeroRangeAxis(stickyZero);
   }
+
+  public void applyWithNotifyChanges(@NotNull final Runnable r) {
```
robinschmid, line 214: the method name implies that it notifies changes. 
Maybe add the boolean that is also present in EChartViewer.applyWithNotifyChanges
Or rename to applyWithoutNotifyChanges?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_smoothing/loess/LoessSmoothingParameters.java
SteffenHeu: changed names in savitzgy golay instead, made more sense

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardController.java
SteffenHeu: need to look that up again, that's odd

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/RegionController.java
SteffenHeu: ok, ill leave it like it is, because otherwise the functionality of this class would depend on the integrationplotviewbuilder staying of that class. and we refactor often :D

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardController.java
SteffenHeu: fixed

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/FeatureDataEntryTask.java
```diff
@@ -0,0 +1,149 @@
...
+          model.getFeatureList().getMemoryMapStorage());
+      chromatogram = (IonTimeSeries<? extends Scan>) model.getPostProcessingMethod().apply(chrom);
+    }
+    return chromatogram;
+  }
+
+  @Override
+  protected void updateGuiModel() {
+    model.featureDataEntriesProperty()
+        .putAll(entries.stream().collect(Collectors.toMap(FeatureDataEntry::file, e -> e)));
```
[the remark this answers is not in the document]
SteffenHeu: no, you would have to put an observable map.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/IntegrationDashboardViewBuilder.java
SteffenHeu: zoming if not connected

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/RegionController.java
robinschmid: Do not understand what you are referring to.
Currently, you create a new RegionController( IntegrationPlotController, Region ) to link the controller to its region. 

The only difference would be to have IntegrationPlotController extend FxCachedViewController instead of FxViewController and then you dont need the RegionController record. So no need to change any view builder. 

But this can also stay like this.

robinschmid:
ready from my side.

PR #2336, opened 2025-01-16
add ddamsmsinfo to extracted lcms pseudo spectra

SteffenHeu:
addition to #2321, so ms2 deepscore still works with lcms dia methods

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_diams2/DiaMs2CorrTask.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_diams2/DiaMs2CorrTask.java
@@ -364,7 +364,7 @@ public class DiaMs2CorrTask extends AbstractTask {
       List<IonTimeSeries<?>> eligibleEics, double[] ms1Rts, double[] ms1Intensities) {
     DoubleArrayList ms2Mzs = new DoubleArrayList();
     DoubleArrayList ms2Intensities = new DoubleArrayList();
-    DoubleArrayList ces = new DoubleArrayList();
+    DoubleArrayList collisionEnergies = new DoubleArrayList();
     MergedMassSpectrum mergedMobilityScan = null; // lazy initialization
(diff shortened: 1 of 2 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_diams2/DiaMs2CorrTask.java
```diff
@@ -416,15 +425,22 @@ public void run() {
 
       ms2Mzs.add(mz);
       ms2Intensities.add(maxIntensity);
+      subSeries.getSpectra().stream().map(Scan::getMsMsInfo).filter(Objects::nonNull)
+          .map(MsMsInfo::getActivationEnergy).filter(Objects::nonNull)
+          .mapToDouble(Float::doubleValue).average().ifPresent(ces::add);
```
robinschmid, line 430: this is not very descriptive. Had to check what ces was. 
Maybe parts of this could be reuseable methods? 

Streaming the non null MsMsInfo seems like a thing that is done more often. Even above for the activation method. 
`getSpectra().stream().map(Scan::getMsMsInfo).filter(Objects::nonNull)`
Could be turned into 
`ScanUtils.streamMsMsInfos(Collection<? extends Scan> scans)`

And maybe give ces a more descriptive name. Now it makes sense that I read it but initially not. energies or fragmentEnergies or something like this.

robinschmid: Otherwise for those long streams it might be better to force formatting at some points to underline behavior?

```java
subSeries.getSpectra().stream() //
      .map(Scan::getMsMsInfo).filter(Objects::nonNull) //
      .map(MsMsInfo::getActivationEnergy).filter(Objects::nonNull) //
      .mapToDouble(Float::doubleValue).average() //
      .ifPresent(ces::add);
```

robinschmid (dismissed):
Looks good and should work great. 
Only one optional comment.

robinschmid (approved):

PR #2342, opened 2025-01-20
Fix MS2 pairing in polarity switching files

### LINKED ISSUE #2329: gnps.mgf export is ignoring ionmode
### Basic information

* My operating system and version: Win 
* My mzmine version: 4.4.3

### What happened

When using mzwizard, I encountered an issue while processing files acquired with polarity switching. I set the ion mode to either positive or negative, and the feature list or exported annotations seemed to be filtered properly. For example, there are positive hits when the ion mode is set to positive, and so on.

However, in both cases, the exported gnps.mgf file contains MS2 spectra for both ion modes. What’s even more confusing is that the two exported files are not identical, despite being generated from the same batch (except ion mode setting). It appears there is some "bleeding" of MS2 spectra from the opposite mode into the exported MGF file.

I’m enclosing two MGF files, both generated from the same batch but with different ion mode settings, for your reference.
[mgf_files.zip](https://github.com/user-attachments/files/18411496/mgf_files.zip)

SteffenHeu:
fixes #2329


robinschmid (approved):
Just one potential NPE

robinschmid (approved):

PR #2343, opened 2025-01-20
External (Waters) IMS Calibration permits negative T0 values

AndrewDGillen:
We've noticed that t0 in calibration files from out Waters Cyclic instrument can be below 0. An example calibration file is attached: 

[mob_cal.csv](https://github.com/user-attachments/files/18478064/mob_cal.csv)

A slight change to the REGEX used in interpreting Waters CCS calibration files allows files with a negative t0 values to be used and interpreted by the CCS calibration module, with no other changes apparently required.


SteffenHeu:
Hi @AndrewDGillen,

thanks for the fix!

PR #2347, opened 2025-01-20
Minimum height in InputTypeComponent otherwise error in wizard

robinschmid:
@SteffenHeu you were right the minimum height was important here. I thought it would figure something out but it really shrunk into nothing in the wizard.

This is the minimum height now:
image


SteffenHeu:
I think i'd make it a bit larger still. maybe 200?

PR #2361, opened 2025-01-24
Automatically save a batch file when the wizard is running

SteffenHeu:
Added a module to save the current batch
automatically added in the wizard, if export is enabled


mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/autosave/AutoSaveBatchTask.java
```diff
@@ -0,0 +1,133 @@
...
+  @Override
+  protected void process() {
+    if (!TaskService.getController().isTaskInstanceRunningOrQueued(BatchTask.class)) {
+      // don't error since we don't want to stop processing
+      DialogLoggerUtil.showDialog(AlertType.WARNING, "Cannot save batch",
+          "Cannot save current batch since no batch is running. Please consider saving manually.");
+      return;
+    }
+
+    final BatchTask batchTask = TaskService.getController().getReadOnlyTasks().stream()
```
robinschmid, line 82: stream on this instance might not be safe? Concurrent modification exception?
Better to use getController.getTasksSnapshot() and stream over this

robinschmid: or add a method similar to isTaskInstanceRunningOrQueued(BatchTask.class)) to get the first or last instance of this class

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/autosave/AutoSaveBatchTask.java
```diff
@@ -0,0 +1,133 @@
...
+        savePath = new File(commonFromExport, "batch.%s".formatted(mzbatch));
+      } else if (commonImport != null) {
+        savePath = new File(commonImport, "batch.%s".formatted(mzbatch));
+      }
+    }
+
+    if (savePath == null || savePath.toString().isBlank()) {
+      DialogLoggerUtil.showDialog(AlertType.WARNING, "Cannot save batch",
+          "The %s module cannot automatically determine a path to export the batch file to. Please save the batch manually.",
+          false);
```
robinschmid, line 108: maybe better to just error() on this task and let the batch handle the dialog?

mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterUtils.java
```diff
@@ -323,4 +333,44 @@ public static Stream<Parameter<?>> streamThisAndEmbeddedDeep(final Parameter<?>
     return paramStream;
   }
 
+  @Nullable
+  public static <S extends MZmineProcessingStep<?>, T extends Collection<S>> File extractCommonExportPath(
+      T batch) {
+    final List<File> allExportPaths = batch.stream().map(MZmineProcessingStep::getParameterSet)
+        .<File>mapMulti((paramSet, c) -> streamParametersDeep(paramSet).forEach(p -> {
+          if (Objects.requireNonNull(p) instanceof FileNameSuffixExportParameter fnp) {
+            if (fnp.getType() == FileSelectionType.SAVE && fnp.getValue() != null) {
```
robinschmid, line 342: Can be simplified. You can use the deep stream of a specific class. This already does the instanceof, cast, nonnull
```java
streamParametersDeep(paramSet, FileNameSuffixExportParameter.class).forEach(fnp -> {
```

Also no need to check for `fnp.getType() == FileSelectionType.SAVE ` as this is always given for this parameter.

mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterUtils.java
```diff
@@ -323,4 +333,44 @@ public static Stream<Parameter<?>> streamThisAndEmbeddedDeep(final Parameter<?>
...
+              if (fnp.getValue().getName()
+                  .contains(".")) { // if there is a . in the file name, interpret as file
```
robinschmid, line 344: this check is not correct.  . ist a valid character in path names. At the start for hidden folders I think and also in the middle just to write something like mz200.00 (stupid but ok). 

Also very often we will not enter the filename with format like gnps.csv because gnps export adds the formats automatically.

Maybe just always use the parent?
This parameter value should never be just a folder but always a prefix of a file name.

mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterUtils.java
```diff
@@ -323,4 +333,44 @@ public static Stream<Parameter<?>> streamThisAndEmbeddedDeep(final Parameter<?>
...
+              } else {
+                c.accept(fnp.getValue());
+              }
+            }
+          }
+        })).filter(Objects::nonNull).toList();
+
+    return allExportPaths.stream().collect(Collectors.groupingBy(p -> p, Collectors.counting()))
+        .entrySet().stream().sorted(Entry.comparingByValue(Comparator.reverseOrder()))
+        .map(Entry::getKey).findFirst().orElse(null);
```
robinschmid, line 355: super confusing but ok like this. 
Just found a solution to not do reverse:
Instead of sorted.findFirst you could do max()

Those things I sometimes remember and sometimes they are somewhere hidden :D

robinschmid: This counting parents is the same logic like the method below - maybe thats a good to have in the FileAndPathUtils class?
getMostCommonParentDirectory or something like that

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/autosave/AutoSaveBatchTask.java
```diff
@@ -0,0 +1,133 @@
...
+
+    final String mzbatch = ExtensionFilters.getExtensionName(ExtensionFilters.MZ_BATCH);
+
+    File savePath = parameters.getOptionalValue(AutoSaveBatchParameters.savePath).orElse(null);
+    if (savePath == null || savePath.toString().isBlank()) {
+//      Set.of(SiriusExportModule.class, GnpsFbmnExportAndSubmitModule.class,
+//          CSVExportModularModule.class, LegacyCSVExportModule.class,
+//          ProjectMetadataExportModule.class);
+      final File commonFromExport = ParameterUtils.extractCommonExportPath(queueCopy);
+      final File commonImport = ParameterUtils.extractCommonRawFileImportFilePath(queueCopy);
```
robinschmid, line 97: so this is now majority wins and not really an extractCommon.

robinschmid (approved):
Lookes great to me. 
Mostly some small comments. Can be merged after that

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/autosave/AutoSaveBatchTask.java
SteffenHeu: The idea is to not let the batch fail

robinschmid:
Just talked with ansgar maybe we can name the suffix autosave_batch.mzbatch

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/autosave/AutoSaveBatchTask.java
robinschmid: Ah ok. Makes sense. 
Then maybe add a comment not that some smart ass will change that in the future

SteffenHeu: Are you talking about yourself? :D

robinschmid (approved):
Great

PR #2363, opened 2025-01-24
message in Task.error is nullable because it is in exception.getErrorMessage

robinschmid:


taskcontroller/src/main/java/io/github/mzmine/taskcontrol/AbstractTask.java
```diff
@@ -107,7 +109,8 @@ public void cancel() {
   }
 
   @Override
-  public void error(@NotNull String message, @Nullable Exception exceptionToLog) {
+  public void error(@Nullable String message, @Nullable Exception exceptionToLog) {
+    message = requireNonNullElse(message, "");
```
SteffenHeu, line 113: This is not done in the dataset tasks

robinschmid: good catch. Added handling to all extending classes

PR #2364, opened 2025-01-24
Reuse rows and features in alignment if original featurelist will be removed

robinschmid:
If original feature list is set to be removed we can reuse the rows and features directly. 

Cohort study with 500 samples and 21.000 aligned features went from 20 s --> 10 s. Memory should also greatly improve.

2x speed up

- [x] test for gc
- [x] test for lc


robinschmid:
I used #2367 to compare the content of the feature lists before and after alignment and they equal for both reusing and copying features - this is in LC-MS. Will now also check GC-EI-MS

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/align_join/JoinRowAlignScorer.java
```diff
@@ -134,8 +134,13 @@ public void scoreRowAgainstBaseRows(final List<FeatureListRow> baseRowsByMz,
...
 
     // find all rows in the aligned rows that might match
-    final List<FeatureListRow> candidatesInAligned = FeatureListUtils.getCandidatesWithinRanges(
-        mzRange, rtRange, mobilityRange, baseRowsByMz, true);
+    List<FeatureListRow> candidatesInAligned = null;
+    try {
+      candidatesInAligned = FeatureListUtils.getCandidatesWithinRanges(mzRange, rtRange,
+          mobilityRange, baseRowsByMz, true);
+    } catch (Exception exception) {
+      System.out.println(exception);
```
SteffenHeu, line 142: fix

PR #2371, opened 2025-01-27
add batch check for raw files and flist parameter setting

SteffenHeu:
In case the feature list or raw file selection is all over the place for batches. If the first steps (e.g. if files are already imported) are something different than "last batch" but all consecutive are, no warning is shown, as that is a common case. Only if it happens in the middle of the batch

grafik


robinschmid (approved):
Seems to work great

PR #2376, opened 2025-01-28
Featurelist renaming

robinschmid:


robinschmid:
Merged with @ansgarkorf  and @SteffenHeu

PR #2389, opened 2025-01-30
Track step memory only if GC is performed

robinschmid:
Memory usage only makes sense if GC was triggered

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchTask.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchTask.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2024 The mzmine Development Team
+ * Copyright (c) 2004-2025 The mzmine Development Team
  *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
@@ -25,7 +25,6 @@
 
 package io.github.mzmine.modules.batchmode;
 
-import static io.github.mzmine.gui.preferences.MZminePreferences.runGCafterBatchStep;
 import static io.github.mzmine.main.ConfigService.getPreference;
 import static java.util.Objects.requireNonNullElse;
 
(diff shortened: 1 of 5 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/timing/StepTimeMeasurement.java
```diff
@@ -27,18 +27,30 @@
...
-    this(stepNumber, duration.toMillis() / 1000.0, name,
-        ConfigService.getConfiguration().getUsedMemoryGB());
+  /**
+   * @param performGcAndMeasureMemory memory measurements are only precise when combined with GC
+   *                                  before, if active this will perform a GC
+   */
+  public StepTimeMeasurement(final int stepNumber, final String name, final Duration duration,
+      final boolean performGcAndMeasureMemory) {
+    if (performGcAndMeasureMemory) {
+      System.gc();
```
SteffenHeu, line 42: this is absolutely hidden away from the logic of the actual batch task^^. This belongs into the task in my optionen

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchTask.java
```diff
@@ -456,13 +467,6 @@ private void processQueueStep(int stepNumber) {
     if (!createdFeatureLists.isEmpty()) {
       previousCreatedFeatureLists = createdFeatureLists;
     }
-
-    Duration duration = Duration.between(start, Instant.now());
-
-    if (requireNonNullElse(getPreference(runGCafterBatchStep), false)) {
-      System.gc();
-    }
-    stepTimes.add(new StepTimeMeasurement(stepNumber + 1, method.getName(), duration));
```
SteffenHeu, line 465: I think the placement here was way better, because the necessary logic was still in the batch task

SteffenHeu (commented):
looks good except one thing

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchTask.java
robinschmid: I put the measurement outside of the function so that all local vars in that method may be cleared by GC.

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/timing/StepTimeMeasurement.java
robinschmid: I think the class was hard to use. At least the memory measurement only makes sense if GC was performed. 
Maybe we could just change from constructor to factory methods to make it more obvious:
StepTimeMeasurement.garbageCollectAndMeasure();

At least I would not want to track memory if GC was not performed as this is purely random.

PR #2404, opened 2025-02-07
Bugfix: added a check to stop empty features during GC alignment

wchu207:
See comments for a description of the bug. I would have had the cloner return null, but it's guaranteed to return non-null and I didn't want to mess with the interface, since that might require changes elsewhere.

I also reenabled row bindings here. It probably slows things down, but I've been getting incorrect exported results -- particularly, average MZ for the row.


ansgarkorf:
Hi, thanks for spotting this bug. I looked at your suggested fix, however this leads to a 0.0 mz row with a single feature with 0.0 intensity. We are investigating this deeper and will come up with a fix.

PR #2410, opened 2025-02-12
GNPS GC exports

robinschmid:
- use PseudoSpectrum in ADAP mgf and msp export
- use the LC-MS export for GNPS also for GC (allows merging of spectra)

Maybe we should remove the GNPS GC-MS export? And maybe also the ADAP msp and mgf export. They may be exchanged with the Export scans (feature) module - maybe build a wrapper module with simpler parameters for GC-EI-MS export


SteffenHeu:
What did the adap gc export write as PEPMASS and what does the gnps export write now?

robinschmid:
Default was AS_IN_FEATURETABLE and this is row.getAverageMz
There are also two other options for highest mz or highest intense signal. 
Overall the mz in GC-EI-MS is quite random so maybe not important to give those options. People cannot rely on an mz not to switch to another signal just because there might be another signal with similar intensity or higher mz.

PR #2411, opened 2025-02-12
Lower chromatographic threshold in GC-EI wizard

robinschmid:
GC-EI generates a lot of fragment peaks with the same mz across the whole RT range


SteffenHeu:
Usually a lower threshold leads to many more nosie features being introduced. how does that affect the feature detection performance? @ansgarkorf

robinschmid:
The problem is that this filter was never really designed for GC-EI. In LC-MS a specific m/z should only elute in a narrow RT range. In GC-EI, we get the same m/z fragments throughout the whole chromatogram. Second problem is potential baseline on some chromatograms - especially for low res GC-EI Q-MS. 

Currently, the wizard sets it to the max of 90% and this will remove many features in GC MS.
We could increase the absolute minimum that we set (currently 0.3) but not sure. I got 0.63 now with those settings.

PR #2412, opened 2025-02-12
Added support for library search with Kovats retention index for GC-MS

wchu207:
3 main parts to this:

1. RIType for features, calculated by RICalculationModule
2. RI added DBEntryField, uses RIRecord type to handle different GC column types
3. Additions to spectral library search, to optionally use the retention indices


robinschmid:
Thanks for your pull request. I will take a look at it and try to provide feedback. 

Are you already using this module / custom mzmine version in your work? 
Do you have some example files and maybe a batch file that you can share with us?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/RIDiffType.java
```diff
@@ -0,0 +1,69 @@
...
+  @NotNull
+  @Override
+  public String getUniqueID() {
+    // Never change the ID for compatibility during saving/loading of type
+    return "ri_diff";
+  }
+
+  @Override
+  public @NotNull String getHeaderString() {
+    return "RI range";
```
robinschmid, line 58: Is this supposed to be a range or a difference?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/RIScaleType.java
```diff
@@ -0,0 +1,27 @@
+package io.github.mzmine.datamodel.features.types.annotations;
+
+import io.github.mzmine.datamodel.features.types.abstr.StringType;
+import io.github.mzmine.datamodel.features.types.modifiers.AnnotationType;
+import javafx.util.StringConverter;
+import javafx.util.converter.DefaultStringConverter;
+import org.jetbrains.annotations.NotNull;
+
+public class RIScaleType extends StringType implements AnnotationType {
```
robinschmid, line 9: Are there a defined number of values?
It might be nice to define an Enum class and define all possible values for this type and then change it to DataType<RIScale>

wchu207: The scales represent a linear interpolation between retention time values for straight-chain alkanes. Octane by definition has RI = 800, Nonane has RI = 900, Decane has RI = 1000, and so on

The retention times for the straight-chain alkanes are given in separate files. This type is just stores the filename, so you can check and maybe reuse it downstream if necessary.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/RIType.java
```diff
@@ -0,0 +1,116 @@
...
+
+import static io.github.mzmine.datamodel.features.types.modifiers.BindingsType.*;
+import static io.github.mzmine.datamodel.features.types.modifiers.BindingsType.MIN;
+
+/**
+ * Retention index type
+ * This is frequently rounded, because everyone uses rounded versions
+ * Rounding is therefore necessary for the endpoints of ranges to be placed at integer values
+ */
+public class RIType extends IntegerType {
```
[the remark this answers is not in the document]
wchu207: All spectral libraries seem to round it too, so I set it up to do that for individual features' RI and then again when averaging. That way, tolerance ranges wouldn't fall between integer values, but come to think of it, it's probably better to just average it when it's being used.

wchu207:
> Thanks for your pull request. I will take a look at it and try to provide feedback.
> 
> Are you already using this module / custom mzmine version in your work? Do you have some example files and maybe a batch file that you can share with us?

I don't think I can provide you with the scans of the breath samples since that's medical information, but here's the batch file & reference files used to construct the retention index scales.

[RI batch file.zip](https://github.com/user-attachments/files/18787919/RI.batch.file.zip)

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/RIDiffType.java
robinschmid: Ok just asked because the unique ID said diff and this said range.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/AdvancedSpectralLibrarySearchParameters.java
```diff
@@ -46,6 +47,11 @@ public class AdvancedSpectralLibrarySearchParameters extends SimpleParameterSet
   public static final OptionalParameter<RTToleranceParameter> rtTolerance = new OptionalParameter<>(
       new RTToleranceParameter());
 
+  public static final OptionalParameter<RIToleranceParameter> riTolerance = new OptionalParameter<>(
+          new RIToleranceParameter("RI tolerance [±]", "Retention index tolerance for spectral library entries to be matched against a feature."));
```
robinschmid, line 51: Better remove the ± here to match all the other tolerances that imply +-.
If you want you can add this to the description though.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/AdvancedSpectralLibrarySearchParameters.java
```diff
@@ -46,6 +47,11 @@ public class AdvancedSpectralLibrarySearchParameters extends SimpleParameterSet
...
+
+  public static final BooleanParameter ignoreWithoutRI = new BooleanParameter("Ignore library entries without RIs", "");
```
robinschmid, line 53: add a description of what ignore means here. Like exclude entries without RI from search or ignore the RI filter then and still match?

mzmine-community/src/main/java/io/github/mzmine/util/RIColumn.java
```diff
@@ -0,0 +1,12 @@
+package io.github.mzmine.util;
+
+public enum RIColumn {
```
robinschmid, line 3: I have never really used RI matching. 
But I know there are so many different parameters available in databases like column, carrier gas, ramp type, column/phase type. How many of these parameters are important for a proper matching?

mzmine-community/src/main/java/io/github/mzmine/util/RIRecord.java
```diff
@@ -0,0 +1,173 @@
...
+// This holds information from the RI field in spectral libraries for gas chromatography
+//   The RI field is of form "s=RI/CI/n_samples n=RI/CI/n_samples p=RI/CI/n_samples" (CI denotes confidence interval)
+//   s denotes semipolar, n denotes nonpolar, and p denotes polar
+//   If there are 0 samples for s or n or p, then that part is skipped
+//   If there is one sample, then n_samples and CI are skipped
+//   Unclear if non-NIST other formats differ
+
+
+public class RIRecord
+{
```
robinschmid, line 45: please use the code style that we supply with mzmine source code:
https://mzmine.github.io/mzmine_documentation/contribute_intellij.html#code-formatter 

This makes it easier to review and avoid unnecessary changes of style by reformatting

mzmine-community/src/main/java/io/github/mzmine/util/RIRecord.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.EnumMap;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Logger;
+
+import static java.util.Map.entry;
+
+
+// This holds information from the RI field in spectral libraries for gas chromatography
+//   The RI field is of form "s=RI/CI/n_samples n=RI/CI/n_samples p=RI/CI/n_samples" (CI denotes confidence interval)
```
robinschmid, line 37: is this imported from the NIST RI library?
Which file is this specifically?

I know they provide much more information about each individual method so maybe we can load more info and turn this into something like an RILibrary that can help any other compound annotation module to also match RI

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+
+
+/**
+ * The task will be scheduled by the TaskController. Progress is calculated from the
+ * finishedItems/totalItems
+ */
+public class RICalculationTask extends AbstractFeatureListTask {
+
+  record RIScale(Date date, String fileName, PolynomialSplineFunction interpolator){
+    public RIScale(Date date, String fileName, PolynomialSplineFunction interpolator){
```
robinschmid, line 70: default record constructor can be removed. It is added automatically

robinschmid: Also this RIScale class is used outside of the Task  - so better to push it into its own file

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+    this.suffix = parameters.getValue(RICalculationParameters.suffix);
+  }
+
+  @Override
+  protected void process() {
+    linearScales = new ArrayList<>(Arrays.stream(parameters.getValue(RICalculationParameters.alkaneFiles))
+            .sequential()
+            .map(this::processScale)
+            .filter(Objects::nonNull)
+            .toList());
```
robinschmid, line 105: you can also remove the new ArrayList<> at the start and instead let the stream create an ArrayList. 
```java
    .collect(CollectionUtils.toArrayList())
// or
    .collect(Collectors.toCollection(ArrayList::new))
```

Also sequential is not needed - the stream is already sequential because the input array has a sequence order.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+
+    linearScales.sort(new Comparator<RIScale>() {
+        @Override
+        public int compare(RIScale o1, RIScale o2) {
+            return o1.date.compareTo(o2.date);
+        }
```
robinschmid, line 111: you can easily define comparators like this:
```java
linearScales.sort(Comparator.comparing(RIScale::date));
```
You could even put it into the stream part above after the filter as .sorted(Comparator.comparing(RIScale::date))

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+      if (shouldAddSummary) {
+        outputList.addFeatureType(new RIMaxType());
+        outputList.addFeatureType(new RIMinType());
+        outputList.addFeatureType(new RIDiffType());
+        outputList.addFeatureType(new RIScaleType());
+      }
+
+
+      for (RawDataFile file : outputList.getRawDataFiles()) {
+        Date sampleDate = extractDate(file.getAbsolutePath());
```
robinschmid, line 172: interesting idea to extract the date from the filename. Maybe we could either extract from filename or from the injection date that may or may not be present in the Sample metadata table under Project/Sample metadata:

```java
      final MetadataTable metadata = ProjectService.getMetadata();
      final DateMetadataColumn dateColumn = metadata.getRunDateColumn();

      for (RawDataFile file : outputList.getRawDataFiles()) {
        final LocalDateTime date = metadata.getValue(dateColumn, file);
        // date.toLocalDate() // date without time
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+        for (RIScale _riScale : linearScales) {
+          if (sampleDate.after(_riScale.date) || sampleDate.equals(_riScale.date)) {
+            final RIScale riScale = _riScale;
```
robinschmid, line 175: no need to create this final var. You can just use _riScale and rename the loop var to riScale. 
You can even make the loop var final.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+        outputList.addFeatureType(new RIMinType());
+        outputList.addFeatureType(new RIDiffType());
+        outputList.addFeatureType(new RIScaleType());
+      }
+
+
+      for (RawDataFile file : outputList.getRawDataFiles()) {
+        Date sampleDate = extractDate(file.getAbsolutePath());
+        for (RIScale _riScale : linearScales) {
+          if (sampleDate.after(_riScale.date) || sampleDate.equals(_riScale.date)) {
```
robinschmid, line 174: sampleDate may be null - then skip and maybe count the files or track the names of files that did not have a date. 
Either log those files at the end - or find a strategy to which scale to use if no matching dates were found.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+      ri = (float) riScale.interpolator.getPolynomials()[riScale.interpolator.getPolynomials().length - 1].value(rt - knots[knots.length - 1]);
+      hasRI = true;
+    }
+
+    else if (shouldExtrapolate && rt < knots[0]) {
+      ri = (float) riScale.interpolator.getPolynomials()[0].value(rt - knots[0]);
+      hasRI = true;
+    }
+
+    if (hasRI) {
```
robinschmid, line 212: could remove the whole hasRI boolean and instead just check ri!=null

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+      feature.set(RIType.class, ri);
+      feature.set(RIScaleType.class, FilenameUtils.getName(riScale.fileName));
+    }
+  }
+
+  private Date extractDate(String fileName) {
+    // Extract date from file name
+    SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd");
+    Matcher matcher = Pattern.compile(".*(\\d{4}([-/])\\d{2}([-/])\\d{2}).*").matcher(fileName);
```
robinschmid, line 221: I think many people also start filenames without -
so 20241231 as in yyyyMMdd.
We could add this format here as well.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+      try {
+        Date date = extractDate(file.getName());
+        String fileName = file.getAbsolutePath();
+
+        if (date == null || fileName == null) {
+          return null;
+        }
+
+        CSVReader reader = new CSVReader(new BufferedReader(new FileReader(file)));
+        List<String[]> data = reader.readAll();
```
robinschmid, line 134: buffered reader and csv reader are never closed... Usually good to put them into a try-with-resources.

But there is an even better option already available in mzmine. You can create a small POJO class like a Java record and we use Jackson to read the csv. You can annotate the fields with @JsonProperty to give them any name:
```java
  record ExperimentalRIEntry(@JsonProperty("Carbon #") int nCarbons, @JsonProperty("RT") float rt) {
  }
```

And then use this function to read the CSV into a list of these objects:
```java
    final List<ExperimentalRIEntry> entries = CsvReader.readToList(file,
        ExperimentalRIEntry.class, ',');
```

Here is a test case that we could add. For this maybe add a test file into the test/resources folder and then generate a test file corresponding to the read function:
```java
import static org.junit.jupiter.api.Assertions.*;

import io.github.mzmine.modules.dataprocessing.norm_ri.RICalculationTask.ExperimentalRIEntry;
import io.github.mzmine.util.io.CsvReader;
import java.io.File;
import java.io.IOException;
import java.util.List;
import org.junit.jupiter.api.Test;

class RICalculationTaskTest {

  @Test
  void readLadder() throws IOException {
    File file = new File("D:\\tmp\\RI.batch.file\\RI batch file\\Alkane ladders\\2023-08-30.csv");
    final List<ExperimentalRIEntry> entries = CsvReader.readToList(file,
        ExperimentalRIEntry.class, ',');

    assertNotNull(entries);
    assertEquals(12, entries.size());
    assertEquals(5, entries.getFirst().nCarbons());
    assertEquals(4.315, entries.getFirst().rt(), 0.0001);
  }
}
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+          return null;
+        }
+        List<Double> x = new ArrayList<Double>();
+        List<Double> y = new ArrayList<Double>();
+        for (String[] pair : data.subList(1, data.size())) {
+          y.add(Double.parseDouble(pair[0]) * 100);
+          x.add(Double.parseDouble(pair[1]));
+        }
+        PolynomialSplineFunction interpolator = new LinearInterpolator().interpolate(
+                x.stream().mapToDouble(Double::doubleValue).toArray(),
```
robinschmid, line 146: also good like this but just a hint we sometimes use fastutil collections internally for primitive lists. 
```java
DoubleArrayList x = new DoubleArrayList();
x.toDoubleArray()
// also allows x.doubleStream()
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  private Date extractDate(String fileName) {
+    // Extract date from file name
+    SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd");
+    Matcher matcher = Pattern.compile(".*(\\d{4}([-/])\\d{2}([-/])\\d{2}).*").matcher(fileName);
+    if (matcher.find()) {
+      try {
+        return dateFormat.parse(matcher.group(1));
```
robinschmid, line 224: The java Date API is outdated and was replaced by LocalDate, Instant, LocalDateTime, and DateTimeFormatter etc. 

Maybe better to adjust and use those classes. We also use them in the DateMetadataColumn.

Here is how you can convert a String to LocalDate - this also shows how to convert the different formats
```java
    assertEquals("2024-12-31", LocalDate.parse("2024-12-31").toString());
    assertEquals("2024-12-31", LocalDate.parse("20241231", DateTimeFormatter.BASIC_ISO_DATE).toString());
```

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/RITolerance.java
```diff
@@ -0,0 +1,79 @@
...
+ * or relative (percent) but as rest of MZmine codebase, it assumes that rt values (other than the
+ * tolerance given in constructor) are in minutes in methods such as getToleranceRange or
+ * checkWithinTolerance
+ */
+public class RITolerance {
+
+  private final int tolerance;
+  private final RIColumn column;
+
+  public RITolerance(final int rtTolerance, RIColumn type) {
```
robinschmid, line 43: make sure to rename the int rtTolerance to float riTolerance

robinschmid (changes requested):
This looks great so far. 
Dont be discouraged by so many comments. I just try to also point to some classes and design decisions we took in mzmine. 

Your current design is very much designed to a strict naming patter of raw data files and also for the RI ladder files and the RI values in NIST? libraries. 

From what library files are you loading the RI values currently? 
Can you share an example library file (can be made up). 
Can you share some raw data - does not have to be from those clinical samples but maybe the ladder or something else that is not restricted?

I think we can make this a bit more flexible by using a date column in the project metadata table. 

Later we could think about creating some sort of compound library classes in mzmine to store the RI values for specific structures. This way people could extract and add RI values for structures, load them from different libraries and use them across different annotation tools. 
Like library matching or CSV compound database search could then just ask the RI library in mzmine for all RI entries for a particular SMILES or inichi or inchikey. 

Many options to build on this. 

Do you have an email - I could invite you to our mzmine community slack and maybe we could meet to discuss more.

wchu207:
Hi, sorry for the delayed response. My personal email is willchu207@gmail.com. 

Reference libraries contain information regarding carrier gas, column type, and so on. But that's all kind of implied by the alkane ladder, which includes the reference times for the straight-chain alkanes for those parameters So we just select the parameters during the library search. We could add that information to RIType, but that would primarily be for creating your own libraries, I suspect.

I'll look through the requested changes by the end of the week. 

I've included some alkane ladders. I'll look for a library that I can share.
[Alkane scales.zip](https://github.com/user-attachments/files/18875743/Alkane.scales.zip)

wchu207:
Here's a couple in-house libraries. NIST has a very similar format when exported, it's just that retention indexes are something like SemiStdPolar={value} instead of s={value}.

[libraries.zip](https://github.com/user-attachments/files/18875788/libraries.zip)

mzmine-community/src/main/java/io/github/mzmine/util/RIRecord.java
wchu207: Oh yeah, this is custom-made. When I export MSP files from NIST, this is all I can get. In MSSearch this is under the field "Experimental RI median (plus-minus) deviation"

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/RIScaleType.java
robinschmid: Thanks for clarifying that this is a file

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,227 @@
...
+        outputList.addFeatureType(new RIMinType());
+        outputList.addFeatureType(new RIDiffType());
+        outputList.addFeatureType(new RIScaleType());
+      }
+
+
+      for (RawDataFile file : outputList.getRawDataFiles()) {
+        LocalDate sampleDate = metadataTable.getValue(metadataTable.getRunDateColumn(), file).toLocalDate();
+        for (RIScale riScale : linearScales) {
+          if (sampleDate != null && sampleDate.isAfter(riScale.date()) || sampleDate.isEqual(riScale.date())) {
```
robinschmid, line 160: there are missing () sampleDate.isEqual may throw null pointer exception. 
```java
if(sampleDate != null && (sampleDate.isAfter(riScale.date()) || sampleDate.isEqual(riScale.date())))
```
Or just handle null separately outside the loop. 
```java
if(sampleDate==null) {
    logger.fine("No date for sample "+file.getName());
    continue;
}
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
```diff
@@ -0,0 +1,227 @@
...
+    if (ri != null) {
+      feature.set(RIType.class, ri);
+      feature.set(RIScaleType.class, FilenameUtils.getName(riScale.fileName()));
+    }
+  }
+
+  private LocalDate extractDate(String fileName) {
+    // Extract date from file name
+    String[] dateComponents = {"\\d{4}", "\\d{2}", "\\d{2}"};
+    final String[] separators = {"-", "/", ""};
```
robinschmid, line 202: Did you try this function?
Maybe best to write a test because it did not work for me. Below is my test.
I will create a separate PR to mzmine and propose a LocalDateParser that handles all file path compatibile date formats. Your second format: yyyy/MM/dd format is not a valid path.

```java
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertNull;

import java.time.LocalDate;
import java.util.regex.Matcher;
import java.util.regex.Pattern;
import org.junit.jupiter.api.Test;

class RICalculationTaskTest {

  @Test
  void testCurrentParser() {
    assertEquals(LocalDate.of(2020, 5, 20), parse("2020-05-20"));
    assertEquals(LocalDate.of(2020, 5, 20), parse("20200520"));
    assertEquals(LocalDate.of(2020, 5, 20), parse("2020.05.20"));
    assertEquals(LocalDate.of(2020, 5, 20), parse("2020-05-20_somthing following"));
    assertEquals(LocalDate.of(2020, 5, 20), parse("Leading2020-05-20"));
    assertEquals(LocalDate.of(2020, 5, 20), parse("Leading2020-05-20_and trailing"));
    assertEquals(LocalDate.of(2020, 5, 20), parse("Leading20200520_and trailing"));
    assertEquals(LocalDate.of(2020, 5, 20), parse("Leading20200520"));
    assertEquals(LocalDate.of(2020, 5, 20), parse("20200520_trailing"));
    assertNull(parse("2020052034234235_some numbers"));
    assertNull(parse("2020-05-2034234235_some numbers"));
  }

  LocalDate parse(String fileName) {
    // Extract date from file name
    String[] dateComponents = {"\\d{4}", "\\d{2}", "\\d{2}"};
    final String[] separators = {"-", "", "\\\\."};

    for (String sep : separators) {
      Matcher matcher = Pattern.compile(String.join(sep, dateComponents)).matcher(fileName);
      if (matcher.find()) {
        try {
          return LocalDate.parse(matcher.group(0));
        } catch (Exception _) {
        }
      }
    }
    return null;
  }

}
```

This fails for the yyyyMMdd pattern at
assertEquals(LocalDate.of(2020, 5, 20), parse("20200520"));

robinschmid: I created a PR but it is not merged yet. It harmonizes date parsing. Feel free to try and comment there as well
https://github.com/mzmine/mzmine/pull/2430

robinschmid:
@wchu207 Thanks for working on this. 
Last week I was at a conference, so I only had time to look at the progress this week. 

Thanks for sharing library files. 
So you create your own library and then use a different format than NIST does when you export their entries?

Can we add the library files or a similar file to the test resources? Then we can write a test for parsing of this field and make sure it stays stable and supported.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_ri/RICalculationTask.java
wchu207: Hm, it did work when I tried itb but testing was brief because this branch only for this pull request. I'll see if I can fix it.

wchu207:
> @wchu207 Thanks for working on this. Last week I was at a conference, so I only had time to look at the progress this week.
> 
> Thanks for sharing library files. So you create your own library and then use a different format than NIST does when you export their entries?
> 
> Can we add the library files or a similar file to the test resources? Then we can write a test for parsing of this field and make sure it stays stable and supported.

Yeah, you can use those files however you want.

mzmine-community/src/main/java/io/github/mzmine/util/RIColumn.java
wchu207: Oh yeah forgot to reply to this comment directly

https://github.com/mzmine/mzmine/pull/2412#issuecomment-2669785296

mzmine-community/src/main/java/io/github/mzmine/util/RIRecord.java
wchu207: Just saw this. Thanks, I'll do that

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/RITolerance.java
```diff
@@ -0,0 +1,79 @@
...
+import io.github.mzmine.util.RIColumn;
+import io.github.mzmine.util.RIRecord;
+
+/**
+ * RTTolerance allows specifying retention time tolerance it is either absolute (seconds or minutes)
+ * or relative (percent) but as rest of MZmine codebase, it assumes that rt values (other than the
+ * tolerance given in constructor) are in minutes in methods such as getToleranceRange or
+ * checkWithinTolerance
+ */
+public class RITolerance {
```
robinschmid, line 38: Fix the javadoc I changed the text

wchu207:
I've accepted the pull request. Somehow this has broken RIMaxType, RIMinType, RIDiffType. I suspect that the way GCAlignerModule evaluates row bindings is involved somehow, so I'll look into that.

I'll work on the documentation after I fix that, hopefully tomorrow.

wchu207:
I've identified the bug. RICalculationModule was modified to add the summary row types via `outputList.addRowBinding(new Type().createDefaultRowBindings())` instead of adding them as feature types. Makes sense. Except GCAlignerModule creates a wholly new feature list row and naively adds them as row types, omitting the feature type listeners.

I don't have an elegant fix for this. Using the summary types as feature types would be redundant and clutter things. Modifying GCAlignerModule to add the bindings would lead to the same bindings being added once per pre-existing list. We can't check if the bindings are already present since they're unique objects. One compromise would be to check if the type has been added, and if not, always add the listeners for that type, but that would be equivalent to adding listeners in `addRowType` and presumably there's a reason we don't do that.

robinschmid:
Oh you are right. My fault. 
This is something that we never did before. Although binding a row type to another feature type should be possible somehow.

For now you can just add them as feature types again and this will resolve it. They are only added if the user requests that so no big issue

wchu207:
Oh I just checked in and saw you requested changes wrt a javadoc. Fixed.

Also, I noticed you changed the method for selecting the scale in RICalculationTask to a search for the nearest one. The previous method's purpose wasn't obvious -- it would sort the list to newest-first, then find the first scale that was made before the sample was taken; i.e. find the nearest scale in the past.

We designed it this way because sometimes the column would degrade to the point of needing to be replaced, after which we would take new measurements for the straight-chain alkanes. This creates scales that are only applicable for runs that occur after the column was changed.

Do you mind if I revert the changes? Or maybe include an option to switch between these modes?

robinschmid:
We just pushed a new release and I am about to merge your PR into mzmine and this will generate a dev release. 
Therefore, I added some changes like more error messages and logs and wanted to notify you.
During this process I found that the module would silently run on a dataset where the samples were measured before the RI ladder. I know people should do it the opposite way but I thought it would be nice to handle this missing case by just using the closest ladder.

But you are right it makes sense to always use the previous and only in this special case just use the closest ladder. 
I will change this.

robinschmid:
you can now also add the time to the file names and it will use it to decide which scale to use. 
image

PR #2419, opened 2025-02-15
optimise ims features by memory mapping mobilograms

SteffenHeu:
Comparison of approaches to save memory for IMS features. 

- Regular is the current approach, saving each mobilogram's scans in a list in ram, memory mapping the intensity and mz values to the disc.
- Rectangular makes all mobilograms consecutive (no left-out mobility scans, missing are replaced with 0 intensity). Needs additional processing after ims expanding, smoothing, resolving and gap filling. Indices of start and end mobility scans were stored in ram. (code not included here). Saves about 33kB per feature
- Mapping indices of mobility scans to disc using a MemoryLayout and a list of indices. Needs more disc accesses, performance impact during ims expanding, smoothing, resolving and gap filling. Not sure why ms2 paring is so much faster. Saves about 32kB per feature

Processing time comparison:
grafik

Memory usage comparison:
grafik


robinschmid:
Very interesting tests. 
So mostly it was the IMS expander that took a hit in speed.

How many samples did you process?

SteffenHeu:
I think 40

mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/impl/StorageUtils.java
```diff
@@ -234,6 +235,19 @@ public static MemorySegment storeValuesToIntBuffer(@Nullable final MemoryMapStor
...
   }
 
+  public static MemorySegment allocateSegment(@Nullable final MemoryMapStorage storage,
+      @NotNull final MemoryLayout layout, int numValues) {
+
+    MemorySegment buffer;
+    if (storage != null) {
+      buffer = storage.allocateMemorySegment(layout, numValues);
+    } else {
+      buffer = MemorySegment.ofArray(new int[(int) (layout.byteSize() / 4 * numValues)]);
```
robinschmid, line 245: why /4?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/series/IonMobilogramTimeSeriesToRtMobilityHeatmapProvider.java
```diff
@@ -123,38 +131,41 @@ public String getToolTipText(int itemIndex) {
...
       }
     }
     if (isUseSingleColorPaintScale) {
       javafx.scene.paint.Color base = javafx.scene.paint.Color.BLACK;
 //          MZmineCore.getConfiguration().isDarkMode() ? javafx.scene.paint.Color.BLACK
 //              : javafx.scene.paint.Color.WHITE;
-      paintScale = new SimpleColorPalette(new javafx.scene.paint.Color[]{base, color}).toPaintScale(
-          PaintScaleTransform.LINEAR, Range.closed(0d, max));
+      paintScale = new SimpleColorPalette(base, color).toPaintScale(PaintScaleTransform.LINEAR,
+          Range.closed(1d, max));
```
robinschmid, line 150: maybe comment why it is 1 now? was 0 before

mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/impl/StorageUtils.java
SteffenHeu: because an int has 4 bytes

SteffenHeu: this is done due to memory alignment. creating an byte[] array has no alignment, int[] has 4 byte alignment and my layout just had int types. doubles would require an 8 byte alignment, though. so the safest way would be to create a double array and divide by 8.

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/series/IonMobilogramTimeSeriesToRtMobilityHeatmapProvider.java
SteffenHeu: only for testing. reverted.

mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/impl/StorageUtils.java
```diff
@@ -234,6 +236,27 @@ public static MemorySegment storeValuesToIntBuffer(@Nullable final MemoryMapStor
...
+    } else {
+      // must be at least 1 large. If layout.byteSize * numValues < 8 bytes, a double[0] would be created.
+      final long minSize = Math.ceilDiv(layout.byteSize() * numValues,
+          OfDouble.JAVA_DOUBLE.byteSize());
+      if (minSize > Integer.MAX_VALUE) {
+        throw new IllegalArgumentException(
+            "The size of the memory segment is too large. %d entries of %d B per entry (%d B)".formatted(
+                numValues, layout.byteSize(), minSize));
+      }
+      buffer = MemorySegment.ofArray(new double[(int) minSize]);
```
robinschmid, line 254: Seems better to allocate memory segment of the correct size instead of just allocating doubles. You got the layout + numValues so this should be easy.  

Now with arena it feels bad that the MemoryMapStorage is nullable in so many places. Otherwise we could have a storage that by default just keeps in memory. There is already the fallbackAllocator.  

Is there some way to use an Arena here? As far as I understand - we would need an arena that we keep a reference too. Like we do with MemoryMapStorage. Then when a feature list is deleted all memory segments of this arena are closed. 

Or is there a different way to allocate a MemorySegment of byte size and then have the GC take care of it?

robinschmid: Your solution works, but just thinking if there are other options of creating the MemorySegment of correct size without the array. 
Arena seems the only option right?

robinschmid: from the docs I am not sure if we can use ofArray, put double[] and then internally just put ints and floats into the memory. 
It seems like the same memory address is used for the MemorySegment and the array. 

This could break at some point in the future

image

robinschmid: It is more that this use of the ofArray method is not really what it was designed for. 
It keeps the backing double array alive - and while this array is never used - it still sounds unsafe to put int / float into a heap memory region that was reserved for a double array. 
I guess it works for now, as we do not intent to use the double array.

Also dont really get the alignment thing. Something like Arena.ofAuto().allocate(numBytes) should be aligned and then it depends on the actual MemoryLayout used. But this is also forced aligned or developers have to use the unaligned types. 
Only problem I see is where to get the Arena from. Global would be wrong and ofAuto is also wrong as there should be one arena for all memory segments created by this.

SteffenHeu: Yes, `Arena.ofAuto().allocate(MemoryLayout, long)` would be aligned. But we don't have an arena here. As you said, it wouldnt make sense to create one here, so the obvious point is to put it in the MemoryMapStorage and making it notnull everywhere. But that is a large refactor and not the point of this PR.

mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/impl/SimpleIonMobilogramTimeSeries.java
```diff
@@ -263,15 +256,15 @@ public List<Frame> getSpectra() {
...
   }
 
   @Override
   public IonMobilogramTimeSeries copy(@Nullable MemoryMapStorage storage) {
     double[][] data = DataPointUtils.getDataPointsAsDoubleArray(getMZValueBuffer(),
         getIntensityValueBuffer());
-    return IonMobilogramTimeSeriesFactory.of(storage, data[0], data[1], mobilograms, frames,
-        summedMobilogram.copy(storage));
+    return IonMobilogramTimeSeriesFactory.of(storage, data[0], data[1],
+        mobilograms.storedMobilograms(), frames, summedMobilogram.copy(storage));
```
robinschmid, line 267: Maybe better to provide a constructor to copy IonMobilogramTimeSeriesFactory and reuse mobilograms directly?
Also in the other copy methods.

mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/impl/SimpleIonMobilogramTimeSeries.java
```diff
@@ -284,15 +277,15 @@ public IonMobilogramTimeSeries copyAndReplace(@Nullable final MemoryMapStorage s
       @NotNull final double[] newIntensityValues) {
     MemorySegment intensityValues = StorageUtils.storeValuesToDoubleBuffer(storage,
         newIntensityValues);
-    return new SimpleIonMobilogramTimeSeries(mzValues, intensityValues, storage, mobilograms,
-        frames, summedMobilogram);
+    return new SimpleIonMobilogramTimeSeries(mzValues, intensityValues, storage,
+        mobilograms.storedMobilograms(), frames, summedMobilogram);
```
robinschmid, line 281: Similar here. We are reusing most objects like the mzValues so maybe reusing the already stored mobilograms makes sense

SteffenHeu: the StoredMobilograms / MappedStoredMobilograms is specific to the IonMobilogramTimeSeries and the MappedStoredMobilograms can only be created after the mobilograms have already been stored to a MemorySegment. Therefore it cannot be altered.

robinschmid:
squash merged

PR #2426, opened 2025-02-24
add full csv export to wizard

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/BaseWizardBatchBuilder.java
```diff
@@ -403,6 +405,24 @@ protected static void makeAndAddDdaExportSteps(final BatchQueue q, final boolean
...
+
+    param.setParameter(CSVExportModularParameters.featureLists,
+        new FeatureListsSelection(FeatureListsSelectionType.BATCH_LAST_FEATURELISTS));
+    param.setParameter(CSVExportModularParameters.fieldSeparator, ",");
+    param.setParameter(CSVExportModularParameters.idSeparator, ";");
+    param.setParameter(CSVExportModularParameters.omitEmptyColumns, true);
+    param.setParameter(CSVExportModularParameters.filter, FeatureListRowsFilter.ALL);
+
+    File fileName = FileAndPathUtil.eraseFormat(exportPath);
+    fileName = new File(fileName.getParentFile(), fileName.getName() + "_full_feature_table.csv");
```
robinschmid, line 419: maybe we can name it `_quant_full` like in the mol net GNPS export. 
The mol networking export may be set to export both the simple and comprehensive format and this way we would only have a single comprehensive file exported then. 

Currently the FileNameWithSuffixParameter here actually calls it `quant_modular` so maybe we should also adjust this? Use the same suffix here and in the parameter?

robinschmid (dismissed):
I like this change. 
Maybe we should harmonize the suffixes that this export uses.

robinschmid (approved):

PR #2428, opened 2025-02-24
Option for TIC plot without sample lines

robinschmid:
Sample TIC lines are kind of useless if there are too many samples or if the mz range is too wide. 
I propose an option to skip lines for n>x samples. Per default it is off. 
In the XIC quick - currently it turns off for >50 samples. Can also make it always with lines in the quick and only allow this option in the setup dialog.


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/chromatogram/TICVisualizerParameters.java
```diff
@@ -53,21 +55,26 @@ public class TICVisualizerParameters extends SimpleParameterSet {
...
+      "Plot type", "Type of Y value calculation (TIC = sum, base peak = max)",
+      TICPlotType.values());
 
   /**
    * m/z range.
    */
   public static final MZRangeParameter MZ_RANGE = new MZRangeParameter();
 
+  public static final OptionalParameter<IntegerParameter> ticMaxSamples = new OptionalParameter<>(
+      new IntegerParameter("Simplify above n samples",
```
SteffenHeu, line 74: "Draw only features above n samples"
"Omit full chromatograms above n samples"

or something else that makes it clear

robinschmid: It now reads
Omit EIC lines >n samples

PR #2439, opened 2025-02-28
Access actual log file via reflection - open correct file in menu

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/workspace/WorkspaceMenuHelperImpl.java
```diff
@@ -138,18 +135,15 @@ public void loadConfiguration() {
...
-     * There doesn't seem to be any way to obtain the log file name from the logging FileHandler, so
-     * it is hard-coded here for now
-     */
-    final Path logFilePath = Paths.get(
-        FileUtils.getUserDirectory() + File.separator + "mzmine_0_0.log");
-
+    final File logFile = ConfigService.getConfiguration().getLogFile();
+    if (logFile == null) {
+      return;
+    }
```
SteffenHeu, line 141: just in case this fails it might be good to either open the folder path where the file usually is, or show a message to the user, as he explicitly requested the log file.

mzmine-community/src/main/java/io/github/mzmine/util/logging/LoggerUtils.java
```diff
@@ -0,0 +1,53 @@
...
+              for (final Handler handler : parent.getHandlers()) {
+                downstream.accept(handler);
+              }
+            }
+          }).filter(FileHandler.class::isInstance).map(FileHandler.class::cast) // find file handlers
+          .map(LoggerUtils::reflectiveAccessToLogFile) // find first file
+          .filter(Objects::nonNull).findFirst().orElse(null);
+    } catch (Exception ex) {
+      // silent for now
+      ex.printStackTrace();
```
SteffenHeu, line 32: either log the stacktrace ore remove its printing

mzmine-community/src/main/java/io/github/mzmine/util/logging/LoggerUtils.java
```diff
@@ -0,0 +1,53 @@
...
+      Field filesField = fh.getClass().getDeclaredField("files");
+      filesField.setAccessible(true);
+      File[] logFiles = (File[]) filesField.get(fh);
+      if (logFiles.length > 0 && logFiles[0] != null) {
+        return logFiles[0];
+      }
+    } catch (NoSuchFieldException | IllegalAccessException e) {
+      // silent for now
+      e.printStackTrace();
+      logger.log(Level.WARNING, "error during logger setup: " + e.getMessage(), e);
```
SteffenHeu, line 49: same

mzmine-community/src/main/java/io/github/mzmine/util/logging/LoggerUtils.java
robinschmid: ah good catch. This was supposed to be silent but I had an exception and needed the stack trace.

mzmine-community/src/main/java/io/github/mzmine/util/logging/LoggerUtils.java
robinschmid: I will actaully keep the log here just in case. So we know why the log was null
```java
logger.log(Level.WARNING, "Cannot get logger via reflection: " + e.getMessage(), e);
```

mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/workspace/WorkspaceMenuHelperImpl.java
robinschmid: We log the log file path now in the beginning when opening mzmine. 
Actually this was non-null in the end and I was able to remove the null checks.

PR #2445, opened 2025-03-04
Jdk 23, remove string templates

SteffenHeu:


robinschmid (approved):
Looks good and works great

PR #2450, opened 2025-03-05
user service update

SteffenHeu:
use restriction interface in wizard
update gui on license change
update packages


mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/factories/WorkflowWizardParameterFactory.java
```diff
@@ -83,11 +74,6 @@ public final boolean equals(Object o) {
...
 
-  /**
-   * Check if this workflow is available with the license and should be displayed in the current
-   * wizard.
-   * Todo for the future: might be confusing to enable a workspace and have all possible
-   *  workflows here. Maybe rename and limit to possible with licence and workspace in the future
-   */
-  public abstract boolean isAvailableWithLicense(@Nullable MZmineUser user);
+  @Override
+  public abstract @NotNull Set<@NotNull UserActiveService> getUnlockingServices();
```
robinschmid, line 78: I guess this can be removed and then just rely on ServiceRestricted bringing in the method

SteffenHeu: this one already originates from the ServiceRestricted, i just wanted to make it clear in this factory that you have to override the method.

PR #2454, opened 2025-03-07
Fix custom lipid classes

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_lipidid/common/identification/LipidFragmentationRule.java
```diff
@@ -228,4 +233,42 @@ public static LipidFragmentationRule loadFromXML(XMLStreamReader reader)
       return null;
     }
   }
+
+  public static Result validate(@Nullable LipidFragmentationRule rule) {
+    StringBuilder errors = new StringBuilder();
+    if (rule == null) {
+      errors.append("Rule is null\n");
```
robinschmid, line 240: was this for debugging or is that a real issue if the rule is null?

robinschmid (approved):
Looks good to me.

PR #2393, opened 2025-01-31
IonTimeSeries.subseries use copy of scans or a sublist view

robinschmid:
IonTimeSeries.subseries used to just do list.sublist which creates a view and keeps the original list alive. This could lead to a chromatogram.getScans list to be kept alive by sub features. 

However, FeatureResolverTask already uses a FeatureFullDataAccess  that uses a single scans list. This was actually beneficial memory wise as all features from the same sample use the same backing list to create sublists, saving memory.

PR #2455, opened 2025-03-07
Reuse frame lists in resolver and IMS expanding

robinschmid:
This is linked to #2393 and may improve memory usage for IMS processing by reusing frame lists if they are a continuous sublist of the old frames list


robinschmid:
checked and was merged easily with all other datamodel branches. See #2642

robinschmid:
squash merged

PR #2460, opened 2025-03-11
baseline refactor - make num samples relative

SteffenHeu:
make the number of baseline samples relative to hopefully make it more transferable between data sets

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/AbstractBaselineCorrectorParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/AbstractBaselineCorrectorParameters.java
@@ -29,7 +29,6 @@ import io.github.mzmine.parameters.Parameter;
 import io.github.mzmine.parameters.impl.SimpleParameterSet;
 import io.github.mzmine.parameters.parametertypes.BooleanParameter;
 import io.github.mzmine.parameters.parametertypes.PercentParameter;
-import org.jetbrains.annotations.Nullable;
 
 public class AbstractBaselineCorrectorParameters extends SimpleParameterSet {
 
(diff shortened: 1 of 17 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/AbstractBaselineCorrectorParameters.java
```diff
@@ -28,20 +28,21 @@
...
+  public static final PercentParameter samplePercentage = new PercentParameter(
+      "Percentage of baseline samples", """
       The approximate number of samples taken from the chromatogram to fit the baseline.
-      The actual number might be slightly different for each chromatogram, depending on other parameters.
+      The actual number might be slightly different for each chromatogram, depending if peak removal is activated.
       Too low values may fail to approximate the baseline correctly, too high values may put
-      too much weight on chromatographic signals and distort the baseline. Default: 50.
-      """, 50, 2, Integer.MAX_VALUE);
+      too much weight on chromatographic signals and distort the baseline. Default: 10%.
+      """, 0.05);
```
robinschmid, line 45: default says 10% but is set to 5%. 
I think 5 % seems good but both good to me

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_baselinecorrection/AbstractBaselineCorrectorParameters.java
```diff
@@ -58,4 +59,18 @@ public AbstractBaselineCorrectorParameters(String onlineHelpUrl, Parameter<?>...
   public AbstractBaselineCorrectorParameters(Parameter<?>[] parameters, String onlineHelpUrl) {
     super(onlineHelpUrl, parameters);
   }
+
+  @Override
+  public int getVersion() {
```
robinschmid, line 64: getVersion and getVersionMessage needs to be moved to BaselineCorrectionParameters class. 
Currently, we only save the parameters version to the step - not for each embedded parameterset.  
This is something to consider as well - but might be more challenging to read the version of the embedded parameters from xml?

robinschmid (dismissed):
I like the percentage setting for samples. 
Just minor things can be merged after that.

PR #2461, opened 2025-03-13
Region extraction from Volcano and PCA plot

SteffenHeu:
Add ROI selection and extraction to PCA Loadings and Volcano plot. 
This means you can extract siginificantly up/downregulated features and e.g. build libraries with it.

grafik
grafik


mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYChart.java
```diff
@@ -524,4 +529,50 @@ public void addRegression(ColoredXYDataset dataset, int series) {
...
+   * added to a region. Region selection can be finished by
+   * {@link SimpleXYZScatterPlot#finishPath()}.
+   */
+  @Override
+  public void startRegion() {
+    isDrawingRegion.set(true);
+
+    if (currentRegionListener != null) {
+      removeChartMouseListener(currentRegionListener);
+    }
```
robinschmid, line 544: The exact same code is repeated in KendrickMassPlotChart SimpleXYZScatterPlot and here.
Maybe it would be better to create a class of its own `RegionSelection` and just hold an instance here in this plot. Instead of implementing the interface AllowsRegionSelection here.

This class could just hold all the properties and related functions.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotRegionExtractionTask.java
```diff
@@ -0,0 +1,114 @@
...
+        .map(ds -> (VolcanoDatasetProvider) ds.getValueProvider()).toList();
+
+    List<FeatureListRow> rows = new ArrayList<>();
+    for (VolcanoDatasetProvider ds : datasets) {
+      for (int i = 0; i < ds.getValueCount(); i++) {
+        final RowSignificanceTestResult testResult = ds.getItemObject(i);
+        final int finalI = i;
+        if (regions.stream().anyMatch(
+            region -> region.contains(ds.getDomainValue(finalI), ds.getRangeValue(finalI)))) {
+          rows.add(testResult.row());
```
robinschmid, line 102: we have to create a copy of the feature list rows right?
Either change on the top to create a copy of all rows and then filter:
FeatureListUtils.createCopy(flist, " extracted", getMemoryMapStorage(), true);

Or copy each row here?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCALoadingsExtractionTask.java
```diff
@@ -0,0 +1,101 @@
...
+        .map(ds -> (PCALoadingsProvider) ds.getValueProvider()).toList();
+    final List<Path2D> regions = param.getValue(PCALoadingsExtractionParameters.regions).stream()
+        .map(RegionSelectionListener::getShape).toList();
+
+    List<FeatureListRow> rows = new ArrayList<>();
+    for (Path2D region : regions) {
+      for (PCALoadingsProvider loadings : loadingProviders) {
+        for (int i = 0; i < loadings.getValueCount(); i++) {
+          if (region.contains(loadings.getDomainValue(i), loadings.getRangeValue(i))) {
+            rows.add(loadings.getItemObject(i));
```
robinschmid, line 88: copy row here or above for all

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/statistics/TTestConfigurationComponent.java
```diff
@@ -72,13 +72,24 @@ public TTestConfigurationComponent() {
     // todo check if this actually works or throws type exceptions.
     groupACombo = new ComboBox<>();
     groupBCombo = new ComboBox<>();
+    metadataCombo.setEditable(true);
+    groupACombo.setEditable(true);
+    groupBCombo.setEditable(true);
```
robinschmid, line 77: I guess this is for the case when new metadata columns were added?
For users being able to select them?

I have a weird issue when trying to write upper case with shift key in the combo box it opens the quick access. Same for you?

SteffenHeu: Serves that purpose too, but main reasion is that you can add something if you did not load the data/metadata yet, for example when setting up the batch.

robinschmid (changes requested):
Looks good. Just the rows need to be copied. 
Maybe also for other modules?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/statistics/TTestConfigurationComponent.java
robinschmid: Oh true its the parameter component. I was just looking that it was used in the tab.

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYChart.java
SteffenHeu: moved everythin into RegionSelectionWrapper. That class now just needs an EChartviewer

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/statistics/TTestConfigurationComponent.java
SteffenHeu: yeah i have the same issue and i cannot figure out why

PR #2467, opened 2025-03-17
add unknown imaging pattern and fix npe in image viewer

SteffenHeu:
johannes made a python tool which set the imaging pattern to ` <cvParam cvRef="IMS" accession="IMS:1000411" name="one way"/>` which does not exist in the imzml parser. it's techically a fly-back. to prevent similar issues, i added a UNKNOWN pattern and a fallback value in the imaging viewer.

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_imzml/ImagingParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_imzml/ImagingParameters.java
@@ -181,6 +181,9 @@ public class ImagingParameters {
         lateralWidth = maxNumberOfPixelX * pixelWidth;
       }
     }
+    if(pattern == null) {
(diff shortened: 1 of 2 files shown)


robinschmid (approved):
All done

SteffenHeu:
> Maybe you can tell him that the correct term is flyback?
> 
> remark: 091123 Added attributes for imaging laser shot method. Attribute "one way" (IMS:1000411) obsoleted -> replaced by new attribute "fly back"
> 
> [Term] id: IMS:1000413 name: flyback def: "The scanning always happens in the same direction. As soon as the end of the sample is reached, the stage is positioned at the starting edge to begin the next run." [COMPUTIS:IMS] is_a: IMS:1000041 ! Scan Pattern
> 
> https://github.com/imzML/imzML/blob/master/imagingMS.obo

I did, but it was not possible to change it easily in the parser. And since every scan has x and y coordinates, we are not dependent on it anyway.

PR #2500, opened 2025-03-31
Metadata table

### LINKED ISSUE #2440: metadata import duplicates columns
metadata import duplicates columns

SteffenHeu:
- fixes #2440 (column duplication on import)
  - MetadataColumn.equals and hash set only use the title of the column now. Description is not always available during import but lead to the columns not being equal. Expected behaviour is for the columns to just be merged. the first imported table defines the description now
- fix reordering of columns on editing a single cell
- created a wrapper to wrap around the metadata table in the view model.

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/MetadataRow.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/MetadataRow.java
@@ -1,0 +1,91 @@
+package io.github.mzmine.modules.visualization.projectmetadata;
+
+import io.github.mzmine.datamodel.RawDataFile;
+import io.github.mzmine.javafx.dialogs.DialogLoggerUtil;
+import io.github.mzmine.modules.visualization.projectmetadata.table.MetadataTable;
+import io.github.mzmine.modules.visualization.projectmetadata.table.columns.MetadataColumn;
+import java.util.logging.Logger;
+import javafx.beans.property.SimpleObjectProperty;
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.value.ObservableValue;
+import java.util.Objects;
+
+/**
+ * Represents a single row in the TableView, linked to a RawDataFile.
+ */
+public class MetadataRow {
+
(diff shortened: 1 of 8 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/MetadataRow.java
```diff
@@ -0,0 +1,91 @@
...
+   */
+  public ObservableValue<String> getCellValueProperty(MetadataColumn<?> column) {
+    // We use SimpleStringProperty to wrap the value.
+    // It reads directly from the underlying sourceTable.
+    // NOTE: This doesn't automatically update if the underlying table changes
+    // externally. For full two-way binding *observability*, the underlying
+    // MetadataTable would need to support listeners, which is much more complex.
+    // This implementation focuses on displaying and *editing* via the TableView.
+    Object value = sourceTable.getValue(column, this.rawDataFile);
+    return new SimpleStringProperty(Objects.requireNonNullElse(value, "").toString());
```
robinschmid, line 49: maybe it makes sense to create SimpleDoubleProperty und SimpleObjectProperty for the other column types via switch?

This should fix the sorting of numbers.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/MetadataTableModel.java
```diff
@@ -0,0 +1,202 @@
...
+      public String fromString(String s) {
+        final var converted = column.convertOrElse(s, null);
+        if (converted == null) {
+          DialogLoggerUtil.showErrorDialog("Cannot convert input",
+              "The input %s is invalid for the column %s of type %s. Please adhere to the format %s.".formatted(
+                  inQuotes(s), inQuotes(column.getTitle()), inQuotes(column.getType().toString()),
+                  inQuotes(column.exampleValue().toString())));
+          return null;
+        }
+        return s;
```
robinschmid, line 190: maybe return converted.toString here as converted may be in a different format than input s

PR #2501, opened 2025-03-31
Iterative exclusion MS2 pairing

SteffenHeu:
depends on #2500 
- Added as an optional parameter in the advanced ms2 grouping parameters
- Takes a metadata column that contains the name of a raw data file (file extensions are stripped) and pairs it to that file during the regular ms2 grouping step
- MS2s of multiple files can be paired to a **single** file
- Regular ms2 paring processing is applied (rt filter, mz filter...)
image

note: does probably not cover chimerity assessment during library creation (probably, untested, next pr)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_groupms2/GroupMS2Processor.java
```diff
@@ -95,6 +104,14 @@ public class GroupMS2Processor extends AbstractTaskSubProcessor {
   private GroupedMs2RefinementProcessor refineTask;
   private @NotNull String description = "";
 
+  /**
+   * Defines the column name that contains the file name of the data file that MS2s from other files
+   * shall be assigned to. Eg. File A would be a MS1 only file and files B and C would be MS2 only
+   * files. Files B and C would contain "File A" in their row in the mainQuantFileColumn.
+   */
+  private final String mainQuantFileColumnName = DEFAULT_QUANT_FILE_COLUMN_NAME;
```
robinschmid, line 112: has to use the parameter from GroupMS2AdvancedParameters

robinschmid: added a commit to do this https://github.com/mzmine/mzmine/pull/2501/commits/52f1729f648dce800c38a89d66a2d4704c16a414

robinschmid:
I rebased on the master with #2500 included and then added two commits for a typo and to use the parameter instead of a constant column name

robinschmid:
Looks good to me. 
One challenge is how to select only the "MS1" samples for chromatogram builder.

SteffenHeu:
Could add an option "skip files without ms1 scans" in the chromatogram builder

Sanchezillana:
I find the new grouping parameter very promising, but I was wondering: has it been tested in setups where the reference file (used for quantification and alignment) contains only MS1 scans, and the MS2 scans come from separate files which also contain MS1 scans?
I'm currently working with iterative DDA data acquired on both Thermo and Agilent platforms, with plenty of data. In our workflow, the first injection(s) are MS1-only (used for robust quantification), while the subsequent injections are DDA-MS2 runs with dynamic exclusion (but also include MS1 scans). I'd be happy to help test this or contribute feedback.

PR #2512, opened 2025-04-04
Update IMS expander + add integration test

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderParameters.java
```diff
@@ -41,16 +41,17 @@
...
 public class ImsExpanderParameters extends SimpleParameterSet {
 
   public static final FeatureListsParameter featureLists = new FeatureListsParameter();
 
-  public static final OptionalParameter<MZToleranceParameter> mzTolerance = new OptionalParameter<>(
-      new MZToleranceParameter("m/z tolerance",
-          "m/z tolerance for peaks in the mobility dimension. If enabled, the given "
-              + "tolerance will be applied to the feature m/z. If disabled, the m/z range of the "
-              + "feature's data points will be used as a tolerance range."));
+  public static final MZToleranceParameter mzTolerance = new MZToleranceParameter("m/z tolerance",
```
robinschmid, line 50: There is an option to load the old optional mztol and unpack to reuqired parmaeter:
https://github.com/SteffenHeu/mzmine3/pull/26

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderSubTask.java
```diff
@@ -112,7 +114,7 @@ public double getFinishedPercentage() {
         + createdRows / (double) expandingTraces.size() * 0.5;
   }
 
-  @Override
+  // does not work, but why?
```
robinschmid, line 117: I guess this needs to be reverted to @Override

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderSubTask.java
```diff
@@ -135,32 +137,148 @@ public void run() {
...
+
+              if (traceMzRange.lowerEndpoint() > mz) {
+                // all further traces have an even higher start-mz
+                break;
+              }
+
+              // this assumes that the range width at a similar m/z is also of almost equal width.
+              // e.g. a feature with m/z 600 and a feature with m/z 601:
+              // the range of the feature at m/z 600 may not be wider than the range of the feature at m/z 601
+              if (traceMzRange.upperEndpoint() < mz) {
```
robinschmid, line 170: I think you can just add this clause and the comment above is resolved. Because the firstTraceIndex should only skip a trace if exactly that trace was too narrow. This runs in a loop so if multiple traces are < mz then multiple will be skipped. 
```java
if(traceMzRange.upperEndpoint() < mz) {
    if(traceIndex==firstTraceIndex) {
        firstTraceIndex++;
    }
    continue;
}
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderSubTask.java
```diff
@@ -135,32 +137,148 @@ public void run() {
...
+        }
+        processedFrames.getAndIncrement();
+
+        desc = flist.getName() + ": expanding traces for frame " + processedFrames.get() + "/"
+            + totalFrames + " m/z range: " + RangeUtils.formatRange(traceMzRange, mzFormat);
+      }
+    } catch (MissingMassListException e) {
+      // allow traces to be released
+      expandingTraces = null;
+      e.printStackTrace();
```
robinschmid, line 188: remove e.printStackTrace 
For the exceptions I still ask my self how to log it all the time.  
I guess here it should be enough to just do error("Error in IMS expander",  e) and to remove all logging here?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderSubTask.java
```diff
@@ -135,32 +137,148 @@ public void run() {
...
+
+    // allow traces to be released
+    expandingTraces = null;
+
+    setStatus(TaskStatus.FINISHED);
+  }
+
+  // works but 5 times slower
+  //  @Override
+  public void run2() {
```
robinschmid, line 222: maybe remove or rename and add comment that this is an alternative but slower.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderTask.java
```diff
@@ -146,11 +143,9 @@ public void run() {
     final List<? extends FeatureListRow> rows = new ArrayList<>(flist.getRows());
     rows.sort((Comparator.comparingDouble(FeatureListRow::getAverageMZ)));
 
-    // either we use the row m/z + tolerance range, or we use the mz range of the feature.
     final List<ExpandingTrace> expandingTraces = new ArrayList<>(rows.stream().map(
         row -> new ExpandingTrace((ModularFeatureListRow) row,
-            useMzToleranceRange ? mzTolerance.getToleranceRange(row.getAverageMZ())
-                : row.getFeature(imsFile).getRawDataPointsMZRange())).toList());
+            mzTolerance.getToleranceRange(row.getAverageMZ()))).toList());
```
robinschmid, line 148: sad, this was actually a nice use of the actual data mz range. Although this is sometimes quite narrow. 
This is only so that the ranges are kept at similar widths?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_maldispotfeaturedetection/MaldiSpotFeatureDetectionTask.java
```diff
@@ -244,9 +244,7 @@ private ParameterSet createExpanderParameters() {
     final ParameterSet expanderParameters = ConfigService.getConfiguration()
         .getModuleParameters(ImsExpanderModule.class).cloneParameterSet();
     expanderParameters.setParameter(ImsExpanderParameters.useRawData, false);
-    expanderParameters.getParameter(ImsExpanderParameters.mzTolerance).setValue(true);
-    expanderParameters.getParameter(ImsExpanderParameters.mzTolerance).getEmbeddedParameter()
-        .setValue(mzTolerance);
+    expanderParameters.getParameter(ImsExpanderParameters.mzTolerance).setValue(mzTolerance);
```
robinschmid, line 247: maybe better to also set the other parameters? Currently it is only used for the sub tasks but just asking if this should be still defined. 
`mobilogramBinWidth, maxNumTraces, handleOriginal`

robinschmid (commented):
Looks good to me. Just a few minor things

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_imsexpander/ImsExpanderTask.java
SteffenHeu: yes

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_maldispotfeaturedetection/MaldiSpotFeatureDetectionTask.java
SteffenHeu: I can add them yes, but only the tolerance and the raw data parameter are used in the sub task

robinschmid: ok no need then

Issue #2507, opened 2025-04-02
Add all subtypes to CSV export


robinschmid:
CSV export could check if a DataType has ModularDataType values and then export all DataType sub columns together with the already defined subcolumns in the SubColumnFactory. 

Like LocalCSVCompoundDatabaseMatch only exports a few fields and SpectralLibraryMatch as well.

PR #2522, opened 2025-04-08
Improve modular csv export

SteffenHeu:
- CSV modular export now also includes types that are not listed in the feature table #2507 
- Adds InchiKey to feature table


mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_csv/CSVExportModularTask.java
```diff
@@ -287,6 +299,63 @@ private void exportFeatureList(ModularFeatureList flist, BufferedWriter writer)
...
+      }
+
+      for (ModularDataModel row : rows) {
+        final Object entry = row.get(mainType);
+        if (entry == null) {
+          continue;
+        }
+
+        final Object first;
+        if (entry instanceof List list) {
```
robinschmid, line 336: could be more general Collection and then c.stream().findFirst.

robinschmid (approved):

PR #2526, opened 2025-04-08
Bugfix: Changed NistMspParser to clear metadata fields between library entries

wchu207:


SteffenHeu:
Good catch, thank you

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/parser/NistMspParser.java
```diff
@@ -100,7 +100,7 @@ public boolean parse(AbstractTask mainTask, File dataBaseFile, SpectralLibrary l
               // add and push
               addLibraryEntry(entry);
               // reset
-              fields = new EnumMap<>(fields);
+              fields = new EnumMap<>(DBEntryField.class);
```
SteffenHeu, line 103: Can you use clear() in both cases? Should be more efficient that to create a new map

wchu207: All done!

PR #2527, opened 2025-04-09
Fixes

SteffenHeu:
- adapt color palette when swithcing theme
- add single row prediction to formulas
- fix tic plot not showing spectrum if feature data set is selected

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
@@ -57,6 +57,7 @@ import io.github.mzmine.parameters.parametertypes.submodules.ParameterSetParamet
 import io.github.mzmine.util.ExitCode;
 import io.github.mzmine.util.StringUtils;
 import io.github.mzmine.util.color.ColorUtils;
+import io.github.mzmine.util.color.SimpleColorPalette;
 import io.github.mzmine.util.files.FileAndPathUtil;
 import io.github.mzmine.util.web.Proxy;
 import io.github.mzmine.util.web.ProxyType;
@@ -65,8 +66,10 @@ import io.mzio.users.gui.fx.UsersController;
 import java.io.File;
 import java.text.DecimalFormat;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import javafx.application.Platform;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.SimpleBooleanProperty;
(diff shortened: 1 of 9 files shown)


mzmine-community/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
```diff
@@ -453,6 +458,27 @@ private void adjustColorsToThemeDarkMode(final Themes theme) {
...
+        cloned.set(i, inverted);
+      }
+    }
+
+    final List<SimpleColorPalette> palettes = getParameter(
+        MZminePreferences.defaultColorPalette).getPalettes();
+    // check if an inverted palette already exists. if not, use the inverted one.
+    final Optional<SimpleColorPalette> match = palettes.stream().filter(p -> p.equals(cloned, false)).findFirst();
+    match.ifPresentOrElse(colors -> setParameter(MZminePreferences.defaultColorPalette, colors),
+        () -> setParameter(MZminePreferences.defaultColorPalette, cloned));
```
robinschmid, line 480: could be 
```java
setParameter(MZminePreferences.defaultColorPalette, match.orElse(cloned));
```
but also ok like this.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_formulaprediction/ResultWindowController.java
```diff
@@ -167,8 +174,24 @@ private void addIdentityClick(ActionEvent ae) {
...
     }
 
-    SimpleFeatureIdentity newIdentity = new SimpleFeatureIdentity(formula.getFormulaAsString());
-    featureListRow.addFeatureIdentity(newIdentity, false);
+    // make the formula column visible in the feature table
+    final Integer result = ((MZmineDesktop) DesktopService.getDesktop()).getAllTabs().stream()
+        .filter(tab -> tab instanceof FeatureTableTab).map(FeatureTableTab.class::cast)
+        .filter(tab -> tab.getFeatureList() == featureListRow.getFeatureList())
+        .map(FeatureTableTab::getFeatureTable).findFirst().map(table -> {
+          if (!table.getFeatureList().getFeatureTypes().contains(FormulaListType.class)) {
```
robinschmid, line 182: maybe you could just add a RowTypeListener in the FeatureTableFX?
Or we do something more general like a listener that evaluates if new types were added to a table and then refresh it?

Also the check should be featureList.getRowTypes.contains.
Currently FeatureTypes are checked

robinschmid (dismissed):
Looks good to me

robinschmid (approved):

PR #2539, opened 2025-04-14
add tab subtitles

SteffenHeu:
Add subtitles to the tab. introduces an utility method, because Tab#getText cannot be overridden

image


javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFlows.java
```diff
@@ -80,4 +85,18 @@ public static Region newTextFlowInAccordion(final String title, boolean expanded
...
+      switch (child) {
+        case Text t -> c.accept(t.getText());
+        case Label t -> c.accept(t.getText());
+        case TextField tf -> c.accept(tf.getText());
+        case TextArea ta -> c.accept(ta.getText());
+        case Parent p -> c.accept(getText(p));
+        default -> {
+        }
+      }
+    }).collect(Collectors.joining());
```
robinschmid, line 100: maybe join with a space or allow to choose joining?

robinschmid (approved):
Looks good. 
Just one question for the getText utility. 
Maybe this method should be named getChildrenText()?
Can be merged after your decision.

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFlows.java
robinschmid: Ok this explains why it is in the FxTextFlow. 
Makes sense.

PR #2577, opened 2025-04-29
Provide unmodifiable view in featureList.getRows

robinschmid:
Now has two methods one to retrieve view and one for modifiable copy

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/FeatureList.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/FeatureList.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2024 The mzmine Development Team
+ * Copyright (c) 2004-2025 The mzmine Development Team
  *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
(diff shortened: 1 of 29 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/FeatureList.java
```diff
@@ -203,10 +204,20 @@ default boolean hasRowType(DataType type) {
   public FeatureListRow getRow(int row);
 
   /**
-   * Returns all feature list rows
+   * All modification to the rows should be done through the provided methods or internally in this
```
SteffenHeu, line 207: ... to the row order ...
otherwise it also sounds like e.g. adding a data type or so.

robinschmid: you mean **All modification to the row order**?
But this really means all mutations like clear, add, setAll, remove.
I will specify this more

robinschmid:
```java
/**
   * An unmodifiable view of the list of rows. All mutations are made by the {@link FeatureList}
   * instance internally. Like sorting {@link FeatureList#applyDefaultRowsSorting()}, deletion,
   * setAll, add, clear.
   *
   * @return an unmodifiable view of rows
   */
```

PR #2600, opened 2025-05-12
Fixes

SteffenHeu:
- Allow correlation of MS1 features with MS1 scans by DIA correlation (e.g. in source fragmentation)
- Fix single spectrum library search (broken since spectra merging)
- fail batch on default if some "parameter was updated" warnings appear
  - add option to ignore
- save module name (gui name) in batch file


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_diams2/DiaMs2CorrTask.java
```diff
@@ -755,8 +755,14 @@ private Map<IsolationWindow, List<Scan>> extractIsolationWindows(
       }
     }
 
+    if (windowScanMap.isEmpty()) {
+      windowScanMap.put(new IsolationWindow(Range.closed(Double.MIN_VALUE, Double.MAX_VALUE), null),
```
robinschmid, line 759: maybe good to put a comment why you are doing this. 
This is supposed to be a range that encloses the whole range right?
Double.MIN_VALUE is tricky, most often I get confused by it. Usually it should be 0 or Double.NEGATIVE_INFINITY or Range.all in this case. 

MIN_VALUE is just the smallest value > 0

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/RowsSpectralMatchTask.java
```diff
@@ -115,21 +115,29 @@ public class RowsSpectralMatchTask extends AbstractTask {
   // 13C, H, 2H or Cl
   private boolean needsIsotopePattern;
   private int minMatchedIsoSignals;
-  private double scanPrecursorMZ;
 
-  public RowsSpectralMatchTask(ParameterSet parameters, @NotNull Scan scan,
-                               @NotNull Instant moduleCallDate) {
+  // only used for single spectrum searches
+  private final Double singleScanPrecursorMZ;
+  private final boolean isSingleScanSearch;
```
robinschmid, line 121: isSingleScanSearch is unused

robinschmid (approved):
Looks good to me. Just one comment

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_diams2/DiaMs2CorrTask.java
SteffenHeu: Range.all does not have bounds, which would not work here. Using 0

PR #2606, opened 2025-05-16
Add library entry ID to feature table and exports

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/DatabaseEntryIdType.java
```diff
@@ -0,0 +1,41 @@
...
+package io.github.mzmine.datamodel.features.types.annotations;
+
+import io.github.mzmine.datamodel.features.types.abstr.StringType;
+import org.jetbrains.annotations.NotNull;
+
+public class DatabaseEntryIdType extends StringType {
+
+  @Override
+  public @NotNull String getUniqueID() {
+    return "database_entry_id";
```
robinschmid, line 34: maybe we could call it just EntryIdType and entry_id as unique id?
Then we could use it for any kind of String ID even if it is not a database

robinschmid (dismissed):
Just an idea to make it a more general type.

robinschmid (approved):

PR #2607, opened 2025-05-16
Feature list export to xml

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_xml/ExportFeaturesToXMLModule.java
```diff
@@ -0,0 +1,30 @@
...
+import io.github.mzmine.taskcontrol.Task;
+import io.github.mzmine.util.MemoryMapStorage;
+import java.time.Instant;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class ExportFeaturesToXMLModule extends TaskPerFeatureListModule {
+
+  public ExportFeaturesToXMLModule() {
+    super("Export features to XML", ExportFeaturesToXMLParameters.class,
```
robinschmid, line 18: Maybe we could rename the module to **Export feature data**  and then just add a ComboParameter with for now only XML as Format option. 
Then it is easy to add different exports in the future without the need to change documentation etc.

robinschmid (approved):
Looks good to me.

PR #2655, opened 2025-05-31
Simplify scan, remove ranges, save memory

robinschmid:
- guava Range is a memory killer so thought to change it to primitives

Together with MSMSInfo shift to primitives it saves a lot. See Ranges, Scan, MassList

## Before
image
 
## After
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/Scan.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/Scan.java
@@ -162,7 +162,7 @@ public interface Scan extends MassSpectrum, Comparable<Scan> {
   /**
    * @return The actual scanning range of the instrument
    */
-  @NotNull Range<Double> getScanningMZRange();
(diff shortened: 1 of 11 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/Scan.java
```diff
@@ -162,7 +162,7 @@ default boolean hasInjectionTime() {
   /**
    * @return The actual scanning range of the instrument
    */
-  @NotNull Range<Double> getScanningMZRange();
+  @Nullable Range<Double> getScanningMZRange();
```
SteffenHeu, line 165: Making something nullable that was not null before has downstream implications. not sure i want to merge this pr without adapting those usages.

robinschmid: The children like SimpleScan actually annotated this as Nullable. So this is the actual behavior. 
This should have been nullable in the first place

PR #2666, opened 2025-06-11
Merge fragmentation methods separately

robinschmid:
- merges each method separately
- only merges across methods if selected in advanced
- uses mixed as method then

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/MergedMassSpectrum.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/MergedMassSpectrum.java
@@ -138,7 +138,7 @@ public interface MergedMassSpectrum extends Scan {
         case SAME_ENERGY -> "Same energy merged";
         case UNDEFINED_ENERGY -> "Undefined energy merged";
         case ALL_ENERGIES -> "All energies merged";
-        case ALL_METHODS -> "All methods merged";
+        case ALL_METHODS -> "All fragmentation methods merged";
         case ALL_MSN_TO_PSEUDO_MS2 -> "MSn to pseudo MS2 merged";
         case CORRELATED_MS1 -> "CORRELATED MS"; // required by SIRIUS?
         case PASEF_SINGLE -> "Single PASEF";
@@ -154,7 +154,7 @@ public interface MergedMassSpectrum extends Scan {
         case SAME_ENERGY -> "SAME_ENERGY";
         case UNDEFINED_ENERGY -> "UNDEFINED_ENERGY";
         case ALL_ENERGIES -> "ALL_ENERGIES";
-        case ALL_METHODS -> "ALL_METHODS";
+        case ALL_METHODS -> "ALL_FRAGMENTATION_METHODS";
         case ALL_MSN_TO_PSEUDO_MS2 -> "ALL_MSN_TO_PSEUDO_MS2";
         case CORRELATED_MS1 -> "CORRELATED_MS";
         case PASEF_SINGLE -> "PASEF_SINGLE";
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_scan_merge_select/options/MergedSpectraFinalSelectionTypes.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_scan_merge_select/options/MergedSpectraFinalSelectionTypes.java
@@ -58,7 +58,8 @@ public enum MergedSpectraFinalSelectionTypes implements UniqueIdSupplier {
     return !isEmptySelection(types) &&
         // at least one of the merging types
         (types.contains(ACROSS_ENERGIES) || types.contains(EACH_ENERGY) //
-            || types.contains(MSN_TREE) || types.contains(MSN_PSEUDO_MS2));
+            || types.contains(ACROSS_FRAGMENT_METHODS) || types.contains(MSN_TREE)
+            || types.contains(MSN_PSEUDO_MS2));
   }
 
   public static boolean containsSampleDefinition(
@@ -100,7 +101,7 @@ public enum MergedSpectraFinalSelectionTypes implements UniqueIdSupplier {
       case EACH_SAMPLE -> "Each sample";
       case ACROSS_ENERGIES -> "Across energies";
       case EACH_ENERGY -> "Each energy";
-      case ACROSS_FRAGMENT_METHODS -> "Across methods";
+      case ACROSS_FRAGMENT_METHODS -> "Across fragmentation methods";
       case MSN_TREE -> "MSn tree";
       case MSN_PSEUDO_MS2 -> "MSn to pseudo MS2";
     };
--- a/mzmine-community/src/main/java/io/github/mzmine/util/scans/merging/SpectraMerger.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/util/scans/merging/SpectraMerger.java
@@ -164,8 +164,8 @@ public class SpectraMerger {
         scansToMerge = resultsByMethod.stream().mapMulti((r, consumer) -> r.acrossSamples()
             .forEach(node -> node.streamScanByEnergy().forEach(consumer)));
       }
-      // do the actual merging only on MS2
-      final List<Scan> toMerge = scansToMerge.filter(scan -> scan.getMSLevel() <= 2).toList();
+      // do the actual merging on the whole list of scans
+      final List<Scan> toMerge = scansToMerge.toList();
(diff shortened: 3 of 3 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/MergedMassSpectrum.java
```diff
@@ -134,6 +138,7 @@ public String toString() {
         case SAME_ENERGY -> "Same energy merged";
         case UNDEFINED_ENERGY -> "Undefined energy merged";
         case ALL_ENERGIES -> "All energies merged";
+        case ALL_METHODS -> "All methods merged";
```
SteffenHeu: "methods" is unspecific

All fragmentation methods merged
All excitation methods merged

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_scan_merge_select/options/MergedSpectraFinalSelectionTypes.java
```diff
@@ -95,6 +100,7 @@ public String toString() {
       case EACH_SAMPLE -> "Each sample";
       case ACROSS_ENERGIES -> "Across energies";
       case EACH_ENERGY -> "Each energy";
+      case ACROSS_FRAGMENT_METHODS -> "Across methods";
```
SteffenHeu: again, just methods is unspecific

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_scan_merge_select/options/MergedSpectraFinalSelectionTypes.java
```diff
@@ -46,13 +46,18 @@ public enum MergedSpectraFinalSelectionTypes implements UniqueIdSupplier {
...
+
   // MSn
   MSN_TREE, MSN_PSEUDO_MS2;
 
   public static boolean isActive(Collection<MergedSpectraFinalSelectionTypes> types) {
     return !isEmptySelection(types) &&
-           // at least one of the merging types
-           (types.contains(ACROSS_ENERGIES) || types.contains(EACH_ENERGY) //
+        // at least one of the merging types
+        (types.contains(ACROSS_ENERGIES) || types.contains(EACH_ENERGY) //
```
SteffenHeu, line 60: add across methods here?

mzmine-community/src/main/java/io/github/mzmine/util/scans/ScanUtils.java
```diff
@@ -301,6 +302,85 @@ public static List<Float> extractCollisionEnergies(MassSpectrum spectrum) {
...
+        } else {
+          yield ActivationMethod.UNKNOWN;
+        }
+      }
+      case SpectralLibraryEntry entry -> {
+        final Object m = entry.getOrElse(DBEntryField.FRAGMENTATION_METHOD, null);
+        if (m instanceof ActivationMethod a) {
+          yield a;
+        } else if (m != null) {
+          yield m;
```
SteffenHeu, line 328: use to string here, so it's an actual string and not any object? then maybe also trim?

SteffenHeu: or even better add some patterns to ActivationMethod and create a from string method? then we can properly deal with it in downstream processing

mzmine-community/src/main/java/io/github/mzmine/util/scans/ScanUtils.java
```diff
@@ -2054,6 +2136,23 @@ public static Map<String, List<Scan>> splitBySample(final List<Scan> scans) {
...
+   * spectral libraries). Includes all source scans of {@link MergedMassSpectrum} in listing all
+   * fragmentation methods.
+   *
+   * @param scans list of spectra to be grouped
+   * @return maps fragmentation method to list of scans with keys: Either unique
+   * {@link ActivationMethod#getName()} (other strings possible) or {@link ActivationMethod#MIXED}
+   * if multiple activation methods were merged. null == {@link ActivationMethod#UNKNOWN}. Might use
+   * other string values if loaded for spectral library.
+   */
+  public static Map<String, List<Scan>> splitByFragmentationMethod(List<Scan> scans) {
```
SteffenHeu, line 2151: having the enum here would also be beneficial

mzmine-community/src/main/java/io/github/mzmine/util/scans/merging/SpectraMerger.java
```diff
@@ -108,13 +115,84 @@ public SpectraMerger(final List<MergedSpectraFinalSelectionTypes> finalScanSelec
...
+        }
+      } else {
+        // not merged across energies so take every individual energy
+        // can always use across samples here
+        // in case no merging across samples happened the variable points to for each samples instead
+        scansToMerge = resultsByMethod.stream().mapMulti((r, consumer) -> r.acrossSamples()
+            .forEach(node -> node.streamScanByEnergy().forEach(consumer)));
+      }
+      // do the actual merging only on MS2
+      final List<Scan> toMerge = scansToMerge.filter(scan -> scan.getMSLevel() <= 2).toList();
```
SteffenHeu: why?

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralLibraryEntryFactory.java
```diff
@@ -346,6 +347,9 @@ public void addScanSpecificFields(@NotNull final SpectralLibraryEntry entry,
       }
     }
 
+    // extract fragmentation method from spectrum - may be mixed if multiple
+    entry.putIfNotNull(DBEntryField.FRAGMENTATION_METHOD, extractFragmentationMethodShort(spec));
+
```
SteffenHeu, line 352: so before it was sometimes a String, sometimes an ActivationMethod. Is it fully uniform now?

mzmine-community/src/main/java/io/github/mzmine/datamodel/MergedMassSpectrum.java
robinschmid: fragmentation now

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_scan_merge_select/options/MergedSpectraFinalSelectionTypes.java
robinschmid: I think you are right. Good catch. Will test it.

mzmine-community/src/main/java/io/github/mzmine/util/scans/ScanUtils.java
robinschmid: Problem is that spectral libraries read non standard fragmenation method information as String. And not all of them can be converted to proper enum constants. Then most of them would just be enum.Unknown or something like that. 

So I think there is no other way than using String here for the grouping of scans. 
The library spectra may still be grouped if the method reads something random like: FRAGMENTATION_30

mzmine-community/src/main/java/io/github/mzmine/util/scans/ScanUtils.java
robinschmid: Not really. We are not using the key in any way and the fragmentation methods may be defined as random strings in spectral library entries.

mzmine-community/src/main/java/io/github/mzmine/util/scans/merging/SpectraMerger.java
robinschmid: can also merge all. Not sure why I opted to only merge MS2

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralLibraryEntryFactory.java
robinschmid: There is not really a way to make it uniform in the spectral library entry as you can always set what ever. 
But I think most places now use String

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_scan_merge_select/options/MergedSpectraFinalSelectionTypes.java
robinschmid: works

robinschmid:
Add test with data with two methods and library building

PR #2671, opened 2025-06-12
prevent feature lists selection exception

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/FeatureListsSelection.java
```diff
@@ -92,22 +92,6 @@ public void resetEvaluatedSelection() {
   @NotNull
   public ModularFeatureList[] getMatchingFeatureLists() {
 
-    if (evaluatedSelection != null) {
```
robinschmid, line 95: maybe we should do the same for raw data files

robinschmid (approved):

PR #2678, opened 2025-06-13
Faster pseudo spectrum preview

robinschmid:
- adds FeatureList combobox to preview
- adds listeners to events
- faster preview
- removes XYAnnotations that showed width of peaks (duplicated number of data points to draw) - too slow. Could change the XY to XYZ dataset in the future and handle the RT width in the renderer
- adds label to show that parameters have changed

## NOW
image

## BEFORE
image


javafx-framework/src/main/java/io/github/mzmine/javafx/properties/PropertyUtils.java
```diff
@@ -99,4 +101,11 @@ public static void onChange(final Runnable operation, final ObservableList<?> ob
     });
   }
 
+  /**
+   * @param <T> elements of list and binding
+   * @return create a new property for the first element in a list
+   */
+  public static <T> ObservableValue<T> firstElementProperty(ObjectProperty<List<T>> listProp) {
```
SteffenHeu, line 110: this is only unidirectional, right?

PR #2690, opened 2025-06-22
Add documentation links + readout script

SteffenHeu:
- Add a script that prints all MZmineModule instances that do not have a link to the online documentation
- Add some doc links that exist but were not listed

### DIFF
--- a/mzmine-community/src/main/resources/dependency/dependency-licenses.json
+++ b/mzmine-community/src/main/resources/dependency/dependency-licenses.json
@@ -131,6 +131,19 @@
             ]
         },
         {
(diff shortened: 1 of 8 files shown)


mzmine-community/src/main/resources/dependency/dependency-licenses.json
```diff
@@ -130,6 +130,19 @@
                 }
             ]
         },
+        {
+            "moduleName": "com.fasterxml.jackson.datatype:jackson-datatype-jdk8",
```
robinschmid, line 134: how did this get added?

PR #2705, opened 2025-07-08
Add checkbox to pseudo spectrum pane

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/util/javafx/WindowsMenu.java
```diff
@@ -74,12 +75,18 @@ public WindowsMenu() {
...
+    menuBar.setUseSystemMenuBar(true);
+    menuBar.getMenus().add(new WindowsMenu());
+
+    switch (rootNode) {
+      case BorderPane bp -> bp.setTop(menuBar);
+      case Pane p -> {
+        Pane rootPane = (Pane) rootNode;
+        rootPane.getChildren().add(menuBar);
+      }
+      default -> {}
```
robinschmid, line 89: maybe throw exception here?

mzmine-community/src/main/java/io/github/mzmine/util/FeatureListUtils.java
```diff
@@ -138,6 +165,26 @@ public static void bindSelectedFeatures(WeakAdapter weak, FeatureTableFX table,
     });
   }
 
+  public static void bindSelectedFeatures(WeakAdapter weak, FeatureTableFX table,
+      ObjectProperty<List<Feature>> selectedFeatures, ObservableBooleanValue isUpdateEnabled) {
```
robinschmid: What is the difference of this method and the one without isUpdateEnabled? 
The first ListChangeListener will always update the selected values. 

Maybe just add the check for update enabled to the first list listener?

```java
if (weak.isDisposed() || (isUpdateEnabled!=null && isUpdateEnabled.get() != true)) {
    return;
}
```

Then you could remove the listener on the update property. And you could just overload and forward the call from the method without the ObservableBooleanValue

mzmine-community/src/main/java/io/github/mzmine/util/FeatureListUtils.java
```diff
@@ -150,6 +197,27 @@ public static void bindSelectedRawDataFiles(WeakAdapter weak, FeatureTableFX tab
     });
   }
 
+  public static void bindSelectedRawDataFiles(WeakAdapter weak, FeatureTableFX table,
+      ObjectProperty<List<RawDataFile>> selectedFiles, ObservableBooleanValue isUpdateEnabled) {
+    weak.addListChangeListener(table, table.getSelectionModel().getSelectedCells(), c -> {
+      if (weak.isDisposed() || !isUpdateEnabled.get()) {
```
robinschmid, line 181: Ok now I see that is actually what you did for the other methods. So this was just missing in the other method. 

Still would be good to avoid code duplication by overloading the method. You can just pass in null or new BooleanProperty(true) to this method from its version without boolean. 

Same for feature, row, rawFiles method

robinschmid (changes requested):
Looks good. Just one comment with code duplication and a missing check in the selectedFeatures method

mzmine-community/src/main/java/io/github/mzmine/util/javafx/WindowsMenu.java
robinschmid: still maybe throw exception?

PR #2711, opened 2025-07-09
Added support for converting retention time values to temperature values

DomiDotso:
Retention time values are now able to be converted to retention time values based on a programmed temperature ramping rate. This will aid with thermal desorption and decomposition studies using direct analysis in real time coupled with HRMS. Additional support will be needed to add the option to display temperature to the visualization modules. (please help)

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/transform_rttotemperature/RTtoTemperatureModule.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/transform_rttotemperature/RTtoTemperatureModule.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2004-2024 The MZmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+ * OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.dataprocessing.transform_rttotemperature;
+
+import io.github.mzmine.datamodel.MZmineProject;
+import io.github.mzmine.datamodel.RawDataFile;
+import io.github.mzmine.modules.MZmineModuleCategory;
+import io.github.mzmine.modules.MZmineProcessingModule;
+import io.github.mzmine.parameters.ParameterSet;
+import io.github.mzmine.taskcontrol.Task;
+import io.github.mzmine.util.ExitCode;
+import io.github.mzmine.util.MemoryMapStorage;
+import org.jetbrains.annotations.NotNull;
+
+import java.time.Instant;
+import java.util.Collection;
+
+public class RTtoTemperatureModule implements MZmineProcessingModule {
+
+    private static final String MODULE_NAME = "Retention Time → Temperature";
+    private static final String MODULE_DESCRIPTION =
+            "Converts scan retention times into oven temperatures based on the user-defined program.";
+
+    @Override
+    public @NotNull String getName() {
+        return MODULE_NAME;
+    }
+
+    @Override
+    public @NotNull String getDescription() {
+        return MODULE_DESCRIPTION;
+    }
+
+    @Override
+    public @NotNull ExitCode runModule(@NotNull MZmineProject project,
+                                       @NotNull ParameterSet parameters,
+                                       @NotNull Collection<Task> tasks,
+                                       @NotNull Instant moduleCallDate) {
+        // Use one shared storage for all tasks in this module invocation
+        MemoryMapStorage storage = MemoryMapStorage.forRawDataFile();
+
+        // Retrieve the selected raw data files via the static parameter key
+        RawDataFile[] files = parameters
+                .getValue(RTtoTemperatureParameters.dataFiles)
+                .getMatchingRawDataFiles();
+
+        for (RawDataFile file : files) {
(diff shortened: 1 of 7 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/transform_rttotemperature/RTtoTemperatureModule.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright (c) 2004-2024 The MZmine Development Team
```
SteffenHeu, line 2: please update to 2025 and mzmine

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/transform_rttotemperature/RTtoTemperatureModule.java
```diff
@@ -0,0 +1,91 @@
...
+import io.github.mzmine.util.ExitCode;
+import io.github.mzmine.util.MemoryMapStorage;
+import org.jetbrains.annotations.NotNull;
+
+import java.time.Instant;
+import java.util.Collection;
+
+public class RTtoTemperatureModule implements MZmineProcessingModule {
+
+    private static final String MODULE_NAME = "Retention Time → Temperature";
```
SteffenHeu, line 43: please use the code formatter of the project: mzmine-intellij-code-formater.xml

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/transform_rttotemperature/RTtoTemperatureTask.java
```diff
@@ -0,0 +1,162 @@
...
+        double FinalTemp = parameters.getValue(RTtoTemperatureParameters.finaltemp);
+        double rate = parameters.getValue(RTtoTemperatureParameters.tempramp);
+
+        // 2) Read the Celsius vs. Kelvin flag
+        boolean useKelvin = parameters
+                .getParameter(RTtoTemperatureParameters.displayKelvin)
+                .getValue();
+
+        // 3) Fetch scans & precompute temperatures
+        List<Scan> scans = dataFile.getScanNumbers(1);
```
SteffenHeu, line 82: The better way to do this is to use a ScanSelectionParameter and use the scans of the scan selection here

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/transform_rttotemperature/RTtoTemperatureTask.java
```diff
@@ -0,0 +1,162 @@
...
+                logger.info("Conversion canceled after "
+                        + processedScans + " of " + totalScans + " scans");
+                return;
+            }
+            Scan s = scans.get(i);
+            if (s instanceof SimpleScan) {
+                ((SimpleScan) s).setRetentionTime((float) temps[i]);
+            }
+            processedScans++;
+        }
```
SteffenHeu, line 111: Usually we try to not alter the raw data. The "mzmine way" would be to introduce a new data type, e.g., `DesorptionTemperatureType extends FloatType` (or DoubleType, but i'd say the accuracy of float is enough, especially since a float (the rt) is used in the calculation of the temperature) and then set this to the features.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/chromatogram/TICPlot.java
```diff
@@ -167,6 +171,31 @@ public TICPlot() {
...
+      unitMenu.getItems().add(item);
+    }
+
+// 3) Show the menu on right-click anywhere on the chart:
+    this.addEventHandler(MouseEvent.MOUSE_CLICKED, e -> {
+      if (e.getButton() == MouseButton.SECONDARY) {
+        unitMenu.show(this, e.getScreenX(), e.getScreenY());
+      }
+    });
+
```
SteffenHeu, line 198: This basically overrides the current context menu, or at least causes both to be shown, or? 
In general, this does not check if the RT has actually been replaced by a temperature or not, so it is kind of hacky :D.
I would prefer if the TIC plot would not be changed and we added a new option to display a desorption plot. (see comment)

DomiDotso:
> Hi @DomiDotso,
> 
> great that you are contributing!
> 
> I have a suggestion, though: In general, we try to not alter the raw data after it has been imported. By overriding the RT with a temperature, that is being done, making it at least a little bit hacky.
> 
> If I were to implement this in "the mzmine way", I would create a new DataType (which can be added to the feature table). Then, you can calculate the temperature at which the features are the most intense at, and set it to the feature.
> 
> Regarding the visualization: Here the question is, what the "goal" is. Is it just to inspect the data, without exporting the full desoprtion chromatograms or so? Then it might be better to create a specific visualizer for that. We try to follow the MVCI (Model, view, controller, interactor) principle for all new visualisers. As example, you can check out the `IntegrationDashboard `or the `StatisticsDashboard`, which implement the necessary interfaces. When opening it, you could then use parameter set to let the user define which ramp he/she ran. By opening a feature table in the dashboard or by just binding to the selected feature in an already open feature table, you can then update the plot and display the temperature on the x axis. To create the plot, we have the `SimpleXYChart `class, along with the `ColoredXYDataset`, which lets you better control the color. To get the data in there, the easiest way would be to use the `AnyXYProvider`. To get the feature's chromatogram you can follow this route:
> 
> ```java
>       ModularFeatureList flist; // get via FeatureListsParameter from the parameter set
>       List<ModularFeatureListRow> rows = flist.getRows();
>       ModularFeatureListRow row = rows.getFirst();
>       final ModularFeature bestFeature = rows.getFirst().getBestFeature();
>       final IonTimeSeries<? extends Scan> featureData = bestFeature.getFeatureData();
> 
>       featureData.getIntensity(index);
>       featureData.getRetentionTime(index);
> ```
> 
> The AnyXYProvider can then compute the temperature values in the visualiser on the fly.
> 
> Please let me know what you think about that.
> 
> Again, great to see a community contribution.
> 
> Best, Steffen

Hello Steffen, 

Thank you for getting back to me so quickly. 

I plan on making the appropriate edits as soon as possible. In regards to the scope of my project we are using temperature programmed decomposition with direct analysis in real time coupled with HRMS (orbitrap) to study the decay of polymers as a function of temperature. For data processing we have been manually extracting each EIC record for a small portion of the species of interest, and converting the retention times into temperature values in order to determine the decay of a particular species as a function of temperature as well as that species contribution to the total mass fraction of plastic by normalizing, integrating, and accounting for ionization efficiency. Needless to say, it is a processing method that takes quite a bit of time, but I believe that mzMine will help immensely. 

By implementing the proposed module above, I was able to generate the shapes (EICs) as a function of temperature for all of the species of interest. For the additional steps, I plan on developing further modules for use with mzMine if it is at all possible. The area calculation as well as the normalization modules available will make my processing 100% more efficient, not to mention I would be able to gather all of the EICs rather than a select few. It would also be really cool to use the 3D plots with a temperature label as well, currently I am using Originpro and it does not like 3D plots at all.

I will attach an excel file displaying what I am hoping to accomplish.

Let me know what you think!

One last thing, is there a way I can export the complete EIC record for each species in the feature list all at once? 
[PETG_White_60C.csv](https://github.com/user-attachments/files/21166367/PETG_White_60C.csv)

SteffenHeu:
Hi @DomiDotso,

you can export everything from the feature table to an XML file using this export: https://mzmine.github.io/mzmine_documentation/module_docs/io_export_feature_data/export-feature-data.html 
It is more complex to parse the xml than a csv file, though.
If you only need to export the temperature chromatograms, writing a specific export for that may be the best solution without altering the raw data. As an alternative, you could make an alternative class to the IonTimeSeries that stores the temperature values per feature

Sounds interesting! Please let us know, if you need more suggestions.

PR #2716, opened 2025-07-11
Harmonize statistics

robinschmid:
- adds DataTable and FeatureDataTable to harmonize the preparation of feature abundance table
- Missing value imputation, abundance measure, scaling all bundled into AbundanceDataTablePreparationConfig
- StatisticUtils.extractAbundancesPrepareData applies the extraction and preparation
- All stats modules now use this unified method to prepare data
- added missin value imputation to all stats modules
- added new global LOD imputer to rely on a global limit of detection as the minimum intensity
- added significance and fold-change filter to RowsFilterTask

### Modified modules
- ANOVA
- TTEST
- Volcano: visual and extraction
- PCA: visual and extraction
- RowsFilter


javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxComboBox.java
```diff
@@ -92,4 +95,40 @@ public static <T> ComboBox<T> createComboBox(String tooltip, T[] values,
       Property<T> selectedItem) {
     return createComboBox(tooltip, List.of(values), selectedItem);
   }
+
+  /**
+   * A combobox with auto complete for its items. This is useful for parameters when the input text
+   * may be used although items do not define this option. This may be different from
+   * SearchableComboBox but not testet.
+   */
+  public static <T> ComboBox<T> newAutoCompleteComboBox(String tooltip) {
```
SteffenHeu: tooltip nulllable?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFields.java
```diff
@@ -137,4 +142,77 @@ public static <T> TextFormatter<T> attachDelayedTextFormatter(final TextField te
...
+   * Automatically bind auto completion to a text field based on a list, e.g., ObservableList or
+   * static
+   *
+   * @param textField the target
+   * @param options   options in an ObservableValue
+   * @return AutoCompletionBinding of the string representation
+   */
+  public static AutoCompletionBinding<String> bindAutoCompletion(TextField textField,
+      List<?> options) {
+
```
SteffenHeu: what about an option with a supplier or a function for the options? that would mean we can always update from the metadata table on typing.

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/DataTable.java
```diff
@@ -0,0 +1,120 @@
...
+  default Spliterator<double[]> spliterator() {
+    return Spliterators.spliterator(iterator(), getNumberOfFeatures(),
+        Spliterator.SIZED | Spliterator.ORDERED | Spliterator.IMMUTABLE);
+  }
+
+
+  /**
+   * Stream all values in no guaranteed order
+   */
+  default @NotNull DoubleStream streamValues() {
```
SteffenHeu, line 135: whats this used for?

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeatureListRowAbundances.java
```diff
@@ -0,0 +1,165 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public interface FeatureListRowAbundances {
+
+  /**
+   * @param trackMissingValues creates a {@link MissingValuesImputedFeatureListRowAbundances} to
+   *                           track the original missing values so that if zero values are imputed
+   *                           later the original missing values ratio is kept
+   */
+  static FeatureListRowAbundances of(FeatureListRow row, double[] values,
```
SteffenHeu: what is values?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/SignificanceTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+public interface SignificanceTest {
+
+  /**
+   * Applies prechecks. Use {@link #test(List)} to skip tests for multiple tests
+   *
+   * @return The tests p value
+   */
+  default double checkAndTest(List<double @NotNull []> data) {
+    applyPreChecks(data);
```
SteffenHeu, line 39: result of applyPreChecks is ignored. If it is not important, change method to void?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/UnivariateRowSignificanceTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+  @Override
+  public RowSignificanceTestResult test(FeatureListRow row) {
+    final int rowIndex = featureRowIndexMap.get(row);
+    final double[] groupAAbundance = groupAData.getFeatureData(rowIndex, false);
+    final double[] groupBAbundance = groupBData.getFeatureData(rowIndex, false);
+
+    try {
+      final double p = test.checkAndTest(List.of(groupAAbundance, groupBAbundance));
+      return new TTestResult(row, column.getTitle(), p);
```
SteffenHeu: if the test result works for all different tests, we should rename it from ttestresult

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/UnivariateRowSignificanceTest.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+    var that = (UnivariateRowSignificanceTest<?>) obj;
+    return Objects.equals(this.test, that.test) && Objects.equals(this.column, that.column)
+        && Objects.equals(this.groupA, that.groupA) && Objects.equals(this.groupB, that.groupB);
+  }
+
+  public StorableTTestConfiguration toConfiguration() {
+    return new StorableTTestConfiguration(test, column().getTitle(), groupA().toString(),
+        groupB().toString());
+  }
```
SteffenHeu: some for this, should be UnivariateSignificanceTestConfiguration or so

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/anova/AnovaTest.java
```diff
@@ -25,58 +25,61 @@
...
 
-  private static final Logger logger = Logger.getLogger(AnovaTest.class.getName());
-  private final AtomicDouble finishedPercentage = new AtomicDouble(0d);
   private final List<List<RawDataFile>> groupedFiles;
   private final MetadataColumn<?> groupingColumn;
-  private List<AnovaResult> result = List.of();
+  // the data table and its raw data file indexes
+  private final FeaturesDataTable dataTable;
+  private final Map<RawDataFile, Integer> dataTableFileIndexMap;
+  private final Map<FeatureListRow, Integer> rowIndexMap;
```
SteffenHeu: do these need to be fields? They come from the data table and are passed into the same method the data table is passed into

SteffenHeu (commented):
first part

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/utils/scaling/ScalingFunction.java
```diff
@@ -34,10 +35,10 @@
  * statistical analysis. A statistician recommended "AutoScaling" as a default, which scales
  * intensities to the standard deviation of the row.
  */
-public interface ScalingFunction extends Function<RealVector, RealVector> {
+public interface ScalingFunction extends Function<RealVector, RealVector>, DataTableProcessingStep {
```
SteffenHeu, line 38: if the interface can be used to scale a data table and a vector, it's not really clear where it is done, or? The imputer now only works on the data table, so make it similar here?

Overall, i would believe the vector functions to be optimized, so if possible i would prefer to use those.

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/statistics/AbundanceDataTablePreparationConfigParameter.java
```diff
@@ -0,0 +1,66 @@
...
+    getEmbeddedParameters().setAll(measure, imputationFunction);
+  }
+
+  @Override
+  public AbundanceDataTablePreparationConfigParameter cloneParameter() {
+    final AbundanceDataTablePreparationConfigSubParameters embeddedParametersClone = (AbundanceDataTablePreparationConfigSubParameters) getEmbeddedParameters().cloneParameterSet();
+    final AbundanceDataTablePreparationConfigParameter copy = new AbundanceDataTablePreparationConfigParameter(
+        getName(), getDescription());
+    copy.setEmbeddedParameters(embeddedParametersClone);
+    copy.setValue(this.getValue());
```
SteffenHeu: redundant to set both. the setEmbedded internally uses setValue

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotRegionExtractionParameters.java
```diff
@@ -26,64 +27,121 @@
...
+    return 2;
+  }
+
+  @Override
+  public @Nullable String getVersionMessage(int version) {
+    return switch (version) {
+      // this changes behavior a bit but should be fine. There will be very few batches that actually rely on this module.
+      // For a while people can just use the same old version to reproduce old results
+      case 2 ->
+          "Added support for missing value imputation. This has changed the default of missing value imputation to global limit of detection.";
```
SteffenHeu: changed -> introduced

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotRegionExtractionParameters.java
```diff
@@ -26,64 +27,121 @@
...
-      RowSignificanceTestModules.TWO_GROUP_TESTS);
   public static final TTestConfigurationParameter config = new TTestConfigurationParameter(
-      "TTest configuration", "Configure the T-Test.");
+      "TTest configuration", "Configure the t-test.");
 
   public static final RegionsParameter regions = new RegionsParameter();
 
+  public static final AbundanceDataTablePreparationConfigParameter dataPreparationConfig = new AbundanceDataTablePreparationConfigParameter();
+
+  private static final AbundanceMeasureParameter LEGACY_ABUNDANCE_PARAMETER = new AbundanceMeasureParameter(
```
SteffenHeu: not static?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/FoldChangeSignificanceRowFilter.java
```diff
@@ -0,0 +1,193 @@
...
+  public boolean applySignificancePFilter() {
+    return applySignificancePFilter;
+  }
+
+
+  /**
+   * @param rowIndex row index in the data tables
+   * @return true if all requirements are met
+   */
+  public boolean matches(int rowIndex) {
```
SteffenHeu, line 115: there is the rowToIndexMap right? maybe also provide a method where you can just check a FeatureListRow?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/FoldChangeSignificanceRowFilter.java
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * @param dataTable             a data table that is already zero value/missing value imputed
+   *                              etc.
+   * @param grouping              column and two groups
+   * @param maxValueP             maxmimum p value. use <=0 to deactivate
+   * @param minLog2FoldChange     use 0 to deactivate, either signed or absolute value
+   * @param foldChangeFilterSides FC may be applied to Math.abs(value) to filter both up and down
+   *                              regulation or as a signed filter to only retain one side.
+   */
+  public FoldChangeSignificanceRowFilter(FeaturesDataTable dataTable,
```
SteffenHeu, line 66: Would it be better to just give the rows and the config here and let the filter do the data preparation?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RsdFilterParameters.java
```diff
@@ -34,32 +40,93 @@
...
+  @Override
+  public int getVersion() {
+    return 2;
+  }
+
+  /**
+   * Create a filter for a specific data table. Will automatically select only the grouped data
+   * files
+   */
+  RsdFilter createFilter(List<FeatureListRow> rows, List<RawDataFile> rawFiles) {
```
SteffenHeu, line 110: i feel like it should be the other way round. the RSD filter should create a filter when passed a parameter set. otherwise we will bloat the parameter set with more and more methods. same for the fold change filter

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RsdFilterParameters.java
```diff
@@ -34,32 +40,93 @@
...
+import java.util.List;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 public class RsdFilterParameters extends SimpleParameterSet {
 
+
+  public static final AbundanceMeasureParameter abundanceMeasure = new AbundanceMeasureParameter();
+
+  public static final MissingValueImputationParameter missingValueImputation = new MissingValueImputationParameter();
```
SteffenHeu, line 54: should this be an option for the RSD filter? the closer features are to the global LOD, the more likely they are to be retained

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/metadata/Metadata2GroupsSelection.java
```diff
@@ -0,0 +1,104 @@
...
+import io.github.mzmine.util.StringUtils;
+import java.util.List;
+import java.util.Map;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public record Metadata2GroupsSelection(@NotNull String columnName, @NotNull String groupA,
+                                       @NotNull String groupB) {
+
+  public static Metadata2GroupsSelection NONE = new Metadata2GroupsSelection("", "", "");
```
SteffenHeu: final?

mzmine-community/src/test/java/stats/StatsUtilsTest.java
```diff
@@ -72,14 +70,15 @@ void test() {
...
-    final Array2DRowRealMatrix matrix = new Array2DRowRealMatrix(missingData);
-    final RealMatrix zero = StatisticUtils.imputeMissingValues(matrix, false, new ZeroImputer());
-    Assertions.assertEquals(0, zero.getEntry(1, 1));
-    Assertions.assertEquals(0, zero.getEntry(2, 3));
-
-    final RealMatrix oneFifth = StatisticUtils.imputeMissingValues(matrix, false,
-        new OneFifthOfMinimumImputer());
-    Assertions.assertEquals((double) 1 / 5, oneFifth.getEntry(1, 1));
-    Assertions.assertEquals((double) 2 / 5, oneFifth.getEntry(2, 3));
+    // TODO needs tests for DataTable or FeatureDataTable
```
SteffenHeu: ;)

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeaturesDataTable.java
```diff
@@ -0,0 +1,149 @@
...
+
+public final class FeaturesDataTable implements ModifiableDataTable {
+
+  public static final FeaturesDataTable EMPTY = new FeaturesDataTable(List.of(),
+      new FeatureListRowAbundances[0]);
+
+  private final @NotNull List<RawDataFile> dataFiles;
+  private final @NotNull FeatureListRowAbundances[] dataRows;
+
+  public FeaturesDataTable(@NotNull List<RawDataFile> dataFiles,
```
SteffenHeu, line 56: downstream we now always deal with indices and not the actual row. I think this makes it more difficult. i saw at some point that there are rowToIndex maps used externally. Maybe we should just keep track of that in here?

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeaturesDataTable.java
```diff
@@ -0,0 +1,149 @@
...
+   * @return map of the data file to index
+   */
+  public @NotNull Map<RawDataFile, Integer> getDataFileIndexMap() {
+    return CollectionUtils.indexMap(getRawDataFiles());
+  }
+
+  /**
+   * @return map of the row to index
+   */
+  public @NotNull Map<FeatureListRow, Integer> getFeatureRowIndexMap() {
```
SteffenHeu: ah ok the map is created on demand. maybe we create it in any case and allow operations by passing the rows?

SteffenHeu (changes requested):
second part

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxComboBox.java
robinschmid: made it nullable everywhere and also cleaned up the class

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFields.java
robinschmid: added a function with supplier. 
Metadata should get a simple listener

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/DataTable.java
robinschmid: easy method to stream all values. for example if you want to calculate the summaryStats or min, max, mean. 
The DataTableUtils uses for loops to optimize away the stream but in general this can be useful.

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeatureListRowAbundances.java
robinschmid: * @param values             is the feature values for all samples in the order of the samples
   *                           {@link RawDataFile} list of the parent {@link DataTable}

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/SignificanceTest.java
robinschmid: void. 
Was mostly changed to boolean because I wanted to use exhaustive switch

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/UnivariateRowSignificanceTest.java
robinschmid: UnivariateRowSignificanceTestResult

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/UnivariateRowSignificanceTest.java
robinschmid: renamed to UnivariateRowSignificanceTestConfig

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/utils/scaling/ScalingFunction.java
robinschmid: I think it is fine to keep both the apply(RealVector) and the processInPlace / process(DataTable) function?

robinschmid: I will definitely add the centering to the config and already apply it to the data table

robinschmid: I added the centering to the AbundanceDataTablePreparationConfig.
In the PCAUtils I still keep the applyCentering just to be sure the input is centered. Is that ok?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/statistics/AbundanceDataTablePreparationConfigParameter.java
robinschmid: good catch!

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotRegionExtractionParameters.java
robinschmid: true otherwise could give issues

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/FoldChangeSignificanceRowFilter.java
robinschmid: I think it is better to do the preparation outside of the filter. Has nothing to do with the filter itself. 
Also then we would need to pass the featureList, rows, rawFIles, config, maybe some filters

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RsdFilterParameters.java
robinschmid: I prefer it this way because of 3 things:
1. You can more easily generate the filter from the parameterSet.parameter instead of calling a static factory:

```java
parameters.getParameter(someStaticParameter).createFilter();
```

2. all information is in the ParameterSet class and it is easier to reach and it is easier to see if we add a new parameter that we also have to change this method

3. this way the filter class is clean of parameters implementation details and really just shows you what the state and function is

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RsdFilterParameters.java
robinschmid: Good question. 
As far as I read how other tools do this they apply these steps:

1. filter out features > max missing values
2. impute missing values (LOD might help to include more smaller features; while 0-value will give a greater penalty for missing features)
3. apply RSD filter

Especially with the option to impute 0 values you can give greater penalty. Previously missing features would not imply penalty as they were filtered out.

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeaturesDataTable.java
robinschmid: changed by putting the index maps into the FeatureDataTable and now allowing direct access via FeatureListRow and RawDataFile

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeaturesDataTable.java
robinschmid: changed by putting the index maps into the FeatureDataTable and now allowing direct access via FeatureListRow and RawDataFile

mzmine-community/src/test/java/stats/StatsUtilsTest.java
robinschmid: moved to FeatureFindingTest to work on real data

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeatureListRowAbundances.java
SteffenHeu: so the abundances? :D

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/utils/scaling/ScalingFunctions.java
```diff
@@ -25,14 +25,22 @@
...
+
+  public final static ScalingFunctions[] valuesExcludeNone = new ScalingFunctions[]{AutoScaling,
+      ParetoScaling, RangeScaling, MeanCentering};
 
   public ScalingFunction getScalingFunction() {
     return switch (this) {
       case AutoScaling -> new AutoScalingFunction();
       case ParetoScaling -> new ParetoScalingFunction();
       case RangeScaling -> new RangeScalingFunction();
+      case MeanCentering -> new MeanCenterScalingFunction();
```
SteffenHeu, line 45: not sure if i want the centering work as a scaling function. as far as i understood, it is best practice to apply actual scaling before doing a PCA. This may give the impression that it is statistically sound to not apply scaling but just centering.

valuesExcludeNone is used for the pca dashboard. exclude centring from that field?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterParameters.java
```diff
@@ -163,9 +178,20 @@ public RowsFilterParameters() {
     return IonMobilitySupport.SUPPORTED;
   }
 
+  @Override
+  public @Nullable String getVersionMessage(int version) {
+    return switch (version) {
+      case 3 -> """
```
SteffenHeu, line 184: this means that there is an empty line, or?

mzmine-community/src/test/java/stats/StatsUtilsTest.java
```diff
@@ -70,16 +68,4 @@ void test() {
     Assertions.assertEquals(0, unitVar.getEntry(2, 3));
   }
 
-  @Test
-  void testImputation() {
```
SteffenHeu, line 74: imputation test is removed, but the test in the feature finding test does not test the imputation

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeatureListRowAbundances.java
robinschmid: ok makes sense. renamed to abundances

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/utils/scaling/ScalingFunctions.java
robinschmid: Makes sense to remove it from PCA. I added it to the scaling functions enum as this was also present in other tools in the same selection. Like MetaboAnalyst also lists None and Mean centering there - but that is the preprocessing for many different tools so makes sense that they put it. 

I removed centering from the PCA options

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterParameters.java
robinschmid: What do you mean? Looks good to me:

```java
return switch (version) {
      case 3 -> """
          "%s" has changed internally. Missing value imputation was added.
          "%s" was added as an additional filtering option.""".formatted(cvFilter.getName(), foldChangeFilter.getName());
      default -> null; };
```

mzmine-community/src/test/java/stats/StatsUtilsTest.java
robinschmid: only this one is tested:
GLOBAL_LIMIT_OF_DETECTION

Will add tests for a SimpleDataTable for each algorithm

robinschmid: added tests for all ImputationFunctions and ScalingFunctions

SteffenHeu (approved):

PR #2724, opened 2025-07-15
Minimum samples filter also for metadata groups

robinschmid:
- adds minimum samples filter for groups
- harmonizes the filter against all raw data files
- improves layout alignment in VolcanoPlot

Relies on the other PR


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterParameters.java
```diff
@@ -59,6 +60,9 @@ public class RowsFilterParameters extends SimpleParameterSet {
   public static final OptionalParameter<MinimumSamplesParameter> MIN_FEATURE_COUNT = new OptionalParameter<>(
       new MinimumSamplesParameter(), false);
 
+  public static final OptionalParameter<MinimumSamplesInMetadataParameter> MIN_FEATURE_IN_GROUP_COUNT = new OptionalParameter<>(
```
SteffenHeu, line 63: disable when the parameter was not loaded

robinschmid: good catch. 
Also applied the same now to the RSD filter and significance filter

mzmine-community/src/main/java/io/github/mzmine/parameters/CompositeParametersParameter.java
```diff
@@ -0,0 +1,59 @@
...
+package io.github.mzmine.parameters;
+
+import javafx.scene.Node;
+import org.w3c.dom.Element;
+
+/**
+ * Parameter that combines multiple parameters that are only internally used to handle sub values,
+ * create components, save, load values...
+ */
+public abstract class CompositeParametersParameter<ValueType, EditorComponent extends Node> extends
```
SteffenHeu, line 36: not through the full pr yet, but does it make sense to implement `EmbeddedParameterSet`?

mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterUtils.java
```diff
@@ -439,4 +443,61 @@ public static <M extends MZmineProcessingModule, S extends MZmineProcessingStep<
     return FileAndPathUtil.getMajorityFilePath(
         allImportedFiles.stream().map(File::getParentFile).toList());
   }
+
+
+  /**
+   * Loads parameters in order. Can be useful to save and load parameters in order when designing
+   * other composite parameters.
+   */
+  public static void loadParametersInOrder(Element xmlElement, Parameter<?>... parameters) {
```
SteffenHeu: I think depending on the order of the parameters is not a good idea, that way we cannot remove or add a new parameter. Why hot use the logic of the general parameter set?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/MinimumSamplesInMetadataComponent.java
```diff
@@ -0,0 +1,68 @@
...
+    if (value == null) {
+      metadata.setValue("");
+      minSamples.setValue(MinimumSamplesFilterConfig.DEFAULT.minSamples());
+      return;
+    }
+    metadata.setValue(value.columnName());
+    minSamples.setValue(value.minSamples());
+  }
+
+  public MinimumSamplesFilterConfig getValue() {
```
SteffenHeu, line 68: notnull?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/MinimumSamplesInMetadataComponent.java
```diff
@@ -0,0 +1,68 @@
...
+    super(minSamples, metadataBox);
+    this.minSamples = minSamples;
+    this.metadata = metadata;
+    setVgap(FxLayout.DEFAULT_SPACE);
+    setHgap(FxLayout.DEFAULT_SPACE);
+
+    setValue(value);
+  }
+
+  public void setValue(MinimumSamplesFilterConfig value) {
```
SteffenHeu: nullable

mzmine-community/src/main/java/io/github/mzmine/parameters/CompositeParametersParameter.java
robinschmid: My idea was to hide the internal parameter set and make this only protected. 

The internal parameterset is mostly there so that you can compose multiple parameters into a single parameter with a single return value like a filter. 
Then we can reuse the logic for load and save.

robinschmid: It would be hard to also get the value from the internal parameters as we do not have static parameters

mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterUtils.java
robinschmid: yeah makes sense my thought process was to easy. 
Now just uses the same method to save and load parameters like SimpleParameterSet in ParameterUtils

PR #2738, opened 2025-07-17
Add simple string mapper to DataType 

SteffenHeu:
Allows mapping from DataType values from simple strings as commonly used in CSV values. Not every type must implement this function.
Also adds UniqueIdSupplier to EnumDataType

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/otherdectectors/PolarityTypeType.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/otherdectectors/PolarityTypeType.java
@@ -49,7 +49,7 @@ public class PolarityTypeType extends EnumDataType<PolarityType> implements Null
     NoTextColumn {
 
   public static final Function<@Nullable String, @Nullable PolarityType> mapper = s -> {
-    if (s == null) {
+    if (s == null || s.isBlank()) {
       return null;
     }
     return PolarityType.parseFromString(s);


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/abstr/EnumDataType.java
```diff
@@ -43,4 +58,40 @@ public T[] enumValues() {
...
+      writer.writeCharacters(idSupplier.getUniqueID());
+    } else {
+      throw new IllegalArgumentException(
+          "Wrong value type for data type: " + getUniqueID() + " value class: " + value.getClass());
+    }
+  }
+
+  @Override
+  public @Nullable Function<@Nullable String, @Nullable T> getMapper() {
+    return s -> UniqueIdSupplier.parseOrElse(s, enumValues(), null);
```
robinschmid, line 95: maybe make this function a private final field?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/abstr/StringType.java
```diff
@@ -83,4 +84,8 @@ public Object loadFromXML(@NotNull XMLStreamReader reader, @NotNull MZmineProjec
     return text;
   }
 
+  @Override
+  public @Nullable Function<@Nullable String, @Nullable String> getMapper() {
+    return s -> s;
```
robinschmid: public static final field?

mzmine-community/src/main/java/io/github/mzmine/datamodel/PolarityType.java
```diff
@@ -25,14 +25,15 @@
...
+import io.github.mzmine.datamodel.utils.UniqueIdSupplier;
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 /**
  * Represents the polarity of ionization.
  */
-public enum PolarityType {
+public enum PolarityType implements UniqueIdSupplier {
```
robinschmid, line 36: 1. In PolarityTypeType, overwrite the getMapper and point it to PolarityType.parseFromString to allow more string values. 

2. in parseFromString, add missing cases that match unique ID:
```java
case "any" -> ANY;
case "neutral", "n" -> NEUTRAL;
case "unknown" -> UNKNOWN;
```

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/MobilityUnitType.java
```diff
@@ -124,4 +100,15 @@ public Object evaluateBindings(@NotNull BindingsType bindingType,
...
   }
+
+  @Override
+  public @Nullable Function<@Nullable String, @Nullable MobilityType> getMapper() {
+    return s -> {
+      if (s == null) {
+        return null;
+      }
+      return Arrays.stream(MobilityType.values()).filter(t -> t.getUnit().equals(s)).findFirst()
+          .orElse(null);
```
robinschmid: why map by the unit which is not unique - only because the getFormattedString uses unit?
Maybe then safer to return null for the mapper?

If we keep a mapper, then it should definitely also handle conversion from the unique ID enum. Maybe through a static parseFromString method in MobilityUnit and handle multiple cases there like it is done in PolarityType?

robinschmid: but still as the toFormattedString is not unique this might be better null

robinschmid (commented):
minor things

Nice with the test.

robinschmid (approved):

PR #2746, opened 2025-07-21
Module to Merge spectral libraries while retaining metadata

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/MergeLibrariesModule.java
```diff
@@ -0,0 +1,73 @@
...
+import io.github.mzmine.util.MemoryMapStorage;
+import java.time.Instant;
+import java.util.Collection;
+import javax.print.attribute.standard.MediaSize.NA;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class MergeLibrariesModule implements MZmineProcessingModule {
+
+  public static final String NAME = "Merge spectral libraries";
```
robinschmid: Maybe change name to: Merge & export spectral libraries
Then it will also easily show up in the quick search wiht the export keyword. 
Also then it is more clear that you can also just export libraries?

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/MergeLibrariesTask.java
```diff
@@ -0,0 +1,140 @@
...
+      error("No spectral libraries selected.");
+      return;
+    }
+
+    try {
+      newFile.getParentFile().mkdirs();
+    } catch (final Exception e) {
+      logger.log(Level.SEVERE, "Failed to create directory", e);
+      error(e.getMessage());
+      return;
```
robinschmid, line 121: Use utility method
```java
    if (!FileAndPathUtil.createDirectory(newFile.getParentFile())) {
      error("Cannot create directory %s.".formatted(newFile.getParentFile().getAbsolutePath()));
      return;
    }
```

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/MergeLibrariesTask.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    final AtomicInteger entryId = new AtomicInteger(0);
+    try (var w = Files.newBufferedWriter(newFile.toPath(), WriterOptions.REPLACE.toOpenOption())) {
+      for (final SpectralLibrary lib : libs) {
+        for (SpectralLibraryEntry entry : lib.getEntries()) {
+          final SpectralDBEntry copy = new SpectralDBEntry((SpectralDBEntry) entry);
+          copy.putIfNotNull(DBEntryField.ENTRY_ID, entryId.incrementAndGet());
```
robinschmid: only change the ENTRY_ID if user demands this? 
Like by adding another boolean parameter?

ENTRY_ID can be something meaningful like the GNPS ID or similar from other libraries.

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/MergeLibrariesTask.java
```diff
@@ -0,0 +1,140 @@
...
+          exportedEntries++;
+        }
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+
+    if (removeAndImport) {
+      FxThread.runLater(() -> ProjectService.getProject()
+          .removeSpectralLibrary(libs.stream().toArray(SpectralLibrary[]::new)));
```
robinschmid: collection.stream.toArray should be
collection.toArray

Here `libs.toArray(SpectralLibrary[]::new)`

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/SpectralLibrarySelectionType.java
```diff
@@ -41,4 +45,12 @@ public String toString() {
     return stringValue;
   }
 
+  @Override
+  public @NotNull String getUniqueID() {
+    return switch (this) {
+      case ALL_IMPORTED -> "All imported libraries";
+      case SPECIFIC -> "Specific libraries";
```
robinschmid: The uniqueID should be the previous name and not the toString to match the batch files. 
No preference really but
```java
case ALL_IMPORTED -> "ALL_IMPORTED";
case SPECIFIC -> "SPECIFIC";
case AS_SELECTED_IN_MAIN_WINDOW -> "SELECTED_IN_MAIN_WINDOW";
```

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/MergeLibrariesParameters.java
```diff
@@ -0,0 +1,62 @@
...
+import io.github.mzmine.parameters.parametertypes.selectors.SpectralLibrarySelectionParameter;
+import io.github.mzmine.parameters.parametertypes.selectors.SpectralLibrarySelectionType;
+import java.util.List;
+
+public class MergeLibrariesParameters extends SimpleParameterSet {
+
+  public static final SpectralLibrarySelectionParameter speclibs = new SpectralLibrarySelectionParameter(
+      new SpectralLibrarySelection(SpectralLibrarySelectionType.AS_SELECTED_IN_MAIN_WINDOW, List.of()));
+
+  public static final FileNameParameter newLibraryFile = new FileNameParameter(
```
robinschmid: could use the suffix export parameter

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/MergeLibrariesTask.java
SteffenHeu: mhh true, but we only give it a "regular" number when exporting, which would lead to duplicate IDs. 
How about indexing the existing IDs first, and in case of duplicates applying just a prefix like this:  ´oldLibName_oldID´

robinschmid: This could also be interesting. 
I think giving an option to just renumber the ENTRY_ID should be good as well.

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/MergeLibrariesParameters.java
SteffenHeu: thinking about it: when people export a library they probably just want to put the library name and not just give a suffix, or?

robinschmid: I think in those cases they will not use the redefine outputs method and the suffix is only applied then. 
So I think using the suffix parameter just provides a save filename suffix in case people use this module in their workflow and want to redirect the output somewhere.

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/MergeLibrariesTask.java
```diff
@@ -0,0 +1,175 @@
...
+
+    try (var w = Files.newBufferedWriter(newFile.toPath(), WriterOptions.REPLACE.toOpenOption())) {
+      for (final SpectralLibrary lib : libs) {
+        for (SpectralLibraryEntry entry : lib.getEntries()) {
+
+          final String currentId = entry.getOrElse(DBEntryField.ENTRY_ID, null);
+          final boolean isDuplicate = currentId == null || duplicateIds.contains(currentId);
+
+          final String newEntryId = idHandling.getNewEntryId(entry, isDuplicate,
+              () -> String.valueOf(entryId.incrementAndGet()));
```
robinschmid: Best to add here also a suffix. 
Otherwise it is harder to do substring searches later. 
Something like:
libraryname_1_id

Otherwise libraryname_1 would match against many other library IDs

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/IdHandlingOption.java
```diff
@@ -0,0 +1,79 @@
...
+    };
+  }
+
+
+  @Override
+  public String toString() {
+    return switch (this) {
+      case KEEP_ALL -> "Keep existing IDs";
+      case AVOID_DUPLICATES -> "Avoid duplicates";
+      case NEW_ID_WITH_LIBRARY_NAME -> "New IDs with old library name";
```
robinschmid, line 76: I think I would add an option to just create new ids without the old library name. So just a renumbering like:  
newname_1_id

But we can add this later as well

PR #2751, opened 2025-07-21
Redefine outputs now keeps previous file format and adds format automatically

### LINKED ISSUE #2717: Low impact - ¿Easy Fix? - Batch mode/Redefine outputs/missing extension
### Extension ".csv" missing when "redefine outputs" it's used on an existing batch.

* Win11, i12700k, 32 GB RAM.
* My mzmine version: 4.7.8

### What happened

Normally, when using "mzwizard" to create a batch, the "path output" is also automatically configured. However, when modifying the path of an existing batch with "redefine outputs," some file names do not end with the default extension before "redefine outputs."

For example, when creating a batch for UHPLC/Orbitrap in DDA, parameters are created based on given conditions. By default, the "export all annotations to CSV file" section is created, and thus the created file has a ".csv" extension. When modifying the "output" using "redefine outputs" for that same batch, the path of the aforementioned file name does change, but it is saved without an extension (the ".csv" extension is not created).

As far as you may have noticed, this only happens with the annotations file and the file to be exported to "MetaboAnalyst."

The problem is solved by manually adding the extension to the file once it's created, but adding that ".csv" to these two sections would save a small step.

I've attached screenshots; I don't think adding a batch file is necessary. Cheers!
New batch created:
Image
Batch modified using "redefine output"
Image

robinschmid:
fixes #2717

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_xml/ExportFeaturesDataParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_xml/ExportFeaturesDataParameters.java
@@ -1,3 +1,28 @@
+/*
+ * Copyright (c) 2004-2025 The mzmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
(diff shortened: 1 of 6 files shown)


robinschmid:
> Looks good, also use in ProjectExport task?

Already uses the correct parameter type without a suffix as the file format is specific in this case:
```java
public static final FileNameSuffixExportParameter projectFile = new FileNameSuffixExportParameter(
      "Project file", "File name of project to be saved", extensions, null);
```

PR #2759, opened 2025-07-23
Retention time calibration adjustments

robinschmid:
- fail early if there are no RT correction candidates
- stable coloring 
- different shapes to differentiate the series better, great if the colors start repeating themselves
A couple of examples:
image
image
image


mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/JFreeChartUtils.java
```diff
@@ -25,12 +25,229 @@
...
+        // vertical rectangle
+        new Rectangle2D.Double(-delta / 2, -delta, size / 2, size),
+        // horizontal ellipse
+        new Ellipse2D.Double(-delta, -delta / 2, size, size / 2),
+        // right-pointing triangle
+        createTriangle(size, TriangleDirection.RIGHT),
+        // left-pointing triangle
+        createTriangle(size, TriangleDirection.LEFT),
+        // star shape
+        createDefaultStar(size, 0, 0),
```
SteffenHeu: to the bottom or remove? :D

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_rtcalibration2/ScanRtCorrectionTask.java
```diff
@@ -390,6 +394,14 @@ public void run() {
     goodStandards.sort(Comparator.comparingDouble(rtMeasure::getRt));
     final List<RtStandard> monotonousStandards = removeNonMonotonousStandards(goodStandards,
         referenceFlistsByNumRows, rtMeasure);
+
+    if (monotonousStandards.isEmpty()) {
+      logger.warning(
+          "No monotonous standards found. No retention time correction will be appplied. The task finishes with success to not break batch processing.");
```
SteffenHeu: Maybe use the DialogLoggerUtil here, so GUI users get better feedback? Can use the non blocking one

SteffenHeu (commented):
minor only

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_rtcalibration2/ScanRtCorrectionTask.java
robinschmid: Added a auto closing dialog here. 

And also added a TextFlow to the top of AbstractPreviewPane to allow error messages:

### RT width too narrow
image

### parameters not matching
image

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/JFreeChartUtils.java
robinschmid: Moved stars all the way down so should only appear when there are 14 datasets:
Also made them a bit smaller. I think I like the stars more than the hexagons as they are at least easy to spot. But for me good like this:

image

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/SimpleXYChart.java
```diff
@@ -326,6 +326,13 @@ public void addDatasets(Collection<? extends ColoredXYDataset> datasets) {
     });
   }
 
+  public void setDatasets(@NotNull List<@NotNull DatasetAndRenderer> datasets) {
+    applyWithNotifyChanges(false, () -> {
+      removeAllDatasets();
+      datasets.forEach(ds -> addDataset(ds.dataset(), ds.renderer()));
```
SteffenHeu, line 332: since foreach is not blocking, is it the best method to execute this on the java fx thread?

PR #2762, opened 2025-07-24
More options how to renumber IDs

robinschmid:
- Added options for renumbering
- moved duplicate handling into the function to avoid endless loop where the libName_original ID is actually a duplicate

Also changed the use of library.getName to not contain the size of the library. This also removes the size from the filename if library is converted to LibraryRawDataFIle. The feature list still contains the size. 
Reason is to have better ID and USI from filename.

image


mzmine-community/src/main/java/io/github/mzmine/modules/io/export_merge_libraries/IdHandlingOption.java
```diff
@@ -41,29 +47,45 @@ public enum IdHandlingOption implements UniqueIdSupplier {
       case KEEP_ALL -> "keep_all";
       case AVOID_DUPLICATES -> "avoid_duplicates";
       case NEW_ID_WITH_LIBRARY_NAME -> "new_id_with_library_name";
+      case RENUMBER_WITH_DATASET_ID -> "renumber_with_dataset_id";
+      case RENUMBERED_WITH_FILENAME -> "renumbered_with_filename";
```
SteffenHeu: call these NEW_ID_ instead of renumber or rename NEW_ID_WITH_LIBRARY_NAME to renumber? so we have the same pattern

SteffenHeu: the new options also need a description in the parameter

robinschmid: image

PR #2763, opened 2025-07-25
Fix metadata reader for date columns

robinschmid:
- if one value in a column was a date the column was read as a date column
- was expecting an exception but new date parser was not throwing


mzmine-community/src/test/resources/metadata/metadata_wide_type_mismatch.tsv
```diff
@@ -0,0 +1,6 @@
+Filename	ATTRIBUTE_Group	ATTRIBUTE_NumberCol	ATTRIBUTE_Group2	run_date	sample_id	date2
+a.mzML	A	12	A	2021-08-31T15:33:15	A	20240125
+b.mzML	A	0.5	B	NODATE	20241005	2025-10-30
+c.mzML	c	NONUMBER	B	2021-08-31T19:23:46	1	30.12.2020
```
SteffenHeu: maybe test the new parser also with an empty field?
i can see in ProjectMetadataColumnParameters that empty strings are not passed currently, but we may always change in the future. 
Current DateTimeUtils.parse would return null for an empty field and  date column would then throw. -> column will be parsed as text.

robinschmid: empty values are covered in some of the other tests but added an empty value here to the last date2 column.
And more checks to the conversion.

PR #2767, opened 2025-07-28
Auto update selected FeatureLists, RawDataFile, Libraries in components

robinschmid:
- listen to changes to selected feature lists etc without listener to not create potential leak
- auto shuts down listener when the dialog is closed
- only updates the component and keeps a ReadOnlyProperty of the currently selected files/featurelists/libraries
- triggers updates to previews like retention time correction
- mass detection, local minimum and other also get that update but they do not use the selected files or featurelists but all - so does not matter there

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/FeatureListsComponent.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/FeatureListsComponent.java
@@ -34,7 +34,10 @@ import io.github.mzmine.parameters.parametertypes.MultiChoiceParameter;
 import io.github.mzmine.parameters.parametertypes.StringParameter;
 import io.github.mzmine.project.ProjectService;
 import io.github.mzmine.util.ExitCode;
+import java.util.List;
 import javafx.animation.PauseTransition;
+import javafx.beans.property.ReadOnlyObjectProperty;
+import javafx.beans.property.ReadOnlyObjectWrapper;
 import javafx.collections.FXCollections;
 import javafx.scene.control.Button;
 import javafx.scene.control.ComboBox;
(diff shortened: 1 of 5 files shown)


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilesComponent.java
```diff
@@ -47,6 +52,7 @@ public class RawDataFilesComponent extends GridPane {
   private final Button detailsButton;
   private final Label numFilesLabel;
   private RawDataFilesSelection currentValue = new RawDataFilesSelection();
```
SteffenHeu: annotate notnull? used in here as if not null

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/FeatureListsComponent.java
```diff
@@ -49,6 +54,7 @@ public class FeatureListsComponent extends HBox {
   private final Button detailsButton;
   private final Label numPeakListsLabel;
   private FeatureListsSelection currentValue = new FeatureListsSelection();
```
SteffenHeu: annotate notnull? used in here as if not null but can be set to null in setValue

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilesComponent.java
```diff
@@ -102,22 +107,35 @@ public RawDataFilesComponent() {
...
+      updateNumFiles();
+      autoUpdate.playFromStart();
+    });
+    autoUpdate.playFromStart();
   }
 
   void setValue(@Nullable RawDataFilesSelection newValue) {
-    if(newValue==null) {
+    if (newValue == null) {
       currentValue = null;
```
SteffenHeu: used in updateNumFiles as if notnull

SteffenHeu (dismissed):

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/FeatureListsComponent.java
robinschmid: Somehow handled as NotNull but the setValue did set this to null. 
Changed and made NotNull with the default selection. 
I guess this value never reached null before.

PR #2771, opened 2025-07-29
Add min samples in one group parameter

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/MinimumSamplesFilter.java
```diff
@@ -39,11 +39,12 @@
  *
  * @param minSamples   min samples
  * @param column       the column or null
+ * @param group        if present then only filter for this group
  * @param files        all data files
  * @param groupedFiles grouped by column or null if there is no column
  */
 public record MinimumSamplesFilter(@NotNull AbsoluteAndRelativeInt minSamples,
-                                   @Nullable MetadataColumn<?> column,
+                                   @Nullable MetadataColumn<?> column, @Nullable String group,
```
SteffenHeu: The matches method needs to be updated or not?

SteffenHeu: Ah ok, it is managed by the config. I think the usage of this filter class just became confusing, because now the groupedFiles are either all groups or a single group. 

Currently it is possible to create the filter with an invalid (?) configuration, e.g. giving a group but no groupedFiles or the other way round. The group is not used inside this filter, but i agree we should have it to know whats going on. 

My suggestion would be to override the constructor (is it possible for records?) and add checks that if group != null then groupedFiles.size == 1 and groupedFiles != null. Or make the constructor package private (again does that work for records?), so the filter can only be created by a valid config.

SteffenHeu (commented):
usability enhancements

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/MinimumSamplesFilter.java
robinschmid: changed to a class now with package private constructor

PR #2776, opened 2025-07-29
fix feature values null in MsnTreeFeatureDetectionTask

SteffenHeu:
odd coincidence in the msn tree feature detection:
it is possible that the precursor ion is not found in the ms2 spectra at all but the ms2s still contain information

in that case you cannot open the feature table because height is null and we sort by height

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_msn_tree/MsnTreeFeatureDetectionTask.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_msn_tree/MsnTreeFeatureDetectionTask.java
@@ -113,8 +113,7 @@ public class MsnTreeFeatureDetectionTask extends AbstractTask {
         ScanUtils.getMSnFragmentTrees(dataFile, mzTol));
     trees.sort(Comparator.comparingDouble(PrecursorIonTree::getPrecursorMz));
     List<Range<Double>> mzRangesSorted = trees.stream()
-        .mapToDouble(PrecursorIonTree::getPrecursorMz)
-        .mapToObj(mzTol::getToleranceRange).toList();
+        .mapToDouble(PrecursorIonTree::getPrecursorMz).mapToObj(mzTol::getToleranceRange).toList();
 
     extractorFunction = new ExtractMzRangesIonSeriesFunction(dataFile, scanSelection,
         mzRangesSorted, ScanDataType.MASS_LIST, this);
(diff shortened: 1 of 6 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_msn_tree/MsnTreeFeatureDetectionTask.java
```diff
@@ -140,6 +139,11 @@ public void run() {
       // need to set mz if data was empty
       if (!hasData) {
         f.setMZ(mstree.getPrecursorMz());
+        f.setHeight(0f);
+        f.setArea(0f);
+        f.setRT(0f);
+        // should we discard the feature in that case? it is possible that the precursor ion is
+        // not found in the ms2 spectra at all but the ms2s still contain information
```
robinschmid, line 150: the eic is build from MS1 spectra usually. Comment points at MS2 spectra. 

I think this could point to an issue that the mz tolerance is too narrow or that the precursor mz values are not recalibrated but the spectra are.  

I would keep the feature as it still has the mz and the MS2 / MSn spectra. 

Maybe we can just log a potential issue? Maybe just set a boolean state and just log it once at the end of the task that there were some empty chromatograms and this may mean the mz tol is too narrow or the mz are not calibrated

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_msn_tree/MsnTreeFeatureDetectionTask.java
```diff
@@ -140,6 +139,11 @@ public void run() {
       // need to set mz if data was empty
       if (!hasData) {
         f.setMZ(mstree.getPrecursorMz());
+        f.setHeight(0f);
+        f.setArea(0f);
+        f.setRT(0f);
```
robinschmid, line 147: RT could be set to one of the MS2 scans RT

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_msn_tree/MsnTreeFeatureDetectionTask.java
robinschmid: maybe we could show these messages that are not real errors as a notification in the right corner instead of a dialog:
```java
NotificationService.show(NotificationType.INFO,  ....
```

robinschmid (approved):

PR #2795, opened 2025-08-04
Scan visualizer show original spectrum instead of denormalized

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/datamodel/impl/masslist/SimpleMassList.java
```diff
@@ -115,6 +120,19 @@ public static MassList loadFromXML(XMLStreamReader reader, @Nullable MemoryMapSt
...
+    return new SimpleMassList(storage, spec.mzs(), spec.intensities());
+  }
+
+  /**
+   * Expects to still have the reader at the xml element of this mass list with the local name
+   * {@link #XML_ELEMENT}
+   *
+   * @return the spectral arrays
+   */
+  protected static @NotNull SimpleSpectralArrays loadSpectralArrays(XMLStreamReader reader)
```
SteffenHeu: not a fan of extracting a method for this. This method working relies on the writer still being at the start position of the element and then reads all sub elements until the end of the XML_ELEMENT has been reached. But it only processes two. 
so as soon as we add a new element that is not related to the spectral arrays, we also need to put it into this method any way or that we need to remove this method then.

copying the load intensities/mzs in sub methods is not the nicest way, but at least it's still extensible.

robinschmid: True. I thought it is better to avoid code duplication. 
I reverted the change and copied the code to SimpleFactorMassList.loadFromXML

robinschmid:
Also not sure. Was just the best solution I could think of now. 
I mean we definitely need to keep the factor of the normalization somehow and I thought the mass list is the best place. 
We could also put the factor in every mass list but I think having a separate class is better?

SteffenHeu:
After discussion: Denormalization in mass detection stage is a problem for DIA workflows. Better move to scan merging step. (make sure it is not used for merging of ims files in dia step)

PR #2796, opened 2025-08-04
FeatureTableFX, better graphical column width on first show

robinschmid:
- fixed min width for graphical column types to have them show in full size when they are first added 

Maybe good to try other graphical types and charts in the table how they behave


SteffenHeu:
I feel like the issue is something specific with the feature shape chart.
When i process an IMS dataset and enable the shape columns, the mobilograms are shown correctly out of the box:
image

robinschmid:
should work better with the preferred size and new PR for graphical columns

PR #2805, opened 2025-08-10
Feature table rework + annotation filters

robinschmid:
- FXML --> MVC
- Moved menu to the bottom
- Added link to documentation, parameter dialog, quick column selection
- Removed old flexible filter combobox based on data types.
- Added RowTypeFilter to allow custom filters

## TODO

- [x] add filter to RowsFilterParameters to allow similar filtering

## added filters

Annotation filters go through all FeatureAnnotations and sometimes specific columns. 

- substructure matches by SMILES, INCHI, SMARTS
- formula, e.g.: >C6Br
- formula ranges: C6Br - C20Br3
- Lipids many formats see below
- Ion identity ID
- Adducts

## Lipid filter

*   PC only define class
*   C for any lipid class
*   C20 for 20 C and any number of double bonds and oxygen
*   C20:2 for 20 C and 2 double bonds
*   PC20:2 also specifying the lipid class as PC
*   C>20:2 for more than 20 C and exactly 2 double bonds
*   C>20:>2 also more than 2 double bonds
*   PC18:2_18:0 defining chains
*   PC>18:>2_>18:0 defining ranges as well
*   Ranges PC20:2 - PC40:6

## Examples
image

### Any lipid
image

### class
image

### double bonds
image

### All oxidized:
image
image

## Examples substructure amide
image

## Example Formula range
image


javafx-framework/src/main/java/io/github/mzmine/javafx/components/converters/OptionsStringConverter.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.Collection;
+import javafx.scene.control.ComboBox;
+import javafx.util.StringConverter;
+
+/**
+ * Uses the options of a combobox or other list of options to convert to actually selected element.
+ * <p>
+ * The issue is that if a ComboBox is editable then the text entered is used as value and not the
+ * actual value object that may be selected. Need to set the string converter in this case.
```
SteffenHeu: Add: Note: only required for combo boxes that do not contain < String> as type

javafx-framework/src/main/java/io/github/mzmine/javafx/components/converters/OptionsStringConverter.java
```diff
@@ -0,0 +1,69 @@
...
+    return t == null ? "" : t.toString();
+  }
+
+  @Override
+  public T fromString(String s) {
+    for (T option : options) {
+      if (option == null) {
+        continue;
+      }
+      if (option.toString().equals(s)) {
```
SteffenHeu, line 69: in some cases the options don't cover the input value. so we have to make sure that this converter is not applied to all combo boxes. e.g. the metadata group selection would fail if this is applied, because in a batch the metadata is not import -> no options present

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxComboBox.java
```diff
@@ -124,8 +132,23 @@ public static <T> ComboBox<T> newAutoCompleteComboBox(@Nullable String tooltip,
...
    * @return AutoCompletionBinding of the string representation
    */
-  public static AutoCompletionBinding<String> bindAutoCompletion(ComboBox<?> combo) {
+  public static <T> AutoCompletionBinding<T> bindAutoCompletion(ComboBox<T> combo,
+      boolean autoGrow) {
+    if (autoGrow) {
+      // auto grow with input
+      applyAutoGrow(combo);
+    }
+    getSetEditable(combo);
```
SteffenHeu: see comment in converter.

javafx-framework/src/main/java/io/github/mzmine/javafx/components/formatters/FormatDoubleRangeStringConverter.java
```diff
@@ -0,0 +1,214 @@
...
+ */
+public class FormatDoubleRangeStringConverter extends StringConverter<Range<Double>> {
+
+  private final NumberFormat numberFormat;
+
+  /**
+   * Constructs a {@code FormatDoubleStringConverter} with the default locale and format.
+   */
+  public FormatDoubleRangeStringConverter() {
+    this(Locale.getDefault());
```
SteffenHeu, line 52: we set the default locale to english, right?

javafx-framework/src/main/java/io/github/mzmine/javafx/validation/IconValidationDecoration.java
```diff
@@ -0,0 +1,119 @@
...
+import org.kordamp.ikonli.javafx.FontIcon;
+
+public class IconValidationDecoration extends AbstractValidationDecoration {
+
+  private static final Image REQUIRED_IMAGE = new Image(
+      org.controlsfx.validation.decoration.GraphicValidationDecoration.class.getResource(
+              "/impl/org/controlsfx/control/validation/required-indicator.png")
+          .toExternalForm()); //$NON-NLS-1$
+
+  private static final String SHADOW_EFFECT = "-fx-effect: dropshadow(three-pass-box, rgba(0,0,0,0.8), 10, 0, 0, 0);"; //$NON-NLS-1$
```
SteffenHeu: could these be added to the css?

javafx-framework/src/main/java/io/github/mzmine/javafx/validation/TooltipFixGraphicDecoration.java
```diff
@@ -0,0 +1,71 @@
...
+import org.controlsfx.control.decoration.GraphicDecoration;
+
+/**
+ * Tooltip of control hides tooltip of error message this class fixes this
+ */
+public class TooltipFixGraphicDecoration extends GraphicDecoration {
+
+  private Tooltip savedTooltip;
+
+  public TooltipFixGraphicDecoration(Node decorationNode) {
```
SteffenHeu, line 44: is this the node to decorate or the node that is shown as error indicator?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/statsdashboard/StatsDashboardController.java
```diff
@@ -57,7 +58,8 @@ public class StatsDashboardController extends FxController<StatsDashboardModel>
 
   public StatsDashboardController(FeatureTableFX table) {
     super(new StatsDashboardModel());
-    this.table = table == null ? new FeatureTableFX() : table;
+    final FxFeatureTableController tableController = new FxFeatureTableController();
+    this.table = table == null ? tableController.getFeatureTable() : table;
```
SteffenHeu: if it is actually nullable, can you please annotate it as such? (also in the model if not done there)

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterParameters.java
```diff
@@ -167,6 +168,9 @@ public class RowsFilterParameters extends SimpleParameterSet {
       "If checked, all rows with MS2 are retained without applying any further filters on them.",
       true);
 
+  public static final OptionalParameter<RowTypeFilterParameter> ROW_TYPE_FILTER = new OptionalParameter<>(
```
SteffenHeu, line 171: disable if not loaded

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/row_type_filter/filters/FormulaRangeRowTypeFilter.java
```diff
@@ -0,0 +1,119 @@
...
+   * @param query format is min formula - max formula like: C10H20Br2 - C20Br4
+   */
+  FormulaRangeRowTypeFilter(@NotNull RowTypeFilterOption selectedType,
+      @NotNull MatchingMode matchingMode, @NotNull String query) {
+    super(selectedType, matchingMode, query);
+
+    final String[] split = query.split("-");
+    if (split.length != 2) {
+      throw new QueryFormatException(
+          query + " is not a valid formula range, e.g., C10H20Br2 - C20Br4");
```
SteffenHeu, line 58: Is the example even valid? No Hydrogens in the upper bound formula

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxComboBox.java
robinschmid: changed javadoc here. We cannot know what T is here and the StringConverter works for all types:

  /**
   * Also sets a {@link OptionsStringConverter} to be sure that values are converted properly. This
   * is only needed for combobox of type different from String but works also for String ComboBoxes.
   * At this point we do not know the type of T.
   */

javafx-framework/src/main/java/io/github/mzmine/javafx/components/formatters/FormatDoubleRangeStringConverter.java
robinschmid: MZmineCore.init has

    Locale.setDefault(new Locale("en", "US"));

javafx-framework/src/main/java/io/github/mzmine/javafx/components/converters/OptionsStringConverter.java
robinschmid: Good catch. 
I changed this now back to TextField. 
ComboBox behaves weird if the text in the editor is not a valid option in the combobox. 
The textfield has now advanced autoComplete that opens directly on focus:
image
image

javafx-framework/src/main/java/io/github/mzmine/javafx/validation/TooltipFixGraphicDecoration.java
robinschmid: added javadoc this is the decoration itself that is used on a target

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/statsdashboard/StatsDashboardController.java
robinschmid: this was a weird remnant from the inital code. Another constructor was calling this constructor with a null table. 
The controller has @NotNull on the table. 

I also removed the other constructor as it was not used and added the full table view with filters to the stats dashboard

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/row_type_filter/filters/FormulaRangeRowTypeFilter.java
robinschmid: Yes, the idea is that both formulas are completely disconnected from each other:
min formula - max formula

All elements need to be between these two. min only defines the minimum atom counts and max only the maximum atom counts. This is useful if you dont want an upper or lower bound for specific elements

javafx-framework/src/main/java/io/github/mzmine/javafx/validation/IconValidationDecoration.java
robinschmid: Added to css used the same names that were used by jabref.
The original ones where hard to read though. white text on red background. Here is new:

image
image

PR #2817, opened 2025-08-15
Imrove scan selection and other embedded parameters

robinschmid:
- allow using an external parameterset to set the component or get the value from it

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/ParameterSetupPane.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/ParameterSetupPane.java
@@ -354,6 +354,13 @@ public class ParameterSetupPane extends BorderPane implements EmbeddedParameterC
   }
 
   public void setParameterValuesToComponents() {
+    setParameterValuesToComponents(this.parameterSet);
+  }
+
+  /**
+   * May want to apply other parameters to components
+   */
+  public void setParameterValuesToComponents(ParameterSet parameterSet) {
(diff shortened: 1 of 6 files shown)


mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/ParameterSetupPane.java
```diff
@@ -354,6 +354,13 @@ public ParameterSet updateParameterSetFromComponents() {
   }
 
   public void setParameterValuesToComponents() {
+    setParameterValuesToComponents(this.parameterSet);
+  }
+
+  /**
+   * May want to apply other parameters to components
```
SteffenHeu: maybe add to the java doc, that the internal parameter set is not updated by this method. it's the same as for all other parameters, but i think it's even less intuitive for a parameter set

robinschmid: done.
```java
/**
   * May want to apply other parameters to components, similar to other parameters this means that
   * the embedded parameterset in this pane is not updated or changed by this. Still needs to call
   * {@link #updateParameterSetFromComponents()} for this.
   */
```

SteffenHeu: not pushed yet, or? can merge after

robinschmid: not even commited. Was still sleeping on my laptop

PR #2813, opened 2025-08-13
Presets for all modules

robinschmid:
- [ ] ask for permission to overwrite parameters: just in batch mode or in all?

PR #2818, opened 2025-08-15
Use own parameterset in feature table config

robinschmid:
- Change some default and some parameter names
- DataTypeCheckListParameter and component had some quirks

for example this line was important in component because otherwise the selection was kept: 
`checkList.getCheckModel().clearChecks();`

Also included in https://github.com/mzmine/mzmine/pull/2813 for testing of presets

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FxFeatureTableInteractor.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FxFeatureTableInteractor.java
@@ -97,7 +97,8 @@ public class FxFeatureTableInteractor extends FxInteractor<FxFeatureTableModel> 
       if (exitCode == ExitCode.OK) {
         updateWindowToParameterSetValues();
         // set to module
-        ConfigService.getConfiguration().setModuleParameters(FeatureTableFXModule.class, param);
+        ConfigService.getConfiguration()
+            .setModuleParameters(FeatureTableFXModule.class, param.cloneParameterSet());
       }
     });
   }
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableColumnMenuHelper.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableColumnMenuHelper.java
@@ -28,6 +28,7 @@ package io.github.mzmine.modules.visualization.featurelisttable_modular;
 import io.github.mzmine.datamodel.features.ModularFeatureListRow;
 import io.github.mzmine.datamodel.features.types.fx.ColumnID;
 import io.github.mzmine.datamodel.features.types.fx.ColumnType;
+import io.github.mzmine.main.ConfigService;
 import io.github.mzmine.parameters.parametertypes.datatype.DataTypeCheckListParameter;
 import java.util.Comparator;
 import java.util.Map;
@@ -79,6 +80,13 @@ public class FeatureTableColumnMenuHelper extends TableColumnMenuHelper {
   public FeatureTableColumnMenuHelper(FeatureTableFX featureTable) {
     super(featureTable.getTable());
     this.featureTable = featureTable;
+  }
+
+  @Override
+  protected void onPopupClosed() {
+    super.onPopupClosed();
+    ConfigService.getConfiguration().setModuleParameters(FeatureTableFXModule.class,
+        featureTable.getParameters().cloneParameterSet());
   }
 
   /**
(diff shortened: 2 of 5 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FxFeatureTableController.java
```diff
@@ -40,7 +40,9 @@ public class FxFeatureTableController extends FxCachedViewController<FxFeatureTa
   private final FxFeatureTableInteractor interactor;
 
   public FxFeatureTableController() {
-    var params = MZmineCore.getConfiguration().getModuleParameters(FeatureTableFXModule.class);
+    // use a clone, will set parameters to module params after tab close
+    var params = ConfigService.getConfiguration().getModuleParameters(FeatureTableFXModule.class)
+        .cloneParameterSet();
```
SteffenHeu, line 45: i always found it better that it uses the same instance, because it means that a new feature table already has the same layout, even it it was not closed. not sure what the best solution is.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FxFeatureTableInteractor.java
```diff
@@ -95,6 +96,8 @@ public void openParametersDialog() {
       ExitCode exitCode = param.showSetupDialog(true);
       if (exitCode == ExitCode.OK) {
         updateWindowToParameterSetValues();
+        // set to module
+        ConfigService.getConfiguration().setModuleParameters(FeatureTableFXModule.class, param);
```
SteffenHeu: mhh so that means if the columns are updated with the dialog then the changes will be present in the next feature table. but if they are changed using the + menu, the layout is not updated. i think that will lead to cunfusion

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/datatype/DataTypeCheckListParameter.java
```diff
@@ -168,7 +168,7 @@ public void setDataTypeVisible(String typeUniqueId, Boolean val) {
    * @param map Map containing new data types and their values
    */
   public void setDataTypesAndVisibility(Map<String, Boolean> map) {
-    value = new HashMap<>(map);
+    value.putAll(map);
```
SteffenHeu, line 171: clear previous or? put does not remove existing entries, if they are not replaced, or?

robinschmid: Two options. You can decide which one you like better and we can refine this later. 
Explanation in other PR. I think I like the behavior of this PR option better to use the currently set state for unknown columns.

#2820 would use default visible state for unknown columns.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FxFeatureTableController.java
robinschmid: I think it is better to use a different instance for each table. Otherwise it is harder have two tables with different visible columns. 

I added that the global parameterset is updated when the quick menu is closed and when the config window is close. 
So it should be always the latest then. 

But could also change back to using the global parameters. Just think it usually means trouble later on.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FxFeatureTableInteractor.java
robinschmid: I think it is better to use a different instance for each table. Otherwise it is harder have two tables with different visible columns.

I added that the global parameterset is updated when the quick menu is closed and when the config window is close.
So it should be always the latest then.

But could also change back to using the global parameters. Just think it usually means trouble later on.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FxFeatureTableController.java
robinschmid: Just tested open two tables, change columns on one, switch to the other and try to change columns there and it will force the same columns on you. 

I think separating would make sense. Then you could just reopen the table if you want the same layout. If you want different columns then you can do it here. 
But no hard opinion on this.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FxFeatureTableInteractor.java
robinschmid: Just tested open two tables, change columns on one, switch to the other and try to change columns there and it will force the same columns on you. 

I think separating would make sense. Then you could just reopen the table if you want the same layout. If you want different columns then you can do it here. 
But no hard opinion on this.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FxFeatureTableController.java
SteffenHeu: yeah that's true. let's separate then

PR #2830, opened 2025-08-21
Enable denormalization for all mass detectors

robinschmid:


SteffenHeu:
After discussion: Denormalization in mass detection stage is a problem for DIA workflows. Better move to scan merging step. (make sure it is not used for merging of ims files in dia step)

PR #2837, opened 2025-08-26
Add dia ms2 builder with no correlation

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_diams2/no_corr/DiaMs2NoCorrModule.java
```diff
@@ -0,0 +1,76 @@
...
+import java.time.Instant;
+import java.util.Collection;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class DiaMs2NoCorrModule implements MZmineProcessingModule {
+
+  @Override
+  public @NotNull String getName() {
+    return "DIA pseudo MS2 builder (no correlation) (experimental)";
```
robinschmid, line 40: Maybe better to push all algorithm options into a single module. Using ModuleOptionsEnum for the different parameters.
I can imagine that over the time there will be more DIA algorithms like ZT scan etc. And it also makes it clearer to users that there are options available and centralize the documentation effort.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_diams2/no_corr/DiaMs2NoCorrTask.java
```diff
@@ -0,0 +1,288 @@
...
+                RangeUtils.toDoubleRange(mobilityRange));
+            ms2.setMsMsInfo(builtMs2Info);
+            pseudoSpectra.add(ms2);
+          } else {
+
+            final SimplePseudoSpectrum ms2 = new SimplePseudoSpectrum(file, ms2Scan.getMSLevel(),
+                ms2Scan.getRetentionTime(), null, ms2Scan.getMzValues(new double[0]),
+                ms2Scan.getIntensityValues(new double[0]), ms2Scan.getPolarity(),
+                "Pseudo spectrum (uncorrelated)", PseudoSpectrumType.UNCORRELATED);
+            pseudoSpectra.add(ms2);
```
robinschmid, line 209: Why create a pseudo spectrum and not just set the MS2 spectrum here? 
Seems like it just copies all info from MS2 scan.

SteffenHeu: because downstream handling of dia spectra is different than dda spectra. this makes it clear: PseudoSpectrumType.UNCORRELATED

robinschmid: ok makes sense

PR #2840, opened 2025-08-26
fix npe in BioTransformerTask if annotation has no smiles

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_biotransformer/BioTransformerTask.java
```diff
@@ -309,14 +309,16 @@ private FeatureAnnotation getBestAnnotation(@NotNull FeatureListRow row) {
 
     final List<SpectralDBAnnotation> spectralLibraryMatches = row.getSpectralLibraryMatches();
     if (!spectralLibraryMatches.isEmpty() && (smilesSource == SmilesSource.ALL
-        || smilesSource == SmilesSource.SPECTRAL_LIBRARY)) {
-      return spectralLibraryMatches.get(0);
+        || smilesSource == SmilesSource.SPECTRAL_LIBRARY)
+        && spectralLibraryMatches.getFirst().getSmiles() != null) {
+      return spectralLibraryMatches.getFirst();
```
robinschmid, line 314: maybe instead of just using the first, you could stream over the list and find the first one with a structure?

Currently the best way is to use `match.getStructure()` and you can get the canonical smiles from there. Or isomeric smiles in case it has this information. Isomeric smiles will just return canonical non isomerical smiles in case there is no information on stereo centeres etc.  

Same below. 

We should think about a way how we can harmonize this better. 
Maybe everytime a smiles or inchi is set we set all the other ones. 

In spectral lib I already do this for lazy initialization of structure... but maybe better to just initialize it and set the other fields. 

```java
  @Override
  public boolean putIfNotNull(DBEntryField field, Object value) {
    if (field == DBEntryField.SMILES || field == DBEntryField.INCHI) {
      structure = null; // clear and recalculate later
    }
```

SteffenHeu: I think the behaviour is more predictable for the user if it uses the structure that is displayed in the table, so the first one. Can change to getStructure to also take inchi into account.

robinschmid: using structure will be good. 
Always hard how to handle alternative annotations. Especially as the score is sometimes the same or very similar and the ranking is quite random then. 
Maybe you could add an option to use only best annotation or any?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_biotransformer/BioTransformerTask.java
```diff
@@ -215,10 +215,10 @@ public void run() {
       }
 
       final FeatureAnnotation bestAnnotation = getBestAnnotation(row);
-      if (bestAnnotation == null) {
+      if (bestAnnotation == null || bestAnnotation.getStructure().canonicalSmiles() == null) {
```
robinschmid, line 218: also needs to check if bestAnnotation.getStructure is null

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_biotransformer/BioTransformerTask.java
```diff
@@ -233,7 +233,7 @@ public void run() {
         final FeatureListRow row = entry.getValue();
         final FeatureAnnotation bestAnnotation = getBestAnnotation(row);
 
-        if (bestAnnotation == null || !bestSmiles.equals(bestAnnotation.getSmiles())) {
+        if (bestAnnotation == null || !bestSmiles.equals(bestAnnotation.getStructure().canonicalSmiles())) {
```
robinschmid, line 236: same here also needs to check if bestAnnotation.getStructure is null

robinschmid (approved):

PR #2841, opened 2025-08-27
XYProvider now has isComputed state

robinschmid:
- skip the computation in ColoredXYDataset
- precomputed providers can just always return true

Issue was that all datasets would spin up a new task if they were called on the javafx thread even if the data is already computed.


mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/spectra/SingleMobilityScanProvider.java
```diff
@@ -116,4 +116,11 @@ public int getValueCount() {
   public double getComputationFinishedPercentage() {
     return finishedPercentage;
   }
+
+  /**
+   * @return true if computed. Providers that are precomputed may use true always
+   */
+  public boolean isComputed() {
```
SteffenHeu, line 123: not sure why, but this broke the IMS raw data overview. The mobility scan chart always zooms to 0-1 m/z after selecting a new mobility scan

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/providers/impl/IntervalXYProvider.java
```diff
@@ -111,4 +111,11 @@ public int getValueCount() {
   public double getComputationFinishedPercentage() {
     return 0;
   }
+
+  /**
+   * @return true if computed. Providers that are precomputed may use true always
+   */
+  public boolean isComputed() {
+    return true;
```
SteffenHeu, line 119: same here (IMS feature chart)

SteffenHeu (commented):
Change looks good, but it affects some places where i'm not sure why it makes a difference

robinschmid:
This is handled in a bigger PR as all the chart logic is interconnected

PR #2842, opened 2025-08-27
Feature shape column

robinschmid:
- used caching of whole nodes before
- datasets were and are backed by memory mapped data, no calculation needed
- therefore changed to a cell factory and cell class that precreates a chart instance with correct size (fixes the width issue) 
- datasets are set by datasets.setAll
- I tested and optimized to a single draw call

The first cell with index 0 is used for measuring size (as far as I know). And there are many updateItem calls on this cell. I skipped all as the content does not change the size of the cell - the chart defines that already.

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright (c) 2004-2025 The mzmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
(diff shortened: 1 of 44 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
```diff
@@ -0,0 +1,190 @@
...
+    SimpleXYChart<IonTimeSeriesToXYProvider> chart = new SimpleXYChart<>(
+        uf.format("Retention time", "min"), uf.format("Intensity", "a.u."));
+//        uf.format("Retention time", "min"), uf.format("Intensity", "a.u."));
+    chart.setRangeAxisNumberFormatOverride(MZmineCore.getConfiguration().getIntensityFormat());
+    chart.setDomainAxisNumberFormatOverride(MZmineCore.getConfiguration().getRTFormat());
+    chart.setLegendItemsVisible(false);
+    chart.setPrefWidth(GraphicalColumType.LARGE_GRAPHICAL_CELL_WIDTH);
+
+    chart.getChart().setBackgroundPaint((new Color(0, 0, 0, 0)));
+    chart.getXYPlot().setBackgroundPaint((new Color(0, 0, 0, 0)));
```
SteffenHeu, line 177: For me this seems to not be effective, the cells did have a different background color than transparent

robinschmid:
Closed in favor of #2855
This PR has all the changes and is easier to test all the charts work properly once for the big PR.

PR #2845, opened 2025-08-28
Sirius via api

SteffenHeu:
- Export to sirius via api
- Run sirius hob via api
- Import results from existing sirius project
- Rank compound annotations using sirius


mzmine-community/build.gradle
```diff
@@ -224,6 +230,9 @@ dependencies {
 
     // test only
     testImplementation(libs.bundles.mzio.test)
+
+    // sirius api
+    implementation 'io.sirius-ms:sirius-sdk:3.1+sirius6.3.0'
```
robinschmid, line 234: maybe ask what these version numbers refer to.

mzmine-community/build.gradle
```diff
@@ -48,6 +48,7 @@ plugins {
 
     id("maven-publish")
     id("version-catalog")
+    id "com.google.cloud.artifactregistry.gradle-plugin" version("2.2.0")
```
robinschmid: maybe add a comment that this is for SIRIUS otherwise cryptic

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
```diff
@@ -536,4 +537,12 @@ default int compareTo(@NotNull CompoundDBAnnotation o) {
   }
 
   void setStructure(MolecularStructure structure);
+
+  default void enrichMetadata() {
+    MolecularStructure struc = StructureParser.silent()
+        .parseStructure(getSmiles(), getInChI());
```
robinschmid: `SimpleCompoundDBAnnotation.getStructure()`
Although this part does not set all other properties so that is a missed opportunity and should be corrected. 

So maybe we could keep the enrichMetadata method and just call it from the getStructure method in case structure is null.

And I guess the parsers etc should somehow have a finalizing step that just run the enrichMetadata method after parsing is done?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/scores/CsiScoreType.java
```diff
@@ -0,0 +1,26 @@
+package io.github.mzmine.datamodel.features.types.numbers.scores;
+
+import io.github.mzmine.datamodel.features.types.numbers.abstr.ScoreType;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
+import org.jetbrains.annotations.NotNull;
+
+public class CsiScoreType extends ScoreType {
```
robinschmid: rename to SiriusCsiFingerIdScoreType
Or SiriusCsiScoreType 
But adding sirius may be nice to make it more findable

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/scores/CsiScoreType.java
```diff
@@ -0,0 +1,26 @@
...
+
+  private static final NumberFormat FORMAT = new DecimalFormat("#.##");
+
+  @Override
+  public NumberFormat getFormat() {
```
robinschmid, line 13: maybe also replace export format?

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/export/ExportToSiriusParameters.java
```diff
@@ -0,0 +1,75 @@
...
+import io.github.mzmine.parameters.ParameterSet;
+import io.github.mzmine.parameters.impl.SimpleParameterSet;
+import io.github.mzmine.parameters.parametertypes.OptionalParameter;
+import io.github.mzmine.parameters.parametertypes.StringParameter;
+import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsParameter;
+import io.github.mzmine.parameters.parametertypes.selectors.FeatureListsSelection;
+import io.github.mzmine.util.FeatureUtils;
+import java.util.List;
+
+public class ExportToSiriusParameters extends SimpleParameterSet {
```
robinschmid: maybe we can name this differently?
Kind of clashes with the SiriusExportModule. 

Maybe 
- SiriusApiAddRowsModule?
- SiriusApiFingerIdModule
- SiriusApiResultsImportModule
- SiriusApiRankAnnotationsModule (currently RankAnnotationsBySiriusModule)

This way its clear that this is all api based. 
And we could (in theory) still have an import of Sirius summary files or other sirius result files.

robinschmid: the package already says siriusapi but we use the module classes often in different places so have SiriusApi in each name may make it easier to spot that they belong together.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/export/SiriusApiExportRowsParameters.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.List;
+
+public class SiriusApiExportRowsParameters extends SimpleParameterSet {
+
+  public static final FeatureListsParameter flist = new FeatureListsParameter(1, 1);
+
+  public static final OptionalParameter<StringParameter> rowIds = new OptionalParameter<>(
+      new StringParameter("Row IDs",
+          "If selected, only specific features will be exported. Otherwise, the complete feature list is exported. The ids of the rows to run CSI:FingerID for."),
+      false);
```
robinschmid, line 48: Describe the expected format here in the description. 
Maybe best to use the same one that is used in FxFeatureTableFilterMenu that uses commas and ranges as in 1,5-8,10

```java
List<IndexRange> ranges = IndexRange.parseRanges(id);
```

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/SiriusNotLoggedInException.java
```diff
@@ -0,0 +1,33 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.tools.siriusapi;
+
+public class SiriusNotLoggedInException extends RuntimeException {
+
+  public SiriusNotLoggedInException() {
+    super("Not logged into Sirius yet, please login first.");
```
robinschmid, line 31: maybe search and replace all to SIRIUS uppercase
https://v6.docs.sirius-ms.io/
https://bio.informatik.uni-jena.de/software/sirius/

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/export/SiriusApiExportRowsModule.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  @Override
+  public @NotNull MZmineModuleCategory getModuleCategory() {
+    return MZmineModuleCategory.FEATURELISTEXPORT;
+  }
+
+  @Override
+  public @NotNull String getName() {
+    return "Export to Sirius (API)";
```
robinschmid: SIRIUS

mzmine-community/src/main/java/io/github/mzmine/util/FeatureUtils.java
```diff
@@ -956,4 +957,16 @@ public static boolean isMrm(@Nullable Feature f) {
...
   }
 
+
+  public static String rowsToIdString(List<? extends FeatureListRow> rows) {
+    return rows.stream().map(FeatureListRow::getID).map(Object::toString)
+        .collect(Collectors.joining(";"));
+  }
+
+  public static List<FeatureListRow> idStringToRows(ModularFeatureList flist, String str) {
+    final Set<Integer> ids = Arrays.stream(str.split(";")).map(Integer::valueOf)
```
robinschmid: I think it might be best to stick to the format of the List<IndexRange> that is used in the FeatureTable filters? 
This way those are compatible and we have not so many different formats.
```java
List<IndexRange> ranges = IndexRange.parseRanges(id);
```

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/Sirius.java
```diff
@@ -0,0 +1,273 @@
...
+      if (sid != null) {
+        return sid.equals(activeSubscriptionId);
+      }
+      return false;
+    }).findFirst().map(s -> {
+      AllowedFeatures allowedFeatures = s.getAllowedFeatures();
+      if (allowedFeatures == null) {
+        return false;
+      }
+      return allowedFeatures.isApi();
```
robinschmid: isApi is Nullable so maybe this has to be isApi == true?
Otherwise this might be a NPE on unboxing

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
```diff
@@ -0,0 +1,297 @@
...
+   * {@link AlignedFeature#getAlignedFeatureId()}.
+   */
+  private static @NotNull Map<Integer, String> getAllAlignedFeatureIds(@NotNull Sirius sirius) {
+    final Map<Integer, String> alreadyImportedFeatures = sirius.api().features()
+        .getAlignedFeatures(sirius.getProject().getProjectId(), null,
+            List.of(AlignedFeatureOptField.NONE)).stream()
+        .filter(af -> af.getAlignedFeatureId() != null && af.getExternalFeatureId() != null)
+        .collect(Collectors.toMap(
+            alignedFeature -> Integer.valueOf(alignedFeature.getExternalFeatureId()),
+            AlignedFeature::getAlignedFeatureId));
```
robinschmid: maybe safer to only add features to this map where the external feature ID is actually an int. People for what ever reason could have added other features to the sirius project externally and the ids may be any string.  

I could imagine two options:

1. Just wrap the Integer.parseInt() into a try and throw an exception if this fails and tell people that they cannot add more features to the sirius project if this should interact with mzmine.
2. Just catch the parsing exception, skip this feature, and log that some features were excluded as they did not come from mzmine.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
```diff
@@ -0,0 +1,297 @@
...
+    final IonIdentity adduct = row.getBestIonIdentity();
+    if (adduct != null) {
+      f.setDetectedAdducts(Set.of(adduct.toString()));
+      f.setCharge(adduct.getIonType().getCharge());
+    }
+
+    f.setMs2Spectra(
+        row.getAllFragmentScans().stream().map(MzmineToSirius::spectrum).filter(Objects::nonNull)
+            .toList());
+    f.setMs1Spectra(List.of(generateCorrelationSpectrum(row)));
```
robinschmid, line 109: maybe also add the best MS1 spectrum?

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
```diff
@@ -0,0 +1,297 @@
...
+
+    return f;
+  }
+
+  public static @Nullable BasicSpectrum spectrum(@Nullable Scan scan) {
+    if (scan == null || scan.getMassList() == null
+        || scan.getMassList().getNumberOfDataPoints() == 0
+        || scan.getMassList().getBasePeakIntensity() == null || (scan instanceof PseudoSpectrum ps
+        && ps.getPseudoSpectrumType() == PseudoSpectrumType.GC_EI)) {
+      return null;
```
robinschmid, line 138: throw missing mass list exception maybe?
Only if massList is null

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
```diff
@@ -0,0 +1,297 @@
...
+      p.setMz(ml.getMzValue(i));
+      spectrum.addPeaksItem(p);
+    }
+
+    final MsMsInfo msMsInfo = scan.getMsMsInfo();
+    if (msMsInfo != null) {
+      spectrum.setCollisionEnergy(String.valueOf(msMsInfo.getActivationEnergy()));
+      spectrum.setPrecursorMz(scan.getPrecursorMz());
+    }
+    spectrum.setMsLevel(scan.getMSLevel());
```
robinschmid, line 150: all this could be extracted to a method `createSpectrum(MassSpectrum scan)` as it is duplicated in generateCorrelationSpectrum. 

Most methods that are not available in MassSpectrum and only in Scan/SpectralLibraryEntry can be used from ScanUtils:
```java
String sourceFileName = ScanUtils.getSourceFile(scan);
Integer scanNumber = ScanUtils.extractScanNumber(scan);
Float energy = ScanUtils.extractCollisionEnergy( scan);
ScanUtils.getMsLevel(scan).ifPresent(spectrum::setMsLevel);
// ScanUtils.getPrecursorMz(scan)
// 
```

I think only one missing is retention time

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
```diff
@@ -0,0 +1,297 @@
...
+    spectrum.setMsLevel(1);
+    spectrum.setName("Correlated MS1 spectrum");
+    return spectrum;
+  }
+
+  public static SearchableDatabase toCustomDatabase(
+      final @NotNull List<CompoundDBAnnotation> compounds, @NotNull Sirius sirius) {
+
+    final Map<String, CompoundDBAnnotation> uniqueCompounds = compounds.stream()
+        .filter(a -> a.getSmiles() != null)
```
robinschmid: a.getSmiles() should better be a.getStructure().isomericSmiles(). Maybe we should even change the getSmiles method to just do this instead? 
This way we also enforce that the correct values from the structure are used.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
```diff
@@ -0,0 +1,297 @@
...
+        .collect(Collectors.toMap(CompoundDBAnnotation::getSmiles, a -> a));
+    final File dbFile = writeCompoundsToFile(uniqueCompounds);
+
+    final SearchableDatabasesApi databases = sirius.api().databases();
+    databases.getCustomDatabases(false, false);
```
robinschmid: remove this line? return type is not used and the line is repeated below

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
```diff
@@ -0,0 +1,297 @@
...
+
+    final SearchableDatabase database = databases.getCustomDatabases(false, false).stream()
+        .filter(db -> db.getDatabaseId().equals(Sirius.mzmineCustomDbId)).findFirst()
+        .orElseGet(() -> {
+          SearchableDatabaseParameters dbParam = new SearchableDatabaseParameters();
+          dbParam.setDisplayName(Sirius.mzmineCustomDbId);
+          dbParam.setLocation(new File(new File(FileAndPathUtil.getMzmineDir(), "sirius_databases"),
+              Sirius.mzmineCustomDbId).getAbsolutePath());
```
robinschmid: Maybe change to the .mzmine/external_resources/ folder?
It is not downloaded but still seems more an external resource:
We also have the resolveIn methods:
```java
// .mzmine/external_resources/sirius/
FileAndPathUtil.resolveInDownloadResourcesDir("sirius/"+Sirius.mzmineCustomDbId);
```

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
```diff
@@ -0,0 +1,297 @@
...
+    file.deleteOnExit();
+
+    if (!file.getParentFile().exists()) {
+      file.getParentFile().mkdirs();
+    }
+
+    try (var bufferedWriter = Files.newBufferedWriter(file.toPath(), StandardCharsets.UTF_8,
+        WriterOptions.REPLACE.toOpenOption())) {
+      CSVWriterBuilder builder = new CSVWriterBuilder(bufferedWriter).withSeparator('\t');
+      final ICSVWriter writer = builder.build();
```
robinschmid: this writer needs to be in a try-with resources and will close the bufferedWriter internally so that does not need to be in a try with resources. 
Maybe best to use the factory:
```java
try(var writer = CSVParsingUtils.createDefaultWriter(file, '\t', WriterOptions.REPLACE)) {
      
    } catch (IOException e) {
    }
```

Both options are good, but we also have this one for a one liner export:
`CsvWriter.writeToFile(file, list, SomeRecord.class);`

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/fingerid/SiriusApiFingerIdParameters.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class SiriusApiFingerIdParameters extends SimpleParameterSet {
+
+  public static final FeatureListsParameter flist = new FeatureListsParameter(1, 1);
+
+  public static final OptionalParameter<StringParameter> rowIds = new OptionalParameter<>(
+      new StringParameter("Row IDs",
+          "The ids of the rows to run Sirius for. If not selected, the whole feature list will be processed."));
```
robinschmid: describe input format, like in the other ApiExport parameters

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/fingerid/SiriusApiFingerIdTask.java
```diff
@@ -0,0 +1,180 @@
...
+      return true;
+    }
+
+    final HiddenParameter<Map<String, Boolean>> optOutParam = ConfigService.getConfiguration()
+        .getModuleParameters(SiriusApiFingerIdModule.class)
+        .getParameter(SiriusApiFingerIdParameters.countWarningOptOut);
+    final String userHash = String.valueOf(CurrentUserService.getUserName().hashCode());
+
+    try (var _ = optOutLock.lockRead()) {
+      if (Boolean.TRUE.equals(optOutParam.getValue().get(userHash))) {
```
robinschmid, line 143: Maybe instead of a Map you could just save a Set<String> of all users that opted out and ask `optOutSet.contains(userHash)` as the opt out only does something in true case.


For the boolean check:
`if(optOutParam.getValue().get(userHash)==true)` should work also for null and Objects.equals(a,b).

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/fingerid/SiriusApiFingerIdTask.java
```diff
@@ -0,0 +1,180 @@
...
+        return true;
+      }
+
+      final TextFlow message = FxTextFlows.newTextFlow(FxTexts.text(
+          "You are trying to compute %d features in Sirius. This may take a considerable amount of time. Do you want to continue?".formatted(
+              rows.size())));
+      final ButtonType result = DialogLoggerUtil.createAlertWithOptOutBlocking("Warning",
+          "Confirmation required", message, "Do not show again", optedOut -> {
+            optOutParam.getValue().put(userHash, optedOut);
```
robinschmid, line 159: only add on true and remove on false? maybe use a `Set<>` instead

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/fingerid/SiriusApiFingerIdTask.java
```diff
@@ -0,0 +1,180 @@
...
+
+      // polarity for fallback adducts
+      // todo: set in a sirius preferences or so? Or just keep these as default adducts
+      final PolarityType polarity = rows.stream()
+          .map(r -> FeatureUtils.extractBestPolarity(r).orElse(null)).filter(Objects::nonNull)
+          .findFirst().orElse(PolarityType.POSITIVE);
+      switch (polarity) {
+        case NEGATIVE -> submission.setFallbackAdducts(List.of("[M-H]-"));
+        default -> submission.setFallbackAdducts(List.of("[M+H]+", "[M+NH4]+", "[M+Na]+"));
+      }
```
robinschmid, line 114: no idea what the fallback is used for, but submission has no defaults for those?
Maybe a question for Markus

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/JobWaiterTask.java
```diff
@@ -0,0 +1,127 @@
...
+      logger.warning(() -> "Sirius job was not successful.");
+      return;
+    }
+
+    onFinished.run();
+  }
+
+  @Override
+  protected @NotNull List<FeatureList> getProcessedFeatureLists() {
+    return Arrays.stream(flist).toList();
```
robinschmid: should this be added as an applied method? This is what this method does. Seems odd for the JobWaiterTask

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/JobWaiterTask.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @Override
+  public String getTaskDescription() {
+    return "Running Sirius on feature list(s) " + Arrays.stream(flist).map(FeatureList::getName)
+        .collect(Collectors.joining(", "));
+  }
+
+  boolean hasFinished(Job job) {
+    final JobState state = job.getProgress().getState();
```
robinschmid: job.getProgress is nullable. Add null checks to this method and to wasSuccessful.
getState is also Nullable so maybe add null as a case to the switches

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/import_annotations/SiriusApiResultsImportTask.java
```diff
@@ -0,0 +1,95 @@
...
+            row);
+        if (siriusAnnotations == null || siriusAnnotations.isEmpty()) {
+          importedFeatures++;
+          continue;
+        }
+
+        final List<CompoundDBAnnotation> current = row.getCompoundAnnotations();
+        if (current != null && !current.isEmpty()) {
+          siriusAnnotations.addAll(current);
+        }
```
robinschmid, line 81: not super important now but maybe we could add a method to the row for this like 
```java
// true - add to start of list, false add to end
row.addCompoundAnnotations(siriusAnnotations, true); 
```

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/rank_annotations/SiriusApiRankAnnotationsTask.java
```diff
@@ -0,0 +1,157 @@
...
+      for (FeatureListRow row : rows) {
+
+        final List<CompoundDBAnnotation> annotations = row.getCompoundAnnotations();
+
+        if (annotations.isEmpty()) {
+          return;
+        }
+
+        // inchi may be there but not the smiles
+        annotations.forEach(CompoundDBAnnotation::enrichMetadata);
```
robinschmid, line 100: might not be needed depending how we decide on the getStructure etc that already does this.
Comment just to tag this line

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/rank_annotations/SiriusApiRankAnnotationsTask.java
```diff
@@ -0,0 +1,157 @@
...
+            });
+
+        task.run();
+
+        final List<CompoundDBAnnotation> rankedAnnotations = SiriusToMzmine.getSiriusAnnotations(
+            sirius, siriusId, row);
+
+        row.setCompoundAnnotations(rankedAnnotations);
+
+        FeatureTableFXUtil.updateCellsForFeatureList(flist);
```
robinschmid: I think this needs to go into the JobWaiterTask.onFinished Runnable that is empty? 
Otherwise this block would also execute if the job failed.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/JobWaiterTask.java
```diff
@@ -0,0 +1,127 @@
...
+  private static final Logger logger = Logger.getLogger(JobWaiterTask.class.getName());
+
+  private final @NotNull Supplier<Job> jobSupplier;
+  private final @NotNull Runnable onFinished;
+  private final @NotNull FeatureList[] flist;
+
+  /**
+   * @param jobSupplier supplier to retrieve the job status from the sirius api.
+   */
+  public JobWaiterTask(@NotNull Class<? extends MZmineModule> callingModule,
```
robinschmid: maybe pass in a `@Nullable Task parentTask` and check if the calling task was cancelled --> in the sleeping loop

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/JobWaiterTask.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import org.jetbrains.annotations.NotNull;
+
+public class JobWaiterTask extends AbstractSimpleTask {
```
robinschmid: maybe rename to SiriusJobWaiterTask as this is specific to sirius

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/Sirius.java
```diff
@@ -0,0 +1,273 @@
...
+      NotificationService.show(NotificationType.INFO, "Starting Sirius",
+          "Trying to start Sirius. This may take a moment.");
+      sirius = SiriusSDK.startAndConnectLocally(ShutdownMode.NEVER, true);
+    }
+
+    int tries = 0;
+    final int maxTries = 10;
+    while (!SiriusSDKUtils.restHealthCheck(sirius.getApiClient()) && tries < maxTries) {
+      tries++;
+      Thread.sleep(50);
```
robinschmid: Just checking, this is always enough time for the server to start up?
10x50ms + the time restHealthCheck takes...

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/Sirius.java
```diff
@@ -0,0 +1,273 @@
...
+  public static String getDefaultSessionId() {
+    return sessionId;
+  }
+
+  private void showGuiForProject() {
+    if (DesktopService.isHeadLess()) {
+      return;
+    }
+    for (GuiInfo gui : sirius.gui().getGuis()) {
+      if (gui.getProjectId().equals(projectSpace.getProjectId())) {
```
robinschmid: NPE potentially as gui.getProjectId() is Nullable
Objects.equals is safer

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/Sirius.java
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+
+  /**
+   * Ranks the current {@link CompoundDBAnnotation}s by exporting to a custom database and using
+   * only that database for possible structures.
+   *
+   * @param row
+   */
+  public void rankCurrentCompoundAnnotations(@NotNull final FeatureListRow row) {
```
robinschmid: this method is unused. Not sure maybe it was supposed to be used by the corresponding module?
Maybe remove or use?

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/SiriusLicenseHasNoApiException.java
```diff
@@ -0,0 +1,33 @@
...
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.modules.tools.siriusapi;
+
+public class SiriusLicenseHasNoApiException extends RuntimeException {
```
robinschmid, line 28: not used

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/SiriusToMzmine.java
```diff
@@ -0,0 +1,264 @@
...
+   * Maps the rows that are already imported to sirius to their aligned feature id.
+   */
+  public static Map<FeatureListRow, String> mapFeatureToSiriusId(@NotNull Sirius sirius,
+      List<? extends FeatureListRow> rows) {
+    sirius.checkLogin();
+    final Map<Integer, String> mzmineIdToSiriusId = sirius.api().features()
+        .getAlignedFeatures(sirius.getProject().getProjectId(), null,
+            List.of(AlignedFeatureOptField.NONE)).stream().collect(Collectors.toMap(
+            alignedFeature -> Integer.valueOf(alignedFeature.getExternalFeatureId()),
+            AlignedFeature::getAlignedFeatureId));
```
robinschmid: can reuse:

```java
MzmineToSirius.getAllAlignedFeatureIds(sirius);
```

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/SiriusToMzmine.java
```diff
@@ -0,0 +1,264 @@
...
+    final AlignedFeature alignedFeature = getSiriusFeatureOrThrow(sirius, siriusFeatureId, row);
+
+    final List<StructureCandidateFormula> structureCandidates = sirius.api().features()
+        .getStructureCandidates(sirius.getProject().getProjectId(),
+            alignedFeature.getAlignedFeatureId(), List.of()).stream()
+        .sorted(Comparator.comparingInt(s -> s.getRank() != null ? s.getRank() : 100)).limit(10)
+        .toList();
+
+    final List<CompoundDBAnnotation> siriusAnnotations = structureCandidates.stream()
+        .sorted(Comparator.comparingInt(StructureCandidateFormula::getRank))
```
robinschmid: the sorting is already done above with better null checks. Remove this line.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableContextMenu.java
```diff
@@ -399,9 +410,29 @@ private void initSearchMenu() {
...
+    sendToSirius.setOnAction(_ -> MZmineCore.runMZmineModule(SiriusApiExportRowsModule.class,
+        SiriusApiExportRowsParameters.of(selectedRows)));
+
+    final MenuItem runFingerId = new ConditionalMenuItem("Send to SIRIUS & Compute",
+        this::siriusApiCheck);
+    runFingerId.setOnAction(_ -> MZmineCore.runMZmineModule(SiriusApiFingerIdModule.class,
+        SiriusApiFingerIdParameters.of(selectedRows)));
+
+    final MenuItem rankUsingFingerId = new ConditionalMenuItem(
+        "Rank Compound annotations using SIRIUS",
```
robinschmid: Compound lower case compound

mzmine-community/src/main/resources/dependency/dependency-licenses.json
```diff
@@ -383,6 +435,19 @@
                 }
             ]
         },
+        {
+            "moduleName": "com.googlecode.concurrent-locks:concurrent-locks",
```
robinschmid, line 439: Always interesting what gets towed in by other dependencies. 
This seems fine. Its old from 2013 but on GH says its feature complete. And it is a 4 class lib. 

https://github.com/npgall/concurrent-locks

mzmine-community/src/main/resources/dependency/dependency-licenses.json
```diff
@@ -2888,13 +3441,34 @@
                 }
             ]
         },
+        {
+            "moduleName": "org.slf4j:jul-to-slf4j",
```
robinschmid, line 3445: ah ok sirius is already towing in the JUL to SLF4j bindings but this seems to be the package that routes all JUL to SLF4j

Maybe it is enough to then create a handler to get the SIRIUS logs and get rid of that log stating that SLF4J is not configured

https://mvnrepository.com/artifact/org.slf4j/jul-to-slf4j

robinschmid (changes requested):
Looks great. 

In the future we could see if we could somehow link to the sirius project in the applied methods of SiriusApiFingerIdModule and others, so that if the project is available we can just import the results instead of recompute. 

But thats a lot of work so just an idea for future directions.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/fingerid/SiriusApiFingerIdTask.java
SteffenHeu: No, no default fallbacks. This a specific suggestion from markus.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/fingerid/SiriusApiFingerIdTask.java
SteffenHeu: The opt out parameter already exists for a different use case. So i will not change the behaviour

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/fingerid/SiriusApiFingerIdTask.java
SteffenHeu: see above

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/JobWaiterTask.java
SteffenHeu: only null if the job state is not queried

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
SteffenHeu: If we change the method to not use the map entry, it creates unclear behaviour, because the result of get(SmilesType.class) and getSmiles will be different

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
SteffenHeu: there it will never be deleted. Not sure if we should delete it by default from the temp folder, but at least it does not create huge files on the main drive.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/Sirius.java
SteffenHeu: Yes because the SiriusSDK.startAndConnectLocally method already waits. Sometimes i still experienced issues however, so we just add a little delay here.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/Sirius.java
SteffenHeu: isApi == true also unboxes or not?

mzmine-community/src/main/resources/dependency/dependency-licenses.json
SteffenHeu: does that mean that using the sirius dependency forces us to route all our jul logs through slf4j?

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
SteffenHeu: the only duplication is the addition of the peaks really.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
SteffenHeu: i couldnt find out how to differentiate between MS1 and correlated in the spectrum types and i don't know if it is required. so currently only either is added.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/rank_annotations/SiriusApiRankAnnotationsTask.java
SteffenHeu: added enrich metadata to loading from csv database

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/Sirius.java
robinschmid: Oh yeah you are right. How weird I thought this would work.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
robinschmid: Ok can be also added later once we know.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
robinschmid: also good like this. 
I like the ScanUtils functions then most methods that take Scan can actually work with MassSpectrum that also includes the SpectralLibraryEntry

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
robinschmid: Currently the path points to .mzmine/sirius_databases/ ... 

I think putting it into the temp folder also sounds good. 
Maybe this already works:
`FileAndPathUtil.createTempFile()`

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
robinschmid: I think the most important question is how we can enforce that the all the structure specific fields are always up to date in case only smiles or inchi are set.
Then there is no mismatch. 
But not sure how to do this best. 
This would mean that the setter would need to check and apply the change?

robinschmid: Or each parser

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/fingerid/SiriusApiFingerIdParameters.java
```diff
@@ -0,0 +1,82 @@
...
+
+public class SiriusApiFingerIdParameters extends SimpleParameterSet {
+
+  public static final FeatureListsParameter flist = new FeatureListsParameter(1, 1);
+
+  public static final OptionalParameter<StringParameter> rowIds = new OptionalParameter<>(
+      new StringParameter("Row IDs", """
+          The ids of the rows to run Sirius for. If not selected, the whole feature list will be processed.
+          Specify row IDs as ranges or lists: 1,5-8,10
+          """));
```
robinschmid: maybe default value of OptionalParameter should be set to false? Like it is done in the other module

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/rank_annotations/SiriusApiRankAnnotationsParameters.java
```diff
@@ -0,0 +1,74 @@
...
+import io.github.mzmine.util.FeatureUtils;
+import java.util.List;
+
+public class SiriusApiRankAnnotationsParameters extends SimpleParameterSet {
+
+  public static final FeatureListsParameter flist = new FeatureListsParameter(1, 1);
+
+  public static final OptionalParameter<StringParameter> rowIds = new OptionalParameter<>(
+      new StringParameter("Row IDs",
+          "The ids of the rows to run Sirius for. If not selected, the whole feature list will be processed."));
```
robinschmid: maybe default of OptionalParameter should be false like in other module? 

Also add the same description for the IDs:

This is from the other class
```java
public static final OptionalParameter<StringParameter> rowIds = new OptionalParameter<>(
      new StringParameter("Row IDs", """
          If selected, only specific features will be exported. Otherwise, the complete feature list is exported.
          The ids of the rows to run CSI:FingerID for.
          Specify row IDs as ranges or lists: 1,5-8,10"""),
      false);
```

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/Sirius.java
robinschmid: Ok makes sense. Was just checking

mzmine-community/src/main/resources/dependency/dependency-licenses.json
robinschmid: AFAIK SLF4J is a facade that unifies all loggers. So any logger calls are unified and get then routed to a main logger framework. So you always still have to choose either log4j, logback, JUL etc

So I think we have the option to define a handler and the forward the logs from SLF4j through JUL. 
But not 100% certain

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/rank_annotations/SiriusApiRankAnnotationsTask.java
robinschmid: Sounds good also seems to cover all important places for now

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/modules/fingerid/SiriusApiFingerIdTask.java
robinschmid: Also fine to use the map. 

Only the question remains if this should be rather handled in the config? So that the value is not overwritten by presets and loading batches. I think it would be easier to handle and saver to put in the config. 

There are many hidden parameters in MZminePreferences that do similar things:
```java
HiddenParameter<Map<String, Boolean>> imsModuleWarnings...
HiddenParameter<Boolean> showQuickStart...
public static final HiddenParameter<Boolean> showTempFolderAlert
```

mzmine-community/src/main/resources/dependency/dependency-licenses.json
SteffenHeu: problem was not that we didnt have a provider, the problem was that we used slf4j 1.7, while 2.x is the current one

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
SteffenHeu: Sorry i thought this referred to the project. THe compound library is actually correct to be placed there. Can change to what you said

mzmine-community/src/main/java/io/github/mzmine/modules/tools/siriusapi/MzmineToSirius.java
SteffenHeu: yes the parser would have to call enrichMetadata. done now in the utils method we usually call.

mfleisch:
I added some additional error handling in this pr:
https://github.com/SteffenHeu/mzmine3/pull/28

robinschmid (approved):
Looks good

Issue #2834, opened 2025-08-25
Add aligned feature list splitter


SteffenHeu:
- Split aligned feature list into individual feature lists per data file
- Keep IDs

PR #2848, opened 2025-08-29
add split feature list module

SteffenHeu:
addresses #2834

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_splitaligned/SplitAlignedFeatureListTask.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_splitaligned/SplitAlignedFeatureListTask.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 2004-2025 The mzmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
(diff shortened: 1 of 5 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_splitaligned/SplitAlignedFeatureListTask.java
```diff
@@ -0,0 +1,141 @@
...
+  @Override
+  protected void process() {
+    final MetadataTable metadata = ProjectService.getMetadata();
+    final List<? extends SamplesGroupedBy<?>> sampleGroups = getGroupedSamples(metadata);
+
+    totalItems = (long) flist.getNumberOfRows() * flist.getNumberOfRawDataFiles();
+
+    for (final SamplesGroupedBy<?> sampleGroup : sampleGroups) {
+
+      final ModularFeatureList resultFlist = FeatureListUtils.createCopy(flist, null,
```
robinschmid, line 89: could be also interesting to allow setting a short name. Maybe an optional string parameter set to somehting like **Aligned split** and then the suffix will be added to that?  
createCopy(flist, null --> this allows to set a full name in this case. 

I think that would be nice because the names are getting quite long

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_splitaligned/SplitAlignedFeatureListTask.java
```diff
@@ -0,0 +1,141 @@
...
+          "_" + (sampleGroup.value() != null ? sampleGroup.valueString()
+              : sampleGroup.files().getFirst().getName()), getMemoryMapStorage(), false,
+          sampleGroup.files(), false, null, null);
+
+      for (final FeatureListRow row : flist.getRows()) {
+        if (sampleGroup.files().stream().noneMatch(row::hasFeature)) {
+          finishedItems.set(finishedItems.get() + flist.getNumberOfRawDataFiles());
```
robinschmid: I think the increment should be sampleGroup.size()? or resultFlist.getNumberOfRawDataFiles(), right?

Or the totalItems could be changed to just numRows x numGroups? Then this would just be a simple increment.

can use `finishedItems.addAndGet`

robinschmid (dismissed):
looks good. Maybe add the naming option. But this can also be done later.

PR #2849, opened 2025-08-29
update dependencies

SteffenHeu:


convention-plugins/java-convention/src/main/kotlin/io.github.mzmine.java-common-conv.gradle.kts
```diff
@@ -46,6 +46,15 @@ configurations.all {
             useVersion(libs.findVersion("log4j.core").get().preferredVersion)  // strict version
             because("patch all transitive dependencies to latest version")
         }
+
+        if (requested.group == "org.apache.xmlgraphics") {
+            if (requested.name != "xmlgraphics-commons") { // different dependency version than batik itself
```
robinschmid, line 51: just checking - this is still required to change the version of this package?

mzmine-community/build.gradle
```diff
@@ -135,21 +135,13 @@ repositories {
 //    maven { url = "https://nexus.nuiton.org/nexus/content/groups/releases/" }
     // For jimzml
     // maven { url = "https://mvnrepository.com/artifact/com.alanmrace/jimzmlparser" }
+    maven {url = "https://repository.everit.biz/nexus/content/groups/public/"}
```
robinschmid, line 140: maybe add a comment here as well that this repo is only required for this batik-xml-commons-ext package. Then we know when we can remove it again.

robinschmid (dismissed):
Is looking good.

PR #2859, opened 2025-09-05
Add plot background to chart theme parameters

robinschmid:
- chart background transparent
- plot data area background same as theme css pane background color
- this way it is easier to harmonize the colors of datasets and dont rely on the background color to be ok. Like in tables the selected row, mouse over, and regular color are already three different colors that make it hard to see all dataset colors.

This will be better to test in combination with the new chart columns.

Also we have to check what good defaults are - not sure what the actual background color is in the css there are so many base / background colors defined. 
Panes are transparent by default, not sure what the background is in our TabPane.

Here some examples of the current state:

image

Here in the table (this screenshot was taken with the columns PR combined just as preview:
image

image

### DIFF
--- a/mzmine-community/src/main/resources/themes/jabref_additions_common.css
+++ b/mzmine-community/src/main/resources/themes/jabref_additions_common.css
@@ -23,6 +23,10 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
 
+.root {
+  -chart-plot-background: -jr-background-alt;
+}
+
 .notification-bar > .pane {
   -fx-background-color: -jr-base;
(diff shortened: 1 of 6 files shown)


mzmine-community/src/main/resources/themes/jabref_additions_common.css
```diff
@@ -23,6 +23,10 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
 
+.root {
+  -chart-plot-background: -jr-background-alt;
```
SteffenHeu, line 27: -fx-control-inner-background

at least for jabref themes

PR #2860, opened 2025-09-05
[on other branch] Single delayed draw, FxXYPlotModel, table column fixed

robinschmid:
This PR may simplify review a bit. 
This merges #2855 onto #2841 and cuts out many provider classes that are now computed = true.

Need to check all plots behave like they should.


SteffenHeu:
- IMS mobility m/z plot from feature list 
the eic chart shows the feature not normalized but the EIC normalized. before both were not normalized

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/PlotCursorConfigModel.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public void setRangeCursorLockedOnData(boolean rangeCursorLockedOnData) {
+    this.rangeCursorLockedOnData.set(rangeCursorLockedOnData);
+  }
+
+  public void setDomainCursorPosition(Double value) {
+    PlotCursorPosition pos = getCursorPosition();
+    if (pos != null) {
+      pos = pos.withDomainValue(value);
```
SteffenHeu: this means that the dataset is not update. it may be possible that the point does not exist in the dataset. e.g. feature shape vs full eic in a plot

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
```diff
@@ -0,0 +1,556 @@
...
+   * Add to first free index
+   */
+  public int addDataset(@Nullable XYDataset dataset, @Nullable XYItemRenderer renderer) {
+    if (dataset == null) {
+      return -1;
+    }
+    for (int i = 0; i <= datasets.size(); i++) {
+      if (datasets.get(i) == null) {
+        datasets.put(i, dataset);
+        if (renderer != null) {
```
SteffenHeu, line 278: what renderer is used when it is null? dataset and no render means nothing is being drawn

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
```diff
@@ -0,0 +1,556 @@
...
+    }
+
+    throw new IllegalStateException("No free index found for dataset but should be within <=size");
+  }
+
+  /**
+   * Add all datasets to respective first free indices. Will only trigger one update for datasets
+   * and one for renderers
+   */
+  public void addDatasets(@NotNull List<? extends XYDataset> newDatasets,
```
SteffenHeu, line 292: What about using the DatasetAndRenderer? I think that doesnt even have to be a ColoredXYDataset in theory. That would already show a mismatch of renderers/datasets at programming time and not at runtime

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
```diff
@@ -0,0 +1,556 @@
...
+    removeDataSet(index);
+    return index;
+  }
+
+  /**
+   * Convenience method to add both datasets and renderers at once
+   *
+   * @param renderer single renderer for all datasets
+   */
+  public void addDatasets(@NotNull SequencedCollection<? extends XYDataset> datasets,
```
SteffenHeu: as far as i am aware having the same renderer for multiple datasets leads to issues with legend genereation. (at least in our simplechart framewortk)

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
robinschmid: Yeah I thought about adding the default renderer like in other implementations like SimpleXYChart

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
robinschmid: SimpleXYPlot and XYZ version actually use the same default renderer internally for all datasets where we do not set a different renderer. 
So I think this is not an issue?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
robinschmid: I also thought about using this class but it is used elsewhere with ColoredXYDataset.getValueProvider
In OtherDataResolverPreviewPane

robinschmid: I had some issues with the existing methods in plot.setRenderer/setDataset(int index, value) that is why I chose to not use a construct like DatasetAndRenderer in the internal data structures but just remodel what they have in XYPlot and make it observable.

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/PlotCursorConfigModel.java
robinschmid: At least in the original plot.crosshair this behavior is possible. The crosshair is either locked on data or free floating. 
This method may also be called from the plot.setDomainCrosshairValue which is used throughout mzmine already. 

The question here would be if we should try to find the closest data point to the new position and then use that.

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
robinschmid: Can you find the code? I could not find it

robinschmid:
Crosshairs now black and white. 
Gray was too invisible

image

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/PlotCursorConfigModel.java
SteffenHeu: The issue i'm expecting is the following:

```
  public PlotCursorPosition withDomainValue(double value) {
    // within float significance return same value - no change triggered
    if (Precision.equalSignificance(value, domainValue, 6)) {
      return this;
    }
    return new PlotCursorPosition(value, rangeValue, zValue, valueIndex, dataset, mouseEvent);
  }
```
This only updates the domain value, but does not cause a change to the valueIndex or the dataset. 
Both are used to determine respective datasets in connected plots. If the value index is not updated, the plot will show the wrong dataset. Same goes for if the dataset is not updated, even though the new point is on a different, e.g. raw file, the wrong file will be shown.

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
SteffenHeu: note the same renderer issue discussed below. should be cloned.

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
SteffenHeu: oh no i didnt mean for maniging it here, i only meant for adding datasets and renderers, so as the parameters for this method

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
SteffenHeu: Which code do you mean?

I think in most cases where   `public synchronized int addDataset(T datasetProvider)` is used, the plots tend to not show legends or only use a single dataset anyway.

robinschmid:
Images also work and removed some blinking that was caused by multiple datasets being calculated at the same time. This was due to multiple changes to the mz range in the parameter components formatting of the value.

image

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/PlotCursorConfigModel.java
SteffenHeu: takaway: the withDomainValue method should update the valueIndex as well and the dataset if possible. 
if there is no respective value in the dataset, go through all datasets in the plot to find the correct data set

robinschmid: Maybe possible to compute the dataset and value index lazily

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
robinschmid: I added methods to FxXYPlot to use List<XYDatasetAndRenderer> 
Added a sealed interface XYDatasetAndRenderer with children SimpleXYDatasetAndRenderer that is the base implementation and the old DatasetAndRenderer has now a TODO note to be renamed to ColoredXYDatasetAndRenderer as it is more specific.

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
robinschmid: see above. Now added a method with XYDatasetAndRenderer. 

The default renderers in the SimpleChart classes are still an issue maybe as they will reuse the renderer. but that was like this before

SteffenHeu:
Found this in my logs after going through the feature table:

```
Exception in thread "JavaFX Application Thread" java.util.ConcurrentModificationException
	at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1606)
	at java.base/java.util.HashMap$EntryIterator.next(HashMap.java:1639)
	at java.base/java.util.HashMap$EntryIterator.next(HashMap.java:1637)
	at org.jfree.chart.plot.XYPlot.getDatasetsMappedToDomainAxis(XYPlot.java:3996)
	at org.jfree.chart.plot.XYPlot.getDataRange(XYPlot.java:4120)
	at org.jfree.chart.axis.NumberAxis.autoAdjustRange(NumberAxis.java:366)
	at org.jfree.chart.axis.NumberAxis.configure(NumberAxis.java:348)
	at org.jfree.chart.plot.XYPlot.configureDomainAxes(XYPlot.java:766)
	at org.jfree.chart.plot.XYPlot.datasetChanged(XYPlot.java:4231)
	at io.github.mzmine.gui.chartbasics.gui.javafx.model.FxXYPlot.datasetChanged(FxXYPlot.java:197)
	at org.jfree.data.general.AbstractDataset.notifyListeners(AbstractDataset.java:205)
	at org.jfree.data.general.AbstractDataset.fireDatasetChanged(AbstractDataset.java:188)
	at io.github.mzmine.gui.chartbasics.simplechart.datasets.ColoredXYDataset.lambda$onCalculationsFinished$0(ColoredXYDataset.java:354)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$4(PlatformImpl.java:419)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run$$$capture(InvokeLaterDispatcher.java:95)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java)
	at --- Async.Stack.Trace --- (captured by IntelliJ IDEA debugger)
	at com.sun.glass.ui.InvokeLaterDispatcher.invokeLater(InvokeLaterDispatcher.java)
	at com.sun.glass.ui.win.WinApplication._invokeLater(WinApplication.java:321)
	at com.sun.glass.ui.Application.invokeLater(Application.java:480)
	at com.sun.javafx.tk.quantum.QuantumToolkit.defer(QuantumToolkit.java:818)
	at com.sun.javafx.application.PlatformImpl.runLater(PlatformImpl.java:417)
	at com.sun.javafx.application.PlatformImpl.runLater(PlatformImpl.java:398)
	at javafx.application.Platform.runLater(Platform.java:171)
	at io.github.mzmine.javafx.concurrent.threading.FxThread.runLater(FxThread.java:55)
	at io.github.mzmine.gui.chartbasics.simplechart.datasets.ColoredXYDataset.onCalculationsFinished(ColoredXYDataset.java:354)
	at io.github.mzmine.gui.chartbasics.simplechart.datasets.ColoredXYDataset.run(ColoredXYDataset.java:342)
	at io.github.mzmine.gui.chartbasics.simplechart.datasets.ColoredXYDataset.handleRunOption(ColoredXYDataset.java:146)
	at io.github.mzmine.gui.chartbasics.simplechart.datasets.ColoredXYDataset.<init>(ColoredXYDataset.java:115)
	at io.github.mzmine.gui.chartbasics.simplechart.datasets.ColoredXYDataset.<init>(ColoredXYDataset.java:133)
	at io.github.mzmine.datamodel.features.types.graphicalnodes.FeatureShapeMobilogramChart.<init>(FeatureShapeMobilogramChart.java:74)
	at io.github.mzmine.datamodel.features.types.FeatureShapeMobilogramType.createCellContent(FeatureShapeMobilogramType.java:73)
	at io.github.mzmine.datamodel.features.types.FeatureShapeMobilogramType.createCellContent(FeatureShapeMobilogramType.java:38)
	at io.github.mzmine.datamodel.features.types.tasks.NodeGenerationThread.run(NodeGenerationThread.java:106)
	at io.github.mzmine.taskcontrol.impl.WrappedTask.run(WrappedTask.java:169)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:328)
```

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/PlotCursorConfigModel.java
robinschmid: First solution just checks for an exact match in the datapoints and otherwise just set the position without item. 

Still have an issue now with the ChartGroup that links the crosshair by mouse movement. 
I think this should be two separate crosshairs maybe? 
Like one that highlights the mouse movement (chartgroup) and one that is fixed where it was clicked or set by binding (current PlotCursorPosition)

I like how the mouse movement crosshair works in the multi MS2 visualizer
image

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlotModel.java
robinschmid: removed the option to set the same renderer for all datasets. Removed the method. 
Introduced methods with XYDatasetAndRenderer

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/PlotCursorConfigModel.java
SteffenHeu: yeah, i think for the chart group use case, we can use a value marker or so

robinschmid: CHartGroup now adds a permanent marker to all the charts. 
Permanent markers are those that are never removed/cleared but only added once. 
They can change visibility and value and this will fire a chart update event through listeners to the properties

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
```diff
@@ -0,0 +1,191 @@
...
+    setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
+
+    plot = createChart();
+    // use stackpane as it is transparent / borderpane is not
+    view = new StackPane(plot);
+
+    graphicProperty().bind(emptyProperty().map(empty -> empty ? null : view));
+  }
+
+  private String getIdentifier(String action) {
```
SteffenHeu: what's this used for?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
```diff
@@ -0,0 +1,191 @@
...
+
+  @Override
+  protected void updateItem(Object o, boolean visible) {
+    // always need to call super.updateItem
+    super.updateItem(o, visible);
+
+    final ModularFeatureListRow row = getTableRow().getItem();
+    // first cell with id 0 seems to be just the measuring cell - no need to put content.
+    // the chart itself already helps measurements
+    if (id == 0) {
```
SteffenHeu, line 97: I understand this was useful for the development, but does this cell id tracking have any relevance to what is done now during production?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
```diff
@@ -0,0 +1,191 @@
...
+        datasets.add(dataset);
+      }
+    }
+
+    if (datasets.isEmpty()) {
+      plot.removeAllDatasets();
+      return;
+    }
+
+    datasets.trimToSize();
```
SteffenHeu: trimming necessary?

mzmine-community/src/main/java/io/github/mzmine/util/RangeUtils.java
```diff
@@ -448,4 +448,43 @@ public static <T extends Number & Comparable<?>> String toString(Range<T> range,
...
+    if (b == null) {
+      return a;
+    }
+    return a.span(b);
+  }
+
+  public static Range<Float> multiplyGrow(Range<Float> range, float factor) {
+    final float length = rangeLength(range);
+    final float diff = (length * factor) - length;
+    return Range.closed(range.lowerEndpoint() - diff, range.upperEndpoint() + diff);
```
SteffenHeu, line 469: This increases the range by two times the factor because the diff is added to both sides. I'd say that's not the expected behaviour

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
```diff
@@ -0,0 +1,191 @@
...
+    plot.applyWithNotifyChanges(false, true, () -> {
+      plot.setDatasets(datasets);
+      try {
+        final XYPlot xyplot = plot.getXYPlot();
+        xyplot.getRangeAxis().setRange(new Range(0, finalMaxHeight), true, false);
+        xyplot.getRangeAxis().setUpperMargin(0.025);
+        xyplot.getDomainAxis().setRange(rtRange, true, false);
+        xyplot.getDomainAxis().setDefaultAutoRange(rtRange);
+      } catch (NullPointerException | NoSuchElementException ex) {
+        // error in jfreechart draw method
```
SteffenHeu: not strictly in the draw call, as the draw is not executed within the runnable we pass here but after the runnable (since we disable the notify). Should we move the try catch into the applyWithNotifyChangesMethod?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxBaseChartModel.java
```diff
@@ -0,0 +1,113 @@
...
+import io.github.mzmine.gui.chartbasics.gui.javafx.DelayedChartDrawAdapter;
+import java.util.function.Consumer;
+import javafx.beans.value.ObservableValue;
+import org.jetbrains.annotations.Nullable;
+import org.jfree.chart.JFreeChart;
+import org.jfree.chart.plot.Plot;
+
+/**
+ * Provides some methods for delayed update calls on charts and plots. This is not always needed as
+ * we use the {@link DelayedChartDrawAdapter} but can still help to limit the number of events if
```
SteffenHeu: i dont see any methods for delayed draws in there. Were they moved?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlot.java
```diff
@@ -0,0 +1,837 @@
...
+  public void setSingleDomainRangeMarker(@Nullable Marker domain, @Nullable Marker range) {
+    plotModel.setSingleDomainRangeMarker(domain, range);
+  }
+
+  /**
+   * Prefer method to set all markers at once or use the apply with notify changes method for single
+   * update calls
+   *
+   */
+  public void setAllDomainMarkers(MarkerDefinition... markers) {
```
SteffenHeu, line 625: maybe encourage use of FxValueMarker in the jdoc in this class?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxValueMarker.java
```diff
@@ -0,0 +1,204 @@
...
+import javafx.beans.property.ReadOnlyFloatWrapper;
+import javafx.beans.property.SimpleBooleanProperty;
+import javafx.beans.property.SimpleFloatProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jfree.chart.plot.ValueMarker;
+
+/**
+ * A {@link ValueMarker} that can be configured in JavaFX. Is invisible if visible false, value
```
SteffenHeu: "An observable ValueMarker"?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/PlotCursorConfigModel.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.function.Supplier;
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.ObjectProperty;
+import javafx.beans.property.SimpleBooleanProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jfree.data.xy.XYDataset;
+
+public class PlotCursorConfigModel {
```
SteffenHeu, line 54: add short description that this class coordinates cursor and crosshair maybe?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/DelayedChartDrawAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+import javafx.animation.PauseTransition;
+import javafx.beans.property.ObjectProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import javafx.util.Duration;
+import org.jetbrains.annotations.NotNull;
+import org.jfree.chart.JFreeChart;
+import org.jfree.chart.event.ChartChangeEvent;
+import org.jfree.chart.event.ChartChangeListener;
+
+public class DelayedChartDrawAdapter implements ChartChangeListener {
```
SteffenHeu, line 59: add desc about accumulating update events

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/FxChartFactory.java
```diff
@@ -0,0 +1,477 @@
...
+import org.jfree.chart.urls.StandardXYURLGenerator;
+import org.jfree.chart.urls.StandardXYZURLGenerator;
+import org.jfree.chart.urls.XYURLGenerator;
+import org.jfree.chart.util.Args;
+import org.jfree.data.xy.IntervalXYDataset;
+import org.jfree.data.xy.TableXYDataset;
+import org.jfree.data.xy.XYDataset;
+import org.jfree.data.xy.XYZDataset;
+
+public class FxChartFactory {
```
SteffenHeu, line 63: add top level desc that this is preferred over the JFreechart chart ffactory

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/datasets/XYDatasetAndRenderer.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import org.jfree.chart.renderer.xy.XYItemRenderer;
+import org.jfree.data.xy.XYDataset;
+
+/**
+ * A more general implementation of {@link XYDataset} and {@link XYItemRenderer} group. A more
+ * specialized implementation is available in {@link DatasetAndRenderer}
+ */
+public sealed interface XYDatasetAndRenderer permits DatasetAndRenderer,
```
SteffenHeu, line 36: good to add this. I know intellij has a way to scan for usages of DatasetAndRenderer and try to replace it by XYDatasetAndRenderer. Did you run that already?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/datasets/DatasetAndRenderer.java
```diff
@@ -25,12 +25,23 @@
...
 import io.github.mzmine.gui.chartbasics.simplechart.providers.PlotXYDataProvider;
 import io.github.mzmine.gui.chartbasics.simplechart.providers.PlotXYZDataProvider;
 import org.jfree.chart.renderer.xy.XYItemRenderer;
 
-public record DatasetAndRenderer(ColoredXYDataset dataset, XYItemRenderer renderer) {
+/**
+ * If possible use {@link XYDatasetAndRenderer} that is used more generally for {@link FxXYPlot}.
+ * This class uses specialized {@link ColoredXYDataset} and is not applicable to all XYPlot.
+ * <p>
+ * TODO need to rename this class to something more specific like ColoredXYDatasetAndRenderer and maybe move to be an inner class of {@link XYDatasetAndRenderer}
```
SteffenHeu, line 38: i agree

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
robinschmid: good catch. 
Removed. Was used for debugging how many cells are repainted etc. 
This way I found out that cell 0 is always the measuring cell and should not be updated

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
robinschmid: Yeah the updateItem method is called too often for the measuring cell but we already have the correct size height of the cell. 
Somehow it just puts many different value in the measuring cell and this was lagging.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
robinschmid: Maybe now it is not needed anymore because of the delayed rendering. 
In regular plot and chart class there are some events that still come through and start a render process although notify=false.

At least for the charts with delayed render this does not happen anymore.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
robinschmid: removed

mzmine-community/src/main/java/io/github/mzmine/util/RangeUtils.java
robinschmid: Good catch. Not sure how I got to this implementation.

robinschmid: now:
```java
    final float length = rangeLength(range);
final float diff = length * (factor - 1) * 0.5f;
    return Range.closed(range.lowerEndpoint() - diff, range.upperEndpoint() + diff);
```

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxBaseChartModel.java
robinschmid: clarified. Those methods are the notify later. 
But added more comment that EChartViewer attaches the DelayedChartDrawAdapter

robinschmid: also added javadoc to both other classes

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/graphicalnodes/ChromatogramFeatureShapeCell.java
robinschmid: removed

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/FxXYPlot.java
robinschmid: added javadoc: 
```java
   * Prefer {@link FxMarker} for markers that may change their visibility or value.
```

robinschmid: also to all other marker methods

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gui/javafx/model/PlotCursorConfigModel.java
robinschmid: Done
```java

/**
 * This class coordinates the crosshair (already available in base JFreeChart in {@link XYPlot}) and
 * the {@link PlotCursorPosition} introduced in some mzmine chart classes.
 * <p>
 * The {@link XYPlot} crosshair is not used and is invisible. Instead, we use
 * {@link #getDomainCursorMarker()} and {@link #getRangeCursorMarker()} to show lines where the user
 * clicks or where the crosshair moves.
 * <p>
 * On click, or on {@link XYPlot#setDomainCrosshairValue(double)}, the
 * {@link #cursorPositionProperty()} is usually changed to the closest data point and the markers
 * are moved as well. The markers may also be bound to other markers in other plots to automatically
 * synchronize their location.
 */
```

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/FxChartFactory.java
robinschmid:
```java

/**
 * This is the preferred alternative to {@link ChartFactory} and generated observable versions of
 * the same charts and plots as {@link FxJFreeChart} and {@link FxXYPlot} that already come with
 * observability, optimized reduced chart draw events, and more.
 */
```

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/datasets/XYDatasetAndRenderer.java
robinschmid: just tried that. Works really bad in this case. 
Many obvious and other hidden issues. 
Many of the plots actually need the ColoredDataset but the refactor function does not see that.

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/simplechart/datasets/DatasetAndRenderer.java
robinschmid: Lets do this in a second PR after this one

Issue #2854, opened 2025-09-03
Spectral Library Searching of negative mode data no longer operational


michael-cowled:
### Basic information

* My operating system and version: Windows Server 2022 Standard V21H2 (but also on Windows 10 Education V22H2)
* My mzmine version: Tried on both 4.7.27 and 4.7.29

### What happened

* After updating to the newest version (from 4.7.8 previously), spectral matching of negative mode data seems to skip. I have confirmed for batches where I have observed matches previously in both positive and negative mode data on V4.7.8, 4.7.27 and 4.7.29 and only 4.7.8 returns results for negative mode, and positive mode matches are still obtained for all modes.
* Batch file can be provided upon request, but I believe the issue to be applicable to all negative mode batches.

michael-cowled:
Hi @SteffenHeu,

Please see attached the data, batch and a subset of the library (which I have confirmed works still on the old version as demonstrated in the screenshots). Recreating the batch from scratch with the wizard was also unsuccessful.

https://drive.google.com/open?id=1dLYJsaQay7udvD9Ee4djj2UQx2RGDCHh&usp=drive_fs

Best regards,
Michael

michael-cowled:
Hi @SteffenHeu,

Oh, sounds like an easy intermediary fix for us then.

The library was created in-house using the Batch spectral library generation module in mzmine v4.7.8 and exported as an .msp file (.json seems to be using a different naming convention for the ion.mode, so that seems to be fine). I'm not sure if this is still occuring for spectral libraries created in the new version, and at least hopefully isn't affecting any of the big public libraries!

PR #2863, opened 2025-09-09
fix spectral library search weak polarity check if spectral library is parsed as neutral

SteffenHeu:
addresses #2854


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_spectral_library_match/RowsSpectralMatchTask.java
```diff
@@ -731,13 +727,15 @@ public int getErrorCount() {
    * @return false if both polarities are defined and do not match, true otherwise.
    */
   public boolean weakPolarityCheck(String entryPolarityString, PolarityType scanPolarity) {
-    if (scanPolarity == null || scanPolarity == PolarityType.UNKNOWN) {
+    if (scanPolarity == null || scanPolarity == PolarityType.UNKNOWN
+        || scanPolarity == PolarityType.ANY || scanPolarity == PolarityType.NEUTRAL) {
```
robinschmid, line 731: This could be simplified to the null tolerant 
```java
if(!PolarityType.isDefined(scanPolarity))
```

robinschmid (approved):
Great this was still important to change :)

Issue #2851, opened 2025-09-02
CLI mode: saving large project fails to copy raw files


kaisengit:
### Basic information

* My operating system and version: Ubuntu 24.04 (docker)
* My mzmine version: 4.7.27

### What happened

* When trying to save a large project via the Linux CLI mode, mzmine fails to copy the raw files into the project dir:

```
[2025-09-02, 15:17:39 UTC] {docker.py:436} INFO - 2025-09-02 15:17:39 FINEST io.github.mzmine.modules.io.projectsave.RawDataFileSaveHandler copyRawDataFilesToZip Copying data file /files/converted_files/42/Eawag_250218_pos_STD50A.mzML to project file.
[2025-09-02, 15:17:41 UTC] {docker.py:436} INFO - 2025-09-02 15:17:41 FINEST io.github.mzmine.modules.io.projectsave.RawDataFileSaveHandler copyRawDataFilesToZip Copying data file /files/converted_files/42/Eawag_250218_pos_STD500A.mzML to project file.
[2025-09-02, 15:17:42 UTC] {docker.py:436} INFO - 2025-09-02 15:17:42 FINEST io.github.mzmine.modules.io.projectsave.RawDataFileSaveHandler copyRawDataFilesToZip Copying data file /files/converted_files/42/Eawag_250218_pos_Chrisb500A.mzML to project file.
[2025-09-02, 15:17:44 UTC] {docker.py:436} INFO - 2025-09-02 15:17:44 FINEST io.github.mzmine.modules.io.projectsave.RawDataFileSaveHandler copyRawDataFilesToZip Copying data file /files/converted_files/42/Eawag_250218_pos_Blind_A.mzML to project file.
[2025-09-02, 15:17:45 UTC] {docker.py:436} INFO - 2025-09-02 15:17:45 FINEST io.github.mzmine.modules.io.projectsave.RawDataFileSaveHandler copyRawDataFilesToZip Copying data file /files/converted_files/42/Eawag_250218_pos_blank.mzML to project file.
[2025-09-02, 15:17:47 UTC] {docker.py:436} INFO - 2025-09-02 15:17:47 WARNING io.github.mzmine.project.impl.MZmineProjectImpl getRelativePath Cannot relativize path /files/$$msdatafiles/Eawag_250218_pos_STD500A.mzML$$ to project file /files/mzmine_output/42/results_pos.mzmine. Files may be located on a different drive.
[...]

Issue #2712, opened 2025-07-10
Failing to open existing project or get feature lists


vgorova:
### Basic information

* My operating system and version: Windows 11
* My mzmine version: 4.5.37

### What happened

I ran HilicNeg analysis in April this year, exported feature lists and other files, but now when I try to open the project and check the feature lists, I am getting this error message (see attached the picture and log file here).

https://drive.google.com/drive/folders/1yiNZGvt6KW_NmLmIXeL2QZoxQNG9kUnu?usp=drive_link

Please find the batch and error log. Is this sufficient information or do you need any raw files? I probably cannot share the actual data..

Thank you for the help!

<img width="390" height="480" alt="Image" src="https://github.com/user-attachments/assets/1f569d98-8ed4-4c15-98e4-b3f65cd65675" />

SteffenHeu:
Yes, you can disable the "Metadata file" checkbox in the import step.

SteffenHeu:
Sorry i thought you were using a batch file, because that one is uploaded in the drive.
For the project: you can rename it from .mzmine to .zip, then extract the raw_data_import_batch.xml, open it in mzmine as a batch file, disable the metadata import, save it as batch file, move it into the .zip (rename to .xml instead of .mzbatch before) and then rename the project again to .mzmine.

vgorova:
> Sorry i dont know what you mean.
> 
> 1. rename project from .mzmine to .zip
> 2. extract the raw_data_import_batch.xml out of the zip file
> 3. in mzmine, open the batch mode, click load, load the "raw_data_import_batch.xml" as a batch file
> 4. in the loaded batch, disable the metadata import in the data import step
> 5. save the batch as a file with the name "raw_data_import_batch". new batches will be called ".mzbatch" instead of .xml. so before moving it into the project .zip file rename it to .xml instead of .mzbatch
> 6. move the new "raw_data_import_batch.xml" into the project zip file
> 7. then rename the project again to .mzmine
> 8. open via load project.

I am sorry, but I am unable to follow. I managed by step 5. 
Then the situation is that I have one .zip file and one extracted one, with the new .mzbatch → .xml file. 

I tried 1) moving this into .zip file - it does only "compressing" but the loading window never conducts the work and the system freezes and I have to hard close it 
or 2) copying the .xml file into extracted zip file, zipping it, then renaming to .mzmine, but then it does not work as project, but mzMine still sees it as "zip". Is there anything I am missing and doing incorrectly? Thank you very much!

[...]

PR #2867, opened 2025-09-10
Copy files from FileNameParameter into mzmine project

SteffenHeu:
Copies files from a FileNameParameter (and Optional file name parameters) into the mzmine project when "Standalone" is used. This fixes an issue where standalone projects cannot be opened due to missing metadata files.

#2851 (partial fix)
fixes future occurrences of #2712

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/projectsave/RawDataFileSaveHandler.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/projectsave/RawDataFileSaveHandler.java
@@ -31,6 +31,8 @@ import io.github.mzmine.modules.MZmineProcessingModule;
 import io.github.mzmine.modules.MZmineProcessingStep;
 import io.github.mzmine.modules.batchmode.BatchQueue;
 import io.github.mzmine.parameters.Parameter;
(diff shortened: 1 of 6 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/io/projectsave/RawDataFileSaveHandler.java
```diff
@@ -310,6 +326,23 @@ private void replaceRawFilePaths(List<BatchQueue> queues) {
             }
             selection.setSelectionType(RawDataFilesSelectionType.SPECIFIC_FILES);
             selection.setSpecificFiles(placeholders);
+          } else if ((parameter instanceof FileNameParameter fnp && fnp.getValue() != null) || (
+              parameter instanceof OptionalParameter<?> op
+                  && op.getEmbeddedParameter() instanceof FileNameParameter && op.isSelected())) {
+            // eg like metadata
+            final FileNameParameter fnp = switch (parameter) {
+              case FileNameParameter p -> p;
```
robinschmid, line 335: maybe add the case for FileNameSuffixExportParameter -> null ? Otherwise files exported in this batch may also be added to the project

PR #2872, opened 2025-09-12
More flexible MSP reader

robinschmid:
- add formats from MassBank exports that are named .msp_RIKE and .msp_NIST
- add to drag and drop import
- handle N as negative polarity as mzmine <4.8 exported N as the polarity which is now handled internally as neutral
- add map to handle alternative keys for DBEntryField, even in msp there are various forms the keys are written
- handle MS level defined as MS and MS2 instead of just an integer

### DIFF
--- a/mzmine-community/src/test/java/io/github/mzmine/util/spectraldb/entry/DBEntryFieldTest.java
+++ b/mzmine-community/src/test/java/io/github/mzmine/util/spectraldb/entry/DBEntryFieldTest.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2004-2025 The mzmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+ * OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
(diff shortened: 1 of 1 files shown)


mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/DBEntryField.java
```diff
@@ -280,15 +345,22 @@ public static DBEntryField forJdxID(String key) {
...
-      case CommentType _ -> DBEntryField.COMMENT;
-      case ClassyFireSuperclassType _ -> DBEntryField.CLASSYFIRE_SUPERCLASS;
-      case ClassyFireClassType _ -> DBEntryField.CLASSYFIRE_CLASS;
-      case ClassyFireSubclassType _ -> DBEntryField.CLASSYFIRE_SUBCLASS;
-      case ClassyFireParentType _ -> DBEntryField.CLASSYFIRE_PARENT;
-      case NPClassifierSuperclassType _ -> DBEntryField.NPCLASSIFIER_SUPERCLASS;
-      case NPClassifierClassType _ -> DBEntryField.NPCLASSIFIER_CLASS;
-      case NPClassifierPathwayType _ -> DBEntryField.NPCLASSIFIER_PATHWAY;
-      case PubChemIdType _ -> DBEntryField.PUBCHEM;
+      case CommentType _ -> io.github.mzmine.util.spectraldb.entry.DBEntryField.COMMENT;
```
SteffenHeu: why the explicit package?

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/DBEntryField.java
```diff
@@ -183,6 +185,69 @@ QUALITY_CHIMERIC, QUALITY_EXPLAINED_INTENSITY(Float.class), QUALITY_EXPLAINED_SI
 
   private static final Logger logger = Logger.getLogger(DBEntryField.class.getName());
 
+  /**
+   * A map of lower case keys
+   */
+  private static final Map<String, DBEntryField> FIELD_ALTERNATIVE_KEYS = HashMap.newHashMap(
```
SteffenHeu, line 191: not sure if it is a good solution to put all the formats into the same map. maybe it's better to have it separated and only have the alternative keys added to all?

robinschmid: I thought of this map rather as a general map that describes all keywords that uniquely identify a field. 
There is already one case where key INSTRUMENT is used in GNPS batch submission to describe the INSTRUMENT_TYPE like qTOF instead of instrument name. 

The function to add new mappings already checks for keys that describe different DBEntryFields and warns us. Currently there no other conflicts. 
In case of conflict in a parser, I think that parser that strays from the default meaning of a key should implement the workaround.

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/DBEntryField.java
robinschmid: this is weird. 
I asked the LLM to generate a draft for the new methods but I think I rejected it as it was bad. 
But maybe I did accept something else then? 
I do have issues lately that intellij would sometimes just not import the class but use the whole reference. But this is not it.

robinschmid: reverted

PR #2912, opened 2025-10-16
Filename parameter avoid empty file 

robinschmid:
- was creating new File("") if textfield was empty
- saving and loading from xml converted this to a relative path


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/filenames/FileNameComponent.java
```diff
@@ -173,8 +173,7 @@ public void setLastFiles(List<File> value) {
 
   public File getValue() {
     String fileName = txtFilename.getText();
-    File file = new File(fileName);
-    return file;
+    return fileName.isBlank() ? null : new File(fileName);
```
SteffenHeu: TextField.getText may be nullable, at least when i go into the method
image

PR #2918, opened 2025-10-17
Add comment column and filter

robinschmid:
- add comment column
- add filter for comment column
- add ANY and ALL matching modes for strings to match all words (separated by string)

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
@@ -418,7 +418,8 @@ public class ModularFeatureListRow extends ColumnarModularDataModelRow implement
     final String manualComment = manual == null ? null : manual.getComment();
 
     // added in mzmine 4.8 the CommentType is now a row type and should replace the manual annotation
-    // for now concatenate the comments from both
+    // for now concatenate the comments from both to support old projects
+    // may remove the concat in future releases
     final String comment = get(CommentType.class);
     final boolean hasManual = StringUtils.hasValue(manualComment);
     final boolean hasComment = StringUtils.hasValue(comment);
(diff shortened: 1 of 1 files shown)


mzmine-community/src/main/java/io/github/mzmine/util/DataTypeUtils.java
```diff
@@ -73,13 +74,19 @@
...
+  public static final List<DataType> DEFAULT_CHROMATOGRAPHIC_ROW = DataTypes.getAll(
       // needed next to each other for switching between RTType and RTRangeType
-      new MZType(), new MZRangeType(), //
-      new HeightType(), new AreaType(), new ManualAnnotationType(), new FeatureShapeType(),
-      new AreaBoxPlotType(), new HeightBoxPlotType());
+      RTType.class, RTRangeType.class, //
+      MZType.class, MZRangeType.class, //
+      HeightType.class, AreaType.class,
+      // added CommentType as default in 4.8 to transition away from ManualAnnotationType
+      CommentType.class,
```
robinschmid, line 86: only change was that CommentType was added.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
```diff
@@ -410,9 +412,22 @@ public boolean hasRowType(Class typeClass) {
...
-    return manual == null ? null : manual.getComment();
+    final String manualComment = manual == null ? null : manual.getComment();
+
+    // added in mzmine 4.8 the CommentType is now a row type and should replace the manual annotation
+    // for now concatenate the comments from both
+    final String comment = get(CommentType.class);
+    final boolean hasManual = StringUtils.hasValue(manualComment);
+    final boolean hasComment = StringUtils.hasValue(comment);
+    if (hasManual && hasComment) {
+      return comment + " " + manualComment;
```
SteffenHeu, line 427: is the row filter the reason for this? i think i would rather put the concatenation into the feature list filter and just return the result of get(CommentType.class) here

robinschmid: done, project load does this. See other PR that removes manualAnnotation by default. 
#2920

PR #2927, opened 2025-10-21
ims: refactor frame ms2 info addition

SteffenHeu:
msms infos were added through an exposed hash set. refactored and added a method to add the msms infos, made the exposed hash set unmodifiable


mzmine-community/src/main/java/io/github/mzmine/datamodel/impl/SimpleFrame.java
```diff
@@ -170,7 +171,7 @@ public double getMobilityForMobilityScanNumber(int mobilityScanIndex) {
   @NotNull
   @Override
   public Set<IonMobilityMsMsInfo> getImsMsMsInfos() {
-    return precursorInfos;
+    return Set.copyOf(precursorInfos);
```
robinschmid: I think a good option istead of the copy is to use an unmodifiable wrapper that avoids the copy. Only issue might be modification exceptions in case multiple threads try to add/read the content.
Collections.unmodifiableSet

So both works for me. I feel like this method does not need to provide multi threading secure implementation of the collection. 
And it may be called a lot to get the precursor mz etc.

robinschmid (approved):

PR #2961, opened 2025-10-30
parallelised formula prediction

SteffenHeu:


gradle/libs.versions.toml
```diff
@@ -9,7 +9,7 @@ beryx-runtime = "2.0.0"
 semver = "1.0.4"
 msdk = "0.0.27"
 # update to 2.9? from 2.5
-cdk = "2.11"
+cdk = "2.12-SNAPSHOT"
```
robinschmid, line 12: Just something to remember. 
Before we merge it would be best if this was released or if we put a local copy of one version somewhere in local libs or so. 
Otherwise this dependency will change too often with undefined consequences

mzmine-community/build.gradle
```diff
@@ -141,6 +141,9 @@ repositories {
 
     // for batik-xml-commons-ext package (svg export)
     maven { url = "https://repository.everit.biz/nexus/content/groups/public/" }
+
+    // for cdk snapshot
+    maven {url = "https://central.sonatype.com/repository/maven-snapshots/"}
```
robinschmid: same for the repo. Maybe best to use local copy of lib

mzmine-community/src/test/java/util/FormulaGenTest.java
```diff
@@ -0,0 +1,110 @@
...
+              MolecularFormulaManipulator.getString(f), mz));
+          if (!mzRange.contains(mz)) {
+            throw new RuntimeException(mz + " not within range " + mzRange.toString());
+          }
+        }
+      };
+      runnables.add(runnable);
+    }
+
+    TaskService.init(masses.length);
```
robinschmid: maybe better to initialize with a fixed number or number of cores?
Update: 
I found a better option. No need to initialize the TaskService you can just use a FixedThreadPoolTask with the number of threads you want.

Also this can simplify the code below to wait for the task to finish and then adding an assertion for the final state. 
So far the assertion was missing. 

```java

    final List<SimpleRunnableTask> tasks = runnables.stream().map(SimpleRunnableTask::new).toList();
    final ThreadPoolTask poolTask = new FixedThreadPoolTask("Main formula task", masses.length,
        tasks);
    final WrappedTask wrappedTask = TaskService.getController()
        .runTaskOnThisThreadBlocking(poolTask);

    Assertions.assertEquals(TaskStatus.FINISHED, wrappedTask.getStatus());
```

SteffenHeu: the test was to check if the parellel generation works. it even workled without the bugfix, so i'll just delete this test as it was not working as expected.

robinschmid (approved):

PR #2968, opened 2025-10-30
Thermo Raw File Parser update. Also, make signing, notarization and stapling for MacOS great again.

JonasLueb:
TRFP is now pulled on demand in build pipeline from https://github.com/CompOmics/ThermoRawFileParser/releases/. The new version is able to be run without mono and other external dependencies, runs as build from CompOmics Repo. 
JDK was updated to version 25. MacOS signing, notarization and stapling was updated accordingly.

### DIFF
--- a/mzmine-community/build.gradle
+++ b/mzmine-community/build.gradle
@@ -485,6 +485,7 @@ task signApp(dependsOn: jpackage) {
             // Remove non-mac executables from external tools to reduce Gatekeeper friction
             try {
                 Path toolsDir = imageDir.resolve('Contents').resolve('external_tools').resolve('thermo_raw_file_parser')
+                // should be not needed anymore as we are downloading the files online now.
                 Files.deleteIfExists(toolsDir.resolve('ThermoRawFileParserLinux'))
                 Files.deleteIfExists(toolsDir.resolve('ThermoRawFileParser.exe'))
             } catch (Exception ignored) {
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
@@ -174,8 +174,7 @@ public class ThermoRawImportTask extends AbstractTask implements RawDataImportTa
     if (Platform.isWindows() && !thermoRawFileParserCommand.endsWith("ThermoRawFileParser.exe")) {
       error("Invalid raw file parser setting for windows. Please select the windows parser.");
       return null;
-    } else if (Platform.isLinux() && !thermoRawFileParserCommand.endsWith(
-        "ThermoRawFileParserLinux")) {
+    } else if (Platform.isLinux() && !thermoRawFileParserCommand.endsWith("ThermoRawFileParser")) {
       error("Invalid raw file parser setting for linux. Please select the linux parser.");
       return null;
     } else if (Platform.isMac() && !thermoRawFileParserCommand.endsWith("ThermoRawFileParser")) {
@@ -236,16 +235,16 @@ public class ThermoRawImportTask extends AbstractTask implements RawDataImportTa
     File parserDirectory = FileAndPathUtil.resolveInExternalToolsDir("thermo_raw_file_parser/");
     if (!parserDirectory.exists()) {
       throw new IllegalStateException(
-          "ThermoRawFileParser directory not found. Expected one of: '<app>/external_tools/thermo_raw_file_parser/', 'external_tools/thermo_raw_file_parser/', '../external_tools/thermo_raw_file_parser/'.");
+          "ThermoRawFileParser directory not found. When running from the IDE run gradle build or gradle test before to download the thermo raw file parser to the external_tools directory."
+              + "Expected one of: '<app>/external_tools/thermo_raw_file_parser/', 'external_tools/thermo_raw_file_parser/', '../external_tools/thermo_raw_file_parser/'.");
     }
 
     if (Platform.isWindows()) {
       return new File(parserDirectory, "ThermoRawFileParser.exe");
     }
-    if (Platform.isLinux()) {
-      return new File(parserDirectory, "ThermoRawFileParserLinux");
-    }
-    if (Platform.isMac()) {
+    if (Platform.isLinux() || Platform.isMac()) {
+      // both platforms have different runners but the same file name
+      // mzmine >4.8.0 download only the matching dependency now
       return new File(parserDirectory, "ThermoRawFileParser");
     }
     throw new IllegalStateException(
--- a/mzmine-community/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/gui/preferences/MZminePreferences.java
@@ -246,11 +246,10 @@ public class MZminePreferences extends SimpleParameterSet {
       new FileNameWithDownloadParameter("Thermo raw file parser location",
           "This is the optional external location to overwrite the internal thermo raw file parsing default. Disable to use the internal parser. macOS currently requires mono installed and the external raw file parser (see download button on the right).",
           List.of(new ExtensionFilter("Executable or zip", "ThermoRawFileParser.exe",
-                  "ThermoRawFileParserLinux", "ThermoRawFileParserMac", "ThermoRawFileParser.zip"),
+                  "ThermoRawFileParser", "ThermoRawFileParser.zip"),
               new ExtensionFilter("zip", "ThermoRawFileParser.zip"),
               new ExtensionFilter("Windows executable", "ThermoRawFileParser.exe"),
-              new ExtensionFilter("Linux executable", "ThermoRawFileParserLinux"),
-              new ExtensionFilter("Mac executable", "ThermoRawFileParserMac")),
+              new ExtensionFilter("Linux / macOS executable", "ThermoRawFileParser")),
           AssetGroup.ThermoRawFileParser));
 
   public static final OptionalParameter<ParameterSetParameter<WatersLockmassParameters>> watersLockmass = new OptionalParameter<>(
--- a/utils/src/main/java/io/github/mzmine/util/files/FileAndPathUtil.java
+++ b/utils/src/main/java/io/github/mzmine/util/files/FileAndPathUtil.java
@@ -789,31 +789,19 @@ public class FileAndPathUtil {
    */
   private static @NotNull File getExternalToolsDir() {
     final File mainDir = FileAndPathUtil.getSoftwareMainDirectory();
-    File toolsDirectory = null;
     if (mainDir != null) {
       File extAtAppRoot = new File(mainDir, "external_tools/");
       if (extAtAppRoot.exists()) {
-        toolsDirectory = extAtAppRoot;
-      }
-    }
-    // Prior behavior when running from project root
-    if (toolsDirectory == null) {
-      File extLocal = new File("external_tools/");
-      if (extLocal.exists()) {
-        toolsDirectory = extLocal;
+        return extAtAppRoot;
       }
     }
     // Dev-run from module dir: parent project root
-    if (toolsDirectory == null) {
-      File extParent = new File("../external_tools/");
-      if (extParent.exists()) {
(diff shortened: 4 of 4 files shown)


mzmine-community/build.gradle
```diff
@@ -469,95 +559,143 @@ task signApp(dependsOn: jpackage) {
...
+                }
+            }
+
+            // Explicitly sign embedded executables (e.g., ThermoRawFileParser)
+            try {
+                Path trfpPath = imageDir.resolve('Contents').resolve('external_tools').resolve('thermo_raw_file_parser').resolve('ThermoRawFileParser')
+                if (Files.exists(trfpPath)) {
+                    getLogger().lifecycle("Signing embedded executable '${trfpPath}'")
+                    // Ensure executable bit
+                    execOps.exec { commandLine(['chmod', '+x', trfpPath.toFile().absolutePath]) }
```
SteffenHeu: the same command is already executed in line 522 or? Comment there also says sign executables

mzmine-community/build.gradle
```diff
@@ -566,6 +704,42 @@ task notarizeApp(dependsOn: signApp) {
 }
 notarizeApp.mustRunAfter(signApp)
 
+// macOS only: optionally fetch TRFP (.NET 8) release asset during build, overlay into image
+// No effect on Windows/Linux builds.
+tasks.register('fetchThermoRawFileParserMac') {
```
SteffenHeu: If all the local files are deleted, and the parser is only downloaded on mac, then windows and linux builds will not run, or?

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
```diff
@@ -227,24 +227,42 @@ public void cancel() {
...
+    }
+    if (parserDirectory == null) {
+      throw new IllegalStateException(
+          "ThermoRawFileParser directory not found. Expected one of: '<app>/external_tools/thermo_raw_file_parser/', 'external_tools/thermo_raw_file_parser/', '../external_tools/thermo_raw_file_parser/'.");
     }
 
     if (Platform.isWindows()) {
-      return FileAndPathUtil.resolveInExternalToolsDir(
-          "thermo_raw_file_parser/ThermoRawFileParser.exe");
+      return new File(parserDirectory, "ThermoRawFileParser.exe");
```
SteffenHeu, line 243: for windows and mac the above logic (230-255) is not required when running from the IDE or the application. I'd prefer to keep the previous      
 ```
return FileAndPathUtil.resolveInExternalToolsDir(
          "thermo_raw_file_parser/ThermoRawFileParser.exe/Linux");
```

where possible and only do the manual resolving in the mac branch

File: mzmine-community/build.gradle (hunk 2 above)
Comment (SteffenHeu) [reply mzmine-community/build.gradle]: Perhaps we can move the parser for mac into external_tools/thermo_raw_file_parser/mac/... for the time we use a different parser on mac than on linux and windows

File: mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java (hunk 3 above)
Comment (JonasLueb) [reply mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java:243]: I could not test this on windows, but it should now work in IDE and Application no matter which system. Are you talking abpout ab the mac "branch" inside the code?

File: mzmine-community/build.gradle (hunk 2 above)
Comment (JonasLueb) [reply mzmine-community/build.gradle]: This task only affects the macos build, see line 710

File: .github/workflows/mac_branch_test_build.yml [hunk 4]
```diff
@@ -93,18 +93,35 @@ jobs:
...
-          security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k abc123 build.keychain
+          KEYCHAIN_NAME=build.keychain
+          KEYCHAIN_PATH="$HOME/Library/Keychains/${KEYCHAIN_NAME}-db"
+          security create-keychain -p abc123 "$KEYCHAIN_NAME"
+          security set-keychain-settings -lut 21600 "$KEYCHAIN_PATH"
+          security default-keychain -s "$KEYCHAIN_PATH"
+          security unlock-keychain -p abc123 "$KEYCHAIN_PATH"
+          security import certificate.p12 -k "$KEYCHAIN_PATH" -P "$MACOS_DEVELOPER_CERTIFICATE_PWD" -A -f pkcs12 -T /usr/bin/codesign -T /usr/bin/security
+          security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k abc123 "$KEYCHAIN_PATH"
+          security list-keychains -d user -s "$KEYCHAIN_PATH" "$HOME/Library/Keychains/login.keychain-db"
```
robinschmid, line 104: this line just lists the keychains and prints details for both the build and the login keychain?

.github/workflows/mac_branch_test_build.yml
```diff
@@ -93,18 +93,35 @@ jobs:
...
+          fi
+          export MACOS_SIGNING_KEYCHAIN="$KEYCHAIN_PATH"
           ./gradlew -p mzmine-community --info --warning-mode all --stacktrace
         env:
           MACOS_DEVELOPER_CERTIFICATE: ${{ secrets.MACOS_DEVELOPER_CERTIFICATE }}
           MACOS_DEVELOPER_CERTIFICATE_PWD: ${{ secrets.MACOS_DEVELOPER_CERTIFICATE_PWD }}
           MACOS_APPLE_ID_PWD: ${{ secrets.MACOS_APPLE_ID_PWD }}
           MACOS_APP_IDENTITY_ID: ${{ secrets.MACOS_APP_IDENTITY_ID }}
           MACOS_APP_IDENTITY_TEAM_ID: ${{ secrets.MACOS_APP_IDENTITY_TEAM_ID }}
+          MACOS_SIGNING_KEYCHAIN: /Users/runner/Library/Keychains/build.keychain-db
```
robinschmid: this is not needed due to this line correct?
`export MACOS_SIGNING_KEYCHAIN="$KEYCHAIN_PATH"`

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
```diff
@@ -227,24 +227,42 @@ public void cancel() {
   }
 
   private File getParserPathForOs() {
-    final Optional<File> prefPath = ConfigService.getPreferences()
-        .getOptionalValue(MZminePreferences.thermoRawFileParserPath);
```
robinschmid, line 231: We will keep this optional parameter right? 
So this should also be checked first

mzmine-community/build.gradle
SteffenHeu: Yeah that's what i mean. The TFRP is only downloaded on MacOS but the binaries it was removed from the external_tools dir. So on windows it will not be downloaded and since the binaries are not there, the import will not work, or?

That's why i suggested we could download the mac version into external_tools/thermo_raw_file_parser/mac/... and keep the linux and windows versions where they are

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
SteffenHeu: sorry i meant it's not required for windows and linux. They just work with
`return FileAndPathUtil.resolveInExternalToolsDir(
         "thermo_raw_file_parser/ThermoRawFileParser.exe/Linux");`
So the more complex resolving logic you added is only required on mac.
Perhaps it would be even better to put the path resolving you added into the FileAndPathUtil.resolveInExternalToolsDir, so we can always just use that method, even on macOS.

.github/workflows/mac_branch_test_build.yml
JonasLueb: This line adds via "-s" the two keychains to the search list. This command has no print, but I removed other unneccesary prints in line 105 - 107, 110

.github/workflows/mac_branch_test_build.yml
JonasLueb: true, removed

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
JonasLueb: true, removed this by accident!

mzmine-community/build.gradle
JonasLueb: yes, removed the redundancy. I tried to spam as much signing as possible to make it work somehow.

mzmine-community/build.gradle
JonasLueb: Ahh, sorry, got it now. I added the same mechanism for linux and windows, to harmonize it and to avoid checking in all those files for any distribution

.github/workflows/mac_branch_test_build.yml
robinschmid: oh ok makes sense

mzmine-community/build.gradle
```diff
@@ -150,6 +153,21 @@ ext {
...
     // UUID for upgrades of this package on Windows. Generated by https://www.uuidgenerator.net 
     win_uuid = "896e9c2d-6db8-4259-a1af-1b5f8112d1e1"
+
+    // ThermoRawFileParser (TRFP) download
+    // Configure in gradle.properties or via env vars to override defaults below if needed:
+    // trfp.linux.url=...
+    // trfp.win.url=...
+    trfpMacUrl = providers.gradleProperty("trfp.mac.url").orNull 
+            ?: System.getenv("TRFP_MAC_URL")
+            ?: "https://github.com/CompOmics/ThermoRawFileParser/releases/download/v.2.0.0-dev/ThermoRawFileParser-v.2.0.0-dev-osx.zip"
```
robinschmid: maybe exchange for
https://github.com/mzmine/ThermoRawFileParser/releases/tag/v2.0.0-dev

Also for the other platforms below

Version name is a bit different with v2.0.0-dev vs v.2.0.0-dev

JonasLueb: Thanks for forking, also wanted to do that today :)

JonasLueb: done

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
```diff
@@ -233,18 +233,20 @@ private File getParserPathForOs() {
       return prefPath.get();
     }
 
+    File parserDirectory = FileAndPathUtil.resolveInExternalToolsDir("thermo_raw_file_parser/");
+    if (!parserDirectory.exists()) {
+      throw new IllegalStateException(
+          "ThermoRawFileParser directory not found. Expected one of: '<app>/external_tools/thermo_raw_file_parser/', 'external_tools/thermo_raw_file_parser/', '../external_tools/thermo_raw_file_parser/'.");
```
robinschmid: When running from the IDE run gradle build or gradle test before to download the thermo raw file parser to the external_tools directory.

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_thermo_raw/ThermoRawImportTask.java
```diff
@@ -233,18 +233,20 @@ private File getParserPathForOs() {
...
+
     if (Platform.isWindows()) {
-      return FileAndPathUtil.resolveInExternalToolsDir(
-          "thermo_raw_file_parser/ThermoRawFileParser.exe");
+      return new File(parserDirectory, "ThermoRawFileParser.exe");
     }
     if (Platform.isLinux()) {
-      return FileAndPathUtil.resolveInExternalToolsDir(
-          "thermo_raw_file_parser/ThermoRawFileParserLinux");
+      return new File(parserDirectory, "ThermoRawFileParserLinux");
```
robinschmid: ThermoRawFileParser

PR #2979, opened 2025-11-03
Skip already imported thermo raw files in batches

robinschmid:
Cannot call batch twice at the moment because files are skipped for mzML extension

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_all/AllSpectralDataImportModule.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_all/AllSpectralDataImportModule.java
@@ -150,6 +150,7 @@ public class AllSpectralDataImportModule implements MZmineProcessingModule {
     final RawDataFileType type = RawDataFileTypeDetector.detectDataFileType(file);
     // we are loading thermo raw files by thermo raw file parser that keeps the .raw extension.
     // MSconvert task can in theory convert thermo .raw so skip the method call instead of changing the way msconvert task works
+    // TODO this will need to change with the data handling parameter that defines which SDK to use for each format.
     final File importedFile = type == RawDataFileType.THERMO_RAW ? file : //
         MSConvertImportTask.applyMsConvertImportNameChanges(file, keepConverted, type);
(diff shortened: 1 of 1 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_all/AllSpectralDataImportModule.java
```diff
@@ -148,8 +148,10 @@ public static ImportFile validateToActualPath(File file, boolean keepConverted)
     file = validateBrukerPath(file);
 
     final RawDataFileType type = RawDataFileTypeDetector.detectDataFileType(file);
-    final File importedFile = MSConvertImportTask.applyMsConvertImportNameChanges(file,
-        keepConverted, type);
+    // we are loading thermo raw files by thermo raw file parser that keeps the .raw extension.
+    // MSconvert task can in theory convert thermo .raw so skip the method call instead of changing the way msconvert task works
+    final File importedFile = type == RawDataFileType.THERMO_RAW ? file : //
+        MSConvertImportTask.applyMsConvertImportNameChanges(file, keepConverted, type);
```
SteffenHeu, line 155: Adapt the applyMsConvertImportNameChanges method to only apply to thermo raw files if the thermo ThermoImportOptions is set to MSCONVERT instead.

robinschmid: We removed the ThermoImportOptions from the configuration...

SteffenHeu: then thermo_raw shouldnt be listed in the supported files anymore, or?

robinschmid: I thought about it this way. 
We did not change the way the MSConvertImportTask works but only the AllSpectralDataImportModule changed. 
Thats why I would change it in the all spectra import. Or we move the method itself to the all import and change it there. 
But seems to be more related to MSConvert task. 

This will be a topic again once we integrate more SDKs.

PR #2989, opened 2025-11-06
add resolver test framework

SteffenHeu:


mzmine-community/src/test/java/resolver_tests/Peak.java
```diff
@@ -0,0 +1,76 @@
...
+
+  boolean matches(@Nullable Peak other) {
+    if (other == null) {
+      return false;
+    }
+    if (!Precision.equalFloatSignificance(topRt(), other.topRt())) {
+      return false;
+    }
+
+    if (rtRange != null && other.rtRange == null || rtRange == null && other.rtRange != null || (
```
robinschmid: not sure in which order this is evaluated actually. I think one is stronger than the other? 
Maybe add () 
or change 
`rtRange != null && other.rtRange == null || rtRange == null && other.rtRange != null`
to XOR
`rtRange == null ^ other.rtRange == null`

mzmine-community/src/test/java/resolver_tests/Peak.java
```diff
@@ -0,0 +1,76 @@
...
+        rtRange != null && other.rtRange != null && !Precision.equalFloatSignificance(
+            rtRange.lowerEndpoint(), other.rtRange.lowerEndpoint())
+            && Precision.equalFloatSignificance(rtRange.upperEndpoint(),
+            other.rtRange.upperEndpoint()))) {
```
robinschmid: the upper bound check should be || !Precision.equal ... and then put `()` around the check.  

Maybe in general flip the check to be positive this seems easier:

```java
return (rtRange == null && other.rtRange == null ) ||
    (rtRange != null && other.rtRange != null &&
    Precision.equalFloatSignificance(rtRange.lowerEndpoint(), other.rtRange.lowerEndpoint())
    Precision.equalFloatSignificance(rtRange.upperEndpoint(), other.rtRange.upperEndpoint()) && 
    );
```

SteffenHeu: i dont like huge return statements

robinschmid: Same thing if you put into an if or just return it, you can also wrap it into an 
if(...)  return true; 
else return false;

PR #2992, opened 2025-11-07
remove usage of javafx internals.

JonasLueb:


javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFields.java
```diff
@@ -226,20 +225,26 @@ public static <T> AutoCompletionBinding<T> bindAutoCompletion(TextField textFiel
...
+    final javafx.beans.property.BooleanProperty triggeredOnFocus = new javafx.beans.property.SimpleBooleanProperty(false);
     // Show all suggestions when the user clicks or when field gains focus
     textField.setOnMouseClicked(_ -> {
       // auto show completion
 //      final String text = textField.getText();
 //      acb.setUserInput(text == null || text.isBlank() ? "*" : text);
       // use this to always show full selection on first click
       acb.setUserInput("*");
+      // clicking should show suggestions regardless of previous focus trigger
+      triggeredOnFocus.set(true);
```
robinschmid, line 240: What does this solve?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/skins/DynamicTextFieldSkin.java
```diff
@@ -65,12 +63,22 @@ public DynamicTextFieldSkin(TextField field, int minColumnCount, int maxColumnCo
...
-    minWidth = minColumnCount == -1 ? 30 : minColumnCount * maxCharWidth;
-    maxWidth = maxColumnCount == -1 ? Double.MAX_VALUE : maxColumnCount * maxCharWidth;
+    // Estimate min/max width using public API by measuring a representative string
+    // (use 'W' as a wide character proxy)
+    if (minColumnCount == -1) {
+      minWidth = 30;
+    } else {
+      measurer.setText("W".repeat(Math.max(0, minColumnCount)));
+      measurer.applyCss();
+      minWidth = measurer.getLayoutBounds().getWidth();
```
robinschmid: I think this is a good solution to get rid of the internals. 
Only issue is that during creation of the skin the stylesheets and scene are not set. 
This is why we listen to field.sceneProperty(). 
Maybe we have to set recalculate the min / maxWidth inside the scene property change. 
And just apply some defaults here in the constructor.

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFields.java
```diff
@@ -226,20 +225,26 @@ public static <T> AutoCompletionBinding<T> bindAutoCompletion(TextField textFiel
 
   private static <T> void autoShowAutoComplete(TextField textField, AutoCompletionBinding<T> acb) {
     acb.setVisibleRowCount(25);
+    // Avoid redundant re-trigger while focused
+    final javafx.beans.property.BooleanProperty triggeredOnFocus = new javafx.beans.property.SimpleBooleanProperty(false);
```
robinschmid: maybe you have to adjust your IDE to automatically add imports for this. 
Also make sure to use the code style in the project. 

[mzmine-intellij-code-formater.xml](https://github.com/user-attachments/files/23415447/mzmine-intellij-code-formater.xml)

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFields.java
JonasLueb: it substitutes the removed pop.isShowing() boolean so that the autocomplete pop up is not being loaded twice after click. (Click triggers acb.setUserInput and then the isNowFocused is triggered...)

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFields.java
JonasLueb: hm, what do you mean with the first part? Shall I import this at the top and only use BooleanProperty in this line as the type hint?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/skins/DynamicTextFieldSkin.java
JonasLueb: just for my understanding of this problem: That issue was also present in the former version, right?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFields.java
robinschmid: yeah this should always just be an `import javafx.beans.property.BooleanProperty;` on top and then here you can just refer to the class. 
Using the whole class path is only needed if there are two classes with the same name used here.

Usually your IDE should autoatically add the imports maybe you have to activate it as a on save linter or something like that.

javafx-framework/src/main/java/io/github/mzmine/javafx/components/skins/DynamicTextFieldSkin.java
robinschmid: The previous version computed a min/max size without the scene by just taking the default font. 

So your solution is definitely improving on this by using the actual style sheet to compute the text width. 
Just needs an adjustment to actually react to the style sheet being set.

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFields.java
robinschmid: ok makes sense

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxTextFields.java
```diff
@@ -226,20 +226,26 @@ public static <T> AutoCompletionBinding<T> bindAutoCompletion(TextField textFiel
 
   private static <T> void autoShowAutoComplete(TextField textField, AutoCompletionBinding<T> acb) {
     acb.setVisibleRowCount(25);
+    // Avoid redundant re-trigger while focused
+    final BooleanProperty triggeredOnFocus = new javafx.beans.property.SimpleBooleanProperty(false);
```
robinschmid: Same is true for `javafx.beans.property.SimpleBooleanProperty` should also import and only use class name.

javafx-framework/src/main/java/io/github/mzmine/javafx/components/skins/DynamicTextFieldSkin.java
```diff
@@ -63,14 +76,20 @@ public DynamicTextFieldSkin(TextField field, int minColumnCount, int maxColumnCo
...
-    final FontMetrics fontMetrics = Toolkit.getToolkit().getFontLoader()
-        .getFontMetrics(field.getFont());
-    final float maxCharWidth = fontMetrics.getCharWidth('W');
-    minWidth = minColumnCount == -1 ? 30 : minColumnCount * maxCharWidth;
-    maxWidth = maxColumnCount == -1 ? Double.MAX_VALUE : maxColumnCount * maxCharWidth;
+      // Recalculate min/max width with actual stylesheets applied
+      // Only recalculate if we actually measured based on column counts (not fixed defaults)
+      if (minColumnCount != -1) {
+        measurer.setText("W".repeat(Math.max(0, minColumnCount)));
+        measurer.applyCss();
```
robinschmid: measurer.applyCss() already happened above. 
So no need to reapply it here nor in the maxColumnCount block below. Once is enough. 

Rest looks good.

PR #3001, opened 2025-11-13
Enable option for default presets for modules

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/minimumsearch/MinimumSearchFeatureResolverParameters.java
```diff
@@ -94,6 +101,62 @@ private static Parameter[] createParams(Setup setup) {
...
+    param.setParameter(PEAK_DURATION, duration);
+    param.setParameter(MIN_NUMBER_OF_DATAPOINTS, minDp);
+
+    return param;
+  }
+
+  @Override
+  public @NotNull List<ModulePreset> createDefaultPresets() {
+    return List.of(new ModulePreset("Orbitrap_UHPLC",
+            MZmineCore.getModuleInstance(MinimumSearchFeatureResolverModule.class).getUniqueID(),
```
SteffenHeu, line 133: not that nice that we have to pass the module here from the parameter set. In the setup dialog we know which storage we use and thus also have the module. But creating another class line ModulePreset just without the module also feels confusing. May be the better solution though. Could call it ParameterSetPreset and map to ModulePreset in the setup dialog.

robinschmid: not sure what you mean. Here we mostly need the unique ID for the preset factory so that the presets are put into a folder:
\.mzmine\presets\modules\uniqueID

You could even just call `String uniqueID = new MinimumSearchFeatureResolverModule().getUniqueID();`

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/minimumsearch/MinimumSearchFeatureResolverParameters.java
```diff
@@ -94,6 +101,62 @@ private static Parameter[] createParams(Setup setup) {
...
+            create(new FeatureListsSelection(FeatureListsSelectionType.BATCH_LAST_FEATURELISTS), "r",
+                OriginalFeatureListOption.KEEP, true, GroupMS2SubParameters.createDefault(),
+                ResolvingDimension.RETENTION_TIME, 0.90, 0.04, 0d, 1E5, 2d, Range.closed(0d, 10d), 5)),
+        //
+        new ModulePreset("TOF_UHPLC",
+            MZmineCore.getModuleInstance(MinimumSearchFeatureResolverModule.class).getUniqueID(),
```
robinschmid, line 139: you could extract a variable for the uniqueID to make it less cluttered and repeated

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/minimumsearch/MinimumSearchFeatureResolverParameters.java
```diff
@@ -94,6 +101,62 @@ private static Parameter[] createParams(Setup setup) {
...
+            create(new FeatureListsSelection(FeatureListsSelectionType.BATCH_LAST_FEATURELISTS),
+                "r", OriginalFeatureListOption.KEEP, true, GroupMS2SubParameters.createDefault(),
+                ResolvingDimension.RETENTION_TIME, 0.90, 0.04, 0d, 1E3, 2d, Range.closed(0d, 10d),
+                5)),
+        //
+        new ModulePreset("Orbitrap_HPLC",
+            MZmineCore.getModuleInstance(MinimumSearchFeatureResolverModule.class).getUniqueID(),
+            create(new FeatureListsSelection(FeatureListsSelectionType.BATCH_LAST_FEATURELISTS),
+                "r", OriginalFeatureListOption.KEEP, true, GroupMS2SubParameters.createDefault(),
+                ResolvingDimension.RETENTION_TIME, 0.90, 0.1, 0d, 1E5, 2d, Range.closed(0d, 10d),
```
robinschmid, line 149: in general extracting variables for the values would make it easier to check that they are the same for all TOF all Orbitrap etc. and would make it easier to change them in the future.
But can also stay like this.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_groupms2/GroupMS2SubParameters.java
```diff
@@ -53,4 +58,37 @@ public GroupMS2SubParameters() {
...
+      @Nullable Integer minSignalsInMs2, boolean limitByIms, boolean combineTims,
+      int imsMinOccurrence, boolean advanced, @Nullable Double timsOutputNoiseLevel,
+      @Nullable Double timsOutputNoiseRelative, @Nullable String iterativeMs2ColName) {
+    final GroupMS2SubParameters param = (GroupMS2SubParameters) new GroupMS2SubParameters().cloneParameterSet();
+    param.setParameter(GroupMS2Parameters.mzTol, mzTol);
+    param.setParameter(GroupMS2Parameters.rtFilter, rtLimits);
+    param.setParameter(GroupMS2Parameters.minimumRelativeFeatureHeight, minRelHeight != null,
+        Objects.requireNonNullElse(minRelHeight, 0.25));
+    param.setParameter(GroupMS2Parameters.minRequiredSignals, minSignalsInMs2 != null,
+        Objects.requireNonNullElse(minSignalsInMs2, 1));
```
robinschmid: use the static default variables introduced in this PR

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_groupms2/GroupMS2SubParameters.java
```diff
@@ -53,4 +58,37 @@ public GroupMS2SubParameters() {
...
+    param.setParameter(GroupMS2Parameters.limitMobilityByFeature, limitByIms);
+    param.setParameter(GroupMS2Parameters.combineTimsMsMs, combineTims);
+    param.setParameter(GroupMS2Parameters.minImsRawSignals, imsMinOccurrence);
+
+    final GroupMs2AdvancedParameters advancedParam = GroupMs2AdvancedParameters.create(
+        timsOutputNoiseLevel, timsOutputNoiseRelative, iterativeMs2ColName);
+    param.setParameter(GroupMS2Parameters.advancedParameters, advanced &&
+        Booleans.countTrue(timsOutputNoiseLevel != null, timsOutputNoiseRelative != null,
```
robinschmid: Maybe good to avoid guava as much as possible. We are using it in many places but I always fear the day we need to remove guava. Hopefully at some point they modularize the project so we can only keep parts but not sure if that will happen.  

We have our own utils class for this:
```java
ObjectUtils.countNull(timsOutputNoiseLevel, timsOutputNoiseRelative, iterativeMs2ColName)>0
```

Would be also good to add the anyNotNull variant. All the other are already there

mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterSet.java
```diff
@@ -297,7 +299,15 @@ default <V, T extends Parameter<V>> Stream<T> streamForClass(Class<T> parameterC
...
 
+  /**
+   *
+   * @return A message that is displayed in an accordion at the top of a parameter dialog.
+   */
   default @Nullable Region getMessage() {
     return null;
   }
+
+  default @NotNull List<ModulePreset> createDefaultPresets() {
```
robinschmid, line 318: add javadoc to link out to the ModulePresetStore where its used. 
And maybe add description that the presets should be generetated with a setAll method on the parameterset

robinschmid (approved):
Looks good to me. Only minor things. Can be merged after

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/minimumsearch/MinimumSearchFeatureResolverParameters.java
SteffenHeu: yes, I understand that.
But if you create the preset in the Parameters of the resolver it is quite obvious that they belong to the specific module. And since the method is called from the parametersetuppane which knows which unique id it belongs to anyway, it feels unneccessary. but to prevent that we would need another wrapper class, which also feels unnecessary and confusing.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/minimumsearch/MinimumSearchFeatureResolverParameters.java
SteffenHeu: i dont think that provides any more clarity, it just causes you having to look in two places: At the call and the definition. If we change eg the TOF preset to use another value, the other will probably still use the variable.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_groupms2/GroupMS2SubParameters.java
SteffenHeu: On one hand you are against code duplication, on the other hand you want to reimplement the utilities from guava :D

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_chromatogramdeconvolution/minimumsearch/MinimumSearchFeatureResolverParameters.java
robinschmid: Can stay like this. 
Was more like there are already 2 TOF presets for UHLPC and HPLC and I guess we should add HILIC as well? 
Then it would make sense to extract a var minHeightTof and use it in all 3. Much easier to change in the future and not to forget some of the presets.

PR #3005, opened 2025-11-17
Precision

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
```diff
@@ -108,8 +131,8 @@ public static BigDecimal round(double value, int sig) {
    * @return A BigDecimal rounded to sig number of significant digits (figures)
    */
   public static BigDecimal round(double value, int sig, RoundingMode mode) {
-    MathContext mc = new MathContext(sig, RoundingMode.HALF_UP);
-    BigDecimal bigDecimal = new BigDecimal(value, mc);
+    MathContext mc = new MathContext(sig, mode);
+    BigDecimal bigDecimal = BigDecimal.valueOf(value).round(mc);
```
robinschmid, line 135: just checking, the behavior is the same right?

mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
```diff
@@ -173,22 +196,30 @@ public static boolean equalDoubleSignificance(final double a, final double b) {
...
+
+    // calculate the allowed difference in significant digits: e.g. 5 sig digits:
+    // 1*10^-5 = 0.00001
+    // then scale to the max of those two values:
+    // 0.00001 * 1234567 = 12.34567
+    // then find the lower power of 10: 12.23467 -> 10
+    // use that as the allowed delta.
+    final BigDecimal scaledSignificance = new BigDecimal("1E-%d".formatted(sigDigits)).multiply(aDecimal.max(bDecimal));
+    final BigDecimal allowedDelta = new BigDecimal("1E%d".formatted((int)Math.log10(scaledSignificance.doubleValue())));
+    return diff.compareTo(allowedDelta) <= 0;
```
robinschmid: I tried with many more cases and the current version allows too much delta. This case fails:
```java
    assertFalse(Precision.equalSignificance(9.99d, 9.98d, 3));
```

Maybe we should use half the diff like in rounding? or 0.500001 to accept rounding errors.  

Also we can fully get rid of BigDecimal as it does not seem to be helpful and it is known to be slow. 

I tried this pure double version (code after the if(a==b)):
```java
final double diff = Math.abs(a - b);
final double absMax = Math.max(Math.abs(b), Math.abs(a));
double maxDiff = Math.pow(10, Math.floor(Math.log10(absMax)) - sigDigits + 1);

// only use half the maxDiff. so if sigDigits was 2 and absMax is 42 then maxDiff will be 0.5
// a bit higher to rather include rounding errors than exclude them
    return diff < maxDiff * 0.500001d;
```

mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
SteffenHeu: yes, i was just hoping for better results

mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
SteffenHeu: refactored it like this:
´´´
    final double diff = Math.abs(a - b);

    final double max = Math.max(Math.abs(a), Math.abs(b));
    final double scaledSignificance = Math.pow(10, sigDigits * -1d) * max;
    final double allowedDelta = Math.pow(10, (int) Math.log10(scaledSignificance));
    return Double.compare(diff,allowedDelta) <= 0;
´´´
adding a factor of 0.500001 to the allowedDelta then causes 

assertFalse(Precision.equalSignificance(9.99000000d, 9.984899999d, 3));

to fail

SteffenHeu: maybe the better option is the remove the significance methods and always require a delta. most of the time you are pretty sure what kind of delta you want to allow in that case

mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
robinschmid: ok

Issue #2981, opened 2025-11-04
Raw data handling sdk conversion parameter


robinschmid:
- [x] first ship a patch in 4.8.x

## For 4.9

- [ ] introduce a subparameter how to handle data conversion / import in all data import
- [ ] preferences and import data module: add comboboxes for all formats to choose data SDK / converter
- [ ] move vendor centroiding to subparameter in all data import

PR #3013, opened 2025-11-20
Add Vendor import param 

SteffenHeu:
Addresses https://github.com/mzmine/mzmine/issues/2981 with more granular control for vendor parameters

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/gui/preferences/VendorImportParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/gui/preferences/VendorImportParameters.java
@@ -27,6 +27,7 @@ package io.github.mzmine.gui.preferences;
 import io.github.mzmine.javafx.components.factories.FxButtons;
 import io.github.mzmine.javafx.util.FxIcons;
 import io.github.mzmine.main.ConfigService;
+import io.github.mzmine.modules.dataprocessing.filter_scan_signals.ScanSignalRemovalModule;
 import io.github.mzmine.parameters.impl.SimpleParameterSet;
 import io.github.mzmine.parameters.parametertypes.BooleanParameter;
 import io.github.mzmine.parameters.parametertypes.ComponentWrapperParameter;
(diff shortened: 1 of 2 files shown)


mzmine-community/src/main/java/io/github/mzmine/gui/preferences/VendorImportParameters.java
```diff
@@ -0,0 +1,122 @@
...
+  public static final ComponentWrapperParameter<Boolean, OptionalModuleParameter<WatersLockmassParameters>> watersLockmass = new ComponentWrapperParameter<>(
+      new OptionalModuleParameter<>("Apply lockmass on import (Waters)",
+          "Apply lockmass correction for native Waters raw data during raw data import via MSConvert.",
+          new WatersLockmassParameters(), DEFAULT_WATERS_LOCKMASS_ENABLED),
+      createJumpToPrefButton("Apply lockmass on import (Waters)"));
+
+  private static final boolean DEFAULT_THERMO_EXCEPTION_SIGNALS = true;
+  public static final ComponentWrapperParameter<Boolean, BooleanParameter> excludeThermoExceptionMasses = new ComponentWrapperParameter<>(
+      new BooleanParameter("Remove calibrant signals (Thermo)",
+          "Internal calibration signals may be present in MS1 and MS2 spectra from Thermo Orbitraps. This option automatically removes those on import.",
```
robinschmid: As far as I understand these can be both reference signals and known interferences that are flagged already in the data format (?). 

Maybe we can rename the parameter to make it applicable to other data formats as well. Maybe we find out that the other SDKs also support something like this or we find other opportunities to remove calibration signals/interferences in the future.   
Then we only have one parameter for this and not one for each format.

Regarding the description, the interferences are actually in all MS1-n scans. The parser sadly only removes the noise signals from MS1-2

Here is an example of the noise. Its mz distribution is crazy wide and weird mass defects:
image

image

Zoom out for more context. All other real signals are much more narrow:
image

robinschmid: This is without the filters with msconvert conversion

SteffenHeu: If we add other vendors to this parameter, we lose the granular control. A reason for the introduction of this parameter set is to ensure reproducility in future. We would then need to enable or disable this parameter depending on what vendor data is being imported to ensure compatibility with old batches.

robinschmid: I see it more like the **Try vendor centroiding** algorithm that also works for all the vendors in the same way. 
Not sure if we need so much granularity. Maybe its easiest to keep the parameter generic for now and if there are other vendor specific filters in the future then we can add them?

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_all/AllSpectralDataImportParameters.java
```diff
@@ -76,6 +68,10 @@ public class AllSpectralDataImportParameters extends SimpleParameterSet {
       ExtensionFilters.MS_RAW_DATA, "Drag & drop your MS data files here.",
       AllSpectralDataImportParameters::validateDistinctPaths);
 
+  public static final ParameterSetParameter<VendorImportParameters> vendorOptions = new ParameterSetParameter<>(
+      "Vendor data options", "Set specific options how to handle vendor data.",
```
robinschmid, line 72: I like the name. I was thinking how to call it and was not able to find a good name but this is quite flexible and descriptive

PR #3016, opened 2025-11-21
allow multiple imported files for RawDataImportTask

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/datamodel/RawDataImportTask.java
```diff
@@ -25,13 +25,14 @@
...
-import org.jetbrains.annotations.Nullable;
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
 
 
 public interface RawDataImportTask extends Task {
 
   /**
-   * @return The imported file or null. Will be null unless this task is finished.
+   * @return The imported files or an empty list. Empty may be caused by the task not being finished or
```
robinschmid: message ends with ...or... holding me in suspense

mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_zip/ZipImportTask.java
```diff
@@ -178,7 +176,7 @@ public void cancel() {
   }
 
   @Override
-  public RawDataFile getImportedRawDataFile() {
-    return getStatus() == TaskStatus.FINISHED ? msdkTask.getImportedRawDataFile() : null;
+  public @NotNull List<RawDataFile> getImportedRawDataFile() {
+    return msdkTask.getImportedRawDataFile();
```
robinschmid: the msdk task returns null on not FINISHED and this is restricted here.

robinschmid (dismissed):
Were you already able to try it with example data? 

Looks good so far. I think the only thing that may not work is when data is already loaded, applying the batch again? Because at this point we are only skipping single files by cleaning up their names to get the expected file names and this is not possible here. 

But this is a niche case and might be hard to implement so I would not do it.

robinschmid (approved):

PR #3018, opened 2025-11-23
Cleanup Precision

robinschmid:
Cleaned up the changes that I tried in precision class. 

I checked and Float.NaN and Float.POSITIVE_INFINITY are perfectly converted to Double so that no extra checks are required. 

Double.compare(a, b) == 0 already covers the case where both values are NaN so I removed the extra checks. 

Added a check for the Double.isInfinite(diff) to capture number overflow.

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
@@ -33,7 +33,7 @@ import org.jetbrains.annotations.Nullable;
 
 public class Precision {
 
-  private static DecimalFormat format = new DecimalFormat("0.0E0");
+  private static final DecimalFormat exponentFormat = new DecimalFormat("0.0E0");
 
   /**
    * Checks if difference of a and b is small equal to max difference
@@ -79,8 +79,23 @@ public class Precision {
    * @return true if Math.abs(a-b) <= maxDelta
    */
   public static boolean equals(double a, double b, double maxDelta, double maxRelativeDelta) {
-    // diff <= max of relative and absolute
-    return Math.abs(a - b) <= Math.max(Math.max(a, b) * maxRelativeDelta, maxDelta);
+    // equal or both are NaN
+    if (Double.compare(a, b) == 0) {
+      return true;
+    }
+    // only one is NaN
+    if (Double.isNaN(a) || Double.isNaN(b)) {
+      return false;
+    }
+    // abs error
+    final double diff = Math.abs(a - b);
+    if (diff <= maxDelta) {
+      return true;
+    }
+
+    final double absoluteMax = Math.max(Math.abs(a), Math.abs(b));
+    final double relativeDifference = Math.abs(diff / absoluteMax);
+    return relativeDifference <= maxRelativeDelta;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
```diff
@@ -192,41 +178,43 @@ public static String toString(double value, int sig, RoundingMode mode, int maxL
...
-    // calculate the allowed difference in significant digits: e.g. 5 sig digits:
-    // 1*10^-5 = 0.00001
-    // then scale to the max of those two values:
-    // 0.00001 * 1234567 = 12.34567
-    // then find the lower power of 10: 12.23467 -> 10
-    // use that as the allowed delta.
+    // Math.floor always rounds to next smaller number so negative values will be like -3.1 -> -4
+    // +1 as we are using floor and in both positive and negative exponents we need +1
     final double max = Math.max(Math.abs(a), Math.abs(b));
-    final double scaledSignificance = Math.pow(10, sigDigits * -1d) * max;
```
robinschmid, line 219: this pow is not needed. The old version does pow(log10(pow)) the new version just does pow(log10)

mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
```diff
@@ -192,41 +178,43 @@ public static String toString(double value, int sig, RoundingMode mode, int maxL
...
     if (Double.isNaN(a) || Double.isNaN(b)) {
       return false;
     }
 
-    if (a == b) {
-      return true;
-    }
     final double diff = Math.abs(a - b);
+    // if one is infinite or if number overflow
+    if (Double.isInfinite(diff)) {
```
robinschmid: to handle number overflow

mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
```diff
@@ -240,18 +228,57 @@ public static boolean equalFloatSignificance(final @Nullable Float a, final @Nul
...
+  /**
+   *
+   * @param a             first value
+   * @param b             second value
+   * @param relativeError relative error that is accepted, e.g., 1E-2, 0.01 of 1%
+   * @return true if the relative difference diff/max is <= relative error. true if both values are
+   * NaN, false if only one is NaN
+   */
+  public static boolean equalRelativeSignificance(final double a, final double b,
+      final double relativeError) {
```
robinschmid, line 265: I checked the methods available in apache commons math Precision. 
They have one that just uses relative value precision and maybe thats actually fine? Then we dont need all the fancy log10, floor, pow10. Maybe forcing it into the decimal system with pow10 is actually bad because we still get all the flaoting point inaccuracies like we want 10 but it may be 9.99999. 

Only issue in apache version is that it returns false if both values are NaN. 
So better to reimplement here.  

to check default float accuracy we could either use 1E-5 or 5E-6 for checks

mzmine-community/src/main/java/io/github/mzmine/util/maths/Precision.java
```diff
@@ -240,18 +228,57 @@ public static boolean equalFloatSignificance(final @Nullable Float a, final @Nul
...
+    // equal or both are NaN
+    if (Double.compare(a, b) == 0) {
```
SteffenHeu, line 267: a = positive and b = negative infinity will not be equal, or? they may cause number overflow later

robinschmid: checked. This this comes down to max and a-b are both INFINITY and devided this is NaN and NaN is always false when compared to any other number

robinschmid: so test result is correct

PR #3026, opened 2025-11-26
ParameterUtils save and load parameters from xml

robinschmid:
- Checks that clone generates a new instance - also for embedded parameters. 
- permutates the values and checks that instances are independent
- saves and loads from xml string, permutates values and checks if loading results in the same value

image

### DIFF
--- a/mzmine-community/src/test/java/io/github/mzmine/parameters/StandardParameterTest.java
+++ b/mzmine-community/src/test/java/io/github/mzmine/parameters/StandardParameterTest.java
@@ -26,7 +26,6 @@
 package io.github.mzmine.parameters;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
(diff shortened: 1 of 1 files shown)


mzmine-community/src/test/java/io/github/mzmine/parameters/StandardParameterTest.java
```diff
@@ -0,0 +1,130 @@
...
+    final Parameter param = test.param();
+    final String xml = ParameterUtils.saveParameterToXMLString(param);
+
+    // use a changed clone to see that values are actually loaded
+    final Parameter clone = test.cloneParamPermutated();
+    ParameterUtils.loadParameterFromString(clone, xml);
+
+    // check value and embedded
+    assertNotSame(param, clone);
+    assertEquals(param.getValue(), clone.getValue());
```
SteffenHeu: you could use Parameter.valueEquals
It has some specific implementations for value types that don't provide a proper equals method

PR #3062, opened 2025-12-18
Update local max mass detector for tof data

SteffenHeu:
- maximum/minimum based mass detector, mainly for TOF data.
- mainly targeted for waters data

### Comparison of commonsignals vs. waters vendor algorithm:
image

### Error investigations:
- mainly signals with low intensities are not matched
image
image

- good overall agreement for signals > 100 counts:
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetectorParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetectorParameters.java
@@ -78,6 +78,7 @@ public class LocalMaxMassDetectorParameters extends SimpleParameterSet {
 
     if (loadedVersion < 2) {
       setParameter(intensityCalculation, AbundanceMeasure.Height);
+      setParameter(minNumberOfDp, 3);
     }
   }
 }


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+    this.noiseLevel = noiseLevel;
+    this.intensityCalculation = intensityCalculation;
+  }
+
+  private static double getMaxMzDiff(final double[] mzs, final double[] intensities,
+      final int numPoints) {
+    for (int i = numPoints - 1; i > 1; i--) {
+      // tof mz value distances are proportional to sqrt(m/z)
+      // so the biggest mass diff will be at the top of the spectrum
+      if (Double.compare(intensities[i], 0) != 0 || !(intensities[i - 1] > 0)) {
```
robinschmid: confusing. 
maybe change into a positive condition here like:
```java
if (Double.compare(intensities[i], 0) == 0 && intensities[i - 1] > 0){
    return Math.abs(mzs[i]-mzs[i-1]);
}
```

robinschmid: maybe use the first 2 datapoints >0 to determine the mz diff?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+        continue;
+      }
+      return Math.abs(mzs[i] - mzs[i - 1]);
+    }
+    return 0.1;
```
robinschmid, line 85: Currently a 0 intensity data point is needed for a result. Otherwise 0.1 mz dinstance which is quite random and wide. 

instead of returning 0.1 maybe we could return the smallest difference between mz instead. 
Its a bit counter intuitive but this value is much closer to the actual mz grid than just 0.1 default.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+      }
+
+      if (thisInt > noiseLevel) {
+        onePointAboveNoise = true;
+      }
+
+      final double mzDelta = thisMz - lastMz;
+
+      // If the gap is too large, we close the current region and start a new one
+      if (mzDelta >= maxDiff) {
```
robinschmid, line 145: I guess this way at most there will be 1 missing data point (0 intensity but not reported). maybe allow two data points at 0? like mzDelta>=maxDiff*2
Results in less segments and may be better for processing

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+        // Only add regions that contain enough data points to form a peak (e.g., > 2 points)
+        if (i - currentRegionStart > MIN_PEAK_RANGE_LENGTH && onePointAboveNoise) {
+          consecutiveRanges.add(new SimpleIndexRange(currentRegionStart, i - 1));
```
robinschmid, line 148: Maybe we should make SimpleIndexRange and other implementations package private. 
Then everything is created in factory
```java
IndexRange.ofExclusive(currentRegionStart, i)
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+        }
+        currentRegionStart = i;
+        onePointAboveNoise = false;
+      }
+
+      lastMz = thisMz;
+    }
+
+    // Add the final region if valid
+    if (numPoints - currentRegionStart > MIN_PEAK_RANGE_LENGTH - 1 && onePointAboveNoise) {
```
robinschmid: why MIN_PEAK_RANGE_LENGTH - 1? 
The simplest example would be if we have 4 data points and they form a single region then the start is 0, num points is 4 and this should be enough to be > MIN LENGTH.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+   * Helper to find all indices in range that are local maxima > noise.
+   */
+  private IntArrayList findLocalMaximaIndices(final double[] intensities, final IndexRange range) {
+    final IntArrayList indices = new IntArrayList();
+    final int start = range.min();
+    final int end = range.maxExclusive();
+
+    for (int i = start; i < end - 1; i++) {
+      final double currentInt = intensities[i];
+//      if (currentInt < noiseLevel) {
```
robinschmid: remove comment. seems like this needs to be removed. Otherwise the mz weighting depends on the noiseLevel

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+
+  /**
+   * Helper to find all indices in range that are local maxima > noise.
+   */
+  private IntArrayList findLocalMaximaIndices(final double[] intensities, final IndexRange range) {
+    final IntArrayList indices = new IntArrayList();
+    final int start = range.min();
+    final int end = range.maxExclusive();
+
+    for (int i = start; i < end - 1; i++) {
```
robinschmid: should loop until <end without the -1 otherwise the question would be why it starts at start and not start+1.
Also below is a check if i==end-1 which is correct for the last data point but currently this is never reached

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+      final DoubleArrayList resultMzs, final DoubleArrayList resultIntensities) {
+
+    if (endIdx - startIdx
+        < MIN_PEAK_RANGE_LENGTH - 2) { // potentially this peak does not start and end at zero.
+      return;
+    }
+
+    final double maxIntensity = intensities[peakMaxIdx];
+
+    final double detectionThreshold;
```
robinschmid: Leads to detection below noiseLevel, so if has to be replaced by this 
```java
final double detectionThreshold = Math.max(noiseLevel, 2 * absMinIntensity);
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+    } else {
+      detectionThreshold = noiseLevel;
+    }
+
+    /*if (maxIntensity < detectionThreshold) {
+      return;
+    }*/
+
+    // M/z weighting still uses the top 40% logic or the more shallow valley
+    final double mzWeightingCutoff = Math.max(maxIntensity * MZ_WEIGHTING_THRESHOLD,
```
robinschmid: just maxIntensity * MZ_WEIGHTING_THRESHOLD. Nothing with Math.max

Just talked about this again. 
Seems like this was supposed to be /2 instead of *MZ_WEIGHTING. 
This leads to double peaks often relying on just the top 1 or 2 datapoints

```java
final double mzWeightingCutoff = Math.max(maxIntensity * MZ_WEIGHTING_THRESHOLD,
        // max of edges+max/2
        Math.max((intensities[startIdx] + maxIntensity),
            (intensities[endIdx] + maxIntensity)) /2);
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+   * @param startIdx        Inclusive start of peak region (valley or range start).
+   * @param endIdx          Exclusive end of peak region (valley or range end).
+   * @param absMinIntensity Absolute minimum intensity of the whole spectrum.
+   */
+  private void processSinglePeak(final double[] mzs, final double[] intensities,
+      final int peakMaxIdx, final int startIdx, final int endIdx, final double absMinIntensity,
+      final DoubleArrayList resultMzs, final DoubleArrayList resultIntensities) {
+
+    if (endIdx - startIdx
+        < MIN_PEAK_RANGE_LENGTH - 2) { // potentially this peak does not start and end at zero.
```
robinschmid: might make it clearer if this is just renamed to MIN_NON_ZERO_DATA_POINTS and then use 2 instead of 4 and just count the non zero points which would be safer criteria to decide if its a good peak

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -0,0 +1,369 @@
...
+        // Move to the next peak
+        activePeakIdx = nextCandidateIdx;
+        leftBoundary = valleyIdx; // Next peak starts from this valley
+      } else {
+        // They are not resolved enough; merge them.
+        // Update the active peak index to be the higher of the two.
+        if (nextCandidateInt > activePeakInt) {
+          activePeakIdx = nextCandidateIdx;
+          // move the left boundary, since it would have been enough of a valley and rise to be a individual peak
+          leftBoundary = valleyIdx;
```
robinschmid: why shift the left boundary? This means they are not merged but only skipped. 
Maybe another check for  if(dropsEnough) needed here?

robinschmid: just checked removing this line and it improves the result as expected. 
Left is current shifting of leftBoundary and right is merging (keeping left boundary)
227.12 is more centered then
image

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetectorParameters.java
```diff
@@ -0,0 +1,80 @@
...
+import io.github.mzmine.parameters.parametertypes.AbundanceMeasureParameter;
+import io.github.mzmine.parameters.parametertypes.DoubleParameter;
+import java.util.Map;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class LocalMaxMassDetectorParameters extends SimpleParameterSet {
+
+  public static final DoubleParameter noiseLevel = CentroidMassDetectorParameters.noiseLevel.cloneParameter();
+  public static final AbundanceMeasureParameter intensityCalculation = new AbundanceMeasureParameter(
```
robinschmid, line 44: add default value HEIGHT.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/BaseWizardBatchBuilder.java
```diff
@@ -320,7 +320,7 @@ protected BaseWizardBatchBuilder(final WizardSequence steps) {
     isNativeIms = allFileTypes.stream().allMatch(
         type -> (type == RawDataFileType.BRUKER_TDF || (type == RawDataFileType.WATERS_RAW_IMS
             && ConfigService.getPreference(MZminePreferences.massLynxImportChoice)
-            == MassLynxImportOptions.NATIVE)));
+            == MassLynxImportOptions.NATIVE_MZMINE_CENTROIDING)));
```
robinschmid: isNativeIms? what does this mean. And why only if the import is native mzmine centroiding? Could be IMS in any case right?
Or maybe also add the other native option here? maybe best just add a method to the MassLynxImportOption.isNative

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
SteffenHeu: returning the smallest difference between two mzs will lead to almost no consecutive datapoints being detected as the difference will be larger most of the time

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
SteffenHeu: One missing data point would lead to a single zero being reported, so that case should be covered or?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
SteffenHeu: in theory the last point may not be a zero and still have some intensity.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
robinschmid: I meant a real missing value that is not reported like 
10, 0, _, 0, 3 ... 
_ would be the not reported value.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
robinschmid: maybe its then better to just count the non-zero datapoints and have a minimum data point parameter instead of the current 4 data points min where 2 zero are expected but not checked.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/BaseWizardBatchBuilder.java
SteffenHeu: native ims data already has frame spectra associated with them

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
robinschmid: can be 5x smallest mz or something like that.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
SteffenHeu: that would be three missing points then though. for two points both zeros would be reported in the raw data. and then it gets pretty far a part

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/BaseWizardBatchBuilder.java
robinschmid: So for NATIVE_WATERS and MSCONVERT both are different to NATIVE mzmine?

PR #3065, opened 2025-12-19
Local minimum detector, review, small changes

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_massdetection/local_max/LocalMaxMassDetector.java
```diff
@@ -214,19 +245,19 @@ private void findAndCentroidPeaks(final double[] mzs, final double[] intensities
       if (dropsEnough && risesEnough) {
         // They are separate peaks.
         // The right boundary for the current peak is the valley.
-        processSinglePeak(mzs, intensities, activePeakIdx, leftBoundary, valleyIdx, minIntensity,
-            resultMzs, resultIntensities);
+        // include valleyIdx in this peak and also as potential new peak - similar to belows use of range.maxExclusive
+        processSinglePeak(mzs, intensities, activePeakIdx, leftBoundary, valleyIdx + 1,
```
robinschmid, line 249: I think the valley index itself should be included in process single peak. At least this would equal treatment compared to the last region that is until range.maxExclusive()

PR #3074, opened 2026-01-07
[on master] Rework IIN full changes

robinschmid:
- test PR with full changes
- new IonType pushed into old package should reduce number of changes


javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxLabels.java
```diff
@@ -87,6 +100,15 @@ public static Label underlined(String name) {
     return label;
   }
 
+  @NotNull
+  public static Label attachProperty(@NotNull Label label,
```
SteffenHeu: bindText? describes more what it does

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxLabels.java
```diff
@@ -73,10 +78,18 @@ public static Label styled(String name, String styleClass) {
     return label;
   }
 
+  public static Label newBoldLabel(ObservableValue<? extends String> name) {
+    return attachProperty(newBoldLabel(""), name);
```
SteffenHeu: in theory this method parameter can be nullable as attachProperty is, but does it make sense to have this method for nullable name?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxListViews.java
```diff
@@ -0,0 +1,102 @@
...
+
+  public static <T> ListView<T> newListView(ObservableList<T> items, boolean enableDeleteKey,
+      final SelectionMode selectionMode) {
+    final ListView<T> view = new ListView<>(items);
+
+    view.getSelectionModel().setSelectionMode(selectionMode);
+
+    if (enableDeleteKey) {
+      view.setOnKeyPressed(event -> {
+        if (event.getCode() == KeyCode.DELETE || event.getCode() == KeyCode.BACK_SPACE) {
```
SteffenHeu: I feel like backspace is not common for removing entries. only delete?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxButtons.java
```diff
@@ -129,6 +169,15 @@ public static Button createCancelButton(String text, Runnable runnable) {
     return createButton(text, null, FxIconUtil.getFontIcon(FxIcons.CANCEL), runnable);
   }
 
+  /**
+   * Add disableProperty binding
+   */
+  public static <T extends Control> T disableIf(T control,
```
SteffenHeu, line 175: do we need this method if it just calls the one in FxControls?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/rework/NewIonTypeType.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.xml.stream.XMLStreamReader;
+import javax.xml.stream.XMLStreamWriter;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Specifies an ion type.
+ * Todo: remove {@link IonAdductType}, rename this?
+ */
+public class NewIonTypeType extends DataType<IonType> {
```
SteffenHeu: unused

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/IonIdentityListType.java
```diff
@@ -64,7 +64,7 @@ public class IonIdentityListType extends ListWithSubsType<IonIdentity> implement
   private static final List<DataType> subTypes = List.of(new IonNetworkIDType(),
       // start with netID
       new IonIdentityListType(), // add self type to have a column
-      new SizeType(), new NeutralMassType(), new PartnerIdsType(), new MsMsMultimerVerifiedType(),
+      new SizeType(), new NeutralMassType(),
```
SteffenHeu, line 67: removing will probably cause integration tests to fail because we have less columns then

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/io/IonLibraryPresetStore.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Override
+  public @NotNull PresetGroup getPresetGroup() {
+    return KnownPresetGroup.ION_LIBRARY_PRESET;
+  }
+
+  @Override
+  public @NotNull List<IonLibraryPreset> createDefaults() {
+    return List.of();
```
SteffenHeu, line 50: The mzmine defaults are not returned here because they would not be added as soon as the user has at least one own library,right?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonIdentity.java
```diff
@@ -83,161 +62,30 @@ public class IonIdentity implements Comparable<IonIdentity> {
   public IonIdentity(IonType ionType) {
     super();
     this.ionType = ionType;
-    this.adduct = ionType.toString(false);
     molFormulas = FXCollections.observableArrayList();
```
SteffenHeu: must be observable?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonIdentity.java
```diff
@@ -316,40 +108,16 @@ public void setNetwork(IonNetwork net) {
...
    *
    * @return
    */
   public int getScore() {
     if (network == null) {
-      return partner.size();
+      return 0;
     }
-    return network.size() + (getMSMSMultimerCount() > 0 ? 1 : 0) + (getMSMSModVerify() > 0 ? 1 : 0);
+    return network.size();
```
SteffenHeu, line 120: javadoc does not match anymore

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonLibraries.java
```diff
@@ -0,0 +1,102 @@
...
+   */
+  public static boolean isInternalLibrary(@Nullable IonLibrary library) {
+    return library != null && isInternalLibrary(library.name());
+  }
+
+  /**
+   * @return true if input is an internal library that may never be changed
+   */
+  public static boolean isInternalLibrary(@Nullable String name) {
+    return name != null && (name.toLowerCase().contains("mzmine default"));
```
SteffenHeu: is it checked if the user creates a library with this in the name?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonNetwork.java
```diff
@@ -219,58 +253,51 @@ public double getNeutralMass() {
     return neutralMass == null ? calcNeutralMass() : neutralMass;
   }
 
-  @Override
-  public IonIdentity put(FeatureListRow key, IonIdentity value) {
-    IonIdentity e = super.put(key, value);
-    if (key.getID() < lowestID || lowestID == -1) {
-      lowestID = key.getID();
+  public IonIdentity put(FeatureListRow row, IonIdentity value) {
```
SteffenHeu, line 248: should this method check if the netural mass of the row matches the current neutral mass of the network? say in 50 ppm or 0.1 Da? edit: ah we even have a mz tolerance in this class

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonNetworkNode.java
```diff
@@ -0,0 +1,32 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.datamodel.identities.iontype;
+
+import io.github.mzmine.datamodel.features.FeatureListRow;
+
+public record IonNetworkNode(FeatureListRow row, IonIdentity ion) {
```
SteffenHeu: both are notnull i guess?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonSearchRow.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * A row for IonType matching, This class makes the search independent of {@link FeatureListRow} and
+ * search could be used also for other mz values.
+ *
+ * @param mz           the mz
+ * @param signedCharge the signed charge like -1 used to filter potential adducts for this row
+ * @param absCharge    the absolute charge Math.abs() to apply some filters
+ * @param polarity     the polarity if known
+ */
+public record IonSearchRow(double mz, @Nullable Integer signedCharge, @Nullable Integer absCharge,
```
SteffenHeu, line 45: i dont really understand what this class does

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonType.java
```diff
@@ -303,280 +324,209 @@ public String getMassDiffString() {
...
+          e);
+      return formula;
+    }
+    // add for n molecules the M formula
+    for (int i = 2; i <= molecules; i++) {
+      FormulaUtils.addFormula(result, formula);
+    }
+
+    // add first then remove
+    stream(false).filter(IonPart::isAddition).forEach(ion -> ion.addToFormula(formula, false));
```
SteffenHeu, line 376: good idea to use foreach in a stream?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
```diff
@@ -106,11 +106,10 @@ public interface CompoundDBAnnotation extends Cloneable, FeatureAnnotation,
...
   @NotNull
   static List<CompoundDBAnnotation> buildCompoundsWithAdducts(
-      CompoundDBAnnotation neutralAnnotation, IonNetworkLibrary library) {
+      CompoundDBAnnotation neutralAnnotation, IonLibrary library) {
     final List<CompoundDBAnnotation> annotations = new ArrayList<>();
-    for (IonType adduct : library.getAllAdducts()) {
-      if (adduct.isUndefinedAdduct() || adduct.isUndefinedAdductParent() || adduct.getName()
-          .contains("?")) {
+    for (IonType adduct : library.ions()) {
+      if (adduct.isUndefinedAdduct() || adduct.isUndefinedMass()) {
```
SteffenHeu, line 112: In IonType the M+? still exists. Is that fully covered by undefined adduct and undefined mass now?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_ion_identity_networking/addionannotations/AddIonNetworkingParameters.java
```diff
@@ -146,6 +146,23 @@ public Map<String, Parameter<?>> getNameParameterMap() {
     var nameParameterMap = super.getNameParameterMap();
     // we use the same parameters here so no need to increment the version. Loading will work fine
     nameParameterMap.put("m/z tolerance", getParameter(MZ_TOLERANCE));
+    // {@link LegacyIonLibraryParameterSet} is loaded directly by the new library parameter
+    nameParameterMap.put("Ion identity library", fullIonLibrary);
```
SteffenHeu: getParameter

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_ion_identity_networking/ionidnetworking/IonNetworkingParameters.java
```diff
@@ -153,12 +153,28 @@ public Setup getSetup() {
...
+  }
+
   @Override
   public Map<String, Parameter<?>> getNameParameterMap() {
     // parameters were renamed but stayed the same type
     var nameParameterMap = super.getNameParameterMap();
     // we use the same parameters here so no need to increment the version. Loading will work fine
     nameParameterMap.put("m/z tolerance", getParameter(MZ_TOLERANCE));
+    // {@link LegacyIonLibraryParameterSet} is loaded directly by the new library parameter
+    nameParameterMap.put("Ion identity library", fullIonLibrary);
```
SteffenHeu: getParameter

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/MZTolerance.java
```diff
@@ -37,6 +37,7 @@
 public class MZTolerance {
 
   public static final MZTolerance FIFTEEN_PPM_OR_FIVE_MDA = new MZTolerance(0.005, 15);
+  public static final MZTolerance WIDE_25_PPM_OR_8_MDA = new MZTolerance(0.01, 25);
```
SteffenHeu: 8 != 10

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxLabels.java
robinschmid: I think Nullable in the factories is an ok default and just needs to be handled. 
In this case it is kind of hard to call the method with null as you would need to cast null. newBoldLabel(String/ObservableValue) both have just one argument. I think that is fine as usually we would call it with a value and empty case could also be "" for string.

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxListViews.java
robinschmid: removed

mzmine-community/src/main/java/io/github/mzmine/util/FormulaUtils.java
```diff
@@ -774,29 +918,56 @@ public static IMolecularFormula cloneFormula(@Nullable final IMolecularFormula f
...
-    IMolecularFormula molecularFormula = FormulaUtils.neutralizeFormulaWithHydrogen(
+    // check if formula already charged
+    IMolecularFormula molecularFormula = createMajorIsotopeMolFormulaWithCharge(
         annotation.getFormula());
+
+    if (molecularFormula == null) {
+      return null;
+    }
+
+    // TODO add mechanism to check if formula is already charged ion formula
```
SteffenHeu: whats the plan here?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/factories/FxButtons.java
robinschmid: I added this mostly for visibility because I was not sure if I would find it again in FxControls. 
But I can also remove this delegate method if you want

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/iin/rework/NewIonTypeType.java
robinschmid: removed. Was used as an intermediate when I still had both Ion type implementations

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/io/IonLibraryPresetStore.java
robinschmid: The defaults will always be there - but they are not stored. This is why they are not added as defaults here. 
They are added to the list of libraries in the 
`GlobalIonLibraryService.getIonLibrariesUnmodifiable()`

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonIdentity.java
robinschmid: most likely not. Removed

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonIdentity.java
robinschmid: adjusted. This might be something to change again once we know how to handle the MSMS verification in datatypes

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonLibraries.java
robinschmid: This is check in the IonLibraryEditController but next commit will also add a check in the constructor of the SimpleIonLibrary. 
This constructor will replace the name of the library with unnamed library if the mzmine default is in its name. This might happen if the library is loaded from a parameter and this might be the active default library or an older version of the library if we change the libraries in the future.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonNetwork.java
robinschmid: Those checks are usually done in the tasks that create the network or add more ions to a network. 
The mztolerance in the network is still from the initial implementation and the user can actually use different tolerances for main ions or additional ions. 

I removed the mz tolerance now from this class as it was misleading from the beginning to have a mutable instance of mztolerance that may change due to different method calls.

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/ionidentity/IonLibraryComponent.java
```diff
@@ -0,0 +1,151 @@
...
+ */
+public class IonLibraryComponent extends BorderPane implements ParameterComponent<IonLibrary> {
+
+  private final StringProperty selectedName = new SimpleStringProperty();
+  private final StringProperty numIonsText = new SimpleStringProperty();
+  private final StringProperty conflictWithLocalText = new SimpleStringProperty();
+  private final StringProperty updateWithGlobalTooltip = new SimpleStringProperty();
+  private final BooleanBinding hasConflictWithLocal = conflictWithLocalText.isNotEmpty();
+
+  private final ObjectProperty<IonLibrary> library = new SimpleObjectProperty<>();
```
SteffenHeu: i dont see that the the library is updated anywhere in this component. probably needs to be done in handleSelectedNameChanged?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/ionidentity/IonLibraryComponent.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  private void exchangeIonListForGlobalVersion() {
+    // TODO implement way to get the global library with that name
+  }
+
+  private void handleSelectedNameChange(String newName) {
+    final IonLibrary currentLibrary = getValue();
+    if (currentLibrary != null && newName.equals(currentLibrary.name())) {
+      // TODO
```
SteffenHeu: needs to be implemented to actually change.

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/ionidentity/IonLibraryComponent.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  @Override
+  public IonLibrary getValue() {
+    return library.getValue();
```
SteffenHeu, line 177: either the parameter or the component must clone the value at the point the ion library is selected. probably the parameter. For example if a usere selects a library for a batch, processes and then alters the global ion library and the parameter just stores a reference, the library in the parameter will not reflect the actual batch parameters anymore. 
Or does the global ion library only return clones anyway?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/GlobalIonLibrariesInteractor.java
```diff
@@ -0,0 +1,128 @@
...
+  public void applyToGlobalIons() {
+    final List<IonLibrary> libraries = List.copyOf(model.getLibraries());
+    final List<IonType> types = List.copyOf(model.getIonTypes());
+    final List<IonPart> parts = List.copyOf(model.getParts());
+    final List<IonPartDefinition> partDefinitions = List.copyOf(model.getPartsDefinitions());
+
+    final GlobalIonLibraryService global = GlobalIonLibraryService.getGlobalLibrary();
+    final int currentVersion = global.getVersion();
+    final int retrivalVersion = model.getRetrivalVersion();
+    // TODO add check that if the versions mismatch than we need to apply safer merging of states
```
SteffenHeu, line 92: tag

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/IonPartCreatorPane.java
```diff
@@ -0,0 +1,133 @@
...
+  private final FilterableListView<IonPartDefinition> partListView;
+
+  // additional properties for the list view
+  private final ObjectProperty<IonPartSorting> listSorting = new SimpleObjectProperty<>(
+      IonPartSorting.ALPHABETICAL);
+
+
+  public IonPartCreatorPane(ObservableList<IonPartDefinition> parts) {
+    setPadding(DEFAULT_PADDING_INSETS);
+    final Label title = newBoldTitle("List of ion building blocks");
```
SteffenHeu: Maybe add a info box above the ionPartDefinitionPane that describes what a building block is and how it relates to the ion types.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/IonTypeCreatorPane.java
```diff
@@ -0,0 +1,57 @@
...
+import javafx.scene.layout.BorderPane;
+import javafx.scene.layout.VBox;
+
+class IonTypeCreatorPane extends BorderPane {
+
+  public IonTypeCreatorPane(ObservableList<IonType> types,
+      ObservableList<IonPartDefinition> partsDefinitions) {
+    setPadding(DEFAULT_PADDING_INSETS);
+
+    final Label title = newBoldTitle("All currently defined ion types:");
```
SteffenHeu: similar to the ion part creator pane, add an info box that explains the relation to the ion library and the building blocks.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/IonTypeDefinitionPane.java
```diff
@@ -0,0 +1,213 @@
...
+      }
+    });
+  }
+
+  private Node createIonTypeByStringPane() {
+    var lbParsingResult = newBoldLabel(
+        parsedIonType.map(ion -> ion.toString(IonTypeStringFlavor.FULL_WITH_MASS))
+            .orElse("Cannot parse input"));
+
+    final TextField inputText = newTextField(10, parsedIonTypeString,
```
SteffenHeu, line 122: Currently the input `M+[13]C-[12]C` is parsed correctly to a mass diff of 1.00335 but it still says that 1 part is unknown. The issue is that the charge is unknown. in such a case the error message should be more descriptive

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/IonPartDefinitionPane.java
```diff
@@ -0,0 +1,336 @@
...
+import javafx.util.Duration;
+import org.controlsfx.validation.ValidationSupport;
+import org.jetbrains.annotations.NotNull;
+import org.openscience.cdk.interfaces.IMolecularFormula;
+
+/**
+ * Pane to show a {@link IonPartDefinition} that acts as a definition of alternative names for ion
+ * parts for parsing. Like ACN is a chemical name synonym for a formula.
+ */
+public class IonPartDefinitionPane extends BorderPane {
```
SteffenHeu, line 77: If i define an ion part called "13C" with the formula `[13]C` everything seems to be correct. but when i want to use it as in `M+13C-C` it is not parsed correctly. So i think the validation needs to also be applied to the name of the ion part definition as it should be possible to use it in an ion type definition, or?

mzmine-community/src/main/java/io/github/mzmine/util/FormulaParser.java
```diff
@@ -0,0 +1,90 @@
...
+
+      // CDK parses charge only from [H]+ not from H+ so we add this behavior here
+      if (FormulaUtils.isUncharged(f)) {
+        final Integer charge = IonTypeParser.parseChargeOrElse(formula, null);
+        f.setCharge(charge);
+      }
+
+      return FormulaUtils.replaceAllIsotopesWithoutExactMass(f);
+    } catch (Exception e) {
+      logger.log(Level.SEVERE, "Cannot create formula for: " + formula, e);
```
SteffenHeu: Not sure if logging this as severe is the correct approach. While into a formula field this is triggered and then logs several severe exceptions

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/global/GlobalIonLibraryService.java
```diff
@@ -0,0 +1,679 @@
...
+
+    // sort
+    merged.sort(IonPartSorting.DEFAULT_NEUTRAL_THEN_LOSSES_THEN_ADDED.getComparator());
+    merged.trimToSize();
+    return merged;
+  }
+
+  private static List<IonType> mergeTypes(final @NotNull List<IonType> first,
+      final @NotNull List<IonType> second) {
+    // TODO maybe remap all parts to the global parts list to unify them
```
SteffenHeu: tag

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/global/GlobalIonLibraryService.java
```diff
@@ -0,0 +1,679 @@
...
+      }
+    }
+
+    // add those from second list check for conflicts
+    for (IonType ion : second) {
+      final IonType potentialConflict = existingNameMap.get(ion.name());
+      if (potentialConflict != null) {
+        // otherwise its equal, nothing to do then
+        if (!Objects.equals(potentialConflict, ion)) {
+          // TODO conflict resolution
```
SteffenHeu: does that below already count as conflict resolution?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/SimpleIonLibrary.java
```diff
@@ -0,0 +1,66 @@
...
+package io.github.mzmine.datamodel.identities.iontype;
+
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A simple ion library used by the parameter. Use {@link #toSearchableLibrary(boolean)} for an
+ * optimized version for searches.
+ */
+public record SimpleIonLibrary(@NotNull String name, @NotNull List<IonType> ions) implements
```
SteffenHeu: i currently dont see that the ions cannot be edited from the outside. (see comment in parameter component) return a copy here or make this an immutable list or so?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/CompoundsByNames.java
```diff
@@ -0,0 +1,69 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.datamodel.identities.iontype;
+
+import java.util.Optional;
+import org.jetbrains.annotations.NotNull;
+
+public class CompoundsByNames {
```
SteffenHeu: IonPartNames?

SteffenHeu: or better IonPartAliases

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonSearchRow.java
robinschmid: This object is used in the SearchableIonLibrary to search for ion pairs. It prepares the feature list row by extracting the charge state and the polarity to get the signedCharge (charge in FeatureListRow is usually absolute value). 
Also it decouples the SearchableIonLibrary from FeatureListRow so that you can search any mz and charge pair. 
For example if you want to search for Spectral signals etc. 
Also we could prepare all the FeatureListRows once and then search them pairwise.

I tried to improve the javadoc further.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/IonType.java
robinschmid: This is not a parallel stream so it is handled on this thread blocking and is the same as a loop or list.forEach. 
The stream has is more convenient to filter

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
robinschmid: `adduct.isUndefinedAdduct()` is the +? 
and the `adduct.isUndefinedMass()` is true if any part of this ion type has an undefined mass. 
For the CompoundDBAnnotation it would not make sense to build an instance for an IonType with undefined mass - we would not be able to calculate the neutral mass from that.

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/tolerances/MZTolerance.java
robinschmid: you are right.

mzmine-community/src/main/java/io/github/mzmine/util/FormulaUtils.java
robinschmid: I was not 100% sure how to best do this. We are not sure if this is the ion or neutral mol formula. 
I changed it now, is it good like this? 
```java
    // if the mz of formula already matches the mz of annotation then the formula is the correct
    // ion formula - otherwise might be neutral molecule formula that needs to be ionized by adduct
    final Double mz = annotation.getPrecursorMZ();
    final double formulaMZ = FormulaUtils.calculateMzRatio(molecularFormula);
    if (mz != null && MZTolerance.FIFTEEN_PPM_OR_FIVE_MDA.checkWithinTolerance(formulaMZ, mz)) {
      return molecularFormula;
    }
```

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/ionidentity/IonLibraryComponent.java
robinschmid: Yeah there were still some TODOs in the component. 
Now it works. 
I created a TextField that has another value class and uses StringConverter to convert to the value. It is different to bindBidirectional as this usually clears the textfield if the string does not convert to a proper value. 

So I use listeners here on the text and value properties

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/ionidentity/IonLibraryComponent.java
robinschmid: True. Good catch. The parameter will now always create a copy

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
SteffenHeu: ok, i was just asking because the check for "?" in the name was required anyway previously

mzmine-community/src/main/java/io/github/mzmine/util/FormulaUtils.java
SteffenHeu: maybe use a smaller tolerance than 0.005/15pm? maybe 0.0005 and 5ppm?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/GlobalIonLibrariesInteractor.java
robinschmid: Need to do this later. Maybe we can talk this through what should happen when and which version of the libraries should be kept.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/IonPartCreatorPane.java
robinschmid: I added the ? button to link to the documentation (will finish the doc page once we finish the PR and components for screenshots.) 

Do you think I should add a permanent info pane on top still? Or is the button + tooltips enoguh?

https://github.com/mzmine/mzmine_documentation/pull/268

image

image

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/IonTypeCreatorPane.java
robinschmid: see comment above, added ? button for documentation with tooltip. 
Can also add an info pane if that is better.

mzmine-community/src/main/java/io/github/mzmine/util/FormulaParser.java
robinschmid: Removed the logs.
There are usually no logs. Was just for debugging as there were some weird behaviors of the formula parser.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/SimpleIonLibrary.java
robinschmid: Internal libraries now use an UnmodifiableIonLibrary and the parameter now uses library.copy() to create a defensive copy (only needed for SimpleIonLibrary.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/CompoundsByNames.java
robinschmid: Used IonPartAliases

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/IonPartDefinitionPane.java
robinschmid: oh if the name starts with a number we would have to wrap the name into (13C) otherwise it is not clear what the count is like +2(13C) 
Usually you could write the formula as [13C] or [13]C

robinschmid: Now works and automatically adds () around the part.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/IonPartCreatorPane.java
SteffenHeu: I'd prefer something similar to the info boxes in the documentation

PR #3077, opened 2026-01-12
Make SpectralDBAnnotation modular

SteffenHeu:
use a ModularDataModel in SpectralDBAnnotation so we don't have to manually implement new types as static fields.
The SpectralDBAnnotation should only contain data that refers to the row/spectral library entry, such as the ppm error between row and entry. "static" info that is the same for all entries, belongs into the entry.

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/CONST.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/CONST.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2025 The mzmine Development Team
+ * Copyright (c) 2004-2026 The mzmine Development Team
  *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
@@ -86,7 +86,7 @@ public class CONST {
   /**
    *
    */
-  public static final String XML_VERSION_ATTR = "v";
+  public static final String XML_VERSION_ATTR = "version";
 
   private CONST() {
   }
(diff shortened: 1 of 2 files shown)


mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralDBAnnotation.java
```diff
@@ -68,39 +78,113 @@ public class SpectralDBAnnotation implements FeatureAnnotation, Comparable<Spect
...
     this.queryScan = queryScan;
     this.entry = entry;
     this.similarity = similarity;
-    this.ccsRelativeError = ccsRelativeError;
-    this.testedPrecursorMz = testedPrecursorMz;
-    this.testedRt = testedRt;
-    this.riDiff = riDiff;
+
+    if (ccsRelativeError != null) {
+      set(CCSRelativeErrorType.class, ccsRelativeError);
```
robinschmid, line 93: maybe it would be a good idea to just handle the null in the ModularDataModel?
Currently it is fine to call set(type, null) when using the columnar model. 
I think the ModularDataModelMap also handles null but we could also add if(value==null) model.remove to remove the key instead of mapping null value to it.

mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/CONST.java
```diff
@@ -84,6 +83,11 @@ public class CONST {
   public static final String XML_OTHER_FILE_DESC_ATTR = "otherdata_filedesc";
   public static final String XML_OTHER_TIME_SERIES_NAME_ATTR = "otherdata_timeseriesname";
 
+  /**
+   *
+   */
+  public static final String XML_VERSION_ATTR = "v";
```
robinschmid: as this is a new property, rename to version so it is clearer in the xml

robinschmid (dismissed):
looks good to me. 
Only an idea to explicitly handle null in the data model.set

PR #3088, opened 2026-01-16
Refactor annotation summary and improve feature annotation handling

SteffenHeu:
---

### Changes introduced in this pull request:

- Added annotation summary component
- Updated annotation handling to support new types like PreferredAnnotationType, SchymanskiLevel, and SumnerLevel.
- Preferred annotation column that covers Lipid matches, Speclib matches, compound db
- Preferred annotation is selected by the evidence given for it
- Ranking of those annotations is done in AnnotationSummary which uses schymanski level as it is more granular

Will need to update integration tests before merging, since this will add new columns to the csv export

Example:
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummary.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummary.java
@@ -193,9 +193,9 @@ public class AnnotationSummary implements Comparable<AnnotationSummary> {
    * https://pmc.ncbi.nlm.nih.gov/articles/PMC3772505/
    */
   @NotNull
-  public SumnerLevel deriveSumnerLevel() {
+  public MsiAnnotationLevel deriveMsiLevel() {
     if (annotation == null) {
-      return SumnerLevel.LEVEL_4;
+      return MsiAnnotationLevel.LEVEL_4;
     }
 
     return switch (annotation) {
@@ -203,27 +203,27 @@ public class AnnotationSummary implements Comparable<AnnotationSummary> {
         if (l.getMsMsScore() != null && l.getMsMsScore() > 0d) {
           switch (l.getLipidAnnotation()) {
             case MolecularSpeciesLevelAnnotation _ -> {
-              yield SumnerLevel.LEVEL_2;
+              yield MsiAnnotationLevel.LEVEL_2;
             }
             case SpeciesLevelAnnotation _ -> {
-              yield SumnerLevel.LEVEL_3;
+              yield MsiAnnotationLevel.LEVEL_3;
             }
             case null, default -> {
             }
           }
         }
-        yield SumnerLevel.LEVEL_4;
+        yield MsiAnnotationLevel.LEVEL_4;
       }
-      case SpectralDBAnnotation _ -> SumnerLevel.LEVEL_2;
+      case SpectralDBAnnotation _ -> MsiAnnotationLevel.LEVEL_2;
       case CompoundDBAnnotation c -> {
         if (rtScore() > 0 || riScore() > 0) {
-          yield SumnerLevel.LEVEL_2;
+          yield MsiAnnotationLevel.LEVEL_2;
         } else if (c.get(SiriusCsiScoreType.class) != null) {
-          yield SumnerLevel.LEVEL_3;
+          yield MsiAnnotationLevel.LEVEL_3;
         }
-        yield SumnerLevel.LEVEL_4;
+        yield MsiAnnotationLevel.LEVEL_4;
       }
-      default -> SumnerLevel.LEVEL_4;
+      default -> MsiAnnotationLevel.LEVEL_4;
     };
   }
 
@@ -231,11 +231,11 @@ public class AnnotationSummary implements Comparable<AnnotationSummary> {
    * https://pubs.acs.org/doi/10.1021/es5002105
    */
   @NotNull
-  public SchymanskiLevel deriveSchymanskiLevel() {
+  public SchymanskiAnnotationLevel deriveSchymanskiLevel() {
     final double ipScoreMatchThreshold = 0.75;
 
     if (annotation == null) {
-      return SchymanskiLevel.LEVEL_5;
+      return SchymanskiAnnotationLevel.LEVEL_5;
     }
 
     return switch (annotation) {
@@ -243,30 +243,31 @@ public class AnnotationSummary implements Comparable<AnnotationSummary> {
         if (l.getMsMsScore() != null && l.getMsMsScore() > 0d) {
           switch (l.getLipidAnnotation()) {
             case MolecularSpeciesLevelAnnotation _ -> {
-              yield SchymanskiLevel.LEVEL_2a;
+              yield SchymanskiAnnotationLevel.LEVEL_2a;
             }
             case SpeciesLevelAnnotation _ -> {
-              yield SchymanskiLevel.LEVEL_2b;
+              yield SchymanskiAnnotationLevel.LEVEL_2b;
             }
             case null, default -> {
             }
           }
         }
         // lipids are hard to judge by isotope scores so default to level 5
-        yield SchymanskiLevel.LEVEL_5;
+        yield SchymanskiAnnotationLevel.LEVEL_5;
       }
       case SpectralDBAnnotation s ->
-          riScore() > 0 || rtScore() > 0 ? SchymanskiLevel.LEVEL_1 : SchymanskiLevel.LEVEL_2a;
+          riScore() > 0 || rtScore() > 0 ? SchymanskiAnnotationLevel.LEVEL_1
+              : SchymanskiAnnotationLevel.LEVEL_2a;
       case CompoundDBAnnotation c -> {
         if (c.get(SiriusCsiScoreType.class) != null) {
-          yield SchymanskiLevel.LEVEL_3;
+          yield SchymanskiAnnotationLevel.LEVEL_3;
         } else if (isotopeScore() >= ipScoreMatchThreshold) {
-          yield SchymanskiLevel.LEVEL_4;
+          yield SchymanskiAnnotationLevel.LEVEL_4;
         } else {
-          yield SchymanskiLevel.LEVEL_5;
+          yield SchymanskiAnnotationLevel.LEVEL_5;
         }
       }
-      default -> SchymanskiLevel.LEVEL_5;
+      default -> SchymanskiAnnotationLevel.LEVEL_5;
     };
   }
 
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
@@ -97,8 +97,7 @@ public class AnnotationSummaryType extends DataType<AnnotationSummary> implement
     } else {
       // parent type not set -> is the "summary"/best annotation in the row -> get best annotation and grab summary from there
       column.setCellValueFactory(cdf -> new ReadOnlyObjectWrapper<>(
-          CompoundAnnotationUtils.getBestFeatureAnnotation(cdf.getValue().getValue())
-              .map(a -> AnnotationSummary.of(cdf.getValue().getValue(), a)).orElse(null)));
+          CompoundAnnotationUtils.getBestAnnotationSummary(cdf.getValue().getValue())));
     }
 
     column.setCellFactory(col -> new MicroChartCell());
@@ -203,7 +202,7 @@ public class AnnotationSummaryType extends DataType<AnnotationSummary> implement
       tooltip.setText("""
           Annotation levels:
           %s
-          %s""".formatted(annotationSummary.deriveSumnerLevel(),
+          %s""".formatted(annotationSummary.deriveMsiLevel(),
           annotationSummary.deriveSchymanskiLevel()));
       setTooltip(tooltip);
 
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
@@ -117,16 +117,13 @@ public class MolecularStructureType extends DataType<MolecularStructure> impleme
       return null;
     }
 
-    final DataType<?> mainType = switch (superType) {
-      case PreferredAnnotationType pt -> {
-        FeatureAnnotation annotation = row.getPreferredAnnotation();
-        if (annotation != null) {
-          yield DataTypes.get(annotation.getDataType());
-        }
-        yield superType;
-      }
-      case null, default -> superType;
-    };
+    final DataType<?> mainType;
+    if (superType instanceof PreferredAnnotationType pt
+        && row.getPreferredAnnotation() instanceof FeatureAnnotation a) {
+      mainType = DataTypes.get(a.getDataType());
+    } else {
+      mainType = superType;
+    }
 
     return () -> FxThread.runLater(() -> {
       if (mainType instanceof CompoundDatabaseMatchesType) {
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/modifiers/MappingType.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/modifiers/MappingType.java
@@ -31,6 +31,11 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 /**
+ * Can be used to generate values of data types on demand, e.g. if they are calculated when
+ * requested and the data is not stored in the data model. See
+ * {@link io.github.mzmine.datamodel.features.types.annotations.PreferredAnnotationType} for an
+ * example.
+ *
  * <b>WARNING</b> Types extending this interface MUST NOT call
  * {@link io.github.mzmine.datamodel.features.ModularFeatureListRow#get(DataType)} with themself as
  * an argument, as this is exactly where {@link this#getValue(ModularDataModel)} is called, which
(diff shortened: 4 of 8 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummary.java
```diff
@@ -0,0 +1,303 @@
...
+import java.util.Comparator;
+import java.util.function.Function;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class AnnotationSummary implements Comparable<AnnotationSummary> {
+
+  private static final double maxPpmDiff = 15;
+  private static final double maxRtDiff = 0.3;
+  private static final double maxCcsDev = 0.10;
```
robinschmid, line 51: Maybe higher? 
Not sure what the expected deviation is when you measure on different systems but I would suspect higher from the general CCS scale.
Maybe 2 as max deviation? 
But you know better.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/SchymanskiLevel.java
```diff
@@ -0,0 +1,59 @@
...
+
+  SchymanskiLevel(int numberLevel) {
+    this(numberLevel, "");
+  }
+
+  public int numberLevel() {
+    return numberLevel;
+  }
+
+  // ... existing code ...
```
robinschmid: llm?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/SchymanskiLevel.java
```diff
@@ -0,0 +1,59 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.datamodel.features.annotationpriority;
+
+import org.jetbrains.annotations.NotNull;
+
+public enum SchymanskiLevel {
```
robinschmid: Rename to SchymanskiAnnotationLevel and add paper to doc
https://doi.org/10.1021/es5002105

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/SumnerLevel.java
```diff
@@ -0,0 +1,46 @@
...
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.datamodel.features.annotationpriority;
+
+public enum SumnerLevel {
```
robinschmid: Rename to SumnerAnnotationLevel or MSIAnnotationLevel
Add publication link https://pubmed.ncbi.nlm.nih.gov/24039616/ or https://doi.org/10.1007/s11306-007-0082-2

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummary.java
SteffenHeu: it's a percentage here

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
```diff
@@ -0,0 +1,293 @@
...
+      column.setCellValueFactory(cdf -> {
+        var value = (List<? extends FeatureAnnotation>) cdf.getValue().getValue()
+            .get((DataType<?>) parentType);
+        return new ReadOnlyObjectWrapper<>(
+            value != null && !value.isEmpty() ? AnnotationSummary.of(cdf.getValue().getValue(),
+                value.getFirst()) : null);
+      });
+    } else {
+      // parent type not set -> is the "summary"/best annotation in the row -> get best annotation and grab summary from there
+      column.setCellValueFactory(cdf -> new ReadOnlyObjectWrapper<>(
```
robinschmid, line 99: I guess this is old? 
The PreferredAnnotationType is also a parent type.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
```diff
@@ -114,11 +116,23 @@ public double getPrefColumnWidth() {
     if (!(value instanceof MolecularStructure structure)) {
       return null;
     }
+
+    final DataType<?> mainType = switch (superType) {
```
robinschmid: we now have the switch(superType) and below the if(mainType instanceof) checks. 
Why not add this check also below to the mainType instanceof? Can also use the switch pattern but its the same

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/PreferredAnnotationType.java
```diff
@@ -0,0 +1,109 @@
...
+  @Override
+  public <K> @Nullable K map(@NotNull DataType<K> subType, FeatureAnnotation parentItem) {
+    if (parentItem == null) {
+      return null;
+    }
+    return (K) switch (subType) {
+      case CompoundNameType _ -> parentItem.getCompoundName();
+      case AnnotationSummaryType _ -> parentItem;
+      case PrecursorMZType _ -> parentItem.getPrecursorMZ();
+      case MolecularStructureType _ -> parentItem.getStructure();
```
robinschmid: I think we should also add the Smiles, inchi, inchikey here so that they are saved in the csv export. And the internalID?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/SpectralLibraryMatchesType.java
```diff
@@ -163,6 +166,7 @@ public List<DataType> getSubDataTypes() {
       case InternalIdType _ -> entry.getOrElse(DBEntryField.INTERNAL_ID, null);
       case JsonStringType _ -> entry.getOrElse(DBEntryField.JSON_STRING, null);
       case RIDiffType _ -> match.getRiDiff();
+      case AnnotationSummaryType _ -> null; // created on demand in cell factory
```
robinschmid, line 169: why not just generate the object here?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/fx/DataTypeCellFactory.java
```diff
@@ -78,6 +78,10 @@ public TreeTableCell<ModularFeatureListRow, Object> call(
 //    logger.log(Level.INFO, "Creating cell in DataTypeCellFactory");
     return new TreeTableCell<>() {
 
+      {
+        setWrapText(type instanceof StringType);
```
robinschmid, line 82: weird I thought we would need to set the preferred width on the datatype if we use wrap but somehow it works. 
I had seen some issues with other types where wrap was on

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/modifiers/MappingType.java
```diff
@@ -0,0 +1,39 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ *
+ * @param <T> Must be the same as the {@link DataType <T>}
+ */
+public interface MappingType<T> {
+
+  @Nullable T getValue(@NotNull ModularDataModel model);
```
robinschmid, line 51: Maybe this class is not needed after the refactor? 
Because the datatype is now just a single preferred annotation that is actually set to the datamodel and the regular get will work. 

And the List<FeatureAnnotation> that we need for edits will need to come from somewhere else. 

Otherwise if still needed, good to add some javadoc to method and class what it is used for.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
```diff
@@ -726,4 +719,12 @@ public void addLipidAnnotation(MatchedLipid matchedLipid) {
   public String toString() {
     return FeatureUtils.rowToString(this);
   }
+
+  @Override
+  public <T> @Nullable T get(DataType<T> key) {
+    if (key instanceof MappingType<?> mt) {
```
robinschmid, line 751: might not be needed if MergingType is removed. Just to flag

mzmine-community/src/main/java/io/github/mzmine/util/annotations/CompoundAnnotationUtils.java
```diff
@@ -363,4 +369,154 @@ public static String getAdditionalJson(FeatureAnnotation annotation) {
       default -> null;
     };
   }
+
+  /**
+   *
+   * @param topN The maximum number of annotations per type.
+   * @return A list of all feature annotations in no particular order.
+   */
+  public static @NotNull List<@NotNull FeatureAnnotation> getFeatureAnnotations(
```
robinschmid: rename to getTopNAnnotationsPerType similar to its caller getTopAnnotationsPerType

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/PreferredAnnotationType.java
```diff
@@ -0,0 +1,109 @@
...
+  @Override
+  public @NotNull String getHeaderString() {
+    return "Preferred Annotation";
+  }
+
+  @Override
+  public @Nullable List<? extends FeatureAnnotation> getValue(@NotNull ModularDataModel model) {
+
+    final DataType<?> preferredAnnotationType = CompoundAnnotationUtils.getBestAnnotationType(
+        (FeatureListRow) model);
```
robinschmid: I guess this is already changed with the change of this PreferredAnnotationType class structure. 
Just something to check in the final version. 

This should show the sorted list of all annotations.

mzmine-community/src/main/java/io/github/mzmine/util/annotations/CompoundAnnotationUtils.java
```diff
@@ -363,4 +369,154 @@ public static String getAdditionalJson(FeatureAnnotation annotation) {
...
+    }
+    return getTopAnnotationsPerType(model).stream().map(a -> AnnotationSummary.of(model, a))
+        .min(AnnotationSummary.HIGH_TO_LOW_CONFIDENCE).orElse(null);
+  }
+
+  /**
+   * @return The data type of the best {@link FeatureAnnotation} or null if no annotation exists.
+   */
+  public static @Nullable DataType<?> getBestAnnotationType(@Nullable FeatureListRow row) {
+    return getBestFeatureAnnotation(row).map(a -> DataTypes.get(a.getDataType())).orElse(null);
```
robinschmid: now that each class has FeatureAnnotation.getDataType(), maybe this method is not needed because the caller should just do getBestFeatureAnnotation.
getBestFeatureAnnotation is already a bit more complex to compute and I can see myself using this method here to get the type and also call other methods that all do the same logic of creating the whole list and sorting all annotations. 
So maybe best to keep it to the most needed functions?

mzmine-community/src/main/java/io/github/mzmine/util/annotations/CompoundAnnotationUtils.java
```diff
@@ -363,4 +369,154 @@ public static String getAdditionalJson(FeatureAnnotation annotation) {
...
+    return getTopAnnotationsPerType(model).stream()
+        .collect(Collectors.toMap(FeatureAnnotation::getDataType, a -> a));
+  }
+
+  public static @Nullable AnnotationSummary getBestAnnotationSummary(
+      @Nullable final FeatureListRow model) {
+    if (model == null) {
+      return null;
+    }
+    return getTopAnnotationsPerType(model).stream().map(a -> AnnotationSummary.of(model, a))
```
robinschmid, line 467: I think it might be better to use all annotations to find the actual best one by AnnotationSummary? 
Currently it just uses the first annotation for each type but if this sorting should be based on annotation summary then this is different to how sorting is done currently in the other types. So maybe best to work with all annotations.

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralLibraryEntry.java
```diff
@@ -185,4 +190,33 @@ default Integer toInteger(Object v) {
    */
   MolecularStructure getStructure();
 
+  @Nullable
+  default IsotopePattern calculateIsotopePattern() {
```
robinschmid, line 194: why duplicate code from feature annotation here?
SpectralDBAnnoation calls `entry.calculateIsotopePattern` but could just call `FeatureAnnotation.super.calculateIsotopePattern()`?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummary.java
robinschmid: ok that makes sense.
Then maybe best to capture this in the name or javadoc

robinschmid: RT and RI diff are absolute values though right?

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralLibraryEntry.java
SteffenHeu: because the isotope pattern is specific to the library entry, not the SpectralDBAnnotation.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/SchymanskiLevel.java
SteffenHeu: yeah i let it refactor from class to enum

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
SteffenHeu: yeah, but we can still leave it, or? Maybe there's a usecase in the future

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/MolecularStructureType.java
SteffenHeu: my reason was that if i check for preferred type below, i still need to get the type of the annotation in that case and then i would need to duplicate the code that opens the compound or spectral mirror tab in there.
i can make this swtich to an if, probably more concise

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/PreferredAnnotationType.java
SteffenHeu: i changed the modular csv export instead, so it correctly indexes the preferred annotations and all relevant sub types are exported.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/SpectralLibraryMatchesType.java
SteffenHeu: it's not stored in the row or the type anyway, so it does not matter where it's created. Also it is not exported. so i figured i might just put it into the cell factory.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/modifiers/MappingType.java
SteffenHeu:
```java
/**
 * Can be used to generate values of data types on demand, e.g. if they are calculated when
 * requested and the data is not stored in the data model. See
 * {@link io.github.mzmine.datamodel.features.types.annotations.PreferredAnnotationType} for an
 * example.
 *
 * <b>WARNING</b> Types extending this interface MUST NOT call
 * {@link io.github.mzmine.datamodel.features.ModularFeatureListRow#get(DataType)} with themself as
 * an argument, as this is exactly where {@link this#getValue(ModularDataModel)} is called, which
 * would lead to a stack overflow. Should such a case be necessary, we recommend to directly use
 * {@link
 * io.github.mzmine.datamodel.features.columnar_data.ColumnarModularDataModelRow#get(DataType)}
 * instead.
 *
 * @param <T> Must be the same as the {@link DataType <T>}
 */
```

mzmine-community/src/main/java/io/github/mzmine/util/annotations/CompoundAnnotationUtils.java
SteffenHeu: Good point. My logic was that it's debatable, because the annotations are sorted from best to worst by default. if the user selects a specific annotation in the lipids by himself making it the first, i would say we should also take that one into account for the preferred annotation. 
Otherwise you change the best lipid annotation and then the preferred annotation is still the lipid you did not want.

robinschmid: Ok makes sense

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralLibraryEntry.java
robinschmid: Ok makes sense. 
So was more to have the same method in the entry.

PR #3092, opened 2026-01-17
Add structure rendering options to preferences

### LINKED ISSUE #3060: [Bug]: Library Match Export
### mzmine version/build

4.8.5

### Operating system

Windows

### Observed issue behavior

When exporting a library match, the structure visualization is not working properly in the pdf or emf. (See pdf and the screenshot)

[baptifoline_match.pdf](https://github.com/user-attachments/files/24195087/baptifoline_match.pdf)

Image

### Steps to reproduce

See pictures

robinschmid:
This PR combines the other two #3090 and #3091 and we can decide what we take into the next release. 

Added parameters to the preferences to control the visualization of structures. 
Default option is to keep a fixed bond length and label size but can also set to fit the size of the pane. 

image

This PR contains additional improvements for double bonds and other bond types by modifying the default sizes.

Left is the current version (height of cells is a bit higher left maybe because I changed the min width and height of the structure pane)
right is the PR version. 
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/StructureRenderParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/StructureRenderParameters.java
@@ -23,9 +23,8 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
 
-package io.github.mzmine.gui.preferences;
+package io.github.mzmine.modules.visualization.molstructure;
 
-import io.github.mzmine.modules.visualization.molstructure.Structure2DRenderConfig;
 import io.github.mzmine.modules.visualization.molstructure.Structure2DRenderConfig.Sizing;
 import io.github.mzmine.parameters.impl.SimpleParameterSet;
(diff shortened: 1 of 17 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/StructureRenderParameters.java
```diff
@@ -0,0 +1,71 @@
...
+import io.github.mzmine.parameters.parametertypes.DoubleParameter;
+import io.github.mzmine.parameters.parametertypes.submodules.ParameterSetParameter;
+import java.text.DecimalFormat;
+import java.util.List;
+
+/**
+ * Uses {@link ParameterSetParameter} to allow additional parameters that are extra compared to the
+ * ones defined for {@link Structure2DRenderConfig}
+ */
+public class StructureRenderParameters extends SimpleParameterSet {
```
SteffenHeu, line 40: if we keep it, move to modules/visualization/StructureRendererParameters or?

PR #3093, opened 2026-01-17
Restore defaults on specific parameter sets

robinschmid:
only the latest commit is important. 
This describes a way to restore defaults for parameters that use the OptionalModuleComponent (ParameterSetParameter and OptionalParameterSetParameter). Just adds a button to restore the defaults

image


robinschmid:
@SteffenHeu the last two commits added an interface to restore default parameters in a ParameterSetParameter and OptionalParameterSetParameter. 
I think this might be a great pattern if we need defaults on parameters or parametersets in the future. 
But will close the PR for now as we are not adding the parameter to the preferences.

PR #3096, opened 2026-01-20
Fix modal dialog handling

robinschmid:
The dialogs are behaving weird for long texts. I tested this to see if a scroll pane was added automatically but the dialog would just appear off screen and block the main window if it is modal. 

Somehow it had to do mostly with modal dialogs so I changed how they are opened directly on the FX thread if we are on the thread. 

Also the dialog is pushed to the center of the focussed window if it is off screen.


javafx-framework/src/main/java/io/github/mzmine/javafx/dialogs/DialogLoggerUtil.java
```diff
@@ -278,12 +304,18 @@ public static Optional<ButtonType> showDialog(AlertType type, final @Nullable Wi
    */
   private static @NotNull Alert createAlert(final AlertType type, final @Nullable Window owner,
       final String title, final String message, @Nullable final ButtonType... buttons) {
+
+    final TextFlow node = FxTextFlows.newTextFlow(FxTexts.text(message));
+    final ScrollPane scrollPane = new ScrollPane(node);
```
SteffenHeu: shouldnt we do the wrapping in the scroll pane in the inner-most method?

javafx-framework/src/main/java/io/github/mzmine/javafx/dialogs/DialogLoggerUtil.java
```diff
@@ -297,13 +329,77 @@ public static Optional<ButtonType> showDialog(AlertType type, final @Nullable Wi
...
+      } else {
+        centerAlertOnWindow(alert);
+      }
+    });
+
     return alert;
   }
 
+  private static void centerAlertOnWindow(Alert alert) {
+    final Window owner2 = alert.getOwner();
```
SteffenHeu: the owner is not always initialised, but it is notnull in getCurrentScreen

javafx-framework/src/main/java/io/github/mzmine/javafx/dialogs/DialogLoggerUtil.java
SteffenHeu: add overloaded method with textflow, move wrapping into there.

mzmine-community/src/main/java/io/github/mzmine/gui/MZmineGUI.java
```diff
@@ -704,11 +704,17 @@ public void displayMessage(String title, String msg) {
...
       node.getChildren().addAll(FxTexts.text(" "), FxTexts.hyperlinkText(url));
     }
-    DialogLoggerUtil.showMessageDialog(title, node);
+    final ScrollPane scrollPane = new ScrollPane(node);
+    scrollPane.setFitToWidth(true);
+    scrollPane.setFitToHeight(true);
+    scrollPane.setMaxWidth(800);
+    scrollPane.setMaxHeight(800);
+    scrollPane.setPannable(true);
+    DialogLoggerUtil.showMessageDialog(title, scrollPane);
```
SteffenHeu: remove wrapping after changes

javafx-framework/src/main/java/io/github/mzmine/javafx/dialogs/DialogLoggerUtil.java
robinschmid: good catch. 
getCurrentScreen now handles null

javafx-framework/src/main/java/io/github/mzmine/javafx/dialogs/DialogLoggerUtil.java
robinschmid: Moved the creation of Scrollpane into the method with Node content, doing isntanceof TextFlow or TextArea

PR #3098, opened 2026-01-20
Harmonize RawDataFilePlaceholder with WeakReference

robinschmid:
- was no memory leak because batch last files was never cloned
- still made it a palceholder to not accidentally hold on to the references in the future. 
- Also harmonized the placeholder a bit like the feature list placeholder with convenience method to map multiple RawDataFile and Placeholders

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilePlaceholder.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilePlaceholder.java
@@ -112,24 +112,7 @@ public class RawDataFilePlaceholder implements RawDataFile {
    */
   @Nullable
   public RawDataFile getMatchingFile() {
-    final RawDataFile raw = weakRawRef.get();
-    if (raw != null) {
-      return raw;
-    }
-
-    final MZmineProject proj = ProjectService.getProject();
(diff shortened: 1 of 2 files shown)


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilePlaceholder.java
```diff
@@ -100,6 +138,59 @@ public boolean matches(@Nullable final RawDataFile file) {
...
+        return direct;
+      }
+      // find in all files
+      for (RawDataFile file : allFiles) {
+        if (placeholder.matches(file)) {
+          return file;
+        }
+      }
+      return null; // missing file
+    });
```
SteffenHeu, line 153: i get that it's more performant to only retrieve the of all files ones, but i think it just makes it more complex to maintain. I would just use the getMatchingFile method

SteffenHeu: discussion: change single file method to wrap in array and use this one

robinschmid: done

PR #3099, opened 2026-01-21
Export logging.properties to user/.mzmine

robinschmid:
- makes logging properties user facing to edit the log level that should be captured. 
- uses an internal logging file for the default configuration that overwrites some of the user facing ones

@SteffenHeu should we use FINE or FINER as the default log level? 
SO far it was FINEST

### DIFF
--- a/mzmine-community/src/main/resources/logging.properties
+++ b/mzmine-community/src/main/resources/logging.properties
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2004-2022 The MZmine Development Team
+# Copyright (c) 2004-2026 The mzmine Development Team
 #
 # Permission is hereby granted, free of charge, to any person
 # obtaining a copy of this software and associated documentation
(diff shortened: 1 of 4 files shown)


mzmine-community/src/main/resources/logging.properties
```diff
@@ -64,26 +56,3 @@ java.util.logging.FileHandler.limit = 1073741824
...
-# If this property is true, the output should be appended to already existing file
-java.util.logging.FileHandler.append=false
-java.util.logging.FileHandler.level=FINEST
-# Default level for ConsoleHandler. This can be used to
-# limit the levels that are displayed on the console even
-# when the global default has been set to a trace level
-java.util.logging.ConsoleHandler.level=FINEST
-java.util.logging.ConsoleHandler.formatter=java.util.logging.SimpleFormatter
-#java.util.logging.ConsoleHandler.formatter = io.github.mzmine.util.logging.ConsoleFormatter
-java.util.logging.SimpleFormatter.format=%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS %4$-6s %2$s %5$s%6$s%n
```
SteffenHeu, line 89: when making this formatter user editable, we need some kind of parsability check to ensure we still produce valid logs

mzmine-community/src/main/resources/logging.properties
```diff
@@ -33,28 +33,20 @@
...
-org.keycloak.level = SEVERE
-org.apache.http.client.protocol.level = SEVERE
-io.undertow.level = SEVERE
-org.xnio.level = SEVERE
-org.jboss.threads.level = SEVERE
-
-# List of global handlers
-handlers = java.util.logging.ConsoleHandler,io.github.mzmine.util.logging.StatusBarHandler,java.util.logging.FileHandler
+# or change the level logged to file or to console
+java.util.logging.FileHandler.level=FINE
```
SteffenHeu, line 45: i wouldnt say I'm confident that we don't do any kind of logging necessary to properly debug problems in finest level. 

I'm fine with making it editable by the user, but i think i'd rather keep the default as finest for now

robinschmid: Yeah can keep it for now but then try to use the log levels better to at a later time switch the default higher

robinschmid: Makes sense

PR #3101, opened 2026-01-22
Bugfix/filter tooltips

JonasLueb:
bugfix for RowTypeFilter
add tooltips to van krevelen plot

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/vankrevelendiagram/VanKrevelenDiagramXYZDataset.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/vankrevelendiagram/VanKrevelenDiagramXYZDataset.java
@@ -149,13 +149,13 @@ class VanKrevelenDiagramXYZDataset extends AbstractXYZDataset implements Task, X
     };
   }
 
-  FeatureListRow getSelectedRow(final int item) {
+  FeatureListRow getRow(final int item) {
     return filteredRows.get(item);
   }
 
   @Nullable
   String getFormulaString(final int item) {
-    final Object ann = getSelectedRow(item).getPreferredAnnotation();
+    final Object ann = getRow(item).getPreferredAnnotation();
     return ann == null ? null : getFormulaFromAnnotation(ann);
   }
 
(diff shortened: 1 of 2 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/vankrevelendiagram/VanKrevelenToolTipGenerator.java
```diff
@@ -0,0 +1,103 @@
...
+    this.bubbleScaleLabel = bubbleScaleLabel;
+  }
+
+  @Override
+  public String generateToolTip(final XYDataset dataset, final int series, final int item) {
+    if (dataset instanceof VanKrevelenDiagramXYZDataset vkDataset) {
+      final FeatureListRow row = vkDataset.getSelectedRow(item);
+
+      final StringBuilder tooltip = new StringBuilder();
+      tooltip.append("Feature list ID: ").append(row.getID());
```
SteffenHeu: Row ID

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/vankrevelendiagram/VanKrevelenToolTipGenerator.java
```diff
@@ -0,0 +1,103 @@
...
+
+      final String formula = vkDataset.getFormulaString(item);
+      if (formula != null) {
+        tooltip.append("\nFormula: ").append(formula);
+      }
+
+      final String preferredAnnotationName = row.getPreferredAnnotationName();
+      if (preferredAnnotationName != null) {
+        tooltip.append("\n").append(preferredAnnotationName);
+      }
```
SteffenHeu: I'd use the preferred annotation if available and the formula otherwise. 
Currently the same formula is listed twice because the formula also counts as preferred annotation. My update will change that for some time until we make the ResultFormula a FeatureAnnotation. 
So better check if preferred annotation != null, then use that and fomrula otherwise then we are future safe directly.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/vankrevelendiagram/VanKrevelenDiagramXYZDataset.java
```diff
@@ -146,6 +149,20 @@ private String getFormulaFromAnnotation(Object annotation) {
     };
   }
 
+  FeatureListRow getSelectedRow(final int item) {
```
SteffenHeu: maybe only getRow? Selected implies selection in the GUI

PR #3105, opened 2026-01-22
Calculate most intense signal for compound db

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
```diff
@@ -589,4 +593,98 @@ default void enrichMetadata() {
...
+      String formula = compoundDBAnnotation.getFormula();
+      if (formula == null) {
+        MolecularStructure structure = compoundDBAnnotation.getStructure();
+        if (structure == null || structure.formulaString() == null) {
+          continue;
+        }
+        formula = structure.formulaString();
+      }
+
+      final Map<String, Integer> formulaCounts = FormulaUtils.parseFormula(formula);
```
robinschmid: I think I removed this method in the IIN branch because it is not really parsing all possible formulas.  

Best to just parse it to CDK formula. 
Either considering specific isotope notation [13]C as this might be a use case that users provide the exact formula for compound DB. Or alternatively also using the most abundant isotopes. 

In IIN branch there is also a method to then get the number of each element but there is also a method from cdk for this.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchTask.java
```diff
@@ -348,6 +348,10 @@ private List<CompoundDBAnnotation> getCompoundDBAnnotations(final @NotNull Strin
     } else {
       annotations.add(baseAnnotation);
     }
+    if (mzTolerance != null) {
+      annotations.addAll(
+          CompoundDBAnnotation.buildMostIntenseIsotopeRatios(annotations, mzTolerance));
```
robinschmid, line 356: So this is now done by default? 
Maybe add option to activate this. 

In my test it was very slow spending a lot of time on isotope pattern calculation.
Tested with the library building workflow removing the filename column from the wizard.
image

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundannotations/CompoundDBAnnotation.java
```diff
@@ -589,4 +593,98 @@ default void enrichMetadata() {
...
+        final String isotopeComposition = sip.getIsotopeComposition(mostIntenseIndex);
+        if (!isotopeComposition.contains(",")) { // may be multiple formulas (if merged)
+          mainIsotopePeak.put(FormulaType.class, isotopeComposition);
+          mainIsotopePeak.put(CommentType.class, isotopeComposition);
+        } else {
+          mainIsotopePeak.put(CommentType.class, "multiple: " + isotopeComposition);
+
+          // find the most intense individual isotope signal as representative
+          final IsotopePattern highResPattern = IsotopePatternCalculator.calculateIsotopePattern(
+              majorIsotopeIon, 0.005, 0d, adduct.getCharge(), adduct.getPolarity(), true);
```
robinschmid, line 666: maybe it is somehow possible to just calculate the isotope pattern once with little merging and then create a merged pattern from there. 
At least in my tests it was very slow, I cancelled after 10 min for the library building wizard with 4400 compounds

SteffenHeu: will have to see. i think the main time limiting factor is the generation of the isotope pattern with the labelled peaks. I'll try to optimise

robinschmid: As this is only a rough estimate of what the largest signal is - maybe we can just remove isotopes that are very rare and focus on 13C and other more common isotopes. 
Like removing HNO and all other elements

mzmine-community/src/main/java/io/github/mzmine/modules/tools/isotopeprediction/IsotopePatternEstimator.java
```diff
@@ -0,0 +1,342 @@
...
+      if (numOxygen >= 30) {
+        minAbundance = 0.002;
+      }
+      numRelevantElements += numOxygen;
+    }
+
+    return minAbundance;
+  }
+
+  private static boolean allowIsotope(@Nullable IIsotope isotope) {
```
robinschmid, line 336: unused new method. 
Maybe use or discard? Could be used in the main getIsotopes method

mzmine-community/src/main/java/io/github/mzmine/modules/tools/isotopeprediction/IsotopePatternEstimator.java
```diff
@@ -0,0 +1,342 @@
...
+class IsotopePatternEstimator {
+
+  /**
+   * Default intensity of a peak in a calculated isotope pattern. normalized to 0-1.
+   */
+  public static final double DEFAULT_MIN_INTENSITY_IN_PATTERN = 0.1;
+  /**
+   * Still includes N and O isotopes but not [2]H
+   */
+  public static final double DEFAULT_FAST_ABUNDANCE = 0.002;
```
robinschmid, line 63: This needs to be from 0-100
17O is the lowest with 0.04% and this has little impact on the isotope pattern so maybe set the default to 0.2 to just capture 15N and 18O

2H is 0.01% so it would be included here.

PR #3106, opened 2026-01-23
Highlight grouped summary

SteffenHeu:
Make summary more visible

image


mzmine-community/src/main/resources/themes/jabref_additions_common.css
```diff
@@ -45,6 +46,17 @@ Text.error-label {
   -fx-text-fill: -warning-text-color;
   -fx-icon-color: -warning-text-color;
 }
+
+.contrast-label {
+  -fx-text-fill: -contrast-text-color;
+  -fx-icon-color: -contrast-text-color;
+}
+
+.contrast-text {
```
robinschmid, line 55: I think it works using the same style class .contrast-label  if you do

`Text.contrast-label` 

Just like the .warning-label

PR #3107, opened 2026-01-23
report: add cost center

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/util/reporting/jasper/ReportingCustomerParameters.java
```diff
@@ -50,5 +55,29 @@ public void addToMetadata(@NotNull Map<String, Object> jasperParam) {
...
+
+  @Override
+  public int getVersion() {
+    return 2;
+  }
+
+  @Override
+  public @Nullable String getVersionMessage(int version) {
+    return switch (version) {
+      case 2 -> "Added \"Customer cost center\" to the reporting customer parameters.";
```
robinschmid: maybe no version update and warning needed? 
The old parameters will load fine and just map to an empty string. So I feel like less messages that users have to check and resave parameters are better if the default is the same like in the past. 
But open.

PR #3108, opened 2026-01-23
preset: add messages if presets are loaded from old parameters

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/parameters/impl/SimpleParameterSet.java
```diff
@@ -72,6 +79,13 @@ public class SimpleParameterSet implements ParameterSet {
   private String moduleNameAttribute;
   private boolean skipSensitiveParameters = false;
 
+  /**
+   * Error messages populated during loading from xml. This value is not cloned. Do not include in
+   * hashCode or equals.
```
robinschmid, line 83: Great description. very helpful

PR #3111, opened 2026-01-23
[on annotation-summary] Harmonize how subcolumn values are retrieved, CSV export: remove empty columns and duplicates

robinschmid:
- now that SpectralDBAnnotation is modular, move mapping of types from SpectralLibraryMatchesType to annotation and entry
- this is more similar of how CompoundDBAnnotation.get retrieves both match specific and entry specific info
- entry is not modular yet but its planned in the future so use the same get(DataType) method to prepare the transition

### CSV export
- removed duplicate columns like SimilarityType (from Spectral library) and ScoreType (preferred annotation) and the old IonAdductType and IonTypeType
- removed empty columns from PreferredAnnotation and other annotation types
- CompoundAnnotationUtils.getTypeValue returned no value for AnnotationMethod because value was only asked directly from SpectralLibraryMatch.get, now those are handled in all places and then also exported to csv table

## Example table with sorting and the types
[testcsv7.csv](https://github.com/user-attachments/files/24826975/testcsv7.csv)

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/FeatureList.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/FeatureList.java
@@ -30,6 +30,7 @@ import io.github.mzmine.datamodel.ImagingRawDataFile;
 import io.github.mzmine.datamodel.RawDataFile;
 import io.github.mzmine.datamodel.Scan;
 import io.github.mzmine.datamodel.featuredata.IonMobilogramTimeSeries;
+import io.github.mzmine.datamodel.features.annotationpriority.AnnotationSummarySortConfig;
 import io.github.mzmine.datamodel.features.correlation.R2RMap;
 import io.github.mzmine.datamodel.features.correlation.R2RNetworkingMaps;
 import io.github.mzmine.datamodel.features.correlation.RowGroup;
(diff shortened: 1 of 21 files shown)


mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralDBAnnotation.java
```diff
@@ -485,6 +493,48 @@ public Map<DataType, Object> getMap() {
...
+      return value;
+    }
+
+    // moved from SpectralLibraryMatchesType that had some special mapping columns
+    // those columns are neither saved in the match nor in the entry but are computed
+    // some values from entry might have failed return due to mismatching types
+    // therefore some are still mapped here to the corresponding correct type
+    try {
+      return (T) switch (type) {
+        case SpectralLibraryMatchesType _ -> this; // to display the match in the first column
```
SteffenHeu: not a fan of having the gui specific mapping in here. i think i would leave everything gui specific with the SpectralLibraryMatchesType. so annotation summary and spectrallibrarymatchestype

robinschmid: I now moved these to cases back to the SpectramMatchesType.

I looked at the CompoundMatchesType and tried to replicate what is done there. 
Now I saw that the CompoundMatchesType does not have AnnotationSummaryType. 
Maybe we should add this in the next PR

SteffenHeu: true, should add that there

robinschmid: Created a PR #3129

PR #3114, opened 2026-01-26
Spectrum plot - auto select feature list

SteffenHeu:
Automatically select the feature list if the spectrum plot is opened from the feature table


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/spectra/simplespectra/renderers/SpectraItemLabelGenerator.java
```diff
@@ -135,6 +142,16 @@ public String generateLabel(XYDataset dataset, int series, int item) {
     String label = null;
     if (dataset instanceof ScanDataSet) {
       label = ((ScanDataSet) dataset).getAnnotation(item);
+    } else if (dataset instanceof PeakListDataSet pld) {
+      StringBuilder sb = new StringBuilder();
+      sb.append(mzFormat.format(dataset.getXValue(series, item)));
+      sb.append("\n");
+      ModularFeature feature = (ModularFeature) pld.getFeature(series, item);
+      if (feature.get(IonTypeType.class) instanceof IonType ion) {
```
robinschmid: we are never setting the IonTypeType to a feature right? Only to a feature list row?

PR #3118, opened 2026-01-26
on annotation summary: sort preferred annotations

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummary.java
```diff
@@ -136,6 +117,7 @@ public double ccsScore() {
     final Float ccs = annotation.getCCS();
 
     if (averageCCS != null && ccs != null) {
+      final double maxCcsDev = row.getFeatureList().getAnnotationSortConfig().ccsTolerance();
       return getScore(averageCCS / ccs, maxCcsDev);
```
robinschmid: Missed this in the other review. 
Why averageCSS/css 
This needs to be the deviation right?
(averageCSS-css)/css

Same for RI

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummary.java
```diff
@@ -158,6 +140,7 @@ public double riScore() {
     Float ri = CompoundAnnotationUtils.getTypeValue(annotation, RIType.class);
 
     if (averageRI != null && ri != null) {
+      final double maxRiDev = row.getFeatureList().getAnnotationSortConfig().riTolerance();
       return getScore(averageRI / ri, maxRiDev);
```
robinschmid: same here using the difference. Also the maxRiDev is an absolute number right? 
Then it should be 
`getScore(averageRI, ri, maxRiDev)`

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummary.java
SteffenHeu: no this is a percent tolerance, hence deviation instead of difference

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummary.java
SteffenHeu: no, see above

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummaryOrder.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public static final Comparator<@Nullable AnnotationSummary> LIPIDS_FIRST_LOW_TO_HIGH_CONFIDENCE = Comparator.comparing(
+      Function.identity(), nullsFirst(Comparator.<AnnotationSummary>comparingInt(a -> {
+        DataType<?> dt = DataTypes.get(a.annotation().getDataType());
+        return switch (dt) {
+          case LipidMatchListType _ ->
+              ((MatchedLipid) a.annotation()).getMsMsScore() != null ? 1 : 3;
+          case SpectralLibraryMatchesType _ -> 2;
+          case CompoundDatabaseMatchesType _ -> 3;
+          case null, default -> 4;
```
robinschmid: Lipid sorter LOW -> HIGH needs to have reverse numbers right?  
Lipids+MS2 4
SpecLib 3
Comp+lipid 2
default 1
nulls first (already)

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummaryOrder.java
```diff
@@ -0,0 +1,72 @@
...
+  @Override
+  public @NotNull String getUniqueID() {
+    return switch (this) {
+      case SCHYMANSKI -> "schymanski";
+      case MSI -> "msi";
+      case LIPIDS_FIRST -> "lipids_first";
+    };
+  }
+
+  public Comparator<@Nullable AnnotationSummary> getComparator() {
```
robinschmid: rename to getComparatorLowToHigh
and add method for HighToLow. 
Or use enum argument HIGH_TO_LOW etc and just one method.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/annotationpriority/AnnotationSummarySortConfig.java
```diff
@@ -0,0 +1,27 @@
...
+
+import io.github.mzmine.modules.dataprocessing.filter_sortannotations.PreferredAnnotationSortingParameters;
+import io.github.mzmine.parameters.parametertypes.tolerances.MZTolerance;
+import io.github.mzmine.parameters.parametertypes.tolerances.RTTolerance;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ *
+ * @param ccsTolerance ccs tolerance in %
+ * @param riTolerance  ri tolerance in %
```
robinschmid: The RITolerance is usually as absolute difference like for RT. The values are already normalized through the standard ladder and relative tolerance could lead to very wide ranges for large RI and too narrorw in small RI. 

The RITolerance that we already have is also absolute. The parameter and RITolerance itself is not useful here but just changing it to an absolute number would work. 
Also the score calculation would need to be updated then.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/PreferredAnnotationSortingParameters.java
```diff
@@ -0,0 +1,67 @@
...
+  public static final RTToleranceParameter rtTolerance = new RTToleranceParameter(
+      DEFAULT_RT_TOLERANCE);
+
+  public static final double DEFAULT_CCS_TOLERANCE = 0.1;
+  public static final PercentParameter ccsTolerance = new PercentParameter("CCS tolerance",
+      "CCS tolerance used for ranking", DEFAULT_CCS_TOLERANCE, 0d, 1d);
+
+  public static final double DEFAULT_RI_TOLERANCE = 0.1;
+  public static final PercentParameter riTolerance = new PercentParameter("RI tolerance",
+      "RI tolerance used for ranking", DEFAULT_RI_TOLERANCE, 0d, 1d);
```
robinschmid: if changed to absolute RI tolerance then remove the maxValue as null.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAUpdateTask.java
```diff
@@ -116,7 +117,7 @@ protected void process() {
     // data was already prepared
     final List<FeatureListRow> rows = featureDataTable.getFeatureListRows();
 
-    Comparator<@Nullable AnnotationSummary> annotationSummarySorter = AnnotationSummary.LOW_TO_HIGH_CONFIDENCE;
+    Comparator<@Nullable AnnotationSummary> annotationSummarySorter = AnnotationSummaryOrder.SCHYMANSKI_LOW_TO_HIGH_CONFIDENCE;
```
robinschmid: maybe nice to add a default? That would make it easier to change default in the future if we MSI finally releases something. like
`AnnotationSummaryOrder.getDefault().getComparator(LOW_TO_HIGH);`

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/PreferredAnnotationSortingModule.java
```diff
@@ -0,0 +1,29 @@
+package io.github.mzmine.modules.dataprocessing.filter_sortannotations;
```
robinschmid: header missing

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/PreferredAnnotationSortingModule.java
```diff
@@ -0,0 +1,29 @@
...
+import io.github.mzmine.taskcontrol.Task;
+import io.github.mzmine.util.MemoryMapStorage;
+import java.time.Instant;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class PreferredAnnotationSortingModule extends TaskPerFeatureListModule {
+
+  public PreferredAnnotationSortingModule() {
+    super("Sort preferred annotations", PreferredAnnotationSortingParameters.class,
```
robinschmid: I thought this would sort all annotations in all annotation types. 
So that the best speclibmatch and compounddb annotation etc is also sorted to the top. 

User could still apply this early and the config would be saved in the table and hopefully applied by the annotation modules later as well. 
But then users could also re-sort all.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/PreferredAnnotationSortingModule.java
```diff
@@ -0,0 +1,29 @@
...
+        MZmineModuleCategory.FEATURELISTFILTERING, false,
+        "Define how preferred annotations are sorted in a feature list.");
```
robinschmid: Maybe we can add to the description that this is saved to the table and that future annotations will be sorted this way as well.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/PreferredAnnotationSortingModule.java
robinschmid: also in other new classes

mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/FeatureListLoadTask.java
```diff
@@ -384,6 +387,13 @@ private ModularFeatureList readMetadataCreateFeatureList(File file, MemoryMapSto
               .getTextContent());
       flist.getAppliedMethods().addAll(appliedMethods);
       selectedScansMap.forEach(flist::setSelectedScans);
+
+      final FeatureListAppliedMethod preferredAnnoationSorting = ParameterUtils.getLatestModuleCall(
+          appliedMethods, PreferredAnnotationSortingModule.class);
```
robinschmid: Maybe we could save the config object in the FeatureListSaveTask.saveFeatureData then we definitely have the correct version, even if we change the default or if another module changes the sorting.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/PreferredAnnotationSortingParameters.java
SteffenHeu: max value as null?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAUpdateTask.java
SteffenHeu: mzmine sortign is now the default

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/PreferredAnnotationSortingModule.java
SteffenHeu: currently not done, that is the decision of the respective module. Only the preferred annotation is chosen in a different way.

mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/FeatureListLoadTask.java
SteffenHeu: That xml is strictly for data types. We can add another metadata section, but i would rather prefer adding a new enum value to address this, rahter than duplicating code to save this object.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/PreferredAnnotationSortingParameters.java
robinschmid: The maximum bound for the value should be unbounded so that any double works

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/pca_new/PCAUpdateTask.java
robinschmid: great

mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/FeatureListLoadTask.java
robinschmid: Metadata section also sounds good. 
image

What do you mean with adding a new enum? 
This could also use config.toParameterset and use the parameterset to load and save the config. 

Was just thinking it would be nice to save the actual instance and not extract it from the last call of the module.

PR #3123, opened 2026-01-27
Fix size of report structures

robinschmid:
Zoom factor was too small
also added new direct svg rendering

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/util/reporting/jasper/reporttypes/AlignedFeatureReportParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/util/reporting/jasper/reporttypes/AlignedFeatureReportParameters.java
@@ -24,16 +24,15 @@
 
 package io.github.mzmine.util.reporting.jasper.reporttypes;
 
-import io.github.mzmine.gui.chartbasics.graphicsexport.ExportChartThemeModule;
-import io.github.mzmine.gui.chartbasics.graphicsexport.ExportChartThemeParameters;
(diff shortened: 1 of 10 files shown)


mzmine-community/src/main/java/io/github/mzmine/util/reporting/jasper/reporttypes/AlignedFeatureReportParameters.java
```diff
@@ -51,8 +53,15 @@ public class AlignedFeatureReportParameters extends SimpleParameterSet {
       "Chart theme", "Define the chart theme", new ExportChartThemeParameters(), false,
       ExportChartThemeModule.class);
 
+
+  // needs more zoom maybe because of DPI?
+  public static final ParameterSetParameter<StructureRenderParameters> structureRendering = new ParameterSetParameter<>(
```
SteffenHeu: do we need this parameter? is the default not fine for all structures?

robinschmid: I think it is useful and we already have this parameterset providing easy option for users to customize how structures are rendered. 
For example the zoom + bond length already allow to also change the relative size of Atom labels compared to the bonds.
In the future we can add more options like the whole structure black etc.

robinschmid: Do you see any issue with having this option?

PR #3124, opened 2026-01-27
Add export to SVG menu to Structure2DComponent

robinschmid:
- based on #3123
- adds an svg export menu to structure2dcomponent
- by default the svg keeps the same bond lengths and scales the width and height
image

Looks bad without background here so in browser its like this:
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/Structure2DComponent.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/Structure2DComponent.java
@@ -25,7 +25,6 @@
 
 package io.github.mzmine.modules.visualization.molstructure;
 
-import io.github.mzmine.javafx.concurrent.threading.FxThread;
 import io.github.mzmine.main.ConfigService;
 import java.awt.Font;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/Structure2DComponent.java
```diff
@@ -97,7 +122,8 @@ private void onStructureChange(final IAtomContainer mol) {
    * ensures fx thread repaint
    */
   private void repaint() {
-    FxThread.runLater(() -> renderer.drawStructure(this, molecule.getValue(), renderConfig.getValue()));
+    FxThread.runLater(
+        () -> renderer.drawStructure(this, molecule.getValue(), renderConfig.getValue()));
```
SteffenHeu: this is the fx component, is it necessary to force the fx thread here? usually these methods should always be called from the fx thread anyway, or?

robinschmid: The repaint method is not from Canvas it is defined in this class. 
There is a property with the structure that triggers the repaint or resizing of the component. 

In general it is best practice to only update properties on the FX thread but nothing really stops you from doing it from a different thread. 

FxThread.runLater would just be safer. But I can also remove it, I think the property is currently only changed on the fx thread.

robinschmid: removed it

PR #3126, opened 2026-01-28
Rename module to Annotation ranking

robinschmid:
Maybe we should directly call it annotation ranking (more general than preferred annotation ranking). 
And later also use it in ranking the internal lists of other annotation types.


SteffenHeu:
I think each module is responsible for ranking it's annotations. They are free to use this config too, but they are also free to use a more specific logic. So that's also why i called it preferred annotation ranking

PR #3128, opened 2026-01-28
Combined annotation score

robinschmid:
- adds a combined annotation score similar to the one used in the formula prediction dialog
- Uses weights for all scores
- Excludes scores that are unavailable like CCS und RI if the types are not in the feature table row types
- adds scores to tooltip with styling
- AnnotationSummaryType column value factory caches AnnotationSummaries for row+annotation pairs to speed up sorting and avoid recalculation of expensive scores

image

The weights have all defaults and are set in the AnnotationSummarySortConfig:
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
@@ -992,7 +992,8 @@ public class ModularFeatureList implements FeatureList {
   }
 
   @Override
-  public void setAnnotationSortConfig(@NotNull AnnotationSummarySortConfig annotationSortConfig) {
+  public synchronized void setAnnotationSortConfig(
+      @NotNull AnnotationSummarySortConfig annotationSortConfig) {
     this.annotationSortConfig = annotationSortConfig;
     annotationSortConfigVersion++;
   }


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
```diff
@@ -105,7 +143,8 @@ public class AnnotationSummaryType extends DataType<AnnotationSummary> implement
     column.setMinWidth(45);
     column.setPrefWidth(45);
     column.setSortable(true);
-    column.setComparator(AnnotationSummaryOrder.MZMINE.getComparatorLowFirst());
+    // flip sorting so that first click gives correct sorting
+    column.setComparator(AnnotationSummaryOrder.MZMINE.getComparatorHighFirst());
```
SteffenHeu: as discussed, does not fit other scores

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/CombinedScoreWeightsComponent.java
```diff
@@ -0,0 +1,133 @@
...
+import javafx.scene.layout.FlowPane;
+import javafx.scene.layout.Region;
+import javafx.scene.layout.VBox;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class CombinedScoreWeightsComponent extends FlowPane {
+
+  private final NumberFormat numberFormat = new DecimalFormat("0.0");
+  private final @NotNull Property<Double> mz;
```
SteffenHeu: `Property<@ Nullable Double>`
the text formatter's value property 'value' can be null, this leads to NPEs. the get value method must check if all properties do have a value other than null

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/CombinedScoreWeightsComponent.java
```diff
@@ -0,0 +1,133 @@
...
+    ms2.setValue(value.ms2());
+    isotopes.setValue(value.isotopes());
+    mz.setValue(value.mz());
+    rt.setValue(value.rt());
+    ri.setValue(value.ri());
+    ccs.setValue(value.ccs());
+  }
+
+  public CombinedScoreWeights getValue() {
+    return new CombinedScoreWeights(mz.getValue(), rt.getValue(), ri.getValue(), ccs.getValue(),
```
SteffenHeu, line 137: see above, also needs to be nullable then so that the checkValue method works as intended in the parameter

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
robinschmid: reverted. 
Maybe we can find a solution at some point that the type can say what the initial sort order is.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_sortannotations/CombinedScoreWeightsComponent.java
robinschmid: uff totally missed this... not sure how
Added try catch instead of million null checks

PR #3130, opened 2026-01-30
Delayed update of rows in feature table

robinschmid:
-batch remove rows
- delay resolves concurrent modification exception


SteffenHeu (commented):
I tested with only the batch remove change (removeRows over rows.forEach(flist::removeRow) and that already solved the problem. 

Copying the list for treeitem generation is also a good idea.

i think we don't really need the delayed listener, or? I remember something not working in the past when the feature table itself was not the listener, but that may have been a skill issue at the time, though.

PR #3133, opened 2026-01-30
Select/exclude data files by metadata

robinschmid:
- Include multiple groups
- Exclude multiple groups (exclusions are stronger than inclusions) 
- So you can include multiple groups from a column and then exclude the QC and Blanks for example. 
- Or use it in the feature list blank subtraction module

image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilesByMetadataParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilesByMetadataParameters.java
@@ -43,13 +43,13 @@ import org.jetbrains.annotations.Nullable;
 public class RawDataFilesByMetadataParameters extends SimpleParameterSet {
 
   public static final MetadataListGroupsSelectionParameter included = new MetadataListGroupsSelectionParameter(
-      "Include",
-      "Include metadata group names in column. Enter multiple values, one in each text field.",
-      NONE, false);
+      "Include", """
+      Include metadata group names in column. Enter multiple values, one in each text field.
+      Exclude is stronger and will remove data files if excluded.""", NONE, false);
   public static final MetadataListGroupsSelectionParameter excluded = new MetadataListGroupsSelectionParameter(
-      "Exclude",
-      "Exclude metadata group names in column. Enter multiple values, one in each text field.",
-      NONE, false);
+      "Exclude", """
+      Exclude metadata group names in column. Enter multiple values, one in each text field.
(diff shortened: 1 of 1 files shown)


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilesSelection.java
```diff
@@ -113,6 +133,17 @@ public RawDataFile[] getMatchingRawDataFiles() {
...
+      case BY_METADATA -> {
+        List<RawDataFile> matchingDataFiles = ProjectService.getProject().getCurrentRawDataFiles();
+        if (includeMetadataSelection.isValid()) {
+          matchingDataFiles = includeMetadataSelection.getMatchingFiles(matchingDataFiles);
+        }
+        if (excludeMetadataSelection.isValid()) {
+          matchingDataFiles = excludeMetadataSelection.removeMatchingFilesCopy(matchingDataFiles);
+        }
+
+        matchingFiles = matchingDataFiles.toArray(new RawDataFile[0]);
```
SteffenHeu, line 145: you need to add handling for the case that the selection is not valid. otherwise for a group that does not exist, all files are selected since matchingDataFiles is not updated

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/metadata/MetadataListGroupsSelection.java
```diff
@@ -0,0 +1,165 @@
...
+    if (column == null) {
+      return false;
+    }
+
+    final Map<RawDataFile, Object> columnValues = metadata.getData().get(column);
+    // usually few values in groups so ok to stream for each value
+    return columnValues.values().stream().anyMatch(this::matchesValue);
+  }
+
+  private boolean matchesValue(Object value) {
```
SteffenHeu: add nullable annotation please

robinschmid: added a couple more as well

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/selectors/RawDataFilesSelection.java
robinschmid: Great catch. The column values were checked in the isValid method. 
But there were many places that did similar things now. 
I refactored this into an interface with default methods and then the 1Group, 2Groups, ListGroupsSelection just implement the important methods now in a more harmonized way

PR #3136, opened 2026-02-02
Rows filter keep with MS2 does not apply to GC-EI-MS

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterParameters.java
```diff
@@ -166,9 +166,9 @@ public class RowsFilterParameters extends SimpleParameterSet {
       "If checked, the rows that don't contain MS2 scan will be removed.", false);
 
   public static final BooleanParameter KEEP_ALL_MS2 = new BooleanParameter(
-      "Never remove rows with MS2",
-      "If checked, all rows with MS2 are retained without applying any further filters on them.",
-      true);
+      "Never remove rows with MS2", """
+      If checked, all rows with MS2 are retained without applying any further filters on them.
+      GC-EI-MS pseudo MS2 spectra are not considered MS2, because every row has one.""", false);
```
SteffenHeu, line 171: Ok so we change the default to false now?
i thought we add a message to the stats filters that they dont make sense if this filter is enabled

robinschmid: I thought you like it better to be off? 
I am open to both. 
In some workflows this true does not make sense. In annotation workflows it totally makes sense.

SteffenHeu: I'm ok with having it false by default, but i did not expect the change

SteffenHeu:
still needs wizard update right?

PR #3137, opened 2026-02-02
annotation-summary: change text color for annotations with bad scores

SteffenHeu:
change text color for scores where the score was calculated but was zero to differentiate from scores where the type is applicable but no library value was available for matching

image
image
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
@@ -57,6 +57,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.OptionalDouble;
 import javafx.beans.property.Property;
 import javafx.beans.property.ReadOnlyObjectWrapper;
 import javafx.beans.property.SimpleObjectProperty;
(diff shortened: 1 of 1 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
```diff
@@ -301,9 +302,17 @@ private void draw() {
         // If bar is tall, text is on the bar.
         if (score > 0.2) {
           gc.setFill(Color.WHITE); // Assuming bars are colored/dark
+        } else if (annotationSummary.score(scoreType).isPresent()
+            && Precision.equalFloatSignificance(score, 0d)) {
```
robinschmid: above we do: 
`final double score = annotationSummary.score(scoreType).orElse(0d);`
Just keep the optionalScore reference there so we do not have to recalculate the score. 

Maybe it makes sense to flip the colors? 
- Like regular UI color for scores that were actually calculated (and maybe 0)
- Highlight color if the annotation did not provide the necessary info. 

But either way is good with me. Just feels like the exception would be that there was no RT in the library entry.

SteffenHeu: in my understanding not having a RT in the library is better than having an RT in the library that does not match your sample. Since the latter is more critical i would rather highlight that one.

Issue #2835, opened 2025-08-25
Add charge filter to local db search


SteffenHeu:
Option to filter by charge in Local compound db search

PR #3141, opened 2026-02-03
local-comp: add strict charge filtering parameter

SteffenHeu:
adds https://github.com/mzmine/mzmine/issues/2835

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchParameters.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchParameters.java
@@ -186,7 +186,7 @@ public class LocalCSVDatabaseSearchParameters extends SimpleParameterSet {
       new ComboWithStringInputValue<>(HandleExtraColumnsOptions.IGNORE, null));
 
   public static final ComboParameter<ChargeFilterType> chargeFilter = new ComboParameter<>(
-      "Strict charge filtering", """
+      "Charge filter", """
       If enabled, the charge of the compound determined through the "%s" parameter or by importing
       the "%s" type and matched to the determined charge of the feature (isotope pattern must be detected).
       If no isotope pattern is detected, the charge filtering is not executed and compounds are matched


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchTask.java
```diff
@@ -363,6 +366,16 @@ private void checkMatchAndAnnotate(@NotNull CompoundDBAnnotation annotation,
       @Nullable RTTolerance rtTolerance, @Nullable MobilityTolerance mobTolerance,
       @Nullable Double percCcsTolerance, @Nullable RITolerance riTolerance) {
 
+    if (chargeFilter && annotation.getAdductType() != null && row.getRowCharge() != null) {
+      final IonType adduct = annotation.getAdductType();
+      final int absCharge = adduct.getAbsCharge();
+      final int rowCharge = Math.abs(row.getRowCharge());
+      // only apply charge filtering if adduct and charge are determined
+      if (adduct.getPolarity() != row.getRepresentativePolarity() || absCharge != rowCharge) {
```
robinschmid: `!PolarityType.isDefined(row.getRepresentativePolarity())` the static method also handles null (no polarity provided in scans) should always match I would say. It is exceptional but if the info is not provided in the scan then we cannot filter by polarity.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchTask.java
```diff
@@ -363,6 +366,16 @@ private void checkMatchAndAnnotate(@NotNull CompoundDBAnnotation annotation,
       @Nullable RTTolerance rtTolerance, @Nullable MobilityTolerance mobTolerance,
       @Nullable Double percCcsTolerance, @Nullable RITolerance riTolerance) {
 
+    if (chargeFilter && annotation.getAdductType() != null && row.getRowCharge() != null) {
```
robinschmid: I would think that strict charge filtering would be strict. 
So that if the row has no charge state it would not match anything. 
Rows without charge state are often low quality rows eitherway. 

Maybe instead of boolean its easier to turn it into a combo parameter so it is clear what happens? 
Charge filter: Options:
Strict: requires both annotation and row to have charge
Require row charge (annotation without charge ok but row charge required)
Off: match any

SteffenHeu: added an enum.

i added an option to allow annotations without a charge instead of requiring a row charge. i think that makes more sense as that allows to add "wildcard" annotations.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_localcsvsearch/LocalCSVDatabaseSearchParameters.java
```diff
@@ -184,6 +185,13 @@ public class LocalCSVDatabaseSearchParameters extends SimpleParameterSet {
       HandleExtraColumnsOptions.IMPORT_SPECIFIC,
       new ComboWithStringInputValue<>(HandleExtraColumnsOptions.IGNORE, null));
 
+  public static final ComboParameter<ChargeFilterType> chargeFilter = new ComboParameter<>(
+      "Strict charge filtering", """
```
robinschmid: Maybe just Charge filtering or Charge filter now that there are multiple options and one is named Strict?

PR #3142, opened 2026-02-03
Check adduct formula most abundant isotope in compound DB search

robinschmid:
- checks the adduct formula to include ions like Cl and Br that influence the isotope pattern
- excludes Na and other atoms to limit number of pattern calculations

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/util/FormulaUtils.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/util/FormulaUtils.java
@@ -81,7 +81,7 @@ public class FormulaUtils {
    */
   private static final IntSet ATOMS_LOW_ISOTOPE_PROBABILITY = new IntOpenHashSet(
       List.of(Element.H, Element.Na, Element.K, Element.Be, Element.Ca, Element.C, Element.Si,
-          Element.N, Element.P, Element.As, Element.O, Element.S, Element.F, Element.I));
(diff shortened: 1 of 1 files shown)


mzmine-community/src/main/java/io/github/mzmine/util/FormulaUtils.java
```diff
@@ -69,6 +73,41 @@ public class FormulaUtils {
...
 
+
+  /**
+   * Set of ions with low probability that they contribute much to the highest signal in the isotope
+   * pattern. C and Si are very common (Si more in GC-MS) but they will be excluded first and
+   * evaluated later if their number contributes to the max isotope.
+   */
+  private static final IntSet ATOMS_LOW_ISOTOPE_PROBABILITY = new IntOpenHashSet(
+      List.of(Element.H, Element.Na, Element.K, Element.Be, Element.Ca, Element.C, Element.Si,
+          Element.N, Element.P, Element.As, Element.O, Element.S, Element.F, Element.I));
```
SteffenHeu: Sulfur was specifically not put into my pattern check before as 34S abundance is 4.2% which can influence the isotope composition.

robinschmid: Removed from the list. 
Si is still on the list but just because it is handled in a different way together with C

PR #3152, opened 2026-02-04
Stable value isotope pattern caching of matched lipids

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundannotations/FeatureAnnotation.java
```diff
@@ -216,7 +216,15 @@ default IsotopePattern calculateIsotopePattern() {
...
     }
-    if (formula == null) {
+    return calculateIsotopePattern(formula, ionType);
+  }
+
+  /**
+   * Predict pattern with default binning width
+   */
+  public static @Nullable IsotopePattern calculateIsotopePattern(
+      @Nullable IMolecularFormula formula, @Nullable IonType ionType) {
```
SteffenHeu: neutralFormula to clarify that the ion type is added to the formula.

or does this method even have to be public static? Either hide the method which does not work in interfaces, or move it to the IsotopePatternCalculator, since the input there is not specific for an FeatureAnnotation. In the calculator we automatically assume a lot of things then whcih may not be always applicable. 

I would just leave it as it was, since the stable value can also use an instance method, or?

robinschmid: Idea was to have a static method somewhere either IsotopePatternCalculator or FeatureAnnotation to harmonize the default behavior. 
Also to point SpectralLbiraryEntry.calculateIsotopePattern to the same calculation so that the minPatternIntensity and other defaults are handled

robinschmid: I am fine with both places but as it is mostly related to the FeatureAnnotation maybe keep it there?

SteffenHeu: then i'd say the calculator is the best place, we still have the instance method in the FeatureAnnotation. I would not put the logic in the interface if you want to make it a static utility method

PR #3155, opened 2026-02-04
Isotope pattern on demand, precompute all feature annotations when added to row

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/util/annotations/CompoundAnnotationUtils.java
```diff
@@ -96,6 +96,31 @@ public class CompoundAnnotationUtils {
...
+
+  /**
+   * Precalculates the isotope pattern and maybe other properties of a feature annotation to speed
+   * up the feature table. This is usually done in {@link FeatureListRow} methods where annotations
+   * are added.
+   *
+   */
+  public static void precalculateAnnotationValues(@NotNull FeatureAnnotation annotation) {
+    // cache isotope pattern for faster feature table start up
+    annotation.getIsotopePattern();
```
SteffenHeu, line 122: we could also check if it is a compound annotation and apply the ConnectTypeCalculations here

robinschmid: so far they are only used in the manual annotate controller right? 

Or am I missing something? 
Can definitely trigger it here.

PR #3171, opened 2026-02-05
Set all DataType.getPrefColumnWidth to speed up feature table open

robinschmid:
contains PR #3169 

- The calculation of the width of too many columns made the table slow down. 
- Now all columns have a pref width.
- -1 or the default prefWidth of 80.0 will trigger the measurement of multiple values in the column. This is slow. 
- Now uses measurements of the title or of a representative value or fixed width

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/StructureTableCell.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/molstructure/StructureTableCell.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2024 The MZmine Development Team
+ * Copyright (c) 2004-2026 The mzmine Development Team
  *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
@@ -44,7 +44,7 @@ public class StructureTableCell<S> extends SkipMeasurementTreeCell<S, Object> {
 
     setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
     setMinWidth(DataTypes.get(MolecularStructureType.class).getPrefColumnWidth());
-    setHeight(GraphicalColumType.DEFAULT_GRAPHICAL_CELL_HEIGHT);
+    setPrefHeight(GraphicalColumType.DEFAULT_GRAPHICAL_CELL_HEIGHT);
   }
 
   @Override


javafx-framework/src/main/java/io/github/mzmine/javafx/components/skins/tables/OptimizedNestedTableColumnHeader.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  @Override
+  protected TableColumnHeader createTableColumnHeader(TableColumnBase col) {
+    if (col == null || col.getColumns().isEmpty() || col == this.getTableColumn()) {
+      return new OptimizedTableColumnHeader(col);
+    } else {
+      return new OptimizedNestedTableColumnHeader(col);
+    }
+  }
```
SteffenHeu: mhh does this mean we cannot add new columns when the table is open? e.g. Spectral match was on, but no sub type was on and then the user enables them?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/DataType.java
```diff
@@ -246,10 +255,16 @@ public String getFormattedStringCheckType(Object value) {
...
+   * if too many columns use this. So better to define pref width in each type
+   *
    * @return -1 if off, otherwise defines the max column width
    */
   public double getPrefColumnWidth() {
-    return Region.USE_COMPUTED_SIZE; // generally off because columns may be resized usually
+    // define to speed up table open - otherwise calculation of width takes long
+    // if too many types are undefined
+    return Math.min(80,
+        TextLabelMeasurementUtil.measureWidth(getHeaderString() + TableColumns.EXTRA_WIDTH_MARGIN));
```
SteffenHeu: is min correct?
as far as i understood 80 will cause computation and if the column should be wider by the `TextLabelMeasurementUtil.measureWidth(getHeaderString() + TableColumns.EXTRA_WIDTH_MARGIN)` it is computed again with the layout mechanism? 
not sure if the `return Math.min(80,
        TextLabelMeasurementUtil.measureWidth(getHeaderString() + TableColumns.EXTRA_WIDTH_MARGIN));` is a benefit then.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/color/ColorByMetadataParameters.java
```diff
@@ -58,7 +58,7 @@ public class ColorByMetadataParameters extends SimpleParameterSet {
       0d, 1d);
 
   public static final BooleanParameter separateBlankQcs = new BooleanParameter(
-      "Separate blanks & QCs", "Separate coloring for blanks and QCs", true);
+      "Separate blanks & QCs", "Separate coloring for blanks and QCs", false);
```
SteffenHeu, line 61: Why this change?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/skins/tables/OptimizedNestedTableColumnHeader.java
robinschmid: Sorry forgot to either remove those classes or to describe them better. 
This is the last resort option that I tried first. 
This is the only way how to implement your own header it seams. 

Reading through the jfx code I found that setting a prefWidth different from default 80.0 will do the same trick.

robinschmid: I can remove those classes now

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/color/ColorByMetadataParameters.java
robinschmid: This option creates issues with the coloring that we see in the Stats dashboard so maybe better to have the default align with the stats dashboard

javafx-framework/src/main/java/io/github/mzmine/javafx/components/skins/tables/OptimizedNestedTableColumnHeader.java
robinschmid: Done

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/DataType.java
robinschmid: Oh yeah accidentally increased the default here to 80. This was a bad coincidence. 
Will use the other default width that I set in TableColumns. It is 79.

I used Math.min here because some headers are very long. So that the columns will by default not be too large.

PR #3181, opened 2026-02-06
Harmonize copying of config and applied methods

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/util/FeatureListUtils.java
```diff
@@ -824,6 +821,39 @@ public static void copyRows(final FeatureList featureList,
...
+    transferAllButRows(List.of(source), target, transferTypes);
+  }
+  /**
+   * Transfer selected scans, applied methods, annotation sort config, row and feature types
+   *
+   * @param sources        copy from
+   * @param target        copy to
+   * @param transferTypes true then transfer all row and feature types
+   */
+  public static void transferAllButRows(@NotNull List<FeatureList> sources, @NotNull ModularFeatureList target,
```
SteffenHeu: Can we call this copyMetadata?

robinschmid: Makes sense maybe transferMetadata? to keep it similar to the other transfer methods called within?

PR #3192, opened 2026-02-11
allow project load in batch

SteffenHeu:
also permits further processing in batch as you would expect by grouping feature lists
creates a new project only if open project is the first step
otherwise you can add an existing project as an intermediate step

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/ProjectLoadModule.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/ProjectLoadModule.java
@@ -25,7 +25,6 @@
 
 package io.github.mzmine.modules.io.projectload;
 
-import static io.github.mzmine.modules.io.projectload.ProjectLoaderParameters.create;
 import static io.github.mzmine.util.ExitCode.OK;
 
 import io.github.mzmine.datamodel.MZmineProject;
@@ -100,13 +99,15 @@ public class ProjectLoadModule implements MZmineProcessingModule {
       return;
     }
 
-    ProjectLoaderParameters param = ProjectLoaderParameters.create(selectedFile, false, false);
+    // keep libraries in quick open as most people have a single library they want to use for all
+    ProjectLoaderParameters param = ProjectLoaderParameters.create(selectedFile, false, true);
 
     MZmineCore.runMZmineModule(ProjectLoadModule.class, param);
   }
 
   public static void showImportDialog(@NotNull File file) {
-    ProjectLoaderParameters projectLoaderParameters = create(file, false, false);
+    ProjectLoaderParameters projectLoaderParameters = ProjectLoaderParameters.create(file, false,
+        true);
     ExitCode exitCode = projectLoaderParameters.showSetupDialog(true);
     if (exitCode == OK) {
       MZmineCore.runMZmineModule(ProjectLoadModule.class, projectLoaderParameters);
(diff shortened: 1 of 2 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/FeatureListLoadTask.java
```diff
@@ -211,6 +215,16 @@ public void run() {
...
     project.setProjectLoadImsImportCaching(false);
+
+    //  group flists by date created, only use the latest set of feature lists in next batch step
+    final Set<FeatureList> mostRecentStepFeatureLists = Set.copyOf(
+        project.getCurrentFeatureLists().stream().collect(
+                Collectors.groupingBy(flist -> flist.getAppliedMethods().getLast().getModuleCallDate()))
+            .entrySet().stream().max(Entry.comparingByKey()).map(Entry::getValue)
+            .orElse(List.of()));
+    project.getCurrentFeatureLists().forEach(
+        flist -> flist.setExcludedFromBatchLast(!mostRecentStepFeatureLists.contains(flist)));
```
robinschmid, line 228: great idea and good catch

mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/RawDataFileOpenHandler_3_0.java
```diff
@@ -225,29 +226,14 @@ public boolean loadRawDataFiles() throws InterruptedException {
...
-          Thread.sleep(100);
-          if (isCanceled()) {
-            return false;
-          }
-        }
-
-        if (!success.get()) {
+        if (taskStatus != TaskStatus.FINISHED) {
           return false;
         }
```
robinschmid, line 235: I think it is better to just call the task.run on this thread and not rely on the taskcontroller to schedule another task. 
```java

        // run task blocking on this thread
        currentTask = new BatchTask(project, param, Instant.now(), null, this);
        currentTask.run();

        if (currentTask.getStatus() == TaskStatus.ERROR) {
          error(currentTask.getErrorMessage());
          return false;
        }

        if (!currentTask.isFinished()) {
          return false;
        }
```

mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/ProjectLoadModule.java
```diff
@@ -25,27 +25,24 @@
...
  * This class implements BatchStep interface, so project loading can be activated in a batch.
- * 
+ *
  */
 public class ProjectLoadModule implements MZmineProcessingModule {
 
   private static final String MODULE_NAME = "Open project";
-  private static final String MODULE_DESCRIPTION =
-      "This module opens an existing MZmine project. The current workspace will be discarded.";
+  private static final String MODULE_DESCRIPTION = "This module opens an existing mzmine project and discards the current workspace, unless run in batch mode.";
```
robinschmid: add description what happens in batch mode. 
So I guess it is: 
First step in batch mode replaces the current workspace and spectral libraries. If put as an intermediate step in batch mode the new project is merged into the current project.

mzmine-community/src/main/java/io/github/mzmine/modules/io/projectload/version_3_0/RawDataFileOpenHandler_3_0.java
robinschmid: This way by directly running the batch from the project load we can also add a check in the BatchModeModule.runBatchFile to stop a parallel batch run during project import:
```java

    if (MZmineCore.getTaskController().isTaskInstanceRunningOrQueued(ProjectOpeningTask.class)) {
      MZmineCore.getDesktop().displayErrorMessage(
          "Currently loading a project, cannot run a batch until project load is finished.");
      return null;
    }
```

PR #3197, opened 2026-02-18
menu: add corr dashboard to visualisation menu, add rt corr to menu

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/util/javafx/FxMenuUtil.java
```diff
@@ -171,4 +171,20 @@ public static void addSeparator(Menu menu) {
     menu.getItems().add(new SeparatorMenuItem());
   }
 
+  public static void addToSubMenuIfAvailable(@NotNull final Menu mainMenu,
+      @NotNull MenuItem itemToAdd, @NotNull String... subMenuNames) {
```
robinschmid, line 178: Maybe this should throw an exception if the submenu (full path) was not found? Like when we rename a menu then it would be best if we fail early and can fix the name references. 

Currently, if the first subMenuName is not found nothing is added. 
If the first submenu is found but all the following sub menus are not found then the item will be added to the first submenu.

SteffenHeu: i can throw yes. 

i dont understand your second point though, the subMenu is reset to null at every iteration where it is not found

robinschmid: Review while half listening to a talk does not work well it seems. 
True forgot about the orElse. 

Now with the exception the flow is clear. Any submenu is not found = exception.

PR #3198, opened 2026-02-18
Wizard advanced mode

SteffenHeu:
Add option to customize specific parameters in the wizard using the advanced mode.

### DIFF
--- a/.agents/skills/mvci-gui/SKILL.md
+++ b/.agents/skills/mvci-gui/SKILL.md
@@ -8,6 +8,7 @@ description: Use a MVC or MVCI architecture when building complex user interface
 
 ## Model
 
+- Treat the Model as the single source of truth.
 - The Model controls the state. It holds all data in JavaFX properties.
 - Indicate if the value of a property can be null or not with @Nullable or @NotNull annotations in
   the `<>` brackets such as `ObjectProperty<@Nullable X>`.
@@ -37,7 +38,16 @@ description: Use a MVC or MVCI architecture when building complex user interface
 - The ViewBuilder is responsible for the logic that reacts to user input.
 - The ViewBuilder is responsible for logic that reflects the Controller's or Interactor's changes to
   the Model in the View.
-- Event handlers to react to interactions should be put into the view builder.
+- Event handlers to react to user interactions should be put into the view builder.
+- UI control listeners should update one model property (the direct user intent).
+- UI control listeners should not update multiple model properties or perform derived UI
+  recomputation.
+- Derived updates must be triggered by model-property listeners in the ViewBuilder.
+- Prefer bindings from View to Model; if control property is read-only (e.g., selectedItem), use a
+  one-line listener that only writes the matching model property.
+- Avoid calling `onXSelected(...)` methods from other listeners/subscribers; use model properties to
+  drive flow.
+-
 
 ## Interactor
 
@@ -50,4 +60,11 @@ description: Use a MVC or MVCI architecture when building complex user interface
 ## Logic placement decision
 
 - The ViewBuilder handles UI events wiring, the Controller or Interactor handle application-facing
-  commands. 
+  commands.
+
+## Anti-patterns (forbidden)
+
+- `selectedItem listener -> calls onParameterSelected(...)` where `onParameterSelected` sets several
+  model properties.
+- Any UI listener that sets `selected*` plus `propertyX` plus `propertyY` in one callback.
+- Bidirectional sync loops managed by ad-hoc cross-calls instead of model-driven updates.


javafx-framework/src/main/java/io/github/mzmine/javafx/components/util/FxLayout.java
```diff
@@ -59,6 +59,7 @@ public class FxLayout {
   public static final int DEFAULT_SPACE = 5;
   public static final int DEFAULT_ICON_SPACE = 0;
   public static final Insets DEFAULT_PADDING_INSETS = new Insets(5);
+  public static final Insets NO_PADDING_INSETS = new Insets(0);
```
robinschmid: this is already defined in Insets.EMPTY best to just stick to this

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilder.java
```diff
@@ -77,7 +71,20 @@ protected WizardBatchBuilder(WizardSequence steps) {
...
+    return queue;
+  }
+
+  /**
+   * Internal createQueue method to override by implementing classes. PostProcessing is applied in
+   * the final {@link #createQueue()} method.
+   *
+   * @return a batch queue
+   */
+  protected abstract BatchQueue __createQueueInternal();
```
robinschmid: are we doing the __ now? seems not very java like and the method name is already xyInternal

robinschmid: are we doing the __ now? seems not very java like and the method name is already xyInternal

robinschmid: are we doing the __ now? seems not very java like and the method name is already xyInternal

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/ParameterOverride.java
```diff
@@ -0,0 +1,66 @@
...
+ * Represents a single parameter override for a specific module parameter. This class is
+ * serializable so it can be saved/loaded with wizard presets.
+ */
+public record ParameterOverride(String moduleClassName, String moduleName,
+                                Parameter<?> parameterWithValue, ApplicationScope scope) {
+
+  public ParameterOverride(@NotNull String moduleClassName, @NotNull String moduleName,
+      @NotNull final Parameter<?> parameterWithValue, @NotNull ApplicationScope scope) {
+    this.moduleClassName = moduleClassName;
+    this.moduleName = moduleName;
```
robinschmid: we also have MZmineModule.getUniqueID, currently this is by default the class.getSimpleName but this allows us to rename modules and just keep the old name for the uniqueID for load and save. 

Maybe good to use uniqueID here?

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/ParameterOverridesParameter.java
```diff
@@ -0,0 +1,171 @@
...
+
+  @Override
+  public void setValue(List<ParameterOverride> newValue) {
+    this.value = newValue == null ? new ArrayList<>() : new ArrayList<>(newValue);
+  }
+
+  @Override
+  public boolean checkValue(Collection<String> errorMessages) {
+    // Always valid - can have 0 or more overrides
+    return true;
```
robinschmid: do we need to loop over all parameter overwrites and check if those parameters are valid?

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/ParameterCustomizationViewBuilder.java
```diff
@@ -0,0 +1,491 @@
...
+    });
+
+    // --- Model → parameterListView items ---
+    model.getAvailableParameters().addListener(
+        (ListChangeListener<UserParameter<?, ?>>) _ -> parameterListView.getItems()
+            .setAll(model.getAvailableParameters()));
+
+    // --- Wire parameter list selection → controller ---
+    parameterListView.getSelectionModel().selectedItemProperty()
+        .addListener((obs, oldVal, newVal) -> onParameterSelected.accept(newVal));
```
robinschmid: usually the view would just bind the selectedItemProperty to the model via binding or subscription update. 
Then we dont need to forward so many different methods into the view constructor. So usually best to just update the model either in the view or the controller and then react to these changes in each class.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilder.java
SteffenHeu: we also did it for the memory map storage, but i can also change it again. i just think it makes it clearer

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/ParameterOverridesParameter.java
SteffenHeu: yes can add a check here. but i think checking in the gui is more important. also adding a check there

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/ParameterCustomizationViewBuilder.java
SteffenHeu: refactored. now the view builder is a lot more bloated

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilder.java
robinschmid: We can also use it. Just a style decision we have to take but then I think it is mostly used to have the exact same method signature and name twice and you would skip the internal like createQueue() calls __createQueue()

I have never really seen double __ or single _ for private in java libraries. I guess it is more common in languages that do not have full control over visibility. Like in Python or in C/C++ on a file level. Double underscore seems to be even reserved for language and compiler terms in C++ https://stackoverflow.com/questions/224397/why-do-people-use-double-underscore-so-much-in-c

Same in python double underscore are methods defined by language

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/ParameterOverridesParameter.java
```diff
@@ -0,0 +1,176 @@
...
+      String moduleClass = overrideElement.getAttribute(MODULE_CLASS_ATTR);
+      String moduleUniqueId = overrideElement.getAttribute(MODULE_UNIQUE_ID_ATTR);
+      String paramName = overrideElement.getAttribute(PARAM_NAME_ATTR);
+
+      // Load scope with backwards compatibility - default to ALL if not present
+      String scopeStr = overrideElement.getAttribute(SCOPE_ATTR);
+      final ApplicationScope scope = UniqueIdSupplier.parseOrElse(scopeStr,
+          ApplicationScope.values(), ApplicationScope.ALL);
+
+      final MZmineModule module = MZmineCore.getInitializedModules().get(moduleClass);
```
robinschmid, line 113: do we have to use `MZmineCore.getModuleInstance(moduleClass)` so that a module might be initialized at that time (like if it is not already in the config)

And we need to handle module==null

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/ParameterCustomizationViewBuilder.java
robinschmid: This is not what I meant in my comment. 
It was not about putting the whole controller into the ViewBuilder. 
It is still not very easy to know how to handle events in the view and communicate with the controller and we are using many different ways for now, which is fine. 

1. pass lambdas into the constructor of ViewBuilder
2. set consumers or runnables in the model
3. Use a Property<SomeEvent> or Property<SomeState> and enums or sealed interfaces for events to handle more complex state changes. 

So it was fine that the code was passing methods into the constructor. 
Just some of them could have been changed to: 
ViewBuilder control selects a new parameter -> set to model 
Controller subscribes to changes in model and handles the change by changing the model. 

You can leave it like it is now. Just wanted to clarify that this was not about totally changing it. And that I also dont know what the best solution for the reactivity is. I guess it is a mix of the options depending on the complexity of the MVC.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/subparameters/ParameterOverridesParameter.java
SteffenHeu: I mean i can do Class.forName but i considered this to be the safer option. due to the batch mode, we have all modules initialized any way

robinschmid: Dont really know what you mean. 

We have a method `MZmineCore.getModuleInstance(moduleClass)` that should be safer than getting the modules map and calling get on it. 

MZmineCore.getInitializedModules() is currently never used outside of MZmineCore. 
Maybe we could even make this method private?

PR #3206, opened 2026-02-20
Add Rt correction to wizard

SteffenHeu:
### Description

This pull request introduces enhancements and corrections related to runtime correction (RT correction) handling in the wizard functionalities:
- Clears existing RT corrections when initiating new wizard runs.
- Adds a module to remove previously applied RT corrections.
- Introduces an option to enable scan-based RT correction in the wizard.
- Moves the GC EI RT correction step to the EI wizard for better modularity and integrates the scan-based RT correction method into the base wizard.


SteffenHeu:
Added checkbox to chrom builder and rt corr to clear correction
This also means that the clear corrections module shows up every time in a raw file when the chromatogram builder is applied, and also for the rt correction itself.

robinschmid:
could we just remove the RT correction from the applied methods on the raw data files instead of adding a clear RT correction step?  

This would lead to very clean "create batch from applied methods" while currently we would end up with a long list of rt correction, clear correction, rt correction [..] if people just apply the batch multiple times during optimization

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_adapchromatogrambuilder/ADAPChromatogramBuilderParameters.java
```diff
@@ -168,4 +177,29 @@ public Map<String, Parameter<?>> getNameParameterMap() {
...
+    var param = new ADAPChromatogramBuilderParameters().cloneParameterSet();
+
+    param.setParameter(ADAPChromatogramBuilderParameters.dataFiles, files);
+    param.setParameter(ADAPChromatogramBuilderParameters.scanSelection, scans);
+    param.setParameter(ADAPChromatogramBuilderParameters.minimumConsecutiveScans, minRtDataPoints);
+    param.setParameter(ADAPChromatogramBuilderParameters.mzTolerance, mzTolScans);
+    param.setParameter(ADAPChromatogramBuilderParameters.suffix, suffix);
+    param.setParameter(ADAPChromatogramBuilderParameters.minGroupIntensity, minGroupInt);
+    param.setParameter(ADAPChromatogramBuilderParameters.minHighestPoint, minHeight);
+    param.setParameter(ADAPChromatogramBuilderParameters.clearRtCorrection, clearRtCorrection);
```
robinschmid, line 201: also set allowSingleScans

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_remove_scanrtcal/RemoveScanRtCorrectionTask.java
```diff
@@ -0,0 +1,91 @@
...
+
+  @Override
+  protected @NotNull List<RawDataFile> getProcessedDataFiles() {
+    return List.of(files);
+  }
+
+  @Override
+  protected void addAppliedMethod() {
+    // always add applied method for consistency in projects where multiple calibrations are applied.
+    getProcessedDataFiles().forEach(file -> file.getAppliedMethods().add(
```
robinschmid: maybe better to not add an applied step but to remove all RT correction steps?
There should be no side effect and makes the build batch cleaner. 
OR we could also filter those steps in the build batch from applied methods

PR #3217, opened 2026-02-25
Add DefaultOffCustomParameter

SteffenHeu:
Goal is more straightforward handling of advanced parameters

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,8 +25,11 @@ not-bio
 /src/test/resources/rawdatafiles/local
 /src/test/java/import_data/speed/local/tims32subs/
 
+# AI Agent ignores
 codex.md
+AGENTS.md
 claude.md
+.claude/
 
 # External tools fetched locally (do not commit downloaded parser binaries)
 /external_tools/thermo_raw_file_parser/


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/combowithinput/DefaultOffCustomStringParameter.java
```diff
@@ -0,0 +1,86 @@
...
+   * @param offValue          the value returned when {@link DefaultOffCustomOption#OFF} is selected;
+   *                          may be {@code null}
+   * @param includeOff        if {@code false} the {@link DefaultOffCustomOption#OFF} entry is
+   *                          omitted from the combo-box choices
+   */
+  public DefaultOffCustomStringParameter(@NotNull StringParameter embeddedParameter,
+      @Nullable String defaultValue, @Nullable String offValue, boolean includeOff) {
+    super(embeddedParameter, defaultValue, offValue,
+        new DefaultOffCustomStringValue(DefaultOffCustomOption.DEFAULT,
+            embeddedParameter.getValue()),
```
robinschmid: in the other classes this is used:
```java
embeddedParameter.getValue() != null ? embeddedParameter.getValue() : defaultValue
```

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/combowithinput/DefaultOffCustomParameter.java
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Resolves the effective value using the option stored in {@link #value} together with the
+   * {@link #defaultValue} and {@link #offValue} fields.
+   *
+   * @return the resolved value; may be {@code null} when {@link DefaultOffCustomOption#OFF} or
+   * {@link DefaultOffCustomOption#DEFAULT} is selected and the corresponding value is {@code null}
+   */
+  @Nullable
+  public V resolve() {
```
robinschmid: resolveValue? I think I would always search for somehting to get the value in

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/combowithinput/DefaultOffCustomStringParameter.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * A combo parameter with {@link DefaultOffCustomOption} options and a text input field active when
+ * {@link DefaultOffCustomOption#CUSTOM} is selected.
+ * <p>
+ * Both the {@code defaultValue} (returned for {@link DefaultOffCustomOption#DEFAULT}) and the
+ * {@code offValue} (returned for {@link DefaultOffCustomOption#OFF}) are stored in the parameter
+ * itself, so {@link #resolve()} requires no arguments.
+ */
+public class DefaultOffCustomStringParameter extends
```
robinschmid: do we need all the specialized classes or could we just use DefaultOffCustomParameter and handle things similar to OptionalParameter? Like cloning the embedded parameter and saving values etc.  

Because the String, Double, Int, Combo version seem to only implement logic that could be centralized. 

This way we can also use this parameter for more complex parameters like MZTolerance etc.

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/AdvancedParametersParameter.java
```diff
@@ -210,8 +211,30 @@ public <V> V getValueOrDefault(Parameter parameter, V defaultValue) {
       } else {
         return (V) getEmbeddedParameters().getParameter(optional).getEmbeddedParameter().getValue();
       }
+    } else if (parameter instanceof DefaultOffCustomParameter<?, ?, ?> docParam) {
+      return (V) getEmbeddedParameters().getParameter(docParam).resolve();
```
robinschmid: maybe throw an exception here and say we should use the other method? 
Because the defaultValue is not considered here.

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/OffDefaultOn.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Interface for Options of
+ * {@link io.github.mzmine.parameters.parametertypes.ComboParameter<OffDefaultOn>} in
+ * {@link io.github.mzmine.parameters.parametertypes.AdvancedParametersParameter}s to reduce
+ * ambiguity. This is meant to replace a
+ * {@link io.github.mzmine.parameters.parametertypes.BooleanParameter} in an advanced parameter set
+ * to make it clear what the default value is, hence {@link #getValue} may be false or true.
+ */
+public interface OffDefaultOn {
```
robinschmid, line 35: unused

SteffenHeu: yes but required in mzio repo.

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/combowithinput/DefaultOffCustomStringParameter.java
SteffenHeu: yes were were primitives before, will change. the default is nullable anyway

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/combowithinput/DefaultOffCustomStringParameter.java
SteffenHeu: good point, will try

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/AdvancedParametersParameter.java
SteffenHeu: it uses the intended default. my train of thought was that this method feels so clunky with the DefaultOffCustomParameter that you look for the better way. Throwing a runtime exception feels wrong

.claude/settings.local.json
```diff
@@ -0,0 +1,7 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(git ls-tree:*)"
```
robinschmid: claude file best to put on ignore or is this good to have for everyone?

SteffenHeu: yeah this should not be here

PR #3218, opened 2026-02-26
No spectral matching if library import running

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_precursordbsearch/PrecursorDBSearchTask.java
```diff
@@ -89,6 +91,16 @@ public String getTaskDescription() {
 
   @Override
   protected void process() {
+    // have to check if spectral libraries are all imported
+    // cannot run matching if libraries are still imported this is most likely because
+    // user loaded manually and started matching to early
+    if (TaskService.getController()
```
SteffenHeu, line 97: When you drag and drop a library it uses the allspectraldataimport: https://github.com/mzmine/mzmine/blob/ab2b10e61a0ebf81b0cd5e7cca26c8d2a1ee608b/mzmine-community/src/main/java/io/github/mzmine/gui/MZmineGUI.java#L392

So i think we should either use the library import if only libraries are dropped, or also block when the all import is running

robinschmid: The all data import starts the library import task. And it is also added to the task controller

robinschmid: Just tested again also with drag drop and works as intendet

PR #3230, opened 2026-03-09
Centralize module access

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/main/MZmineCore.java
```diff
@@ -403,7 +426,7 @@ public static Collection<MZmineModule> getAllModules() {
    *
    * @return An unmodifiable copy of the currently initialized modules.
    */
-  public static Map<String, MZmineModule> getInitializedModules() {
+  private static Map<String, MZmineModule> getInitializedModules() {
```
SteffenHeu, line 429: do we still need to wrap in copyOf if this is a private method?

robinschmid: In the method that uses this, we stream over the elements so copy is safer here against concurrent modification exception.

PR #3231, opened 2026-03-09
New Intensity normalization module

SteffenHeu:
- Added reference samples parameter to allow interpolation from QCs, updated normalization algorithm (Results will differ from previous results)
- Normalization type was migrated to module options and now stores method-specific parameters.
- Standard compound normalization was migrated into the linear normalizer as a normalization type.
- Stores per-file normalization functions in the applied method to support persistent normalization of newly added features.
- Added metadata normalization

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/AverageIntensityNormalizationTypeModule.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/AverageIntensityNormalizationTypeModule.java
@@ -29,6 +29,8 @@ import io.github.mzmine.datamodel.RawDataFile;
 import io.github.mzmine.datamodel.features.FeatureListRow;
 import io.github.mzmine.datamodel.features.ModularFeature;
 import io.github.mzmine.datamodel.features.ModularFeatureList;
+import io.github.mzmine.javafx.dialogs.NotificationService;
+import io.github.mzmine.javafx.dialogs.NotificationService.NotificationType;
 import io.github.mzmine.parameters.ParameterSet;
 import org.jetbrains.annotations.NotNull;
 
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/FactorNormalizationFunction.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/FactorNormalizationFunction.java
@@ -29,25 +29,26 @@ import io.github.mzmine.parameters.parametertypes.selectors.RawDataFilePlacehold
 import io.github.mzmine.util.XMLUtils;
 import java.time.LocalDateTime;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.w3c.dom.Element;
 
 /**
  * Factor normalization function represented by one global factor.
  */
 public record FactorNormalizationFunction(@NotNull RawDataFilePlaceholder rawDataFilePlaceholder,
-                                          @NotNull LocalDateTime acquisitionTimestamp,
+                                          @Nullable LocalDateTime acquisitionTimestamp,
                                           double factor) implements NormalizationFunction {
 
   public static final String XML_TYPE = "factor";
   private static final String XML_FACTOR_ATTR = "factor";
 
   public FactorNormalizationFunction(@NotNull final RawDataFile referenceFile,
-      @NotNull final LocalDateTime acquisitionTimestamp, final double factor) {
+      @Nullable final LocalDateTime acquisitionTimestamp, final double factor) {
     this(new RawDataFilePlaceholder(referenceFile), acquisitionTimestamp, factor);
   }
 
   public FactorNormalizationFunction(@NotNull final RawDataFilePlaceholder rawDataFilePlaceholder,
-      @NotNull final LocalDateTime acquisitionTimestamp, final double factor) {
+      @Nullable final LocalDateTime acquisitionTimestamp, final double factor) {
     this.rawDataFilePlaceholder = rawDataFilePlaceholder;
     this.acquisitionTimestamp = acquisitionTimestamp;
     this.factor = factor;
@@ -59,7 +60,7 @@ public record FactorNormalizationFunction(@NotNull RawDataFilePlaceholder rawDat
   }
 
   @Override
-  public @NotNull LocalDateTime acquisitionTimestamp() {
+  public @Nullable LocalDateTime acquisitionTimestamp() {
     return acquisitionTimestamp;
   }
 
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/InterpolatedNormalizationFunction.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/InterpolatedNormalizationFunction.java
@@ -30,6 +30,7 @@ import io.github.mzmine.util.XMLUtils;
 import io.github.mzmine.util.maths.Precision;
 import java.time.LocalDateTime;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.w3c.dom.Element;
 
 /**
@@ -37,7 +38,7 @@ import org.w3c.dom.Element;
  */
 public record InterpolatedNormalizationFunction(
     @NotNull RawDataFilePlaceholder rawDataFilePlaceholder,
-    @NotNull LocalDateTime acquisitionTimestamp, @NotNull NormalizationFunction previousFunction,
+    @Nullable LocalDateTime acquisitionTimestamp, @NotNull NormalizationFunction previousFunction,
     double previousWeight, @NotNull NormalizationFunction nextFunction,
     double nextWeight) implements NormalizationFunction {
 
@@ -48,7 +49,7 @@ public record InterpolatedNormalizationFunction(
   private static final String XML_NEXT_FUNCTION_ELEMENT = "nextFunction";
 
   public InterpolatedNormalizationFunction(@NotNull final RawDataFile targetFile,
-      @NotNull final LocalDateTime acquisitionTimestamp,
+      @Nullable final LocalDateTime acquisitionTimestamp,
       @NotNull final NormalizationFunction previousFunction, final double previousWeight,
       @NotNull final NormalizationFunction nextFunction, final double nextWeight) {
     this(new RawDataFilePlaceholder(targetFile), acquisitionTimestamp, previousFunction,
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/NormalizationFunction.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/NormalizationFunction.java
@@ -47,7 +47,7 @@ public sealed interface NormalizationFunction extends UniqueIdSupplier permits
 
   @NotNull RawDataFilePlaceholder rawDataFilePlaceholder();
 
-  @NotNull LocalDateTime acquisitionTimestamp();
+  @Nullable LocalDateTime acquisitionTimestamp();
 
   /**
    * @param mz the mz of the feature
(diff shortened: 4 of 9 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/NormalizedAreaType.java
```diff
@@ -23,22 +22,19 @@
...
+import org.jetbrains.annotations.NotNull;
 
-  private final String name;
+public class NormalizedAreaType extends AreaType {
 
-  NormalizationType(String name) {
-    this.name = name;
+  @Override
+  public @NotNull String getUniqueID() {
+    return "norm_area";
```
robinschmid: maybe we can flip the unique id to area_norm same for height. 
Then the two columns area and area_norm are closer together in exports and easier to find. 

Just my personal preference

mzmine-community/src/main/java/io/github/mzmine/datamodel/AbundanceMeasure.java
```diff
@@ -66,4 +70,24 @@ public Float getOrNaN(@Nullable ModularDataModel featureOrRow) {
...
+      case Area -> "Area";
+      case Height -> "Height";
+    };
+  }
+
+  @Override
+  public @NotNull String getUniqueID() {
+    return switch (this) {
+      case NORMALIZED_AREA -> "normalized_area";
+      case NORMALIZED_HEIGHT -> "normalized_height";
```
robinschmid, line 88: maybe we can use the same unique ID here like in the type. At least for the normalized version. 
The other two have to stay for compatibility.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
```diff
@@ -0,0 +1,216 @@
...
+    if (referenceFiles.isEmpty()) {
+      error("No reference files found for normalization. %s: %s".formatted(
+          normalizationTypeModule.getName(), normalizationTypeModuleParameters.toString()));
+      return;
+    }
+
+    final MetadataTable metadata = ProjectService.getMetadata();
+    final Map<@NotNull RawDataFile, @NotNull NormalizationFunction> fileToFunction = normalizationTypeModule.createReferenceFunctions(
+        referenceFiles, originalFeatureList, metadata, mainParameters,
+        normalizationTypeModuleParameters);
```
robinschmid, line 126: At least for the metadata column normalizer:

catch IllegalStateException that is thrown if RawDataFile has null value mapping in metadata column. 
Fail task with error. 

Could also have the MetadataColumn normalizer throw a specific custom exception type so that we know its an expected exception that killed the task.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerModule.java
```diff
@@ -23,27 +22,28 @@
...
+import org.jetbrains.annotations.Nullable;
 
-public class LinearNormalizerModule implements MZmineProcessingModule {
+public class IntensityNormalizerModule implements MZmineProcessingModule {
 
-  private static final String MODULE_NAME = "Linear normalizer";
-  private static final String MODULE_DESCRIPTION =
-      "Linear normalizer divides the height (or area) of each feature in the feature list by a normalization factor, determined according to the given normalization type.";
+  private static final String MODULE_NAME = "Intensity normalizer";
+  private static final String MODULE_DESCRIPTION = "Linear normalizer scales feature intensities by a normalization factor determined according to the selected normalization type.";
```
robinschmid: Maybe update description to make it more open like. 

Various intensity normalizers that either work with constant factors for each sample or with standard compounds to normalize individual feature intensities. Results are saved in Norm. area and Norm. height columns.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/MetadataColumnNormalizationTypeModule.java
```diff
@@ -0,0 +1,121 @@
...
+            "No valid metadata values available in column: " + columnName));
+
+    final Map<@NotNull RawDataFile, @NotNull NormalizationFunction> functions = new HashMap<>(
+        referenceFiles.size());
+    for (final Entry<@NotNull RawDataFile, @NotNull Double> entry : fileToMetadataValue.entrySet()) {
+      final RawDataFile file = entry.getKey();
+      final double factor =
+          Double.compare(entry.getValue(), 0) == 0 ? 1 : maxMetadataValue / entry.getValue();
+      final LocalDateTime runDate = NormalizationTypeModule.getRunDateOrThrow(metadata, file);
+      functions.put(file, new FactorNormalizationFunction(file, runDate, factor));
```
robinschmid, line 103: Can we make the runDate optional here or have a NormalizationFunction without run date (disallows date interpolation with exception if sample is not covered). 

For the MetadataColumn normalizer it is by default restricted that no interpolation should happen. So the date is not used at all. However, this part will fail for all samples where we cannot read injection date.

Similar thing is true if we apply something like TIC or median normalization on all included raw data files, without filtering for QC or specific type. Then the date is not needed. So maybe better to fail the normalization later when it is applied to the file?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/AverageIntensityNormalizationTypeModule.java
```diff
@@ -0,0 +1,71 @@
...
+        final Float abundance = abundanceMeasure.get(feature);
+        if (abundance != null) {
+          intensitySum += abundance;
+          intensityCount++;
+        }
+      }
+    }
+    if (Double.compare(intensitySum, 0d) == 0) {
+      throw new IllegalStateException(
+          "Sum of feature intensities is 0 for file: " + file.getName());
```
robinschmid, line 69: intensitySum==0 seems to be an ok case. Just no feature for this raw data file. 
I think it is best to just return 0 or 1 as normalization factor. 

This is only an exceptional case if we normalize on QCs. 
Maybe we can just show a notification and log the issue.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
```diff
@@ -0,0 +1,93 @@
...
+  public static final OriginalFeatureListHandlingParameter handleOriginal = new OriginalFeatureListHandlingParameter(
+      "Original feature list",
+      "Defines the processing.\nKEEP is to keep the original feature list and create a new"
+          + "processed list.\nREMOVE saves memory.", false);
+
+  public static final ModuleOptionsEnumComboParameter<NormalizationType> normalizationType = new ModuleOptionsEnumComboParameter<>(
+      "Normalization type", "Normalize intensities by...",
+      NormalizationType.MedianFeatureIntensity);
+
+  public static final HiddenParameter<List<NormalizationFunction>> normalizationFunctions = new HiddenParameter<>(
```
robinschmid, line 73: maybe add javadoc for this one. 
Just to describe that those are stored in applied methods to automatically apply the normalization to features that are added to the feature list. Or after feature has changed integration bounds

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/NormalizationTypeModule.java
```diff
@@ -0,0 +1,67 @@
...
+  @NotNull Map<@NotNull RawDataFile, @NotNull NormalizationFunction> createReferenceFunctions(
+      @NotNull List<@NotNull RawDataFile> referenceFiles, @NotNull ModularFeatureList featureList,
+      @NotNull MetadataTable metadata, @NotNull ParameterSet mainParameters,
+      @NotNull ParameterSet moduleSpecificParameters);
+
+  @NotNull NormalizationFunction createInterpolatedFunction(@NotNull RawDataFile fileToInterpolate,
+      @NotNull NormalizationFunction previousRunCalibration,
+      @NotNull NormalizationFunction nextRunCalibration,
+      @NotNull InterpolationWeights interpolationWeights, @NotNull MetadataTable metadata,
+      @NotNull ParameterSet parameters, @NotNull ParameterSet normalizerParameters);
```
robinschmid, line 52: the argument main parameters `ParameterSet parameters` is never used in any implementation. 


If we can avoid passing two different parameter sets by removing this argument from this method and from the other methods in this class we would improve the decoupling from the main parameter set. 

The main parameter set only provides the abundance measure. Maybe we can just pass it into the constructor of the module or instead of passing parameter set we could pass the abundance measure to the methods that use it like createReferenceFunctions

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurelisttable_modular/FeatureTableFX.java
```diff
@@ -325,6 +326,8 @@ public Class<? extends HeightType> getDefaultAbundanceMeasureType() {
     return switch (parameters.getValue(FeatureTableFXParameters.defaultAbundanceMeasure)) {
       case Height -> HeightType.class;
       case Area -> AreaType.class;
+      case NORMALIZED_HEIGHT -> NormalizedHeightType.class;
+      case NORMALIZED_AREA -> NormalizedAreaType.class;
```
robinschmid: this is used in showCompactChromatographyColumns.  
Maybe check if this column exists and otherwise set the fallback option visible. Like normalized area is not available set area column visible.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/MetadataTableUtils.java
```diff
@@ -123,4 +127,112 @@ public static boolean matchesFilename(final @Nullable String name, @NotNull RawD
...
+    long minPreviousRun = Long.MIN_VALUE;
+    RawDataFile previousRun = null;
+
+    for (final RawDataFile thatFile : otherFiles) {
+      final LocalDateTime thatDate = metadata.getValue(runDateColumn, thatFile);
+      if (thatDate == null) {
+        throw new FileHasNoRunDateException(thatFile);
+      }
+
+      final long diff = runDate.until(thatDate, ChronoUnit.SECONDS);
```
robinschmid: Better use nano seconds here and below. 
Super fast maldi spot analysis is faster than seconds. And no reason to limit it

mzmine-community/src/test/resources/outfiles/mzmine_log.log
```diff
@@ -43,7 +43,7 @@ java.lang.ClassNotFoundException: io.mzio.mzminepro.modules.otherdata.featdet_re
 2025-03-21 10:22:02 FINEST io.github.mzmine.main.MZmineCore getModuleInstance Creating an instance of the module io.github.mzmine.modules.io.export_features_mgf.AdapMgfExportModule
 2025-03-21 10:22:02 FINEST io.github.mzmine.main.MZmineCore getModuleInstance Creating an instance of the module io.github.mzmine.modules.io.export_features_mztabm.MZTabmExportModule
 2025-03-21 10:22:02 FINEST io.github.mzmine.main.MZmineCore getModuleInstance Creating an instance of the module io.github.mzmine.modules.dataprocessing.featdet_massdetection.localmaxima.LocalMaxMassDetector
-2025-03-21 10:22:02 FINEST io.github.mzmine.main.MZmineCore getModuleInstance Creating an instance of the module io.github.mzmine.modules.dataprocessing.norm_linear.LinearNormalizerModule
+2025-03-21 10:22:02 FINEST io.github.mzmine.main.MZmineCore getModuleInstance Creating an instance of the module io.github.mzmine.modules.dataprocessing.norm_intensity.IntensityNormalizerModule
```
robinschmid, line 46: weird how this file sneaked in as it is old. Need to remember to delete it and add gitignore for resources outfiles?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
SteffenHeu: I think since the exceptions have pretty specific descriptions the source of the error should be prett clear without a specific exception type.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/NormalizationTypeModule.java
SteffenHeu: It is not uncommon that this is necessary. New parameters may always be added and then every implementation has to be changed instead of just grabbing or not grabbing another parameter.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/AverageIntensityNormalizationTypeModule.java
SteffenHeu: The problem is at the point of interpolation you don't know what happened. This is not the normalization factor, it is the metric the factor is computed from. And here you don't know what will lead to a factor of 1 because that depends on the other files. And as soon as you start returning 0 here, you will get infinities because you devide by it.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
robinschmid: This is an expected result rather than an exceptional case. 
The only thing was that it would be better to catch the exception and error out the task just with the exception method and not with the whole stack trace. 
Currently WrappedTask.run will automatically log the whole stack trace and AbstractTask.error also logs the full stack trace. 

I think if we know this will happen we can just log exception.message instead. Makes the logs less polluted

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/AverageIntensityNormalizationTypeModule.java
robinschmid: It should be a rare case. So if intensity is zero people would just have to fix their parameters or remove the sample with zero intensities and they can use it again. 
So can stay this way

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/NormalizationTypeModule.java
robinschmid: Can also stay this way

PR #3248, opened 2026-03-20
Update external tools packaging

SteffenHeu:
Use jpackage native options to package the external tools. Changes behaviour on linux, they are now placed in the lib folder. Does not affect windows. Will have to test mac and signing.


mzmine-community/build.gradle
```diff
@@ -302,36 +302,8 @@ application {
...
-        // Ensure external tools are executable and not quarantined on macOS
-        try {
-            def externalToolsDir = EXTERNAL_TOOLS_DESTINATION.asFile
-            def trfpMac = new File(externalToolsDir, "thermo_raw_file_parser/ThermoRawFileParser")
-            if (trfpMac.exists()) {
-                execOps.exec { commandLine(['chmod', '+x', trfpMac.absolutePath]) }
-            }
-            // Remove quarantine attributes recursively (ignore failures on files without the attribute)
-            execOps.exec {
-                commandLine(['xattr', '-r', '-d', 'com.apple.quarantine', externalToolsDir.absolutePath])
```
robinschmid, line 327: I guess we still have to check if this works on macOS?

PR #3250, opened 2026-03-23
Replace groupable list view with TreeView

SteffenHeu:
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/gui/MZmineGUI.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/gui/MZmineGUI.java
@@ -282,9 +282,12 @@ public class MZmineGUI extends Application implements MZmineDesktop, JavaFxDeskt
 
   private static void setupRawDataStrategies(
       @NotNull final GroupableTreeView<RawDataFile> rawDataList) {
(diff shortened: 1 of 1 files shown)


SteffenHeu:
Fixed everything. Manual grouping stays and works now

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/color/ColorByMetadataTask.java
```diff
@@ -146,7 +147,9 @@ protected void process() {
       map.forEach(RawDataFile::setColor);
 
       if (applySorting) {
-        MZmineGUI.sortRawDataFilesAlphabetically(raws);
+        MZmineGUI.sortRawDataFiles(raws.stream()
+            .sorted(Comparator.comparing(RawDataFile::getName, String.CASE_INSENSITIVE_ORDER))
+            .toList());
```
robinschmid, line 152: can we just pass the comparator instead of the sorted list? 
Internally this is just converted to a comparator again

robinschmid: ok this has the option to sort only a few files

PR #3261, opened 2026-03-30
Bugfix for extrapolation with RI calculation, now uses correct offset for large retention times

wchu207:
Before:
image
After:
image

Note: extrapolation is a nonstandard feature that probably no one uses


SteffenHeu:
Thanks 👍

PR #3262, opened 2026-03-30
Intensity normalizer

robinschmid:
split into multiple steps of normalization

1. normalize by metadata each sample (like weight or dilution factor)
2. normalize by internal standards (may change each feature individually)
3. Intra batch correction for example by pooled QCs, interpolate for samples between qcs
4. Inter batch correction (automatically applied if intra batch correction is selected). After normalizing each batch, normalize batches to their median reference sample intensities. 

Functions are saved as multiple steps now and applied one after the other.

## TODO

- [ ] add test for multi batch correction

image


mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
```diff
@@ -410,11 +410,14 @@ public static void recalculateIonSeriesDependingTypes(@NotNull final ModularFeat
...
-    if (normalizer != null) {
-      normalizeAbundances(feature, normalizer);
+    // auto-apply normalization if present, apply all steps in order and accumulate normalization
+    // need to set to null before to start fresh with raw height and area
+    if (feature.getRawDataFile() != null) {
+      feature.set(NormalizedHeightType.class, null);
+      feature.set(NormalizedAreaType.class, null);
+      IntensityNormalizerModule.streamNormalizationFunctionsOfLatestCallForFile(
+              feature.getFeatureList(), feature.getRawDataFile())
+          .forEach(normalizer -> accumulateNormalization(feature, normalizer));
```
SteffenHeu: The idea was that this is not necessary, because the normalization ways always supposed to start with the areaType/heightType.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/MetadataColumnNormalizationTypeModule.java
```diff
@@ -89,18 +91,27 @@ public class MetadataColumnNormalizationTypeModule implements NormalizationTypeM
...
       final RawDataFile file = entry.getKey();
+      // correct sample by factor/median to keep general intensity scales
+      // could also think about using the factor as is
       final double factor =
-          Double.compare(entry.getValue(), 0) == 0 ? 1 : maxMetadataValue / entry.getValue();
+          Double.compare(entry.getValue(), 0) == 0 ? 1 : entry.getValue()/median;
       final LocalDateTime runDate = MetadataTableUtils.getRunDate(metadata, file);
-      functions.put(file, new FactorNormalizationFunction(file, runDate, factor));
+      // divide or multiple by factor
+      functions.put(file, new FactorNormalizationFunction(file, runDate, metadataConfig.mode().isDivide()? 1d/factor : factor));
```
SteffenHeu: Why change the normalisation logic (multiply/divide)? I would say this just increases the risk for user error.

The scale of the intensities ways also kept previously sicne the scale of the raw file with the maxmimum metadata value did not change at all and others were just adjusted to match that one.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/MetadataNormalizationConfig.java
```diff
@@ -0,0 +1,61 @@
...
+package io.github.mzmine.modules.dataprocessing.norm_intensity;
+
+import io.github.mzmine.datamodel.utils.UniqueIdSupplier;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Either multiply or divide value by metadata value. Weight is divided dilution factor is
+ * multiplied
+ */
+public record MetadataNormalizationConfig(@NotNull String metadataColumn, @NotNull Mode mode) {
```
SteffenHeu, line 35: When i write "=1/1000" as my dilution factor i would still multiply by 1000 to de-normalize toward a file that has a dilution of 1?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizationSummary.java
```diff
@@ -0,0 +1,50 @@
...
+package io.github.mzmine.modules.dataprocessing.norm_intensity;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * @param steps the list of normalization steps
+ */
+public record IntensityNormalizationSummary(@NotNull List<IntensityNormalizationSummaryStep> steps, @NotNull
```
SteffenHeu: My approach would have been similar, but i would have used a CompositeNormalizationFunction, which would contain each normalization function as an individual step. The previous functions could then be passed to the next normalization step, which can add it's own normalization on top of that. We can then have a package-private EditableCompositeNormalizationFunction and a immutable public one which is created after the last step. Then we don't have to deal with the input of one method being either the normalized or non normalized values.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/MetadataColumnNormalizationTypeModule.java
robinschmid: Metadata values might be factors or divisors and then people would have to change their metadata. 
I also think it makes it much clearer if the UI already says what happens to the values. So that they divide the abundance or multiply. 

most are divide: mass, cell count, creatinine, 
some are factors: dilution factor, response factor,

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/MetadataNormalizationConfig.java
robinschmid: Both works. People often write dilution factors instead of divisors so often 1000 for a 1/1000 dilution. 
Of course people can go and just change the way they write the metadata. But I think if people see divide or multiple as options they will think about the proper way more often. 
If we always use division, users might not think about it and just pass dilution factors in there. 
So I prefer having a clear UI element

mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/FeatureDataUtils.java
robinschmid: reverted to a single function that may be composed of multiple internally

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizationSummary.java
robinschmid: done. 
I introduced a composite normalziation function and the summary manages the merging of functions. If its all constant factors, they are just multiplied into a new factor function but more complex functions get merged into composite.

mzmine-community/src/main/java/io/github/mzmine/datamodel/AbundanceMeasure.java
```diff
@@ -56,20 +63,46 @@ public Class<? extends DataType<Float>> type() {
...
 
+  /**
+   * Applies a normalization function to the value
+   *
+   * @param featureOrRow          the model
+   * @param normalizationFunction a function
+   * @return the normalized value or raw value if function is null or NaN if feature or value is
+   * null or not a finite number
+   */
+  public float getOrNaN(@Nullable ModularDataModel featureOrRow,
```
SteffenHeu, line 104: This method does not do what the the javadoc says. If there is no function it returns the unnormalized value, but if there is a function it returns the factor. So it either needs to be multiplied still or if return of the factor is intended, i would say the method does not belong into the abundancetype. Also there should be a check that throws an exception if this is not a raw type, or?

SteffenHeu (commented):
a

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/FeatureIntensityNormalizationParameters.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public static final CheckComboParameter<SampleType> sampleTypes = new CheckComboParameter<>(
+      "Reference samples", """
+      Select all sample types that shall be used to calculate the recalibration from.
+      The recalibration of all other samples will be based on the acquisition order, which is
+      determined by the acquisition type column in the metadata (CTRL/CMD + M).
+      """, SampleType.values(), List.of(SampleType.QC));
+
+  public static final ComboParameter<FeatureIntensityNormalizationMode> mode = new ComboParameter<>(
+      "Use intensity", """
```
SteffenHeu: Maybe "Feature abundance normalization mode"? It's a mouthful but more descriptive than use intensity

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizationSimpleSummary.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * This is a storable version that uses weak references in {@link RawDataFilePlaceholder}. The
+ * searchable version is used during normalization building
+ * {@link IntensityNormalizationSearchableSummary}
+ */
+public record IntensityNormalizationSimpleSummary(@NotNull List<NormalizationFunction> functions,
```
SteffenHeu: remove the simple in the name to keep it a bit more concise?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/CompositeListNormalizationFunction.java
```diff
@@ -0,0 +1,120 @@
...
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+/**
+ *
+ * @param functions a list of all functions to be applied. Each factor is multiplied to get the
+ *                  total factor.
+ */
+public record CompositeListNormalizationFunction(
```
SteffenHeu: remove the list from the name to keep it a bit more concise?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
```diff
@@ -107,110 +159,230 @@ public void run() {
...
+
+    // values might be NaN if they are unset
+    if (batchNormMetrics.length != sampleBatches.size()) {
+      logger.info(
+          "Skipping inter batch correction as it seems to be not supported by this batch normalizer: "
+              + normalizationTypeModule.getName());
+      return;
+    }
+
+    logger.info("Normalizing samples inter batches n=" + sampleBatches.size());
```
SteffenHeu: Is this all really info level?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
```diff
@@ -107,110 +159,230 @@ public void run() {
...
+        processedFiles++;
       }
-      final NormalizationFunction interpolatedFunction = normalizationTypeModule.createInterpolatedFunction(
-          fileToInterpolate, previousFunction, nextFunction, result, metadata, mainParameters,
-          normalizationTypeModuleParameters);
-      fileToFunction.put(fileToInterpolate, interpolatedFunction);
     }
+  }
+
+  private void normalizeSamplesBatch(SamplesBatch samplesBatch,
```
SteffenHeu, line 266: add short description that this applies corrections within a single batch (intra batch). ie, metadata, intra sample and intra batch

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/InterpolatedNormalizationFunction.java
```diff
@@ -123,4 +128,11 @@ public void saveToXML(final @NotNull Element functionElement) {
         previousFunction, previousWeight, nextFunction, nextWeight);
   }
 
+  @Override
+  public @NotNull NormalizationFunction withRawFile(@NotNull RawDataFile file) {
```
SteffenHeu: I'm not sure that this makes sense. The weights are specific for the placement of the file between reference files. So why is it supported to change the file for an interpolated function?

SteffenHeu: As far as i can tell this is never called outside of singleInterpolationWeight, but why does this method exist then? in my opinion this encurages future coding mistakes by using a calibration for a file it is not meant to be used for.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/NormalizationFunction.java
```diff
@@ -101,12 +154,17 @@ static void saveAcquisitionTimestamp(final @NotNull Element functionElement,
...
       return null;
     }
 
     return LocalDateTime.parse(acquisitionTimestamp);
   }
 
+  /**
+   * @return Create a copy of the function that points to the new file
+   */
+  @NotNull NormalizationFunction withRawFile(@NotNull RawDataFile file);
```
SteffenHeu: All the normalisations we have are at least file specific or not? what is the intention behind this method in general?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/NormalizationType.java
```diff
@@ -52,6 +48,21 @@ public enum NormalizationType implements ModuleOptionsEnum<NormalizationTypeModu
     this.moduleClass = moduleClass;
   }
 
+  /**
+   * @return All internal normalizers. Currently only standard compound normalzier but could be
+   * extended for lipid normalizer or similar.
+   */
+  public static NormalizationType[] internalSampleNormalizers() {
```
SteffenHeu: intraSample?

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/InterpolationWeights.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  record SingleInterpolationWeight(@NotNull RawDataFile closestRun) implements
+      InterpolationWeights {
+
+  }
+
+  /**
+   * Weights to interpolate normalization between two different two related data files. Data files
+   * might be null if there was a reference on only one side.
```
SteffenHeu: files are annotated as notnull.

SteffenHeu (dismissed):

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
```diff
@@ -48,57 +50,114 @@ public class IntensityNormalizerParameters extends SimpleParameterSet {
...
+          MetadataNormalizationConfig.getDefault()), false);
+
+  // ── Pre-normalization step 2: internal standard compounds
+  /**
+   * Applied after metadata normalization. Each feature is corrected by the nearest/weighted
+   * internal standard compound(s) in m/z and RT space, correcting for extraction efficiency and
+   * matrix effects.
+   */
+  public static final ModuleOptionsEnumComboParameter<NormalizationType> internalStandardization = new ModuleOptionsEnumComboParameter<>(
+      "Sample-internal normalization",
```
SteffenHeu: Intra sample normalization?
Or Sample Correction?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
```diff
@@ -48,57 +50,114 @@ public class IntensityNormalizerParameters extends SimpleParameterSet {
...
+      "Normalize by internal standard compounds before QC drift correction. "
+          + "Corrects for extraction efficiency and matrix effects. "
+          + "Each feature is corrected by the nearest or weighted IS compound(s).",
+      NormalizationType.internalSampleNormalizers(), false);
+
+  // ── Main normalization: QC-based signal drift correction
   public static final ModuleOptionsEnumComboParameter<NormalizationType> normalizationType = new ModuleOptionsEnumComboParameter<>(
-      "Normalization type", "Normalize intensities by...",
-      NormalizationType.MedianFeatureIntensity);
+      "Sample batch correction", """
```
SteffenHeu: Inter sample normalization?
Or just Inter batch correction? (Then i would use Sample Correction for the first one)

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
```diff
@@ -48,57 +50,114 @@ public class IntensityNormalizerParameters extends SimpleParameterSet {
...
+      Inter-batch normalization is performed by aligning the QC median of each batch to the global QC median.""",
+      NormalizationType.intraBatchDriftNormalizers());
+
+  /**
+   * When set, QC drift correction is computed independently per batch (intra-batch correction),
+   * then batches are aligned to the global QC median (inter-batch correction / pooled QC
+   * normalization). Without a batch ID, all QC samples form one continuous reference sequence.
+   */
+  public static final OptionalParameter<MetadataGroupingParameter> batchIdColumn = new OptionalParameter<>(
+      new MetadataGroupingParameter("Batch ID metadata column", """
```
SteffenHeu: Batch correction (by metadata column)

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
```diff
@@ -48,57 +50,114 @@ public class IntensityNormalizerParameters extends SimpleParameterSet {
...
    */
-  public static final HiddenParameter<List<NormalizationFunction>> normalizationFunctions = new HiddenParameter<>(
+  public static final HiddenParameter<IntensityNormalizationSimpleSummary> hiddenNormalizationSummary = new HiddenParameter<>(
       new NormalizationFunctionsParameter());
 
   public IntensityNormalizerParameters() {
-    super(new Parameter[]{featureLists, suffix, normalizationType, featureMeasurementType,
-            handleOriginal, normalizationFunctions},
+    super(new Parameter[]{featureLists, suffix, handleOriginal, featureMeasurementType,
+            batchIdColumn, metadataNormFactorCol, internalStandardization, normalizationType,
```
SteffenHeu, line 120: Maybe list the parameters in the order they are applied in during the correction?

mzmine-community/src/main/java/io/github/mzmine/datamodel/AbundanceMeasure.java
robinschmid: good catch. Forgot to add the value

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/FeatureIntensityNormalizationParameters.java
robinschmid: Maybe: **Normalize by feature abundances** and combo shows median. 
image

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizationSimpleSummary.java
robinschmid: done

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/CompositeListNormalizationFunction.java
robinschmid: done

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
robinschmid: Maybe better fine but this seems like relevant information to maybe find issues why normalization went wrong or lead to unexpected results. 

I changed most of the logs to fine

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
robinschmid: Done
```
  /// Applies normalization to a single samples batch (intra batch) which may be selected by
  /// metadata. If no batching is applied, all samples may be handled as a single batch.
  ///
  /// Applies:
  /// - Metadata normalization by factor or division
  /// - internal standards normalization sample-wise.
  /// - Reference samples (QC)-based intra batch correction
  ///
  /// @param samplesBatch the samples of this batch. Normalizers may set the medianNormMetric of
  /// this batch for later inter batch correction.
  /// @param summary      the summary to save normalization results to
  ```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
robinschmid: They are in order. 
Only batchIdColumn is on top as this influcences many of the other steps and I thought it is nicer to have the two metadata columns next to each other so that the users only have to check the metadata once.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
robinschmid: This is not only the inter batch correction
This combines Intra batch and then inter batch correction (if there are batches).

Another name could be something like
**Reference samples batch correction**

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
robinschmid: I like Intra-sample normalization

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
robinschmid: good idea. Better to reflect the batch correction in the name to point to the other parameter. 

I made it a bit shorter as this selects the column.
**Batch correction metadata column**

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/InterpolationWeights.java
robinschmid: good catch was a left over comment.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
SteffenHeu: but the inter batch correction is only applied if the metadata column is set and it is always the same algorithm

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
SteffenHeu: in what way does it influence the other steps? It only determines if the samples are split into batches or not, or?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/InterpolatedNormalizationFunction.java
robinschmid: Yeah makes sense. In general the dependence of the NormalizationFunction on the raw data file was limiting here because it was not possible to just define a simple factor function for all samples or reuse a function for a different raw data file. 
During interpolation we might only have one reference sample and then the target sample is normalized by exactly the same reference function.

Implemented the changes here in a PR on this branch:
https://github.com/robinschmid/mzmine3/pull/53

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/NormalizationFunction.java
robinschmid: See above. 
Function might come from a single reference function if the target sample was measured after the last QC or before the first QC.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
robinschmid: What do you mean with the last comment? This is the reason I would not call the parameter Inter batch correction.
How would you name this parameter then? 
I think something with batch correction describes best what happens. 

Maybe
Reference samples batch correction 
Inter samples batch correction

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
robinschmid: The splitting into batches by the metadata column happens first. So it makes sense to me that it is the first parameter. Everything else happens afterward. 

Currently the batching has an effect on the **intra-sample normalization** and **sample batch correction**. It acts in both as interpolation boundaries.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
SteffenHeu: Edit: I got confused. Thinking about it, i feel like Inter-Sample is the best description here due to the arguments in the nextparagraph

By the previous comment i mean that for the user the "Batch correction (by metatadata)" parameter basically influences if all samples are split into batches and if yes, potential batch effects are corrected. So calling it "Inter-sample correction" is the most obvious name. Describing that this parameter is also influenced by the batch correction parameter due to splitting the batches and applying the other corrections in the respective batch groups cannot be done in just the parameter name and the attempts to do that just lead to confusing parameter names. So i would move that explanation to the documentation and the tooltip.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
SteffenHeu: i liked it more with the "by" tbh :D but ok like this

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerParameters.java
robinschmid: inter-sample correction it is

PR #3276, opened 2026-04-13
Single library submission, keep last selected

robinschmid:
- fix npe when no library imported
- keeps last selected library
- asks if user really wants to overwrite a library with the same name as new library (only for)

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/row/SpectralLibrarySelectionDialog.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/row/SpectralLibrarySelectionDialog.java
@@ -26,12 +26,14 @@
 package io.github.mzmine.modules.io.spectraldbsubmit.row;
 
 import io.github.mzmine.javafx.components.factories.FxLabels;
+import io.github.mzmine.javafx.dialogs.DialogLoggerUtil;
 import io.github.mzmine.project.ProjectService;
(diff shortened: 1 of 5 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/io/spectraldbsubmit/row/SpectralLibrarySelectionDialog.java
```diff
@@ -126,8 +141,19 @@ public SpectralLibrary fromString(final String string) {
...
       }
-      // create a new library with a placeholder file path
+
       final String name = nameField.getText().strip();
+      final List<SpectralLibrary> libraries = ProjectService.getProject()
+          .getCurrentSpectralLibraries();
+      if (libraries.stream().anyMatch(lib -> lib.getName().equals(name))) {
+        final boolean overwrite = DialogLoggerUtil.showDialogYesNo(AlertType.WARNING,
+            "Overwriting existing library?",
+            "Do you really want to overwrite the existing library, replacing all its content with the new library entries?");
```
SteffenHeu, line 151: This sounds like the actual file will be overwritten. 

"Replace existing library?"
"A library with the name %s already exists. Creating a new library with the same name will remove the current library from the project and add a library only containing the new entries."

robinschmid: Good idea to be more precise. 
Is it good like this? 

image

PR #3282, opened 2026-04-15
Add more complex intensity normalizer tests

robinschmid:
- adds test with multiple normalizers activated with and without batching
- metadata normalization had to be moved out single batch processing because the scaling within a batch led to bad results
- add option to just apply metadata factor as they are (directly) or scaled with is median centering metadata values. The direct option is great if the factors should really be applied as they are, e.g., if they were calculated externally or if they are used to change from raw intensity to something like intensity/weight
- composite functions are now simplified if multiple factor functions in a sequence

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/CompositeNormalizationFunction.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/CompositeNormalizationFunction.java
@@ -26,6 +26,7 @@
 package io.github.mzmine.modules.dataprocessing.norm_intensity;
 
 import io.github.mzmine.util.XMLUtils;
+import java.util.ArrayList;
 import java.util.List;
 import org.jetbrains.annotations.NotNull;
 import org.w3c.dom.Element;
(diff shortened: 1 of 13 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
```diff
@@ -166,6 +166,18 @@ public void run() {
 
     final MetadataTable metadata = ProjectService.getMetadata();
 
+
+    // ── Pass 1: pre-normalization by metadata column (dilution factor, sample weight, …) ──
+    // should happen before separating samples into batches because factor may be scaled to median
```
SteffenHeu, line 171: what is the benefit of scaling the factor to the median?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/CompositeNormalizationFunction.java
```diff
@@ -35,11 +36,65 @@
...
+        hasFactorFunction = true;
+      } else {
+        if (hasFactorFunction) {
+          merged.add(new FactorNormalizationFunction(accumulatedFactor));
+          accumulatedFactor = 1.0;
+          hasFactorFunction = false;
+        }
+        merged.add(function);
+      }
+    }
```
SteffenHeu: as far as i understand this is to keep the normalization steps "in order"?

Say you have
1. factor
2. factor
3. by internal standard compounds
4. factor

only the first 2 are accumulated and then it stays as-is?

This is only required if we have an operation other than multiplication and division in use for normalization, or am i missing something?
It's a good point, but currently the normalization is based on "getNormalizationFactor" so one could ask why we pay attention to that here. I guess it is in case we add addition (in case that should be required) or so in the future?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/MetadataNormalizationConfig.java
```diff
@@ -32,15 +32,16 @@
  * Either multiply or divide value by metadata value. Weight is divided dilution factor is
  * multiplied
  */
-public record MetadataNormalizationConfig(@NotNull String metadataColumn, @NotNull Mode mode) {
+public record MetadataNormalizationConfig(@NotNull String metadataColumn, @NotNull Mode mode, @NotNull Scaling scaling) {
```
SteffenHeu: jdoc that scaling is used for metadata only

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
robinschmid: I think this was in your initial design and I kept it. 
I think best is to just use the raw factor but scaling keeps the general scale of values similar. Which is not the case in the IS normalizer though.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/CompositeNormalizationFunction.java
robinschmid: Factor functions 1 and 2 can be merged because the result is the same. Just factor1*factor2. 
So in the end we only store and use 3 functions. 
But internal standards is a blockage of merging further as it is feature dependent. 

All functions are converted to factors so there is no division happening. So not sure what you refer to.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/MetadataNormalizationConfig.java
robinschmid: Will add that.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/CompositeNormalizationFunction.java
SteffenHeu: i know it's feature dependent, but in the end you still get a factor, so the order does not matter if you do

2* 3 * x * 4 -> 6 * x * 4  or 24 * x. it would only matter if there is some other operation than division or multiplication

robinschmid: all merged now

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/MetadataNormalizationConfig.java
robinschmid: actually removed scaling option

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
robinschmid: scaling option removed

PR #3285, opened 2026-04-17
set initial size of parameter dialogs

SteffenHeu:
with scaling or lower resolutions, some parameter dialogs were too large in height. 
Setting just the height still allows the dialog to be maximized if the user wants

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/EmptyParameterSetupDialogBase.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/EmptyParameterSetupDialogBase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2025 The mzmine Development Team
+ * Copyright (c) 2004-2026 The mzmine Development Team
  *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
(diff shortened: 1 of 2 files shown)


mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/EmptyParameterSetupDialogBase.java
```diff
@@ -159,6 +161,23 @@ protected void parametersChanged() {
     setMinWidth(500.0);
     setMinHeight(400.0);
 
+    showingProperty().subscribe(showing -> {
+      if(!showing) {
+        return;
+      }
+      final Screen primary = Screen.getPrimary();
```
robinschmid: do we need to use the actual screen where the dialog is located? At least I have very different resolution and scaling factors for both screens.
```java
StageWindowSettingsUtil.getCurrentScreen(this);
```

robinschmid (dismissed):

mzmine-community/src/main/java/io/github/mzmine/parameters/dialogs/EmptyParameterSetupDialogBase.java
SteffenHeu: Good question. will check.

robinschmid: maybe you can add a convenience method to actually get the screen as this may return -1 and then just return primary?

PR #3286, opened 2026-04-20
Update ion library selection parameter component

robinschmid:
Components for ion library now looks like this 
image

And when clicking on the button the selection popup looks like this. Selected library on the left and all other available to choose from on the right. Double click, double enter, click outside of popup or X button will close the popup. 

image


javafx-framework/src/main/java/io/github/mzmine/javafx/components/FilterableListView.java
```diff
@@ -165,9 +172,79 @@ private void addKeyEventListener() {
...
+
+    // Allow Up/Down arrows to move list selection while the search field keeps focus.
+    // Also accept Enter on the search field to activate the currently selected item.
+    searchField.addEventFilter(KeyEvent.KEY_PRESSED, event -> {
+      if (event.getCode() == KeyCode.ENTER) {
+        activateSelectedItem();
+//        event.consume();
+      } else if (event.getCode() == KeyCode.DOWN) {
+        moveListSelection(1);
+//        event.consume();
```
SteffenHeu: Why no consume? 
Also this is an event filter not a handler, so it's being handled on the way in, is that correct?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/FilterableListView.java
```diff
@@ -165,9 +172,79 @@ private void addKeyEventListener() {
...
+
+    // Primary mouse click on a non-empty cell activates the item.
+    listView.addEventFilter(MouseEvent.MOUSE_CLICKED, event -> {
+      if (event.getButton() != MouseButton.PRIMARY) {
+        return;
+      }
+      Node node = event.getPickResult().getIntersectedNode();
+      while (node != null && node != listView) {
+        if (node instanceof ListCell<?> cell && !cell.isEmpty()) {
+          activateSelectedItem();
```
SteffenHeu, line 208: this should consume, or? And activation on single click is intentional?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/FilterableListView.java
```diff
@@ -95,10 +99,13 @@ public class FilterableListView<T> extends BorderPane {
   private final ObjectProperty<Consumer<T>> editSelectedAction = new SimpleObjectProperty<>();
   // on create new button
   private final ObjectProperty<Runnable> createNewAction = new SimpleObjectProperty<>();
+  // triggered by Enter on the list or search field, or by a primary mouse click on a non-empty cell
```
SteffenHeu: should be javadoc so it's shown in the context window or?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/FilterableListView.java
SteffenHeu: Double click is better when it triggers an action, or?
When moving the cursor in the list with the arrows it changes the selected item and when clicking once the common behaviour is also to only change the selected item

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/ionidentity/IonLibraryComponentPopoverViewBuilder.java
```diff
@@ -0,0 +1,241 @@
...
+    super(model);
+    this.onCloseRequested = onCloseRequested;
+  }
+
+  @Override
+  public Region build() {
+    final BorderPane root = new BorderPane();
+    root.setPadding(FxLayout.DEFAULT_PADDING_INSETS);
+    root.setPrefWidth(900);
+    root.setPrefHeight(700);
```
SteffenHeu: Maybe smaller depending on the resolution? Like max 1/3 of the actual screen height?

SteffenHeu: Image
Even on my screen this takes up a lot of space. I think we can measure from the button from where it will pop up to the bottom of the screen, or? Then we can set that as pref height and make a minimum height of like 300?

javafx-framework/src/main/java/io/github/mzmine/javafx/components/FilterableListView.java
robinschmid: makes sense, done

javafx-framework/src/main/java/io/github/mzmine/javafx/components/FilterableListView.java
robinschmid: just checked I had the feeling the consume blocked another event. I was trying a couple of things. 
Changed to an EventHandler and consumed events

javafx-framework/src/main/java/io/github/mzmine/javafx/components/FilterableListView.java
robinschmid: Consumed now. 

Up down arrows really just select the item. Usually you skip over multiple until you hit enter key.
Single click or enter key will activate the item. 
Double click or enter key will close popover. 

I prefer the single click. Most people rarely try and use double clicks. 
If they want a different library then they can just click the other library or they can cancel the parameter setup dialog and it returns to the old value.

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/ionidentity/IonLibraryComponentPopoverViewBuilder.java
robinschmid: Not sure I feel like the popover does not really have to stick to the button. The parameter dialogs are all quite different and the button might be somewhere on the screen. 

I would prefer if everything is easy to see because this also acts as a preview of the library if people want to see again what ions are actually included. 

I could use the screen size like you said but I would make the minimum a bit bigger.
We can also fine tune this as we go. 

Made smaller now and to 1/3 of screen size:

image

PR #3289, opened 2026-04-23
DoubleOrAuto parameter, Update other featre selection box

SteffenHeu:
add double or auto parameter
auto select something in the other feature selection pane when the selection changes

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/combowithinput/DoubleOrAutoValue.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/combowithinput/DoubleOrAutoValue.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2004-2026 The mzmine Development Team
+ *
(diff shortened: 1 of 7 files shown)


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/combowithinput/DoubleOrAutoValue.java
```diff
@@ -0,0 +1,64 @@
...
+    return value;
+  }
+
+  @Override
+  public Double getEmbeddedValue() {
+    return manual;
+  }
+
+  public enum DoubleOrAuto implements UniqueIdSupplier {
+    AUTO, MANUAL;
```
robinschmid: Maybe we can extract the existing class from IntOrAuto and just call the enum ValueOrAuto? Or ManualOrAuto?

robinschmid (dismissed):

robinschmid (approved):

PR #3292, opened 2026-04-24
[On rework IIN feature] Global ion library and interfaces 

robinschmid:
- handles the save load and conflict resolution of ion libraries loaded from different sources

Old library loaded that has locally changed:
image


mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/GlobalIonLibrariesInteractor.java
```diff
@@ -59,14 +84,87 @@ protected GlobalIonLibrariesInteractor(GlobalIonLibrariesModel model) {
...
+      return false;
+    }
+
+    if (DialogLoggerUtil.showDialogYesNo("Cascade remove ion building block?", """
+              Do you want to remove all ion types from all libraries that use ion building block:
+              "%s"
+              Otherwise the ion building block will stay defined.
+              
+              This changes:
+              %s""".formatted(def.toString(), matchingLibs.toString()))) {
```
SteffenHeu: I guess the warning should also mention if there is a cloud library in there? (in the future?) Are the ion libraries already tagged as cloud library?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/GlobalIonLibrariesModel.java
```diff
@@ -126,6 +145,12 @@ public IntegerProperty globalIonsVersionProperty() {
   }
 
   public void setGlobalIonsVersion(int globalIonsVersion) {
+    // to track when it was set last
+//    try {
+//      throw new RuntimeException("Setting last model update");
+//    } catch (RuntimeException ex) {
+//      logger.log(Level.WARNING, ex.getMessage(), ex);
+//    }
     this.globalIonsVersion.set(globalIonsVersion);
```
SteffenHeu, line 155: should this only be possible to increment? is a change in any direction fine?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/global/GlobalIonLibraryService.java
```diff
@@ -175,42 +402,140 @@ public void notifyChange() {
...
+        notifyChanges.set(oldNotify);
+      }
       notifyChange();
+      return new ApplyResult.Applied(getVersion());
     }
   }
 
+  /// Snapshot without re-acquiring the read lock (we hold the write lock already).
+  private @NotNull GlobalIonLibraryDTO snapshotWithinLock(int version) {
+    return new GlobalIonLibraryDTO(version, getIonLibrariesUnmodifiable(),
```
SteffenHeu: is this enforceable? like lock.isAcquired or so?

SteffenHeu (commented):
Will test now

SteffenHeu:
image
Has the correlation changed? this is too permissive i think

Also having 2M+Na and M-3H2O+H is a stretch in my opinion without any intermediates like M+H alone
image

[test iin.xml](https://github.com/user-attachments/files/27396224/test.iin.xml)
default comprehensive + was selected from the wizard. maybe it should be default main ions +?

SteffenHeu:
Does it make sense that an "ion" in a library can be not charged?

image

I tried to quickly add 13C and other isotopes as adduct and they are not found (probably because i should have put the +H there as well)
But does it make sense in general that there can be uncharged molecules in a ion library? probably not, or?

I went with the "Define new ion type" dialog. i think there should at least be a hint to say that uncharged "ions" make no sense, especially chemically.

SteffenHeu:
Not for now but for later:
I think it is confusing that there are these two very similar UI elements:
image

Maybe the "Editing ion library" (right) should just be another tab in the "Ion libraries" (left)?

SteffenHeu:
When selecting an ion in the "Define ions" tab, it should automatically populate the text box on the right
image

SteffenHeu:
Then i remove ions and click apply they reapper if they are still used in a library.
1. image
2. 
image
3. 
image

Instead it should ask me to remove the ions from the library, or?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/GlobalIonLibrariesInteractor.java
robinschmid: added a message how many cloud libraries are being changed. 
The libraries have a field origin that says cloud local or builtin. 
Currently there is no cloud implementation

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/fx/GlobalIonLibrariesModel.java
robinschmid: it is only incremented but in the GlobalIonLibraryService.getGlobalLibrary().getVersion() which is used to set it to the model. Or after update the new version is also directly passed

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/global/GlobalIonLibraryService.java
robinschmid: just added a read lock here. The easiest way

robinschmid:
> image Has the correlation changed? this is too permissive i think
> Also having 2M+Na and M-3H2O+H is a stretch in my opinion without any intermediates like M+H alone image
> 
> [test iin.xml](https://github.com/user-attachments/files/27396224/test.iin.xml) default comprehensive + was selected from the wizard. maybe it should be default main ions +?

Correlation of the feature seems fine in your example. No changes there I think. 

The decisions to discard IINs based on rules is implemented in the IonNetworkRefinementTask which now contains a library as main ions. The 2M+Na and 2M+H is currently listed. We have two options:

1. remove more ions from the main ions list 
2. add more filtering options to the refinement something like networks with only 2M (mol>1) and heavily modified neutral loss adducts will be removed.

robinschmid:
Neutral IonType in IonLibrary is possible and handled fine. They are not matched during ion identity. 
IonLibraries with neutral definitions may still be useful in other modules. For example if you want to annotate or search for complex multi neutral loss changes that cannot be captured in just an IonPart but rather in an IonType

Isotopes we still have to add the M modifier to the design as in (M+1) or (M+[13]C2). 
Then an isotope annotation step could just propagate already annotated features to their isotopes like M+H -> (M+1)+H.

Still a few questions open how we want to represent isotopes. But I have some ideas but I talked to people recently and it does not seem like there is a standard notation.

robinschmid:
We could move the Edit library tab into the dashboard. I was just thinking that maybe its good to have them separate as there can be multiple libraries being edited. 
We can also limit this to just one at a time.

robinschmid:
When adding neutral IonType

image

Issue #3279, opened 2026-04-14
[Feature]: cover more input/output paths in the CLI


kaisengit:
### Is your feature request related to a problem?

MZmine's CLI flags handle several paths that are then dynamically adapted when loading a batch:

- ``-i`` / ``--input``: raw data input files
- ``-o`` / ``--output``: output project file (overrides ``ProjectSaveAsModule``)
- ``-l`` / ``--library``: spectral library files
- ``--metadata``: metadata CSV

What is not handled:

- loading a project via `ProjectLoadModule`
- the csv path for the local compound db search

These two are at least the ones I noticed, there might be more.

### Describe the solution you'd like

We would really like the CLI to be consistent with all input/output paths.

Regarding project load, I think the whole `ProjectLoadModule` could be replaced by a new CLI parameter.

For the compound db another path parameter would be necessary.

If there are many path that can be set for the different modules, maybe a central configuration file (e.g. a `yaml`) would even be better suited. 

### Describe alternatives you've considered

Since we are completely working in the CLI environment, this means we automatically edit these paths inside the `.mzbatch` files. Which is cumbersome and prone to errors.

PR #3296, opened 2026-04-30
Allow override of csv database and project import in batch mode

SteffenHeu:
adresses https://github.com/mzmine/mzmine/issues/3279


mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchModeModule.java
```diff
@@ -173,6 +184,18 @@ public static BatchTask runBatchFile(@NotNull MZmineProject project, File batchF
...
+    if (overrideCompDb != null) {
+      if (!newQueue.overrideStepParameter(LocalCSVDatabaseSearchModule.class,
+          LocalCSVDatabaseSearchParameters.dataBaseFile, overrideCompDb)) {
+        return null;
+      }
+    }
+    if (overrideProjectFile != null) {
+      if (!newQueue.overrideStepParameter(ProjectLoadModule.class,
+          ProjectLoaderParameters.projectFile, overrideProjectFile)) {
+        return null;
```
robinschmid, line 196: I think we should log.severe here like for the other cases above?

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
```diff
@@ -362,5 +364,38 @@ public Optional<ParameterSet> findFirst(
...
+   */
+  public <T> boolean overrideStepParameter(@NotNull Class<? extends MZmineProcessingModule> module,
+      @NotNull UserParameter<T, ?> parameter, @NotNull T newValue) {
+    var parameterSets = streamStepParameterSets(module).toList();
+
+    if (parameterSets.size() != 1) {
+      logger.severe("""
+          The parameter %s of module %s cannot be set to value %s because the appears %d times \
+          in the batch. Can only override if it appears exactly once.""".formatted(
+          parameter.getName(), module.getName(), String.valueOf(newValue), parameterSets.size()));
```
robinschmid, line 383: Maybe it is ok to override multiple modules of the same type? 
Like compound database search could be ok multiple times in a batch.
Could maybe just pass the max number of modules as argument

robinschmid (dismissed):
Looks good to me.

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchQueue.java
SteffenHeu: Since this method is used only for the CLI part, i think it would be good to have it tailored for that. If a batch has more than 1 compound db search steps, it probably requires a more complex override scheme than just replacing all steps with a single new database.

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchModeModule.java
SteffenHeu: log is done inside the method if it returns false.

PR #3313, opened 2026-05-18
Ion type parsing, handle and highlight parsing errors

robinschmid:
rename methods to highlight if it throws exception or silent

highlight parsing error in UI
image


robinschmid:
```
    final String prompt = "Format: [2M-H2O+2H]+2 or with charge +(Cu+2)";
    final TextField inputText = newAutoGrowTextField(parsedIonTypeString, prompt, """
            Enter ion types like adducts, in source fragments, and clusters.
            Full format uses M (for molecule), brackets, and charge state: [M+2H]+2 but all three are optional.
            Simple format just uses defined ion building blocks, e.g., -H2O+H results in [M-H2O+H]+
            Define charge of individual parts in (): +(Cu+2)-H equals [M+(Cu+2)-H]+
            Use () to enclose reserved symbols like +- or braces, e.g., +(propan-2-ol)""",
        10, -1);
```

PR #3316, opened 2026-05-19
[on feature branch] Compound dashboard MVC + bind feature tables across dashboards

robinschmid:
- this PR contains feature table bindings between dashboards + the base PR #3317


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundlist/CompoundRowUtils.java
```diff
@@ -0,0 +1,523 @@
...
+  /**
+   * Recursive walk. Recurses into nested compound members first so that an empty nested compound
+   * gets dropped from its parent. Then rewrites {@code compound}'s members, removing both leaf rows
+   * whose id is in {@code detachIds} and nested compounds that just became empty.
+   *
+   * @return true if {@code compound} should be dropped from its parent (became empty after the
+   * rewrite)
+   */
+  private static boolean detachLeavesRecursive(@NotNull final ModularCompoundRow compound,
+      @NotNull final Set<Integer> detachIds, @NotNull final boolean[] changedAccumulator) {
```
SteffenHeu: what is the changedAccumulator for? is a property or atomic value not better than an array?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
```diff
@@ -127,6 +130,21 @@ public boolean getDefaultVisibility() {
             .addTab(new SpectralIdentificationResultsTab(table, s.getClass()));
         case AnalogSpectralLibraryMatchesType a -> MZmineCore.getDesktop()
             .addTab(new SpectralIdentificationResultsTab(table, a.getClass()));
+        case LipidMatchListType _ -> {
+          final LipidAnnotationQCDashboardTab tab = new LipidAnnotationQCDashboardTab();
```
SteffenHeu, line 136: i feel like when opening from a standard feature table, it is totally fine to open this in the same window. a good application for this is when it is opened from another dashboard

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/AbstractBoxPlotType.java
```diff
@@ -112,7 +113,16 @@ public double getPrefColumnWidth() {
 
         tab.getController().groupingColumnProperty().set(selectedColumn);
         tab.getController().abundanceMeasureProperty().set(abundanceMeasure);
-        MZmineCore.getDesktop().addTab(tab);
+        new MZmineWindow().addTab(tab);
+
+        // Bidirectional link so selections sync between source feature table and the new Stats
+        // dashboard's table. The user can disable either direction from the link popover.
+        final FxFeatureTableController sourceCtrl = FxFeatureTableController.controllerFor(table);
```
SteffenHeu, line 127: same here, ony open in new window if this is not a feature table but a dashboard. or make the new window only happen if shift is also pressed or so? We had a reason to move from multiple windows in mzmine 2 to a tab based layout

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/statsdashboard/StatsDashboardController.java
```diff
@@ -69,7 +72,26 @@ public StatsDashboardController() {
...
-    // feature table bindings in view builder
+  }
+
+  /**
+   * Resolve the {@link CompoundRow} that should be displayed for the given selected row. Prefers
+   * the parent compound if the row is a member of one; otherwise returns the row itself when it is
+   * already a compound. Returns {@code null} if neither applies (e.g. a plain feature row that is
+   * not part of any compound).
+   */
+  private static @Nullable CompoundRow resolveCompoundRow(@Nullable FeatureListRow row) {
```
SteffenHeu: private unused method

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_compoundgrouper/CompoundGrouperSubParameters.java
```diff
@@ -0,0 +1,78 @@
...
+      @NotNull final CompoundRepresentativeSelectorOption representativeSelector) {
+    getParameter(COMPONENTIZER).setValue(componentizer);
+    getParameter(REPRESENTATIVE_SELECTOR).setValue(representativeSelector);
+  }
+
+  /**
+   * explicitly apply the default componentizer and representative selector.
+   */
+  public void setAllDefaults() {
+    // explicit defaults so batch wizard output is independent of parameter set defaults
```
SteffenHeu: that should not be the case for the wizard in general, but we usually do it with a static create method that creates a new instance and sets all parameters. in general i like the static pattern more, because it directly takes care of the creation of the parameter set (no additional clone outside the method required)

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_compoundgrouper/SimpleSeederComponentizerModule.java
```diff
@@ -26,11 +27,12 @@ public Class<? extends ParameterSet> getParameterSetClass() {
 
   @Override
   public @NotNull CompoundComponentizerStrategy createStrategy(
-      @NotNull final ParameterSet parameters) {
+      @NotNull final ParameterSet parameters,
+      @NotNull final CompoundRepresentativeSelector representativeSelector) {
```
SteffenHeu, line 31: default rt tol is 0.08 min which is too big. 0.3 min?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/corrgrouping/CorrelateGroupingTask.java
```diff
@@ -293,6 +316,124 @@ Correlation row grouping (metaCorr) has set a minimum number of samples requirem
...
+
+    if (!checkMobilityOverlap(mfa, mfb)) {
+      return -1;
+    }
+
+    // correlate mobilogram
+    return correlate(dataA.getSummedMobilogram(), dataB.getSummedMobilogram());
+  }
+
+  private float correlate(SummedIntensityMobilitySeries a, SummedIntensityMobilitySeries b) {
```
SteffenHeu: we have FeatureCorrelationUtil or not?

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/BaseWizardBatchBuilder.java
```diff
@@ -771,6 +779,20 @@ protected void makeAndAddIinStep(final BatchQueue q) {
...
+   * Compound grouping. Requires Ion Identity Networking and/or Correlation Grouping output as
+   * input.
+   */
+  protected void makeAndAddCompoundGrouperStep(final BatchQueue q) {
+    final CompoundGrouperParameters param = (CompoundGrouperParameters) MZmineCore.getConfiguration()
+        .getModuleParameters(CompoundGrouperModule.class).cloneParameterSet();
+
+    param.setAll(new FeatureListsSelection(FeatureListsSelectionType.BATCH_LAST_FEATURELISTS));
+
+    q.add(new MZmineProcessingStepImpl<>(MZmineCore.getModuleInstance(CompoundGrouperModule.class),
```
SteffenHeu: This needs to set all sub module parameters, especially for the seeder. if i manually set the rt tolerance value once for the module and then create  a batch, it is in the batch

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderFlowInjectLibraryGen.java
```diff
@@ -107,6 +107,9 @@ protected BatchQueue createQueueInternal() {
     // match against own library
     makeAndAddLibrarySearchStep(q, true);
 
+    // compound grouping (requires meta correlation + IIN)
+    makeAndAddCompoundGrouperStep(q);
```
SteffenHeu: Makes no sense for library building with flow inject, or?

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_lipidqc/LipidAnnotationQCDashboardViewBuilder.java
```diff
@@ -260,16 +260,27 @@ private static void selectFirstVisibleRow(@NotNull LipidAnnotationQCDashboardMod
...
+    // is no longer meaningful for this dashboard.
+    final boolean stale =
+        currentRow != null && currentRow.getFeatureList() != model.getFeatureList();
+    if (!stale && currentRow != null) {
+      final boolean inFilter = filteredItems.stream().map(javafx.scene.control.TreeItem::getValue)
+          .filter(Objects::nonNull).anyMatch(row -> row.getID() == currentRow.getID());
+      if (inFilter) {
+        if (!Objects.equals(table.getSelectedRow(), currentRow)) {
+          FeatureTableFXUtil.selectAndScrollTo(currentRow, table);
+        }
```
SteffenHeu: less nested ifs, stick to regular early returns

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurerow4dplot/FeatureRow4DPlotChart.java
```diff
@@ -0,0 +1,251 @@
...
+ * <ul>
+ *   <li>index 0 — the {@link FeatureRow4DPlotDataset} (m/z × RT × color × bubble), rendered with a
+ *       {@link ColoredBubbleDatasetRenderer};</li>
+ *   <li>index 1 — a single-series overlay of the currently selected rows, drawn as outlined
+ *       circles on top of the bubbles.</li>
+ * </ul>
+ * The constructor wires the chart with an empty dataset so the controller can build the view
+ * before data is ready and then push a dataset later via {@link #applyData(FeatureRow4DPlotDataset)}.
+ */
+public class FeatureRow4DPlotChart extends EChartViewer {
```
SteffenHeu, line 45: why is this a new plot class, why not reuse the existing 4d plot? we have region selection and other nice stuff there

mzmine-community/src/main/java/io/github/mzmine/util/FeatureTableFXUtil.java
```diff
@@ -77,48 +77,248 @@ public static void selectAndScrollTo(@Nullable TreeItem<ModularFeatureListRow> r
...
-    VirtualFlow<?> flow = (VirtualFlow<?>) table.getTable().lookup(".virtual-flow");
-    if (flow != null) {
-      final IndexedCell<?> firstCell = flow.getFirstVisibleCell();
-      final IndexedCell<?> lastCell = flow.getLastVisibleCell();
-      if (firstCell != null && lastCell != null && !(itemIndex >= firstCell.getIndex()
-          && itemIndex <= lastCell.getIndex())) {
-        table.scrollTo(itemIndex);
+  /**
+   * Collapse every top-level row in {@code table}'s filtered view except (optionally)
+   * {@code keepExpanded}. Use after navigating to a target row to close compound rows that aren't
```
SteffenHeu: if this is used after navigating, does the table not jump again? you should collapse first and then scroll for the best result, or?

SteffenHeu:
General comments:

- Compound list is not sorted by height (as feature lists are)

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundlist/CompoundRowUtils.java
robinschmid: I thought about it as well. 
I guess the operation does not need to be atomic here but the intent might be clearer then. 
And property is not really good because its not on Fx thread and also does not need all the listener support etc. 

Will change to a AtomicBoolean still.

The return of the method reflects to direct parent und the changeAccumulator stores if any child changed.

robinschmid: done

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/AnnotationSummaryType.java
robinschmid: done, will look for the table owner and open in a new window only if its a another complex dashboard

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/AbstractBoxPlotType.java
robinschmid: done, will look for the table owner and open in a new window only if its a another complex dashboard

I would say that older mzmine versions never had so many different dashboards that work together so multi screen setup is definitely more relevant nowadays

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/statsdashboard/StatsDashboardController.java
robinschmid: good catch. Was fully moved to the feature table

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_compoundgrouper/SimpleSeederComponentizerModule.java
robinschmid: reduced default RT to 0.3 min

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_compoundgrouper/CompoundGrouperSubParameters.java
robinschmid: changed to static setAll method, removed extra methods, now setting all embedded parameters as well.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/BaseWizardBatchBuilder.java
robinschmid: changed to static setAll method, removed extra methods, now setting all embedded parameters as well.

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderFlowInjectLibraryGen.java
robinschmid: I think we might have to use a different algorithm in flow injection / DI that solely uses ion identities? because the correlation does differentiate enough. 
I can remove the module call for now.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/corrgrouping/CorrelateGroupingTask.java
robinschmid: Moved the method over to FeatureCorrelationUtils. 
There was not really a matching method to correlate summed mobilograms. 

A question: 
Is the mobilty axis the same across the whole sample RT and across other samples? Like the start, end, stepsize? 
Or do we need to correlate based on interpolation to the same mobility axis like with MSE?

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_lipidqc/LipidAnnotationQCDashboardViewBuilder.java
robinschmid: unnested if

mzmine-community/src/main/java/io/github/mzmine/modules/tools/batchwizard/builders/WizardBatchBuilderFlowInjectLibraryGen.java
robinschmid: removed from wizard and added comment

robinschmid:
Added the sorting for height in the table.
Also added the default sorting by RT or mz on the setAll like in FeatureList

mzmine-community/src/main/java/io/github/mzmine/util/FeatureTableFXUtil.java
robinschmid: The comment was still old. First design was closing things after selecting/opening the one that was targeted. 
New design already applies the close first then expand, scroll to. 
Select is only done if the row has one unique option.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurerow4dplot/FeatureRow4DPlotChart.java
robinschmid: I looked for existing plots first but none really matched in my opinion. 

- The LipidDashboard KendrickPane 4D plot has a lot of logic and bindings to the LipidDashboard
- KendrickMassPlotAnchorPaneController was also quite specific and old fxml style

Am I missing another one? We can also harmonize this if it makes sense.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_metacorrelate/corrgrouping/CorrelateGroupingTask.java
SteffenHeu: Bruker timstof supports different mobility segments in theory, but I have only seen it once by a user. So normally they should be the same. We have the mobility segments stored in the raw data file.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/featurerow4dplot/FeatureRow4DPlotChart.java
SteffenHeu: Mhhh i had hoped the regular 4d plot would be good. Too bad

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/compoundrowquality/checks/AnnotationAgreementCheck.java
```diff
@@ -0,0 +1,110 @@
...
+  public @NotNull QualityCheckResult evaluate(@NotNull CompoundRow row,
+      @NotNull QualityCheckContext context) {
+    // group by InChIKey first, then SMILES, then formula. keep insertion order so the first
+    // member's annotation appears first in the summary.
+    final Map<String, List<FeatureListRow>> groups = new LinkedHashMap<>();
+    final Map<String, String> displayNameByKey = new LinkedHashMap<>();
+    final List<FeatureListRow> involved = new ArrayList<>();
+
+    for (final CompoundFeatureMember m : row.getCompoundMembers()) {
+      final List<CompoundDBAnnotation> annotations = m.row().getCompoundAnnotations();
```
SteffenHeu: only checks compound annotations? use get/stream allFeatureAnnotations?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/compoundrowquality/checks/AnnotationAgreementCheck.java
```diff
@@ -0,0 +1,110 @@
...
+    final List<String> details = new ArrayList<>(groups.size());
+    for (final Map.Entry<String, List<FeatureListRow>> e : groups.entrySet()) {
+      details.add("%s — %d row%s".formatted(displayNameByKey.get(e.getKey()), e.getValue().size(),
+          e.getValue().size() == 1 ? "" : "s"));
+    }
+    return new DefaultQualityCheckResult(QualityCheckType.ANNOTATION_AGREEMENT, QualityCheckStatus.FAIL,
+        "%d conflicting annotations".formatted(groups.size()), details, involved);
+  }
+
+  private static @Nullable String annotationKey(@NotNull CompoundDBAnnotation a) {
```
SteffenHeu: sounds like a method for CompoundAnnotationUtils, similar to getAnnotationIdentifier

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/compoundrowquality/checks/AnnotationAgreementCheck.java
SteffenHeu: CompoundAnnotationUtils.getTopAnnotationsPerType maybe? so not only compare across rows but also in rows

robinschmid: will change this in by a configuration and parameterset for the compound quality pane

PR #3321, opened 2026-05-25
Fix mzML import crash when optional scan-level CV params are missing or referenced

david-cox-sciex:
Two issues fixed:

1. referenceableParamGroupRef elements inside scanWindow, scan, isolationWindow, selectedIon, and activation were always routed to the spectrum's top-level CV params instead of the correct nested element. This caused scan window limits defined via param group references (as in SCIEX MS Converter output) to be invisible to getScanningMZRange().

2. getDataPointMZRange() dereferenced getMzBinaryDataInfo() after it was already cleared by clearUnusedData(), causing a NullPointerException when both lowest/highest observed m/z CV params and scan window limits were absent.

Added regression tests with minimal mzML files covering both paths.


robinschmid:
Thanks for the commit and thanks for adding small test mzML files. We will check the PR.

robinschmid:
Great. Thanks for the contribution, we tested it and it works as expected.

PR #3322, opened 2026-05-26
Analog search network visualizer

SteffenHeu:
Add analog nodes
add legend
deduplication based on ichi/name/smiles
direct library matches also linked


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/correlation/RowsRelationship.java
```diff
@@ -155,6 +155,13 @@ enum Type {
      */
     MS2Deepscore,
     DREAMS,
+    /**
+     * {@link
+     * io.github.mzmine.modules.dataprocessing.id_spectral_library_analog_search.AnalogSpectralLibrarySearchModule}
+     * Not R2R relationships constructed currently, but we need the enum values for the network
+     * visualizer
+     */
+    ANALOG_COSINE, ANALOG_DREAMS, ANALOG_MS2DEEPSCORE,
```
robinschmid, line 164: maybe we could just use the same edge type or at least the same color like the original DREAMS, COSINE,... edges so that we limit the number of colors, especially as we also have color scaling with the edge score or other properties. 
The nodes are then different color/shape/style and this makes it easy to differentiate. 

Ok I geuss the type is needed so that people can easily remove add the edges in the combo.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/networking/visual/enums/NodeType.java
```diff
@@ -28,7 +28,7 @@
 import java.util.Optional;
 
 public enum NodeType implements ElementType {
-  NEUTRAL_M, ION_FEATURE, SINGLE_FEATURE, NEUTRAL_LOSS_CENTER;
+  NEUTRAL_M, ION_FEATURE, SINGLE_FEATURE, NEUTRAL_LOSS_CENTER, LIB_ANALOG_COMPOUND;
```
robinschmid, line 31: Maybe we could change the shape of the analogs to diamonds, I find those easy to spot between circles.

Also maybe we can update the color scale for the network nodes to the mzmine one.
Currently the analog nodes are kind of in the red orange range of the current color scale.
Maybe we could use a mid gray for them so that they are more like neutral nodes, a bit more like unselected/ghostly then it is clearer they are not really part of the original data.

robinschmid: image

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/AnalogCompoundGrouper.java
```diff
@@ -0,0 +1,312 @@
...
+          original.compoundKey()));
+    }
+    return result;
+  }
+
+  // record every identifier value of the annotation against the given group index; ties go to the
+  // first group seen, matching the deterministic "first wins" assignment for direct matches
+  private static void indexIdentifiers(final SpectralDBAnnotation annotation, final int groupIndex,
+      final Map<IdentifierValue, Integer> index) {
+    for (final CompoundNameIdentifier id : CompoundNameIdentifier.values()) {
```
robinschmid, line 111: Maybe better to limit this to stricter matches? Currently this will also just compare formula and collapse all with the same formula? 
Maybe we make the structure a requirement for the analog matches to actually show up? Or compound name could also be ok. 

Now that we always apply structure harmonization and canonicalization to spectral libraries, maybe it would be enough to check the first block inchikey to group by flat structures? If structure is missing then the compound name has to be the same? 
CAS and formula seem too weak.

What do you think?

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/AnalogCompoundGrouper.java
```diff
@@ -0,0 +1,312 @@
...
+      if (id == CompoundNameIdentifier.INCHI_KEY && normalized.length() >= 14) {
+        index.putIfAbsent(new IdentifierValue(id, normalized.substring(0, 14) + "*"), groupIndex);
+      }
+    }
+  }
+
+  // first identifier (in CompoundNameIdentifier enum order) that hits an existing group's index
+  private static Integer findMatchingGroup(final SpectralDBAnnotation annotation,
+      final Map<IdentifierValue, Integer> index) {
+    for (final CompoundNameIdentifier id : CompoundNameIdentifier.values()) {
```
robinschmid, line 126: same decision here to use only structure inchikey first block or maybe name?
there are a few more places here where CompoundNameIdentifier.values() is streamed

mzmine-community/src/main/resources/themes/graph_network_style.css
```diff
@@ -45,6 +45,23 @@ edge.OTHER {
   fill-mode: dyn-plain;
   fill-color: #ffe519, #ffa63e, #da7e0e, #885313, #492d0d;
 }
+
+edge.ANALOG_MS2Deepscore {
+  size-mode: dyn-size;
+  fill-mode: dyn-plain;
+  fill-color: #19475d;
```
robinschmid, line 52: maybe same color for edges like the score they describe. 
So that DreaMS edges between rows and also to analogs is the same color. 
The node shape/color for analogs already shows its different.  

But can also use different color for those edges.

robinschmid (dismissed):
Only minor things. 
Looks good and works great.

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/AnalogCompoundGrouper.java
SteffenHeu: will remove formula in general form the compound name identifier, cas can stay in my opinion

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/networking/visual/FeatureNetworkGenerator.java
```diff
@@ -519,6 +542,7 @@ private Node getNeutralMolNode(IonNetwork net, boolean createNew) {
         node.setAttribute(NodeAtt.ANNOTATION_SCORE.toString(), scoreForm.format(score));
         node.setAttribute(NodeAtt.EXPLAINED_INTENSITY.toString(),
             scoreForm.format(bestMatch.getSimilarity().getExplainedLibraryIntensity()));
+//        node.setAttribute(NodeAtt.ANNOTATION.toString(), bestMatch.toString());
```
robinschmid, line 545: Why is this removed from neutral mol nodes and compound name and everything else stays?

PR #3326, opened 2026-05-26
Return parsed structure inchi, inchikey, smiles for spectral matches

robinschmid:
often values in the original entries are missing. But the parsed structure is cleaner and has all values

Have to regenerate integrations tests if we want to merge this. 
So maybe best to merge: https://github.com/mzmine/mzmine/pull/3287

auto enrich structure metadata is spectral library entries

remaining decision if default smiles return type should be isomeric or canonical non isomeric. 
It used to be the input smiles which might have been isomeric. Now it is non isomeric but I think its better to provide isomeric by default.


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundannotations/FeatureAnnotation.java
```diff
@@ -169,6 +169,11 @@ void saveToXML(@NotNull XMLStreamWriter writer, ModularFeatureList flist,
 
   @Nullable String getSmiles();
 
+  /**
+   * @return isomeric smiles or the canonical smiles if there is no isomeric smiles
```
SteffenHeu: or null if no smiles exists

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
```diff
@@ -177,7 +178,8 @@ private void exportFeatureList(FeatureList featureList, BufferedWriter writer)
       // Create a list of column DataTypes
       var columns = Stream.of(IDType.class, CompoundNameType.class, IonTypeType.class,
               ScoreType.class, PrecursorMZType.class, MobilityType.class, CCSType.class, RTType.class,
-              FormulaType.class, SmilesStructureType.class, InChIStructureType.class,
+              FormulaType.class, SmilesStructureType.class, SmilesIsomericStructureType.class,
+              InChIStructureType.class,
```
SteffenHeu, line 182: do we need two different smiles as input? if we always assume it is isomeric (we can never be sure anyway) and parse that with the structure parser? would that work?

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralDBEntry.java
```diff
@@ -341,12 +346,24 @@ public MolecularStructure getStructure() {
...
+    enrichMetadata(); // creates new structure if inchi or smiles present
     return structure;
   }
 
+  @Override
+  public void setStructure(final MolecularStructure structure) {
+    if (structure == null) {
+      return;
+    }
+    this.structure = structure;
```
SteffenHeu, line 376: if the input structure is null, should this clear everything? seems unintuitive

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundannotations/FeatureAnnotation.java
robinschmid: added

mzmine-community/src/main/java/io/github/mzmine/modules/io/export_compoundAnnotations_csv/CompoundAnnotationsCSVExportTask.java
robinschmid: This is the annotations export task. Not sure about input. 

Added the isomeric smiles so that we still have the same smiles column as before + isomeric smiles so that we do not break the export format compatibility to other tools, as we just add information and not remove. 

I think it is also nice to have the canonical flat smiles and the isomeric.

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/entry/SpectralDBEntry.java
robinschmid: Same behavior like in CompoundDBAnnotation. 
I think clearing fields might be tough. 
Maybe clearing smiles, isomeric smiles, inchi is ok? 
But entry might only have InchiKey or formula and from those we do not get the structure. So those should definitely be kept. 

Should I set the top 3 to null? I think we could also leave all of them this way. Somehow they should not be null otherwise the call to structure null is weird?

PR #3329, opened 2026-05-28
Chart logics respects plot orientation

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gestures/ChartGestureDragDiffHandler.java
```diff
@@ -111,6 +122,27 @@ public Orientation getOrientation(ChartGestureEvent event) {
     return orient;
   }
 
+  private @Nullable Orientation getAxisOrientation(final @NotNull ChartGestureEvent event) {
```
SteffenHeu, line 129: Is it maybe better to default to the "default" orientation in this method than to return null?

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gestures/ChartGestureDragDiffHandler.java
```diff
@@ -181,4 +205,17 @@ public void accept(ChartGestureEvent event) {
       }
     };
   }
+
+  private double getAxisCoordinate(final @NotNull ChartGestureEvent event,
+      final @NotNull Point2D point, final @NotNull Orientation axisOrientation) {
```
SteffenHeu, line 220: here the orientation is also notnull, and at the callsite the nullabilite of the orientation is not checked

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gestures/ChartGestureDragDiffHandler.java
robinschmid: The idea is that there might be multiple plots and the axis is only present if it was actually clicked/dragged, like if it is the active entity in a plot. This is why the axis orientation might be null.  
The calling method getOrientation actually then falls back to the chart plot orientation if there was no axis targeted by the event.

robinschmid: added javadoc

mzmine-community/src/main/java/io/github/mzmine/gui/chartbasics/gestures/ChartGestureDragDiffHandler.java
robinschmid: checked all the call sites and they use getOrientation() which is NotNull and is either the clicked axis orientation or if a plot is clicked it is the PlotOrientation. 
So never null. 
Only getAxisOrientation might be null but this is handled in getOrientation which is the orientation of the event itself

SteffenHeu: ah sorry i got confused with the mehtods

PR #3363, opened 2026-06-10
Rows filter, by ID, by Compound ID, copy CompoundList

### LINKED ISSUE #3354: [Feature]: Option to save a feature list based on a set of feature IDs
### Is your feature request related to a problem?

Dear mzmine team,

I am currently facing a situation where I would like to subset a feature list based on external information. For example, I might export a feature list, perform an external prioritization step (e.g., RT prediction, PFAS prioritization, etc.), and then obtain a set of prioritized feature IDs.

At the moment, I can subset the feature list by manually pasting a comma-separated list of feature IDs into the filter box at the bottom of the feature list. However, I would like to save this subset as a new feature list so that I can use it with downstream tools such as molecular networking and other analyses.

Would it be possible to add functionality that allows users to create and save a new feature list from a selected set of feature IDs? Alternatively, it would also be very useful to be able to import a CSV file containing feature IDs and use it directly for subsetting.

If this functionality already exists and I have overlooked it, I would greatly appreciate any guidance on how to perform this task.

Many thanks in advance,
Jonathan

### Describe the solution you'd like

Create and save a new feature list from selected feature IDs, allowing the subset to be used directly with downstream tools such as molecular networking.
Import a CSV file containing feature IDs and use it to automatically subset an existing feature list.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

robinschmid:
- first module to copy the CompoundList to the new feature list. Seems to work fine so we could copy this behavior to other modules or even to the general copy feature list method. 
- adds filters for index ranges list of row.getID and compoundID
- fixes #3354


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundlist/CompoundRowUtils.java
```diff
@@ -420,6 +446,56 @@ public static boolean mergeCompoundRows(@NotNull final CompoundList list,
...
+   *
+   * @param compoundRowFilter keep predicate applied to nested compound members, or {@code null} to
+   *                          keep all. The caller applies it to {@code src} itself (top-level
+   *                          rows).
+   * @return the copied compound row, or {@code null} if no member survived (the compound is
+   * dropped)
+   */
+  public static @Nullable ModularCompoundRow copyRowFiltered(@NotNull final ModularCompoundRow src,
+      @NotNull final CompoundList target,
+      @NotNull final Function<FeatureListRow, ModularFeatureListRow> rowMapping,
```
SteffenHeu, line 470: rowMapping should get jdoc

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterTask.java
```diff
@@ -331,12 +357,23 @@ private FeatureList filterFeatureListRows(final FeatureList featureList,
...
+      final boolean rescued =
+          (!allGcEiMS && keepAllWithMS2 && hasMS2) || (keepAnnotated && annotated);
+
+      // The row-ID filter is a strong filter: it respects the keep/remove choice (it is a normal
+      // criterion in isFilterRowCriteriaFailed) and additionally overrides the rescue above. In
+      // keep mode a non-matching id, in remove mode a matching id, blocks the rescue so the row is
+      // removed even when it has MS2 or is annotated.
+      final boolean idFilterBlocksRescue =
+          rowIdRanges != null && rowIdRanges.contains(row.getID()) != removeFailed;
+
```
SteffenHeu: I think this makes the keepAllWithMs2 even less intuitive. It already causes confusion and it overriding some but not all filters makes it even more confusing

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterTask.java
```diff
@@ -365,12 +408,27 @@ private FeatureList filterFeatureListRows(final FeatureList featureList,
...
+          oldToNew.put(row, resetRow);
+        }
       }
     }
 
+    // transfer a remapped copy of the compound list (if any) to the filtered feature list. Members
+    // removed from the feature list are stripped and empty compounds dropped. The compound-ID
+    // filter (if active) is applied to top-level and nested compound rows and respects the
+    // keep/remove choice: keep mode keeps matching compound ids, remove mode keeps non-matching.
+    if (srcCompoundList != null && !srcCompoundRows.isEmpty()) {
```
SteffenHeu, line 449: what happens if the id filter wants to keep a compound but the other filters want to remove a member row?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterTask.java
robinschmid: Maybe have to see how we can make this more obvious. 
I mean the filter by ID only makes sense to me to be the strongest filter. Like if someone says I want IDs 1,2,3 those should always be selected. 
We could also limit the rows filter so if you select IDs to keep or remove that no other option can be active at the same time?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/filter_rowsfilter/RowsFilterTask.java
robinschmid: Yeah maybe we have to only allow either filter by IDs or other filters. So they should be exclusive? 
Because ID filtering seems like the strongest most direct filter right?

SteffenHeu: in that case there is always the option to just run the rows filter with specific ids. All other filters are additive, this one should be too.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/compoundlist/CompoundRowUtils.java
robinschmid: done

PR #3366, opened 2026-06-11
tag potential duplicates in gap-filling

SteffenHeu:
During gap filling it may happen that duplicate features are created for some files due to prior mis-alignment or inconsistencies in feature detection.
This introduces a tag to potentially falsely introduced features. It does not yet provide a cleanup, because i am currently unsure what the best way to do such a cleanup would be, as there may be multiple reasons for it happening in the first place.

image

Equality is checked by the raw mz values, sicne the intensities may be smoothed. Only 50% have to match since the gap filling may be shorter or longer than the true peak.

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/gapfilling/GapFillMzMatchDuplicateType.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/gapfilling/GapFillMzMatchDuplicateType.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2004-2026 The mzmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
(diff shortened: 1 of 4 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/gapfilling/GapFillMzMatchDuplicateType.java
```diff
@@ -0,0 +1,48 @@
...
+public class GapFillMzMatchDuplicateType extends StringType {
+
+  @Override
+  public @NotNull String getUniqueID() {
+    return "gap_fill_mz_duplicate";
+  }
+
+  @Override
+  public @NotNull String getHeaderString() {
+    return "Gap dup (m/z)";
```
robinschmid, line 46: maybe just name it Duplicate gap? or Gap duplicate?

robinschmid (approved):

PR #3379, opened 2026-06-20
Structure similarity networking

robinschmid:
- add a networking edge for structure similarities
- Added Structure similarity (Tanimoto) to Molecular networking module
- 
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/structure_tanimoto/StructureTanimotoNetworkingTask.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/structure_tanimoto/StructureTanimotoNetworkingTask.java
@@ -1,0 +1,212 @@
+/*
+ * Copyright (c) 2004-2026 The mzmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+ * OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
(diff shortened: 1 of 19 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/structure_tanimoto/StructureTanimotoNetworkingTask.java
```diff
@@ -0,0 +1,212 @@
...
+      }
+      final FeatureListRow rowA = rows.get(i);
+      final List<BitSet> fingerprintsA = fingerprintCache.get(rowA);
+      for (int j = i + 1; j < rows.size(); j++) {
+        final FeatureListRow rowB = rows.get(j);
+        final float similarity = TanimotoSimilarity.maxTanimoto(fingerprintsA,
+            fingerprintCache.get(rowB));
+        if (similarity >= minSimilarity) {
+          edges.add(rowA, rowB,
+              new R2RSimpleSimilarity(rowA, rowB, Type.STRUCTURE_TANIMOTO, similarity));
```
SteffenHeu: I feel like the similarity needs to know which two structures/fingerprints caused the similarity. otherwise it is not really traceable and a similarity may be displayed for library match 15 of row a and match 1 for row b and the user has no possibility of knowing

mzmine-community/src/main/java/io/github/mzmine/modules/tools/molecular_similarity/tanimoto/TanimotoSimilarity.java
```diff
@@ -34,23 +34,72 @@
...
+   * @param b fingerprints of the second row's structures
+   * @return the maximum pairwise Tanimoto similarity (0-1)
+   */
+  public static float maxTanimoto(@NotNull final List<BitSet> a, @NotNull final List<BitSet> b) {
+    float max = 0f;
+    for (final BitSet fpa : a) {
+      for (final BitSet fpb : b) {
+        try {
+          final float sim = Tanimoto.calculate(fpa, fpb);
+          if (sim > max) {
```
SteffenHeu: would it make sense to return early if sim == 1f?

mzmine-community/src/main/resources/themes/graph_network_style.css
```diff
@@ -28,7 +28,7 @@ edge.FEATURECORR {
 edge.COSINE {
   size-mode: dyn-size;
   fill-mode: dyn-plain;
-  fill-color: #8be4ff, #008eff, #19475d, #0d3549;
+  fill-color: #8be4ff, #008eff, #1748a2, #0b2698;
```
SteffenHeu, line 31: did you intend to change the colors here?

SteffenHeu (commented):
looks good, just a few questions

mzmine-community/src/main/java/io/github/mzmine/modules/tools/molecular_similarity/tanimoto/TanimotoSimilarity.java
robinschmid: Definitely makes sense.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/group_spectral_networking/structure_tanimoto/StructureTanimotoNetworkingTask.java
robinschmid: I will change to a R2RStructureSimilarity with the fingerprinting and scoring algorithm and the two compared inchiA and inchiB to safe some memory only inchi. I will not put a reference to the FeatureAnnotation because this might be a nightmare to load/save in the project.

mzmine-community/src/main/resources/themes/graph_network_style.css
robinschmid: yeah that was me. All the dark colors were too similar to each other, it was hard to distinguish between the edge types then.

robinschmid:
All done. Also checked the proejct save load

PR #3389, opened 2026-06-30
Search bar in columns list

robinschmid:
image


SteffenHeu:
When i delete the last character the text field loses focus, same after typing the first character. I think it happens when the mouse is above the check list, because it requests focus on change of mouse over.

robinschmid:
Oh did not try interfering with the mouse. 
Good catch. Will try to find a way.

PR #3399, opened 2026-07-10
Wizard option to subset files and preload

robinschmid:
image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_all/ImportSubsetDataMenu.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_all/ImportSubsetDataMenu.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright (c) 2004-2026 The mzmine Development Team
+ *
+ * Permission is hereby granted, free of charge, to any person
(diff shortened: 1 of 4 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/io/import_rawdata_all/ImportSubsetDataMenu.java
```diff
@@ -0,0 +1,179 @@
...
+    final ButtonBase subsetButton = FxIconUtil.newIconButton(FxIcons.FILTER,
+        "Import a random subset of selected files");
+    final IntegerComponent subsetSizeComponent = new IntegerComponent(65, 1, null);
+    subsetSizeComponent.setText(String.valueOf(DEFAULT_SUBSET_SIZE));
+    subsetSizeComponent.setToolTipText("Number of files to import");
+
+    final TextArea subsetTextArea = new TextArea();
+    subsetTextArea.setPrefColumnCount(55);
+    subsetTextArea.setPrefRowCount(10);
+    subsetTextArea.setMinHeight(500);
```
SteffenHeu: I feel like this is too tall by default. I would just remove the min and pref height and rely on the row count

PR #3401, opened 2026-07-11
Loading of legacy ions from project

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/ionidentity/LegacyIonTypeXMLReader.java
```diff
@@ -0,0 +1,101 @@
...
+import io.github.mzmine.modules.io.projectload.version_3_0.CONST;
+import io.github.mzmine.util.ParsingUtils;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import org.jetbrains.annotations.NotNull;
+
+public class LegacyIonTypeXMLReader {
```
SteffenHeu, line 42: It sounds like this class only loads legacy ions, but that is not the case.

Also this is a class having a single method. Then we can also place the logic in the IonType.loadFromXml method.

robinschmid: I added javadoc and renamed the method to say it loads both IonType and legacy. 
The LegacyIonType is package private in a different package than IonType this is why the class is separate.

Issue #3419, opened 2026-08-04
[Bug]: Module TIC/XIC visualizer does not calculate correct m/z for multimers


Adafede:
### mzmine version/build

4.10.6

### Operating system

macOS

### Observed issue behavior

When calculating _m/z_ for multimers (like [2M+H]+), the calculated mass is identical to monomers.

See below

<img width="3736" height="2396" alt="Image" src="https://github.com/user-attachments/assets/ed74d9c6-bf30-41a0-a5cb-6d58936fdfc3" />

<img width="3824" height="2484" alt="Image" src="https://github.com/user-attachments/assets/e6593cb3-e85e-4afc-8280-5952c377881a" />

<img width="3736" height="2396" alt="Image" src="https://github.com/user-attachments/assets/299d05f7-ffad-469a-9b32-7a70c72e06cc" />

<img width="3824" height="2484" alt="Image" src="https://github.com/user-attachments/assets/c070ee27-377b-4448-8279-50fb8125edf6" />

### Steps to reproduce

Enter any formula in the module and choose [M+H]+ and [2M+H]+ to see the same _m/z_

Adafede:
Also, while opening this module I felt like...

_Actually, I would like to tick all of them and see all the traces, maybe?_

robinschmid:
Will close now that the issue is fixed. 
@Adafede I agree it would be nice if users could add multiple traces. Just need to change the parameter type from MZRangeType to a new parameter that allows similar ways to enter mzs and then adds them to a list. 
The "Feature" list below is only for actual detected features from a feature list. So this would not work.

PR #3422, opened 2026-08-04
fix: mass calculation (#3419)

Adafede:
Screenshot 2026-08-04 at 15 52 21


robinschmid (approved):

robinschmid:
Thanks for the contribution. At some point we need to change this to use the new IonType

PR #3423, opened 2026-08-05
Allow custom mzmine_sample_type values

robinschmid:
- added option to manually define mzmine_sample_type
- filter combobox now shows options to select all (also matches yet unknown values) or none
- predefined values qc, blank... are added as well as values found in metadata and user defined values
- used in stats dashboard and in parameters
- sst added as system suitability test
- regex corrected. The old regex did not filter out letters before or after qc or cal because of the [a-z]* leading to chemiCAL to be recognized as cal. 

image

image

image


mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/metadata/SampleTypeFilterParameter.java
```diff
@@ -0,0 +1,196 @@
...
+  public boolean checkValue(Collection<String> errorMessages) {
+    if (!requiresSelection || !value.isEmpty()) {
+      return true;
+    }
+    errorMessages.add("No sample type selected for parameter %s.".formatted(name));
+    return false;
+  }
+
+  @Override
+  public void loadValueFromXML(Element xmlElement) {
```
SteffenHeu, line 156: does this account for loading legacy CheckComboParameters? If not, maybe the better option is to still create a private final CheckComboParameter<SampleType> in the parameter sets that used it and map it to the new parameter instead of doing it here?

SteffenHeu (commented):
looks good, need to check old parameter comapitbility though

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/metadata/SampleTypeFilterParameter.java
robinschmid: yes this should work. 
Also tested here:
https://github.com/mzmine/mzmine/pull/3423/changes#diff-7291f39cd30092ca3fd185ea472163e055e2b8c86b4d1c722a7cbf98e899aa45

The new options also have ALL and NONE, this was not captured in the old style parameter so load save does just add the selected values

robinschmid:
Adjusted to remember last selection

hiding checkboxes when all is selected
image

NONE allows checkboxes and if user starts selecting one NONE is deactivated and changed to list selection
image

PR #3424, opened 2026-08-05
Volcano plot improvements

robinschmid:
- if there were samples loaded but not in the feature list sometimes the plot would not update or load at all
- limit to featureList.getRawDataFiles
- Clear plot if update started
- show message with warnings if wrong metadata selected


mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/MetadataTable.java
```diff
@@ -432,14 +432,30 @@ public static double convertValueToDouble(@Nullable Object value, double stringD
    * @return A list of files associated to the column value or null, if the column value does not
    * exist.
    */
-  public <T> List<RawDataFile> getMatchingFiles(@NotNull MetadataColumn<T> column, @NotNull T value)
+  public <T> List<RawDataFile> getMatchingProjectFiles(@NotNull MetadataColumn<T> column,
```
SteffenHeu, line 439: i'd say the name can stay the same as before, but we can also change it.
Maybe mention in the javadoc that it accounts for all files currently loaded?

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeaturesDataTable.java
```diff
@@ -113,6 +117,10 @@ public FeaturesDataTable(@NotNull List<RawDataFile> dataFiles,
             Arrays.stream(dataRows).map(FeatureListRowAbundances::row).toList());
     this.dataFileIndexMap =
         dataFileIndexMap != null ? dataFileIndexMap : CollectionUtils.indexMapUnordered(dataFiles);
+
+    // unknown keys must not silently map to index 0, see getFeatureIndex and getSampleIndex
+    this.featureRowIndexMap.defaultReturnValue(NO_INDEX);
```
SteffenHeu: The CollectionUtils.indexMapUnordered is only used in this class, so i propose we put the default return value into the method call. Then we are forced to think about that case when using the utils method. i wouldnt have thought about it at the moment.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/UnivariateRowSignificanceTest.java
```diff
@@ -51,13 +50,19 @@ public final class UnivariateRowSignificanceTest<T> implements RowSignificanceTe
 
   public UnivariateRowSignificanceTest(@NotNull FeaturesDataTable dataTable,
       @NotNull SignificanceTests test, MetadataColumn<T> column, T groupA, T groupB) {
-    List<RawDataFile> groupedFilesA = ProjectService.getMetadata().getMatchingFiles(column, groupA);
-    List<RawDataFile> groupedFilesB = ProjectService.getMetadata().getMatchingFiles(column, groupB);
-
+    // limit to actual raw data files to not select files not in feature table
+    final List<RawDataFile> actualRaws = dataTable.getRawDataFiles();
+
+    // this might throw MetadataValueDoesNotExistException
```
SteffenHeu, line 56: so...? either we handle it or we don't do anything with it but just having the comment has no real purpose, or? :D

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/UnivariateRowSignificanceTest.java
```diff
@@ -131,16 +137,6 @@ public T groupB() {
     return groupB;
   }
```
SteffenHeu, line 139: why remove the methods? may still be useful. we can use the raws grom groupA/BData to to retrieve them

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotUpdateTask.java
```diff
@@ -145,14 +176,52 @@ protected void process() {
                 new ColoredXYShapeRenderer(true, ColoredXYShapeRenderer.defaultShape, true)));
       }
     }
+
+    // expected if no features in row in group data files -> no data for row
+    if (missingPValues > 0) {
+      final String msg = """
+          %d of %d features have no p-value and are not shown. This usually means the abundances are constant within both groups - try a different missing value imputation.""".formatted(
```
SteffenHeu: this message does not match the comment. the comment makes more sense?

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
```diff
@@ -134,6 +141,38 @@ public Region build() {
...
+   * nothing to plot at all, the warning replaces the chart instead of overlaying it.
+   */
+  private @NotNull Region createChartPane() {
+    final BooleanBinding hasWarning = model.userWarningProperty().isNotEmpty();
+    final BooleanBinding hasData = Bindings.createBooleanBinding(() -> {
+      final var datasets = model.getDatasets();
+      return datasets != null && !datasets.isEmpty();
+    }, model.datasetsProperty());
+
+    final Label warningLabel = FxLabels.newBoldLabel(model.userWarningProperty());
```
SteffenHeu: what about the XYPlot.setNoDataMessage?

SteffenHeu (commented):
some comments

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/MetadataTable.java
robinschmid: I though this makes it clearer what the difference is. Will add the javadoc.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
robinschmid: oh I did not know this. sounds good.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/UnivariateRowSignificanceTest.java
robinschmid: removed comment added handling in one more place

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/significance/UnivariateRowSignificanceTest.java
robinschmid: was never used. 
Added back with groupAData.getRawDataFiles() which is unmodifiable list

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotUpdateTask.java
robinschmid: Changed the wording. 
Most likely cause is that all samples have no detection in the selected groups and the imputation provides same values. Some people use imputations that add random noise or sample specific noise. Then this would not be true. But so far we dont use those imputation methods. 

%d of %d features have no p-value and are not shown. This usually means the abundances are constant within both groups, for example if there were no detections in the groups and all values are imputed.

mzmine-community/src/main/java/io/github/mzmine/modules/dataanalysis/volcanoplot/VolcanoPlotViewBuilder.java
robinschmid: done

mzmine-community/src/main/java/io/github/mzmine/datamodel/statistics/FeaturesDataTable.java
robinschmid: Good idea. Put the -1 as the default index into the method. much easier.

PR #3425, opened 2026-08-06
Performance improvements for FeatureList.copy by bundling row binding updates e.g. in DuplicatefilterTask

robinschmid:
- row bindings were triggered too often when copying rows and feature lists
- now do it once at featureList.addRow or after changing all features in that row


SteffenHeu (commented):
looks good, can merge after full local integratio ntests run through

PR #3427, opened 2026-08-07
QC RSD columns for area and height and normalized types

robinschmid:
- binding to area or height in features auto calculates the RSD as percentage
- 2 new columns in all integration tests


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/QcAreaRsdType.java
```diff
@@ -0,0 +1,56 @@
...
+import io.github.mzmine.datamodel.features.types.modifiers.MinSamplesRequirement;
+import io.github.mzmine.datamodel.features.types.numbers.abstr.PercentType;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Relative standard deviation (RSD, coefficient of variation) of the {@link AreaType} across all
+ * quality control samples of a feature list. Calculated by {@link QcRsdRowBinding}, which is
+ * created by {@link AreaType#createDefaultRowBindings()}.
+ */
+public class QcAreaRsdType extends PercentType implements MinSamplesRequirement {
```
SteffenHeu: make a mapping type, remove the binding?

robinschmid: mapping type added and also NoDataColumnType as flag for no data column needed

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/preferences/FeatureListPreferences.java
```diff
@@ -0,0 +1,209 @@
...
+ * processing step but influence how values are derived from the feature list, e.g., the sample
+ * types used by {@link AbstractRsdType}.
+ * <p>
+ * A default instance is always created in the feature list constructor and may be replaced by the
+ * "Redefine feature list preferences" module. Preferences are persisted with the project.
+ * <p>
+ * The preferences themselves are immutable, only the internal cache of resolved raw data files is
+ * mutable and lazily populated.
+ */
+public final class FeatureListPreferences {
```
SteffenHeu, line 53: would it make sense to make this a parameter set? In some way or another we need the parameter set anyway and then wwe get the save+load for free

SteffenHeu: For a follow-up PR: Also move the AQS weights in here, or?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/SampleRsdType.java
```diff
@@ -0,0 +1,182 @@
...
+
+  @Override
+  public @NotNull String getUniqueID() {
+    return UNIQUE_ID;
+  }
+
+  @Override
+  public @NotNull String getHeaderString() {
+    // the header only shows the sample type filter, the sub columns carry the titles
+    return "RSD";
```
SteffenHeu, line 76: Since this is the main type and already shows "RSD", do all sub columns have to say RSD as well?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/SampleRsdType.java
```diff
@@ -0,0 +1,182 @@
...
+
+  @Override
+  public Class<ModularDataModel> getValueClass() {
+    return ModularDataModel.class;
+  }
+
+  @Override
+  public boolean getDefaultVisibility() {
+    // the visibility of the sub columns is only applied if this main column is visible
+    return true;
```
SteffenHeu: but since the sub columns are invisible by default, what is the reason of having this one visible?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
```diff
@@ -604,4 +605,31 @@ public String toFullString() {
...
+    return super.getOrDefault(type, defaultValue);
+  }
+
+  @Override
+  public <T> @NotNull T getNonNullElse(DataType<T> type, @NotNull T defaultValue) {
+    if (type instanceof MappingType) {
+      return Objects.requireNonNullElse(get(type), defaultValue);
+    }
+    return super.getNonNullElse(type, defaultValue);
+  }
```
SteffenHeu: should this be moved to the ModularDataModel or ColumnarModularDataModelRow instead?

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
```diff
@@ -812,4 +812,22 @@ public String toString() {
...
+    return super.getOrDefault(type, defaultValue);
+  }
+
+  @Override
+  public <T> @NotNull T getNonNullElse(DataType<T> type, @NotNull T defaultValue) {
+    if (type instanceof MappingType) {
+      return Objects.requireNonNullElse(get(type), defaultValue);
+    }
+    return super.getNonNullElse(type, defaultValue);
+  }
```
SteffenHeu: can be removed if added in ModularDataModel

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/MetadataTable.java
```diff
@@ -88,6 +95,16 @@ public MetadataTable(boolean enableAutoDetection,
...
+   *
+   * @return the current version as modification counter
+   */
+  public long getVersion() {
+    return version.get();
+  }
+
   public Map<MetadataColumn<?>, Map<RawDataFile, Object>> getData() {
     return data;
   }
```
SteffenHeu: With versioning modifications to this may break the versions. I think most of the callsites can circumvent using this method.

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/MetadataTable.java
SteffenHeu: same here, external modification is possible. this is easy to circumvent with a wrapper though

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/SampleRsdType.java
robinschmid: removed from header, kept for unique ID

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/numbers/SampleRsdType.java
robinschmid: good catch. was copy error. default off now

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/preferences/FeatureListPreferences.java
robinschmid: Yes it makes sense to just use the parameters for load save. 
Added a separate parameterset to save and load from xml so that the parameterset of the GUI and the project save load are independent. 

We could add the AQS ranking to both the preferences and the preferences parameters I think this is good. But next PR

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeature.java
robinschmid: documented that get() should handle MappingType. 
Removed all unnecessary overrides. 
Added default into interface for getOrDefault and getNonNullElse

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureListRow.java
robinschmid: removed

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/projectmetadata/table/MetadataTable.java
robinschmid: all metadata operations now internalized and only unmodifiable views leave the class

robinschmid:
Addressed in two larger commits

PR #3432, opened 2026-08-11
Better speed benchmarks and reporting

robinschmid:
- gradle task to run benchmark with fixed memory min and max (currently 32 gb but can be adjusted)
- warm up and real iterations marked
- option to import only once

# Report
- for each step and across steps
- drift analysis to see if later runs throttle because of heat constraints or if jit optimizes the later runs

### DIFF
--- a/mzmine-community/build.gradle
+++ b/mzmine-community/build.gradle
@@ -316,6 +316,48 @@ application {
     ]
 }
 
+/*
+ * Batch speed benchmark, see import_data.speed.BatchSpeedTestMain. Run this instead of starting the
+ * main from the IDE when comparing branches: the heap is fixed here, so every branch is measured with
+ * the same memory. The mzmine preferences (threads, GC after batch step) are not changed, they are
+ * only documented in the results.
+ *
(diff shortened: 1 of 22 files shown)


mzmine-community/build.gradle
```diff
@@ -316,6 +316,58 @@ application {
...
+ * gradlew batchSpeedTest -Pspeed.description=my-branch -Pspeed.batch=D:\some.mzbatch -Pspeed.out=D:\speed
+ * gradlew batchSpeedReport -Pspeed.in=D:\speed.jsonlines
+ *
+ * The heap defaults to 32 GB and can be changed with -Pspeed.heapGb=16. Use the same value for all
+ * branches that are compared.
+ *
+ * The properties are prefixed because plain names like 'description' collide with built in gradle
+ * project properties.
+ */
+def speedTestProperties = ['out', 'description', 'batch', 'iterations', 'warmups', 'inMemory',
```
SteffenHeu: is it possible to put this into a separate file somehow so we dont clutter the build script with test utilities?

robinschmid: moved into its own gradle file and applied in one line. 
Makes sense to keep it cleaner

PR #3433, opened 2026-08-12
Fix lipid ID bug

ansgarkorf:
Ensure multiple fragments are not added to a single signal


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_lipidid/annotation_modules/LipidAnnotationTask.java
```diff
@@ -237,9 +238,10 @@ public void run() {
         }
       }
       if (!possibleRowAnnotations.isEmpty()) {
-        LipidAnnotationUtils.addAnnotationsToFeatureList(row, possibleRowAnnotations,
-            lipidAnalysisType, searchForMSMSFragments, minimumOverallQualityScore, customQcWeights,
-            mzTolerance);
+        // Context-dependent QC scores must not be evaluated while rows are processed in parallel.
+        // Apply the configured threshold after all rows have been annotated and scored.
+        LipidAnnotationUtils.addCandidateAnnotationsToFeatureList(row, possibleRowAnnotations,
```
robinschmid, line 245: one thing we should change as well is the rows.parallelStream().forEach. 
forEach on a parallel stream does not always block the further execution of the calling thread. So better to do something like mapToLong(row -> return some number).sum or count so that the carrier thread will not just continue before work is done.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/id_lipidid/common/identification/fragmentation/LipidFragmentFactory.java
```diff
@@ -295,16 +295,20 @@ private List<LipidFragment> findLipidFragmentFromIonFormula(LipidFragmentationRu
...
+      final @NotNull Scan msMsScan, final @NotNull LipidChainType chainType) {
+    final IMolecularFormula modificationFormula = FormulaUtils.createMajorIsotopeMolFormulaWithCharge(
         rule.getMolecularFormula());
     final List<ILipidChain> chains = LIPID_CHAIN_FACTORY.buildLipidChainsInRange(chainType,
         minChainLength, maxChainLength, minDoubleBonds, maxDoubleBonds, onlySearchForEvenChains);
     final List<LipidFragment> matchedFragments = new ArrayList<>();
     for (final ILipidChain lipidChain : chains) {
       final IMolecularFormula lipidChainFormula = FormulaUtils.cloneFormula(
           lipidChain.getChainMolecularFormula());
+      if (!FormulaUtils.canSubtractFormula(lipidChainFormula, modificationFormula)) {
```
robinschmid: maybe it would be good to change the canSubtract to a subtractOrFail or subtractFormula with Optional<> return type? 
This way it is done in the exact same way. I think there are some special cases we cover with the subtractFormula also looking at the mass number.

PR #3444, opened 2026-08-18
GC-MS: Agilent ChemStation and INFICON HAPSITE importers, headless NIST MSPepSearch search

rvignoneucd:
## What this adds

Native import for two GC-MS formats, and a replacement for the NIST MS Search module that runs headlessly.

### Vendor importers

**Agilent ChemStation** (`.D/DATA.MS`) — streaming reader for the legacy format. m/z is decoded as a big-endian unsigned word in 1/20 mass units, intensity as a 14-bit mantissa plus a 2-bit base-8 exponent. Handles both the `0x01320000` full-file header and the headerless partial LC-MS variant at `0x2F2`. Previously these folders were routed to MSConvert, which targets the newer AcqData layout.

**INFICON HAPSITE** (`.hps`) — reader covering four observed container layouts (native HapsScan, full-scan GPIR, compact GPIR, legacy 1014-byte GPIR). Retention time is recovered by least-squares fitting scan numbers against timestamps in the embedded report, falling back to a per-layout calibration when the report yields fewer than two usable points.

Both are wired into `RawDataFileType`, `RawDataFileTypeDetector`, `ExtensionFilters`, `MSConvertImportTask` and the `AllSpectralDataImportModule` task factories.

### Headless NIST MSPepSearch

Replaces the previous GUI-automation module, which drove `nistms$.exe` through its Windows interface and required the NIST "Automation" option to be enabled. The new module invokes MSPepSearch as a CLI per row and stores hits as spectral library annotations. Licensed `mainlib`/`replib` stay on the user's workstation and are never copied into mzmine.

### Raw Data Overview

A NIST matches tab, compound labels on the chromatogram (off by default, with per-peak visibility and orientation, FMF/RMF filters and a live count), right-click search at a clicked peak apex, and a timeline-style scrub gesture on XY plots.

## Reviewer notes

Flagging the things most likely to matter, including the ones I'd expect pushback on.

**Breaking parameter change.** `NistMsSearchParameters` version moves 3 → 6. Existing batch files containing a NIST step will prompt for reconfiguration; both the library path and the new MSPepSearch directory must be set. `getVersionMessage` explains this to users.

**No vendor binaries are added.** MSPepSearch is not redistributed here. An earlier revision of this branch bundled the runtime; it has been removed and the branch history rewritten, so no `.exe`/`.dll` is present in any commit. Only `external_tools/nist_mspepsearch/README.txt` is tracked, carrying the download link and setup steps; the binaries are gitignored.

Because MSPepSearch is a separate free NIST download and **not** part of a standard NIST MS Search installation, the module exposes a **"MSPepSearch directory"** parameter next to the library picker. Validation failures and the module help both link to https://chemdata.nist.gov/dokuwiki/doku.php?id=peptidew:mspepsearch. A packaged build that ships the runtime under `external_tools/` still resolves it automatically when the parameter is left empty, so a distribution can bundle it without the repository doing so.

**Core GUI coupling — the part I'd most like guidance on.** The NIST matches tab is declared directly in `MainWindow.fxml`, and `MZmineGUI` gained several NIST-specific static accessors. That is module concern living in the shell. It works, but it is not how a module should integrate, and I would rather rework it than have it merged as-is if there is a preferred extension point.

**Known limitation, not a defect.** MSPepSearch's tabular output carries match scores but no peak list, so these annotations have an empty library spectrum and mirror plots show nothing on the library side. The annotation comment records why. Fabricating peaks from the query spectrum would manufacture a perfect self-match, so the arrays stay genuinely empty.

**Unverified against a real MSPepSearch run.** Result parsing keys on `/OUTTAB` columns `Rank`, `MF` and `Name`. If an MSPepSearch build emits different titles, no header is recognized. That case now logs an explicit warning naming the expected columns rather than silently returning zero hits, but I have not confirmed the column titles against output from every MSPepSearch release.

**Behavior change beyond the importers.** `AllSpectralDataImportModule` no longer returns early when every requested file is already loaded. That early return skipped `AllSpectralDataImportMainTask`, and with it metadata import and blank/QC recoloring. `BatchTask.setLastFilesIfAllDataImportStep` already resolves already-loaded files into `BATCH_LAST_FILES` on the normal path, so rerunnable batches still work.

**Base is current.** Rebuilt as a single commit on top of current `master`. The one conflict (`FileNamesComponent.java`, against the wizard file-subsetting change) is resolved: upstream's `buttonBarTopControls` slot and this branch's folder-select button coexist in the same button grid.

## Testing

24 tests across three new test classes, all passing:

- `ChemStationMsParserTest` — 10 tests, including AcqData-precedence and renamed-payload detection
- `HapsiteHpsParserTest` — 7 tests, including one-point and two-point report time fits
- `NistMsSearchTest` — 7 tests

Parser tests build synthetic fixtures; several additionally read real vendor files when `CHEMSTATION_TEST_DIR` or `HAPSITE_TEST_ROOT` is set, and skip otherwise.

```
./gradlew :mzmine-community:test --tests "io.github.mzmine.modules.io.import_rawdata_chemstation.*" --tests "io.github.mzmine.modules.io.import_rawdata_hapsite.*" --tests "io.github.mzmine.modules.dataprocessing.id_nist.*"
```

Compiled and tested on **JDK 26** against current `master` (upstream raised the toolchain from 25 to 26). Verified end to end with a `jpackageImage` build on Windows.

## Platform

NIST MSPepSearch is Windows-only and the module refuses to run elsewhere. Both importers are pure Java and platform-independent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


robinschmid:
Hi, 

thanks for your contribution and for using mzmine. 
Your pull request is very complex and mixes many different parts in one. 
Can you describe what issues you are trying to solve? For example the two formats, is there a way to convert them to mzML or open formats? 

You commited binaries that we cannot check and use for licensing and security reasons. You can always run this code yourself but we cannot include the binaries. 

Best regards

Robin

rvignoneucd:
Hi Robin,

Thank you for the comment! Apologies if I made it overly complicated, yes its a bit much. Here are the takeaways.

I included (but can remove no problem) https://chemdata.nist.gov/dokuwiki/doku.php?id=peptidew:mspepsearch

**"Restrictions and Disclaimers
MSPepSearch program may be redistributed without restriction.

That said, I can pull it and leave instructions where to download it and map that tool.

As for the new format support, we just have 1000s of files we want to be able to process that are .D/chemstation format. It keeps our data processing pipeline cleaner not to carry multiple versions of the same file and if we can read them directly instead of converting everything ill try that.

Hapsite is also a unique format which I had to create a custom importer for.

Ill post an update to this today.

PR #3445, opened 2026-08-18
add test to reproduce structure harmonization error

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/datamodel/structures/StructureParser.java
```diff
@@ -147,25 +151,126 @@ public static boolean containsSmilesSpecialChars(@NotNull String input) {
...
+      return cached;
+    }
+
+    final MolecularStructure parsed = parseStructureWithoutCache(structure, inputType, options);
+    if (parsed == null) {
+      return null;
+    }
+
+    // Derive the clean forms once. Each generator can fail independently. Both smiles flavors are
+    // needed: SmiFlavor.Stereo also changes the traversal, so canonical and isomeric differ even
```
SteffenHeu: What does changes the traversal mean?

mzmine-community/src/main/java/io/github/mzmine/datamodel/structures/StructureParser.java
```diff
@@ -147,25 +151,126 @@ public static boolean containsSmilesSpecialChars(@NotNull String input) {
...
+
+    // decision: keep the derived values instead of discarding them. Generating them for the cache
+    // keys costs roughly 1400 us per structure while an on demand inchiKey() call costs another
+    // ~220 us and formulaString() ~16 us. Since a cached structure is handed to many callers,
+    // storing what was already paid for makes value access about 8x cheaper at zero extra cost.
+    final MolecularStructure mol = new PrecomputedMolecularStructure(parsed.structure(),
+        parsed.formula(), canonicalSmiles, isomericSmiles,
+        inchiStruct != null ? inchiStruct.inchi() : null,
+        inchiStruct != null ? inchiStruct.inchiKey() : null, parsed.monoIsotopicMass(),
+        parsed.mostAbundantMass(), parsed.totalFormalCharge());
```
SteffenHeu: why not simply SimpleMolecularStructre.precomputeValues()?

mzmine-community/src/main/java/io/github/mzmine/datamodel/structures/StructureParser.java
robinschmid: changed to traversal order of atoms and bonds

This means that even if there is no isomeric bond, the output smiles from CANONICAL and ISOMERIC may differ because the traversal order is different

mzmine-community/src/main/java/io/github/mzmine/datamodel/structures/StructureParser.java
robinschmid: yeah makes sense. changed.

PR #3448, opened 2026-08-19
FeatureTable utils refresh all tables weak cache

robinschmid:


mzmine-community/src/main/java/io/github/mzmine/util/FeatureTableFXUtil.java
```diff
@@ -326,16 +337,28 @@ public static void updateCellsForFeatureList(FeatureList flist) {
...
+  /**
+   * @param flist the feature list to look for
+   * @return all existing tables that show this feature list, including tables that are embedded in
+   * a dashboards. Call on the JavaFX thread.
+   */
+  public static @NotNull List<FeatureTableFX> getTablesFor(@Nullable final FeatureList flist) {
+    if (flist == null) {
+      return List.of();
+    }
+    return instances.stream().filter(table -> table.getFeatureList() == flist).toList();
```
SteffenHeu, line 358: not sure how the weak hash map exactly works after a key is GC'ed (if the weak entry exists for a while and may return null here or not).
But WeakHashMap supports null keys, so we should filter for null here anyway, or?

Image

PR #3452, opened 2026-08-19
Checkbox to activate normalized box plots

robinschmid:
image

image

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/AbundanceMeasure.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/AbundanceMeasure.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2004-2026 The mzmine Development Team
+ *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
  * files (the "Software"), to deal in the Software without
@@ -124,6 +125,18 @@ public enum AbundanceMeasure implements UniqueIdSupplier {
     return (float) (value * normalizationFunction.getNormalizationFactor(mz, rt));
   }
 
+
+  /**
+   * @return the normalized counterpart of this raw measure or null if this measure is already
+   * normalized
+   */
+  public @Nullable AbundanceMeasure normalizedVariant() {
+    return switch (this) {
+      case Height -> NORMALIZED_HEIGHT;
+      case Area -> NORMALIZED_AREA;
+      case NORMALIZED_HEIGHT, NORMALIZED_AREA -> null;
+    };
+  }
 
   /**
    * @return true if normalized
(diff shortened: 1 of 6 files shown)


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/fx/MetadataHeaderColumn.java
```diff
@@ -70,8 +78,15 @@ public MetadataHeaderColumn(@NotNull DataType<?> dataType,
 
     final VBox header = FxLayout.newVBox(Pos.CENTER);
     header.setFillWidth(true);
-    final Label title = new Label(dataType.getHeaderString());
-    header.getChildren().add(title);
+
+    final Label title = FxLabels.newLabel(checkBoxSelected.and(checkBoxVisible)
+        .map(normalized -> (normalized ? "Norm. " : "") + dataType.getHeaderString()));
```
SteffenHeu: I really dont like checkboxes or buttons changing their label depending on the state. I never know if the label labels the current state or the one if i change the selection.

why not simply label it "Normalize"? then it's just a simple Yes/No checkbox.

Why a separate label in the first place? Checkboxes have their own text?

mzmine-community/src/main/java/io/github/mzmine/datamodel/AbundanceMeasure.java
```diff
@@ -125,6 +126,18 @@ public float getOrNaN(@Nullable ModularDataModel featureOrRow,
...
+   * @return the normalized counterpart of this raw measure or null if this measure is already
+   * normalized
+   */
+  public @Nullable AbundanceMeasure normalizedVariant() {
+    return switch (this) {
+      case Height -> NORMALIZED_HEIGHT;
+      case Area -> NORMALIZED_AREA;
+      case NORMALIZED_HEIGHT, NORMALIZED_AREA -> null;
+    };
+  }
```
SteffenHeu: i have a similar version in one of my branches, but if the this is called on a normalized variant i simply returned itself again, then we dont have to deal with nullability. the isNormalized is already in place to prefilter

robinschmid: makes sense to return the normalized version then. 
I was also thinking we had something like this.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/fx/MetadataHeaderColumn.java
robinschmid: The checkbox is not visible if the feature table has no normalized area / height. 
Thats why the label and checkbox are split. Could also replace the control completely. 

Should I change it to a combobox then if we have normalzied? 
And then Normalized height box plot / Raw height box plot
We could even leave out the box plot part form the header as the charts already define that. The column menu would still say Height box plot. 

Or instead of checkbox could also use an icon button to toggle and still use the labels Normalized height and Raw height

PR #3464, opened 2026-08-24
Batch fix step duplication

SteffenHeu:
- Remove hashCode and equals implementation of MZmineProcessingStepImpl
  - this fixes not working duplication and subsequent movement in the batch dialog
- Add test for RawDataSavingUtils

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterUtils.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterUtils.java
@@ -214,6 +214,28 @@ public class ParameterUtils {
     return equalValues(a, b, true, true);
   }
 
+  /**
+   * Compares two processing steps by their module and parameter values. Processing-step objects
+   * retain identity equality so equal configurations can coexist in a batch queue.
+   */
+  public static boolean equalValues(@NotNull final MZmineProcessingStep<?> a,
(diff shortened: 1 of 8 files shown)


mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterUtils.java
```diff
@@ -214,6 +214,28 @@ public static boolean equalValues(final ParameterSet a, final ParameterSet b) {
     return equalValues(a, b, true, true);
   }
 
+  /**
+   * Compares two processing steps by their module and parameter values. Processing-step objects
+   * retain identity equality so equal configurations can coexist in a batch queue.
+   */
+  public static boolean equalValues(@NotNull final MZmineProcessingStep<?> a,
+      @NotNull final MZmineProcessingStep<?> b) {
+    return equalValues(a, b, false, false);
```
robinschmid: This might be confusing that equalValues on ParameterSet (just above this method) uses **true, true** and this method has the same name similar signature and uses **false, false**

Maybe make this the default or rename to signal it does not check files and raw data file params.

SteffenHeu: removed, called explicitly

PR #3466, opened 2026-08-25
Main menu racing condition made menu disappear (rare)

robinschmid:


robinschmid:
already fixed upstream

PR #3473, opened 2026-08-28
fix metaboanalyst export for normalized types

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/modules/io/export_features_metaboanalyst/MetaboAnalystExportTask.java
```diff
@@ -221,14 +229,7 @@ private void exportFeatureList(FeatureList featureList, BufferedWriter writer)
...
         if (feature != null) {
-          if(FEATURE_INTENSITY==AbundanceMeasure.Area){
-            final double area = feature.getArea();
-            writer.append(String.valueOf(area));
-          }else{
-            final double height = feature.getHeight();
-            writer.append(String.valueOf(height));
-          }
-
+          writer.append(String.valueOf(abundanceMeasure.getOrNaN((ModularDataModel) feature)));
```
robinschmid, line 232: maybe print empty string instead of NaN? 
At least missing values we keep empty in other exports

robinschmid: Ok null check is done before. 
Just read method name and thought we need to change it

PR #3475, opened 2026-08-31
Standard compound normalizer, use median centered factor

robinschmid:
- uses a median intensity of internal standard across all reference samples to center the normalization factor around 1
- old projects do not report the median abundance of a standard so the old factors remain there


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
```diff
@@ -288,8 +310,20 @@ private void normalizeSamplesBatch(SamplesBatch samplesBatch,
...
+            throw new IllegalStateException(
+                "Resolved internal standards were null. It seems this module does not handle internal standards.");
+          }
+          // reuse the standard rows that were resolved for the whole feature list
+          standardModule.createAllNormalizationFunctionsToSummary(summary, normalizedFeatureList,
+              samplesBatch, metadata, mainParameters, internalStandardParams,
+              internalStandardSelection);
+        } else {
+          throw new IllegalStateException(
+              "Seems like internal standard module %s is not handled.".formatted(
```
SteffenHeu: How about making the InternalStandardSelectingNormalizer a sealed interface, then we can do an exhaustive switch here and dont have to deal with this error message. "Seems like.." sounds like an unprofessional error message that should not reach the user

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/StandardCompoundNormalizationTypeModule.java
```diff
@@ -75,66 +97,157 @@ public List<RawDataFile> getReferenceSamples(@NotNull final FeatureList flist,
...
+   */
+  @Deprecated
   @Override
   public @NotNull Map<@NotNull RawDataFile, @NotNull NormalizationFunction> createReferenceFunctions(
       @NotNull IntensityNormalizationSearchableSummary summary,
       @NotNull final List<@NotNull RawDataFile> referenceFiles,
       @NotNull final ModularFeatureList featureList, @NotNull SamplesBatch samplesBatch,
       @NotNull final MetadataTable metadata, @NotNull final ParameterSet mainParameters,
       @NotNull final ParameterSet moduleSpecificParameters) {
+    throw new UnsupportedOperationException("Use the method with preselected standard selection.");
```
SteffenHeu, line 134: why does this constructor still exist if it just throws an exception?

SteffenHeu:
We've been tinkering with this module so much now, i think it's time for an integration test. Can you do one with the test dataset you used?

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/IntensityNormalizerTask.java
robinschmid: changed message. Sealed now the whole NormalizationTypeModule hierarchy.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/norm_intensity/StandardCompoundNormalizationTypeModule.java
robinschmid: changed inheritance and now there is only one method that has this unsupported operation. 
This method is from the interface itself so no chance of removing it

robinschmid:
Changed the workshop integration test with better isotope finder settings so that the other PR will also have more to test. 
And also added intensity normalization by metadata file, internal standards, and inter sample. 

For this I needed to change all input files like the internal standards.tsv and I just search for it at the input file and in the directory of the batch.

PR #3482, opened 2026-09-02
Change export behavior to overwrite existing files

animesh:


robinschmid:
Hi, Thanks for your contribution. 
Can you describe why you added this change?

animesh:
When i tried to overwrite already exported MGF, it keeps adding spectra, thus the patch.

robinschmid:
We have many users who rely on adding spectra to an existing mgf. 
You could refactor this change and add a parameter around the WriterOptions enum. This enum needs to then implement UniqueIdSupplier and the default needs to be the old behavior to keep compatibility to old batches. 
In the parameters you should then implement the handleLoadedParameters method and handle old versions where the new parameter did not exist to set the value to its default.

robinschmid:
This way the user can define the behavior

animesh:
thanks for the suggestion @robinschmid ,sounds much better 💯 i have tried to add a drop-down  to append/overwrite https://github.com/mzmine/mzmine/compare/master...animesh:mzmine:master?expand=1 see if that looks fine?

PR #3488, opened 2026-09-07
add synonyms type

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/SynonymsType.java
```diff
@@ -0,0 +1,42 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package io.github.mzmine.datamodel.features.types.annotations;
+
+import io.github.mzmine.datamodel.features.types.abstr.StringType;
+import org.jetbrains.annotations.NotNull;
+
+public class SynonymsType extends StringType {
```
robinschmid, line 31: maybe we should make this a ListType<String> 
We never fully defined how we handle synonyms but at least in the mzmine json we handle them as a List<String>. 

Maybe we should also add a case to DBEntryField.convertValue where we parse a potential String to List

SteffenHeu: which separator would you want to use? ;?

robinschmid: Where?
The ListType could just show one or could show number of synonyms? 
And then in the overview in the database match/spectral match we show top 5 synonyms and maybe a button or something next to the synonyms: label to open the rest of the synonyms? 

Sometimes there are so many long synonyms that this would easily overflow. 
No separator alone is quite good for this. I think ; might be best.

SteffenHeu: During parsing. If it is a list, it has to be parsed from string. so we need to split while loading from a compound database and from a spectral database

robinschmid: in the compound db I think the best would be to use ; separator and use the csv reading function where quotes around the string are allowing ; in the value like

value 1;"valueWith;"

This will at least allow quotes and ; in the text

robinschmid: msp uses multiple. Synon: so multiple rows

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/annotations/SynonymsType.java
```diff
@@ -0,0 +1,118 @@
...
+   * Accepts JSON arrays and semicolon-separated names with CSV quoting. Commas belong to names.
+   */
+  public static @Nullable List<String> parse(@Nullable final String text) {
+    if (text == null) {
+      return null;
+    }
+    if (text.isBlank()) {
+      return List.of();
+    }
+    if (text.stripLeading().startsWith("[")) {
```
robinschmid: maybe do text = text.strip() to remove any trailing leading whitespace

mzmine-community/src/main/java/io/github/mzmine/util/spectraldb/parser/NistMspParser.java
```diff
@@ -225,7 +226,15 @@ private void extractMetaData(LibraryParsingErrors errors, Map<DBEntryField, Obje
             value = field.convertValue(content);
           }
           if (value != null) {
-            fields.put(field, value);
+            if (field == DBEntryField.SYNONYMS && fields.get(field) instanceof List<?> previous
+                && value instanceof List<?> synonyms) {
+              // MSP stores additional synonyms on repeated Synon lines.
+              final List<Object> combined = new ArrayList<>(previous);
```
robinschmid: List<String> and then just add all entry by o.toString()? They should already be strings before that

robinschmid (changes requested):

robinschmid:
I will just apply the small changes. And merge. 
We can also change it again if needed.

robinschmid (approved):

PR #3497, opened 2026-09-10
Log batch speed measurements after every step and on failure

robinschmid:
- fixes that the error message in batch always said error in step x+1/19 so error in step 10 would print 11/19
- Logs speed after every step in case mzmine exits
- Logs full speed report also after an error or cancel to get at least some info what took long

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchTask.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchTask.java
@@ -74,6 +74,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Level;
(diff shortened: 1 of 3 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/BatchTask.java
```diff
@@ -308,12 +341,15 @@ private void runBatchQueue() {
...
-              storageBefore, storageAfter));
+      final String stepName = queue.get(stepNumber).getModule().getName();
+      final StepTimeMeasurement stepTime = new StepTimeMeasurement(stepNumber + 1, stepName,
+          duration, runGCafterBatchStep);
+      final StepStorageMeasurement stepStorage = new StepStorageMeasurement(stepNumber + 1,
+          stepName, storageBefore, storageAfter);
+      stepTimes.add(stepTime);
+      stepStorageStats.add(stepStorage);
+      // log each finished step right away so the measurements survive a failing or canceled batch
+      logStepMeasurement(new StepMeasurement(stepTime, stepStorage));
```
SteffenHeu, line 352: but now the full thing is always logged via finishBatchMeasurements. Is it necessary to log a single measurement then?

robinschmid: Idea was to still log it in case mzmine drops, for example because of out of memory or other stopping the process

PR #3359, opened 2026-06-08
isotope finder new algorithms

robinschmid:
- finished and cleaned up
- best to squash merge

PR #3498, opened 2026-09-10
[performance]: better isotope utils

SpeckiJ:
Fixes two small issues:

- Google Range() is not suited for usage in hot-path of the code, i.e. when iterating over huge lists. Complexity of the self-rolled solution is also low. This should bring a ~30% Performance improvement.

- JSONReader creation causes a full classscan on each factory initialization,


robinschmid:
Need to wait for this to merge and then apply the same optimization:
#3359

PR #3501, opened 2026-09-14
add SimpleFloatRange + migrate Gap.java to use it

SpeckiJ:
faster version of Guavas Range, analogous to the existing SimpleDoubleRange/SimpleIntegerRange implementations.

Avoids boxing/unboxing and allows more jitting/inlining magic.


robinschmid (approved):
Looks good to me

PR #3503, opened 2026-09-14
Save IonNetwork

robinschmid:
- Ion network relations removed: not used much and made load save harder, goes against the compound dashboard idea
- relations module was removed, loading old batches will give warning and explanation but skip step 
- IonNetwork is now immutable and BuildingIonNetwork allows modification but is converted to immutable afterward 
- IonNetwork and IonIdentity are saved to an extra xml file in project
- wizard now sets user defined handle old feature list at metaCorr step


mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/ModuleMappingUtils.java
```diff
@@ -45,4 +46,24 @@ public static Map<String, String> getOldModuleNamesMap() {
...
+   * warning instead of failing the whole batch file, because the step cannot be recreated.
+   *
+   * @return removed processing modules, keyed by the class name written to batch files
+   */
+  public static Map<String, RemovedModule> getRemovedModules() {
+    final List<RemovedModule> removed = List.of(//
+        new RemovedModule(
+            "io.github.mzmine.modules.dataprocessing.id_ion_identity_networking.relations.IonNetRelationsModule",
+            "Relations between ion identity networks",
+            "Removed as it was rarely used and made load/save of ions impossible. All other ion identity networking steps are unaffected."));
```
robinschmid, line 60: Removing modules was not covered so far. 
I thought this might be a nice way when an old batch with missing module is loaded to show a user readable message.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/project_io/IonNetworksLoader.java
```diff
@@ -0,0 +1,255 @@
...
+    final String json = reader.getElementText().trim();
+    if (json.isEmpty()) {
+      return List.of();
+    }
+    final IonLibrary library = IonLibraryIO.loadFromJson(json).library();
+    // IonLibraryIO may return an already known library whose ions are ordered differently
+    return IonNetworkXml.canonicalIonTypeOrder(library.ions());
+  }
+
+  private static void readNetwork(@NotNull final XMLStreamReader reader,
```
SteffenHeu, line 125: this reads formulas or?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/project_io/IonNetworksSaver.java
```diff
@@ -0,0 +1,246 @@
...
+   * loader derives the very same order. Instead of trusting that, push the library through the
+   * loader path once and compare: a mismatch would silently give rows the wrong ion type, which is
+   * far worse than failing the save.
+   */
+  private static void verifyIndicesSurviveReload(@NotNull final String libraryJson,
+      @NotNull final List<IonType> indexedTypes) {
+    final List<IonType> reloaded = IonNetworkXml.canonicalIonTypeOrder(
+        IonLibraryIO.loadFromJson(libraryJson).library().ions());
+    if (!reloaded.equals(indexedTypes)) {
+      throw new IllegalStateException("""
```
SteffenHeu: what happens in this case? does project save fail or does it save without ion networks?

mzmine-community/src/main/java/io/github/mzmine/modules/batchmode/ModuleMappingUtils.java
SteffenHeu: We already have PlaceholderModule?

robinschmid: Loading an old batch will now display a readable message:
image

And before the message used the package.className and there was no way to guide the user or explain why or what else to use. 
image


Placeholder module is just for loading projects right?

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/project_io/IonNetworksLoader.java
robinschmid: updated the jdoc and added a record for the network in case we add more fields in the future this makes it clearer. This reads the network ID and all other network related objects like the consensus formulas.

mzmine-community/src/main/java/io/github/mzmine/datamodel/identities/iontype/project_io/IonNetworksSaver.java
robinschmid: Yeah I was also not 100% certain about this one. 
I wanted to reuse the ion library saving as this uses ids for the ion parts and does not repeat the whole information all the time. But the sorting might throw off the results. 
I changed it now so that the regular ion library reader also provides a map of the encountered input order so that the IonType ids in the ion networks never point to the wrong index in the list.

PR #3505, opened 2026-09-16
Make FWHM more robust

robinschmid:
Current version extrapolates FWHM as 2.8547 because first data point is not used and extrapolation goes to far. 

image

- **Search direction**: from each outer edge *inwards* to the apex, stop at the first point ≥ threshold, interpolate against its outer neighbour. Never extrapolates, and interior valleys or zero-intensity dropouts are passed over — the trace is one peak.
- **Both flanks cross 50%** → measured width.
- **One flank crosses** → missing half-width = `max(crossing half-width, observed half-width on the missing side)`.
- **Neither crosses** → extrapolate, with two guards: slope from a least-squares fit over the *whole* flank (not the edge segment), and half-width capped at **2×** the observed half-width.
- **Flank never below 85% of apex** → no extrapolation at all, use the data edge. Too flat to carry width information.
- **TF/AF (5%/10%)** share the search but get **no fallback** — a missing flank is NaN, because mirroring would force the factor to 1.0 and fake a symmetric peak.

Invariant: FWHM ≤ 2× the observed peak width.

**Fixed from the old design**

- First and last data point were never candidates; when a side found none, it fell back to the full feature RT range.
- Picked the *globally* nearest point to the threshold instead of the first crossing → in split peaks the half-height point could land on the far sub-peak's flank. Worse, the result flipped with valley depth: 60% valley → 0.196, 75% valley → 0.366.
- Interpolated between that point and its neighbour toward the apex without checking the pair brackets the threshold → silent extrapolation whenever a flank stayed above 50%, unbounded on flat flanks (measured 1.40 on a feature 0.60 wide).
- Both flanks above threshold → both extrapolated, arbitrary result. Only guards were `fwhm <= 0`, NaN, Inf. 
- Dead code: the "straight drop to 0" branches were unreachable


mzmine-community/src/main/java/io/github/mzmine/modules/tools/qualityparameters/QualityParameters.java
```diff
@@ -25,371 +25,385 @@
...
-      }
-
-      if (dps.getNumberOfValues() < 3) {
+    flist.streamFeatures().forEach(feature -> {
+      final IonTimeSeries<? extends Scan> series = feature.getFeatureData();
+      if (!isInitialized(feature) || series == null || series.getNumberOfValues() < 3) {
         return;
       }
+      final double[] rts = extractX(series);
+      final double[] intensities = extractIntensities(series);
```
SteffenHeu: ScanDataAccess? Will need to limit all methods by an end index then

robinschmid: Good catch, the old code did the same but it does not cover all cases. I removed all use of the double buffer/memory segment methods here and extract the array instead. That is handled in all classes the same way. 
Otherwise I guess I would have to use functions like extractX and loop over the elements.

PR #3509, opened 2026-09-18
Sort library matches after every module call

robinschmid:
- change sorting to AQS instead of MS2 score
- Apply sorting after every module call so that spectral matches and nist matches are sorted together


robinschmid:
The tasks all already sorted. 
Only idea was to change to AQS sorting by default like done in the annotations csv export but we keep MS2 score sorting here

PR #3510, opened 2026-09-18
update feature list summary

SteffenHeu:


mzmine-community/src/main/java/io/github/mzmine/gui/mainwindow/FeatureListSummary.fxml
```diff
@@ -1,81 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
robinschmid, line 1: nice, no fxml anymore

PR #3511, opened 2026-09-18
CSV auto detect format and separator

robinschmid:
- added separator parameter with auto option for read
- auto detects the charset
- auto detects the separator by reading a few lines and finding consistent splits
- uses the default csv format with escaping quotes and separator in value but also has fallback to try without quotes
image


SteffenHeu (commented):
could also add a file that has a BOM header (byte order mark) 
did have a lot of struggle with that in the past

robinschmid:
Should be already there:
image

Definitely have to not let the IDE auto fix things when commiting

PR #3512, opened 2026-09-18
Integration dashboard

SteffenHeu:
add applied method for manual integration
fix ims reintegration

### DIFF
--- a/mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/IonTimeSeriesUtils.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/datamodel/featuredata/IonTimeSeriesUtils.java
@@ -1,6 +1,5 @@
 /*
- * Copyright (c) 2004-2024 The mzmine Development Team
- *
+ * Copyright (c) 2004-2026 The mzmine Development Team
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
  * files (the "Software"), to deal in the Software without
@@ -281,7 +280,7 @@ public class IonTimeSeriesUtils {
       final List<MobilityScan> scans = new ArrayList<>();
       while (access.hasNextMobilityScan()) {
         MobilityScan scan = access.nextMobilityScan();
-        if (mobilityRange != null || mobilityRange.contains((float) scan.getMobility())) {
+        if (mobilityRange == null || mobilityRange.contains((float) scan.getMobility())) {
           final int closestPeakIndex = access.binarySearch(centerMz, DefaultTo.CLOSEST_VALUE);
           if (closestPeakIndex < 0) {
             // empty scan
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/FeatureIntegrationDataCalcTask.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/FeatureIntegrationDataCalcTask.java
@@ -1,6 +1,5 @@
 /*
- * Copyright (c) 2004-2025 The mzmine Development Team
- *
+ * Copyright (c) 2004-2026 The mzmine Development Team
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
  * files (the "Software"), to deal in the Software without
@@ -141,7 +140,7 @@ class FeatureIntegrationDataCalcTask extends FxUpdateTask<IntegrationDashboardMo
       return IonTimeSeries.EMPTY;
     }
 
-    if (file instanceof IMSRawDataFile ims && (FeatureUtils.isMrm(feature))) {
+    if (file instanceof IMSRawDataFile ims && (FeatureUtils.isImsFeature(feature))) {
       final int previousBinningWith = BinningMobilogramDataAccess.getPreviousBinningWidth(flist,
           ims.getMobilityType());
       var chrom = IonTimeSeriesUtils.extractIonMobilogramTimeSeries(
--- a/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/otherdetectors/integrationplot/IntegrationPlotController.java
+++ b/mzmine-community/src/main/java/io/github/mzmine/modules/visualization/otherdetectors/integrationplot/IntegrationPlotController.java
@@ -1,6 +1,5 @@
 /*
- * Copyright (c) 2004-2024 The mzmine Development Team
- *
+ * Copyright (c) 2004-2026 The mzmine Development Team
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
  * files (the "Software"), to deal in the Software without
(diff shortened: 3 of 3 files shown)


mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_manualintegration/FeatureRecord.java
```diff
@@ -0,0 +1,51 @@
...
+ * Identifies a {@link FeatureListRow} for reproducible manual integration. The {@link #rowId()} is
+ * the primary key; {@link #mz()}, {@link #rt()} and {@link #mobility()} serve as a fallback match
+ * when row IDs have shifted (e.g. after re-alignment).
+ *
+ * @param rowId    the row id at capture time
+ * @param mz       the average m/z of the row
+ * @param rt       the average retention time of the row
+ * @param mobility the average mobility of the row, or null for non-IMS data
+ */
+public record FeatureRecord(int rowId, double mz, float rt, @Nullable Float mobility) {
```
robinschmid, line 42: maybe RT should be @Nullable Float because in theory its not required. 
So far there is so many places where we use RT as if its NotNull but maybe we change that in the future.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_manualintegration/ManualIntegrationEntry.java
```diff
@@ -0,0 +1,52 @@
...
+ * @param mzRange       the m/z window used to extract the chromatogram
+ * @param rtRange       the manual RT integration range; null only when {@code deleted} is true
+ * @param mobilityRange the mobility window for IMS extraction, or null for non-IMS data
+ * @param deleted       true if the feature was removed (no integration); replay removes the
+ *                      feature
+ */
+public record ManualIntegrationEntry(@NotNull FeatureRecord feature,
+                                     @NotNull RawDataFilePlaceholder rawFile,
+                                     @NotNull Range<Double> mzRange, @Nullable Range<Float> rtRange,
+                                     @Nullable Range<Float> mobilityRange, boolean deleted) {
```
robinschmid, line 50: We now have pretty good replacement for guava ranges. 
I think we should slowly start removing those and not use them in new places. 
We just removed them from some parts of the gap filler and the memory bar is not racing anymore. 
Only thing is those ranges are currently Range.closed() only, which is fine here. But if we need it we can add other variants as well. 

Just tried if the replacement works easily and it should

```java
public record ManualIntegrationEntry(@NotNull FeatureRecord feature,
                                     @NotNull RawDataFilePlaceholder rawFile,
                                     @NotNull SimpleDoubleRange mzRange,
                                     @Nullable SimpleFloatRange rtRange,
                                     @Nullable SimpleFloatRange mobilityRange, boolean deleted) {

  public ManualIntegrationEntry(@NotNull FeatureRecord feature,
      @NotNull RawDataFilePlaceholder rawFile, @NotNull Range<Double> mzRange,
      @Nullable Range<Float> rtRange, @Nullable Range<Float> mobilityRange, boolean deleted) {
    this(feature, rawFile, SimpleRange.ofDouble(mzRange), SimpleRange.ofFloat(rtRange),
        SimpleRange.ofFloat(mobilityRange), deleted);
  }
}
```

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_manualintegration/ManualIntegrationEntriesParameter.java
```diff
@@ -0,0 +1,238 @@
...
+    this("Manual integrations",
+        "The manual (re-)integrations captured in the integration dashboard.");
+  }
+
+  public ManualIntegrationEntriesParameter(@NotNull String name, @NotNull String description) {
+    this.name = name;
+    this.description = description;
+  }
+
+  private static void appendDoubleRange(@NotNull Element parent, @NotNull String tag,
```
robinschmid, line 77: does it make sense to put these into XMLUtils? 
The parse ranges functions there use nested elements for min and max. 
I like the version here with attributes more I have to say. 
maybe somthing like appendDoubleRangeV2 or something like that.

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_manualintegration/ManualIntegrationEntriesParameter.java
```diff
@@ -0,0 +1,238 @@
...
+
+  @Override
+  public void saveValueToXML(Element xmlElement) {
+    final Document doc = xmlElement.getOwnerDocument();
+    for (final ManualIntegrationEntry entry : value) {
+      final Element entryElement = doc.createElement(ENTRY_ELEMENT);
+      final FeatureRecord feature = entry.feature();
+      entryElement.setAttribute(ROW_ID_ATTR, String.valueOf(feature.rowId()));
+      entryElement.setAttribute(ROW_MZ_ATTR, String.valueOf(feature.mz()));
+      entryElement.setAttribute(ROW_RT_ATTR, String.valueOf(feature.rt()));
```
robinschmid, line 152: in case feature.rt is now nullable, then guard null

mzmine-community/src/main/java/io/github/mzmine/modules/visualization/dash_integration/FeatureIntegrationData.java
```diff
@@ -37,10 +39,13 @@
...
  * @param additionalData additional chromatograms, e.g. from mrm traces. May be empty
+ * @param mzRange        the m/z window used to extract the chromatogram. Needed to reproduce a
+ *                       manual integration.
  */
 public record FeatureIntegrationData(@NotNull RawDataFile file,
                                      @Nullable IntensityTimeSeries feature,
                                      @NotNull IntensityTimeSeries chromatogram,
-                                     @NotNull List<IntensityTimeSeriesToXYProvider> additionalData) {
+                                     @NotNull List<IntensityTimeSeriesToXYProvider> additionalData,
+                                     @NotNull Range<Double> mzRange) {
```
robinschmid, line 49: maybe change to SimpleDoubleRange

robinschmid:
So the new step cannot be applied in a batch?
Is ok if this is the case. Was just thinking that this was more than just reporting/tracking changes - which is fine. Just asking. 

image

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_manualintegration/FeatureRecord.java
SteffenHeu: but if we don't have RT, what would we integrate? In this case it makes sense to have it notnull

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_manualintegration/ManualIntegrationEntriesParameter.java
SteffenHeu: named them appendSimpleXYRange

mzmine-community/src/main/java/io/github/mzmine/modules/dataprocessing/featdet_manualintegration/FeatureRecord.java
robinschmid: maybe we would integrate mobility?

PR #3516, opened 2026-09-21
Sort compound preferred row by ion type ranking

robinschmid:
- this is not the default sorting but the secondary option for defining the representative compound row
- now uses: ion ranking, then AQS, then intensity 
- the default option uses: AQS first


SteffenHeu (dismissed):

mzmine-community/src/main/java/io/github/mzmine/parameters/impl/SimpleParameterSet.java
```diff
@@ -118,6 +125,19 @@ public void setSkipSensitiveParameters(boolean skipSensitiveParameters) {
...
   }
 
+  @Override
+  public void setAsBatchStepParameters() {
+    batchStepParameters = true;
+  }
+
+  /**
+   * @return true if this set is the stored configuration of a batch step, which keeps the selection
+   * of the user instead of starting over whenever the setup dialog is opened
```
SteffenHeu: î dont geht this jdoc. a parameter set in a batch always keeps it's values and does not start over

robinschmid: Oh yeah I moved it from the preferences parameters to the SimpleParameterSet and did not update the doc.

mzmine-community/src/main/java/io/github/mzmine/parameters/ParameterSet.java
```diff
@@ -260,6 +260,18 @@ default IonMobilitySupport getIonMobilitySupport() {
 
   ExitCode showSetupDialog(boolean valueCheckRequired);
 
+  /**
+   * Called when this parameter set becomes the configuration of a batch queue step, either by
+   * adding the step, by loading a queue, or by copying a step. Parameter sets that start over
+   * whenever their setup dialog is opened use this to show the selection of the user instead.
+   */
+  void setAsBatchStepParameters();
```
SteffenHeu, line 272: Ah ok i get it. can you add a link to FeatureListPreferencesParameters#showSetupDialog?

robinschmid: done

PR #3523, opened 2026-09-21
Copy feature list also copies R2RMaps and IonNetwork to redirect relationship references

robinschmid:
- With this feature list copies keep their R2RMaps and IonNetworks like after applying a rows filter or extracting from a PCA plot. 
- also fixes an issue with overlapping regions in volcano and pca plot where rows were extracted multiple times.


mzmine-community/src/main/java/io/github/mzmine/util/FeatureListUtils.java
```diff
@@ -889,21 +897,142 @@ public static ModularFeatureList createCopy(final FeatureList featureList,
...
+   * {@code source} and {@code target} may be the same feature list, for a module that filters rows
+   * in place: removing rows or renumbering their IDs invalidates the relationship map keys just the
+   * same. The relationship maps of the target are therefore replaced, not merged.
+   *
+   * @param source     the feature list the rows were copied from
+   * @param target     the feature list holding the copied rows, may be {@code source} itself
+   * @param rowMapping maps a source row to its row in the target, or to null if that row is gone.
+   *                   Relations of rows that are not mapped are dropped.
+   */
+  public static void transferRowRelationsAndIIN(@NotNull FeatureList source,
```
SteffenHeu, line 1014: same javadoc as the other method, or? can we just use `@see` or `{@link}`?

PR #3526, opened 2026-09-22
Tags

SteffenHeu:
image


mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
```diff
@@ -147,7 +151,8 @@ public class ModularFeatureList implements FeatureList {
    * User defined preferences, never null, defaults to
    * {@link FeatureListPreferences#createDefault()}
    */
-  private volatile @NotNull FeatureListPreferences preferences = FeatureListPreferences.createDefault();
+  private final @NotNull ObjectProperty<@NotNull FeatureListPreferences> preferences = new SimpleObjectProperty<>(
+      FeatureListPreferences.createDefault());
```
robinschmid, line 155: maybe best to not make this a property? Because usually properties should only be set on javafx thread. 
I know we also have other ObservableLists above but I think it might be easier to not use javafx so much in the main objects

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
```diff
@@ -214,7 +219,7 @@ public ModularFeatureList(String name, @Nullable MemoryMapStorage storage, int e
     this.memoryMapStorage = storage;
 
     // only a few standard types
-    addRowType(new IDType());
+    addRowType(new IDType(), new TagDataType());
```
robinschmid, line 222: DataTypes.getAll(IDType.class, TagDataType.class)

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
SteffenHeu: how would we listen to changes then?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/row_type_filter/TagFilterComponent.java
```diff
@@ -0,0 +1,139 @@
...
+import javafx.scene.layout.GridPane;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Checkbox-grid editor for the tag selection of a row-type filter.
+ */
+public final class TagFilterComponent extends GridPane implements ValuePropertyComponent<BitSet> {
+
+  private static final int GRID_WIDTH = 3;
```
robinschmid, line 47: I hope most people will stick to a sane number of tags. 
Maybe a FlowPane instead of grid is flexible enough for any number of check boxes?

Or if grid, we could make it a bit dynamic into more like a square layout?
Like `columns = ceil(sqrt(numTags))` or something like that.Then we have something like 2x2, 3x2, 3x3, 4x3  columns x rows

mzmine-community/src/main/resources/themes/jabref_additions_common.css
```diff
@@ -41,6 +42,17 @@
     -fx-background-insets: 2px;
 }
 
+/* Per-tag colors are supplied as looked-up colors on each checkbox. */
+.tag-check-box > .box,
+.tag-check-box:selected > .box {
+    -fx-background-color: -tag-color;
+    -fx-border-color: -tag-color;
```
robinschmid, line 49: the check marks are very small and hard to see on some colors, especially harder for for people with different vision. 

Maybe the easiest best contrast would be filled box / empty box. Like unselected is just colorful border and when you click it it fills.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
robinschmid: maybe we can just push the updates to all the feature table containing tabs like we do for data changes in the table? 
This way we also do not have to remember to always unsubscribe the listener. 
Currently I think its only unsubscribed if the selected feature list in the table changes?

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/row_type_filter/TagFilterComponent.java
SteffenHeu: the idea is to not change the number of columns. when adding a new tag causes the columns to reorganise it makes it harder to comprehend what is what

mzmine-community/src/main/resources/themes/jabref_additions_common.css
robinschmid: Could look something like this. More flat 
image

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/TagDataType.java
```diff
@@ -0,0 +1,174 @@
...
+      column.setCellFactory(_ -> new TagTreeTableCell(this));
+      // The generic value factory omits null values. Tags need an empty value so an unchecked grid
+      // is visible before the first tag is selected.
+      column.setCellValueFactory(data -> {
+        final ModularFeatureListRow row = data.getValue().getValue();
+        final BitSet tags = row.get(TagDataType.this);
+        return new SimpleObjectProperty<>(tags == null ? new BitSet() : tags);
+      });
+      column.setMaxWidth(50);
+      column.setMinWidth(50);
```
robinschmid, line 120: this seems to be not needed? 
Works fine without and then it also works in presentation mode or with other layouts.

mzmine-community/src/main/resources/themes/jabref_additions_common.css
robinschmid: Tried a few things here:
https://github.com/SteffenHeu/mzmine3/pull/33

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/row_type_filter/TagFilterComponent.java
robinschmid: Tested flowpane here: 
https://github.com/SteffenHeu/mzmine3/pull/33

robinschmid:
Looks good just a few ideas but can be merged

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/ModularFeatureList.java
SteffenHeu: done

mzmine-community/src/main/java/io/github/mzmine/parameters/parametertypes/row_type_filter/TagFilterComponent.java
SteffenHeu: cherrypicked the coloring changes. makes the column layout differ from the filter. will test different approaches.

mzmine-community/src/main/java/io/github/mzmine/datamodel/features/types/TagDataType.java
```diff
@@ -0,0 +1,172 @@
...
+      }
+    }
+    return compare != 0 ? compare : Integer.compare(a.length(), b.length());
+  };
+  private static final Logger logger = Logger.getLogger(TagDataType.class.getName());
+
+  @Override
+  public @NotNull String getUniqueID() {
+    // Never change the ID for compatibility during saving/loading of type.
+    return "row_tags";
```
robinschmid: not sure whats best. 
row_tags is a bit nicer definition. 
tags just naming it tags would make it more general.

